PR #3226, opened 2015-01-02
Reverts CLC-4616 - Updated Roster view to use apiServer print utility

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-4608

This reverts commit 16fd49ca4e56a105b8b437df1b6d61a41bb2b573.


redconfetti:
@christianv or @raydavis , this revert should fix the functionality on the 'Print' button in the Roster LTI tool. It was working before I made this change in favor of a change that Christian suggested in #3221

app/assets/templates/canvas_embedded/roster.html
```diff
@@ -14,7 +14,7 @@
       class="cc-button cc-button-spaced">
         <i class="fa fa-download"></i> Export
       </a>
-      <a class="cc-button cc-button-blue cc-button-spaced" data-ng-click="apiService.util.printPage()">
```
raydavis: @redconfetti , based on @christianv 's original comment and pointer at https://github.com/ets-berkeley-edu/calcentral/pull/3221, I believe you needed to use "api.util.printPage()" rather than "apiService.util.printPage()". Could you give that a try?

redconfetti: Aaah. I see what I did wrong :(

app/assets/templates/canvas_embedded/roster.html
```diff
@@ -14,7 +14,7 @@
...
+      <a class="cc-button cc-button-blue cc-button-spaced" data-ng-click="print()">
```
christianv: @redconfetti indeed, instead of using `print()`, you should be able to use `api.util.priltPage`. In the controller we use `apiService`, in the view we can use `api`.

This is where we expose `apiService` to the view globally:
https://github.com/ets-berkeley-edu/calcentral/blob/de4c82420bd96c7328b113489bd8509682e85a78/app/assets/javascripts/angular/controllers/calcentralController.js#L9

/cc @raydavis

PR #3228, opened 2015-01-02
CLC-4435 Re-enable Spork for JRuby

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-4435

The RubyMine integration is still broken, unfortunately.


johncrossman:
URL and version number are valid. Looks good.

PR #3232, opened 2015-01-05
CLC-4659 Header-based authentication for apis.berkeley.edu feeds

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-4659

We were recently told that our old authentication approach will stop working on January 20. This leaves barely enough time for us to handle it in our normal release cycle, and so as soon as this checks out on master, I'll prepare a QA PR.


ctweney:
Change looks right to me. I'll wait for green Bamboo & Travis before merging.

ctweney:
Bamboo is fine, and so is 1/2 Travis. Thanks for taking care of this change so quickly @raydavis !

PR #3235, opened 2015-01-06
CLC-4674 Handle non-HTML formats by redirecting

ctweney:
https://jira.ets.berkeley.edu/jira/browse/CLC-4674

### DIFF
--- a/app/controllers/bootstrap_controller.rb
+++ b/app/controllers/bootstrap_controller.rb
@@ -6,8 +6,14 @@ class BootstrapController < ApplicationController
   layout 'application'
   caches_action :index, :layout => false
 
+  # index redirects to / if client is being wacky and asking for a format other than HTML.
   def index
-    respond_to :html
+    respond_to do |format|
+      format.html
+      format.all do
+        redirect_to root_path
+      end
+    end
   end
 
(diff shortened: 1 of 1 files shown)


app/controllers/bootstrap_controller.rb
```diff
@@ -6,8 +6,14 @@ class BootstrapController < ApplicationController
   layout 'application'
   caches_action :index, :layout => false
 
+  # index redirects to / if client is being wacky and asking for a format other than HTML.
   def index
-    respond_to :html
+    respond_to do |format|
+      format.html
+      format.all do
```
christianv: @ctweney is this going to show the 404 page when we go to `index.json` or redirect to `/`? Right now it looks like it'll redirect to root, but I'm not sure that's the best user experience.

ctweney: It redirects to /. The 404 page would be a viable alternative. The previous behavior (an HTTP 406 response with no body) is a terrible user experience that could only make REST gurus happy.

christianv: @ctweney thanks, created a follow up at https://jira.ets.berkeley.edu/jira/browse/CLC-4699. Also completely agree that this PR already makes the UX better.

PR #3242, opened 2015-01-09
CLC-4654 - Alt media: caret needs to point downward when expanding

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-4654

Adds jQuery event method to toggler causing the arrow icon to properly toggle between 'icon-arrow-right' and 'icon-arrow-down' classes.

### DIFF
--- a/public/canvas/canvas-customization.js
+++ b/public/canvas/canvas-customization.js
@@ -32,6 +32,14 @@
         var $breadcrumbs = $('nav#breadcrumbs');
         if ($breadcrumbs.data('calcentral-alt-media-notice-applied') !== 'true') {
           $breadcrumbs.append(altMediaPanel);
+          // apply icon toggler - see CLC-4654
+          $('.element_toggler[aria-controls]').on('click', function() {
+            var $icon = $(this).find('i[class*="icon-arrow"]');
+            if ($icon.length) {
+              $icon.toggleClass('icon-arrow-down');
+              $icon.toggleClass('icon-arrow-right');
+            }
+          });
           $breadcrumbs.data('calcentral-alt-media-notice-applied', 'true');
         }
       };
--- a/public/canvas/canvas-customization.js
+++ b/public/canvas/canvas-customization.js
@@ -36,8 +36,7 @@
           $('.element_toggler[aria-controls]').on('click', function() {
             var $icon = $(this).find('i[class*="icon-arrow"]');
             if ($icon.length) {
-              $icon.toggleClass('icon-arrow-down');
-              $icon.toggleClass('icon-arrow-right');
+              $icon.toggleClass('icon-arrow-down icon-arrow-right');
             }
           });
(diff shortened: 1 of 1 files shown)


redconfetti:
@nicolaasmatthijs you want to take a gander at this?

public/canvas/canvas-customization.js
```diff
@@ -32,6 +32,14 @@
         var $breadcrumbs = $('nav#breadcrumbs');
         if ($breadcrumbs.data('calcentral-alt-media-notice-applied') !== 'true') {
           $breadcrumbs.append(altMediaPanel);
+          // apply icon toggler - see CLC-4654
+          $('.element_toggler[aria-controls]').on('click', function() {
+            var $icon = $(this).find('i[class*="icon-arrow"]');
```
nicolaasmatthijs: I'd prefer not to use `class*=` like selectors. Can we either just use `i` if no other icons are available or give the `i` tag a specific class (e.g. `icon-arrow`)?

public/canvas/canvas-customization.js
```diff
@@ -32,6 +32,14 @@
...
+            if ($icon.length) {
+              $icon.toggleClass('icon-arrow-down');
```
nicolaasmatthijs: You can just do `$icon.toggleClass('icon-arrow-down icon-arrow-right');` here

nicolaasmatthijs:
Finished review. Back to @redconfetti for follow-up

public/canvas/canvas-customization.js
redconfetti: The reason I'm using this is because I'm modeling after [the code that Canvas uses to apply the toggle event](https://github.com/instructure/canvas-lms/blob/3402d55a/app/coffeescripts/behaviors/elementToggler.coffee#L102) to these elements. Because my page elements are added after this code is parsed, the events aren't attached to my content.

It's searching specifically within element toggler areas on the page (by class because there may be more than one), and applying this only to the 'i' elements found within those areas that have any variation of the icon arrows (icon-arrow-down, icon-arrow-right). This makes it so that the element togglers may be implemented with the content shown or hidden by default. 

If I hard-code a specific class, I'll only be able to re-use this code with togglers that start with the content hidden.

redconfetti:
@nicolaasmatthijs One TravisCI job failed, the other succeeded. The one that failed had no specific error shown... it just failed to run jscs or jshint with no error. I'm getting no errors locally also.

We good to go? Or you still concerned about the wildcard class selector that I'm using (just like Canvas themselves use)?

nicolaasmatthijs:
@redconfetti : Looks like the Travis build failed. It doesn't seem to be related to this patch though, is it safe to ignore?

nicolaasmatthijs:
> We good to go? Or you still concerned about the wildcard class selector that I'm using (just like Canvas themselves use)?

Nope, that's fine.

Merging ...

PR #3243, opened 2015-01-09
CLC-4725 - Roster Print: Remove SIS ID when printing the roster

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-4725

Applies 'cc-page-roster-student-id' class to all student ids displayed. Hides these student IDs from print media using CSS.

### DIFF
--- a/app/assets/stylesheets/_canvas_roster.scss
+++ b/app/assets/stylesheets/_canvas_roster.scss
@@ -77,10 +77,6 @@
       white-space: pre-wrap;
     }
 
-    .cc-page-roster-student-id {
-      display: none;
-    }
-
     .cc-page-roster-list {
       li {
         margin-top: 15px;
--- a/app/assets/stylesheets/_canvas_roster.scss
+++ b/app/assets/stylesheets/_canvas_roster.scss
@@ -77,6 +77,10 @@
       white-space: pre-wrap;
     }
 
+    .cc-page-roster-student-id {
+      display: none;
+    }
+
     .cc-page-roster-list {
       li {
         margin-top: 15px;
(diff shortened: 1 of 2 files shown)


redconfetti:
@nicolaasmatthijs can you check this one too, resident JS guy?

app/assets/templates/canvas_embedded/roster.html
```diff
@@ -34,7 +34,7 @@
       </div>
       <div class="cc-page-roster-student-name" data-ng-bind="student.first_name"></div>
       <div class="cc-page-roster-student-name"><strong data-ng-bind="student.last_name"></strong></div>
-      <div data-ng-bind="student.student_id"></div>
+      <div class="cc-page-roster-student-id" data-ng-bind="student.student_id"></div>
```
nicolaasmatthijs: Looks like you can just use the `hide` Foundation utility class here (http://foundation.zurb.com/docs/utility-classes.html)

nicolaasmatthijs:
Finished review. Back to @redconfetti for follow-up

app/assets/templates/canvas_embedded/roster.html
redconfetti: If I used the 'hide' class, it would always hide. In this case I'm wanting to give it a specific class, and then specify in the CSS for print media that it doesn't display.

app/assets/stylesheets/_canvas_roster.scss
```diff
@@ -77,6 +77,10 @@
       white-space: pre-wrap;
     }
 
+    .cc-page-roster-student-id {
```
nicolaasmatthijs: Would you mind adding a comment here to explain that `hide` can't be used because of the print CSS?

redconfetti: You know what, I wanted to see if there was a 'cc-hide' class defined, because I know that Christian wants to avoid all Foundation styles except for ones related to [the grid layout](http://foundation.zurb.com/docs/components/grid.html), because it provides us with a responsive design across platforms (mobile, tablet, etc).

See [cc-print-hide class](https://github.com/ets-berkeley-edu/calcentral/blob/b5c7fcda/app/assets/stylesheets/_print.scss#L62)

I just committed an update that simply uses cc-print-hide instead.

nicolaasmatthijs:
Cool. Looks good. Merged.

PR #3244, opened 2015-01-09
CLC-4655 - Alt media: Alt media box isn't aligned with the files panel

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-4655

Adjusted alt-media box to align with newer design.

### DIFF
--- a/public/canvas/canvas-skin.css
+++ b/public/canvas/canvas-skin.css
@@ -95,7 +95,7 @@ button#start_new_course.btn.button-sidebar-wide {
 }
 
 .cc-alt-media-alert-container {
-  padding: 0 1em;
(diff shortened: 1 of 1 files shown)


redconfetti:
@nicolaasmatthijs Just so you know, I'm wanting the left and right to have different padding sizes. In this case the left and top have the same size... but this [cannot be summarized](http://www.w3schools.com/css/css_padding.asp).

redconfetti:
Note: One of the TravisCI instances succeeded, the other was interrupted by a Gem error or something.

nicolaasmatthijs:
Looks good. Merged.

PR #3245, opened 2015-01-11
CLC-4700 - Update gem - google-api-client >= 0.8.2

christianv:
Thanks to @ctweney for reporting the certificate issue - https://github.com/google/google-api-ruby-client/issues/190 which should be fixed in 0.8.2

https://jira.ets.berkeley.edu/jira/browse/CLC-4700

Changelog:
# 0.8.2
* Fixes for file storage and missing cacerts file

# 0.8.1
* Fix logger in rails

# 0.8.0
* Refactored credential storage, added support for redis
* Update gem depdendencies
* Fixed retry logic to allow for auth retries independent of the overall number of retries
* Added `:force_encoding` option to set body content encoding based on the Content-Type header
* Batch requests with the service interface now inherit the service's connection
* `register_discover_document` now returns the API instance
* Added `:proxy` option to set Faraday's HTTP proxy setting
* Added `:faraday_options` option to allow passthrough settings to Faraday connection
* Drop 1.8.x support
* This will be the last release with 1.9.x support


ctweney:
Looks good, thanks for working this in @christianv !

PR #3247, opened 2015-01-12
CLC-4644 Handle undefined future bCourses terms

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-4644

### DIFF
--- a/app/models/canvas/proxy.rb
+++ b/app/models/canvas/proxy.rb
@@ -87,13 +87,26 @@ module Canvas
 
     def self.canvas_current_terms
       terms = []
-      campus_terms = Berkeley::Terms.fetch
-      if (future_term = campus_terms.future) && future_term.name == 'Fall'
+      terms_from_campus = Berkeley::Terms.fetch
+      terms_from_canvas = Canvas::Terms.fetch
+
+      # Get current and next term, and optionally future fall term, from campus data
+      if (future_term = terms_from_campus.future) && future_term.name == 'Fall'
         terms.push future_term
       end
-      terms.push campus_terms.next if campus_terms.next
-      terms.push campus_terms.current
-      terms
+      terms.push terms_from_campus.next if terms_from_campus.next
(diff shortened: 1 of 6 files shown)


app/models/canvas/proxy.rb
```diff
@@ -87,13 +87,26 @@ def self.has_account?(user_id)
 
     def self.canvas_current_terms
       terms = []
-      campus_terms = Berkeley::Terms.fetch
-      if (future_term = campus_terms.future) && future_term.name == 'Fall'
+      terms_from_campus = Berkeley::Terms.fetch
+      terms_from_canvas = Canvas::Terms.fetch
+
+      #Get current and next term, and optionally future fall term, from campus data
```
redconfetti: Could we add a space after the comment hash symbols on line 93 and 100?

redconfetti:
Can you add a simple Rspec test for the new Canvas::Terms model? It just needs to perform the same request that you did to generate the VCR recording, and thus it will actually emulate the request/response from the VCR recording and return the data from the VCR recording response (in Ruby object / hash format).

It's nice to have this so that a developer can get a quick glimpse at the interface for the new model, as well as the data structures one should expect returned from Canvas.

spec/models/canvas/terms_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+  it 'should return an array of terms from Canvas' do
+    expect(@terms).to be_a Array
+    expect(@terms).to_not be_empty
+  end
+
+  it 'should return terms as hashes with id, name, and (unless default term) SIS ID' do
+    @terms.each do |term|
+      expect(term).to be_a Hash
+      expect(term['id']).to be_a Integer
+      expect(term['name']).to be_a String
```
redconfetti: Oh wow. I didn't know that 'be_a' existed. All this time I've been using be_an_instance_of

redconfetti:
party-hard

PR #3253, opened 2015-01-13
CLC-4712 - Course Provisioning: Update course boxes to show course title and number of sections

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-4712

### DIFF
--- a/app/assets/templates/canvas_embedded/canvas_course_sections_form.html
+++ b/app/assets/templates/canvas_embedded/canvas_course_sections_form.html
@@ -1,7 +1,14 @@
 <div class="cc-template-canvas-manage-sections">
   <ul class="cc-template-canvas-manage-sections-margin">
     <li data-ng-repeat="course in currentCourses" class="cc-template-canvas-manage-sections-form-course cc-template-canvas-manage-sections-roster-area">
-      <strong>{{course.title}}</strong><strong data-ng-if="course.course_code"> ({{course.course_code}})</strong>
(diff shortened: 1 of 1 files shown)


raydavis:
See, now I never would've even thought of checking for a pluralize directive.

redconfetti:
@raydavis Yeah, I figured Rails framework provides such helpers, why wouldn't AngularJS.

app/assets/templates/canvas_embedded/canvas_course_sections_form.html
```diff
@@ -1,7 +1,14 @@
 <div class="cc-template-canvas-manage-sections">
   <ul class="cc-template-canvas-manage-sections-margin">
     <li data-ng-repeat="course in currentCourses" class="cc-template-canvas-manage-sections-form-course cc-template-canvas-manage-sections-roster-area">
-      <strong>{{course.title}}</strong><strong data-ng-if="course.course_code"> ({{course.course_code}})</strong>
+      <strong>{{course.title}}</strong>
```
christianv: @redconfetti just fyi - if we're changing lines, we should try to use `ng-bind` instead of `{{}}`

redconfetti: @christianv, I'm not sure why this is always the case, but it seems you ding me for code conventions that I leave in place, not the ones I didn't do in the first place.

I'll fix this soon :disappointed:

PR #3259, opened 2015-01-15
CLC-4682 Find a Person to Add: fix formatting of UID parameter

paulkerschen:
Fixes broken UID search in "Find a Person to Add": https://jira.ets.berkeley.edu/jira/browse/CLC-4682

The API endpoint expects an underscored parameter value, and changing the view code to match seemed less inconsistent than telling the API to expect camelcase.


raydavis:
If the front-end checks pass in Travis-CI, the change is OK by me.

raydavis:
(speaking as the dev who probably caused the bug)

PR #3260, opened 2015-01-15
CLC-4329 Course provisioning: validate "Site Name" as required field

paulkerschen:
Fixes https://jira.ets.berkeley.edu/jira/browse/CLC-4329 along the lines of [this suggestion](http://stackoverflow.com/questions/25471429/angularjs-ng-required-not-working-on-ios) pointed out by @christianv: explicitly show an error message and disable the submit button instead of relying on HTML5 required attribute messaging, which isn't working in Safari.

The extra validity check in the controller code might be redundant, but safety first.

Someone who knows more Angular than me should definitely check this!

### DIFF
--- a/app/assets/javascripts/angular/controllers/pages/canvasCourseProvisionController.js
+++ b/app/assets/javascripts/angular/controllers/pages/canvasCourseProvisionController.js
@@ -92,6 +92,9 @@
     };
 
     $scope.createCourseSiteJob = function() {
+      if ($scope.createCourseSiteForm.$invalid) {
+        return;
+      }
       setErrorText();
       var ccns = selectedCcns();
       if (ccns.length > 0) {
--- a/app/assets/templates/canvas_embedded/course_provision.html
+++ b/app/assets/templates/canvas_embedded/course_provision.html
@@ -121,7 +121,7 @@
             </ul>
           </div>
 
-          <form data-ng-submit="createCourseSiteJob()" class="cc-page-course-provision-form cc-page-provision-site-labels-form">
+          <form name="createCourseSiteForm" data-ng-submit="createCourseSiteJob()" class="cc-page-course-provision-form cc-page-provision-site-labels-form">
             <div class="row">
               <div class="medium-6">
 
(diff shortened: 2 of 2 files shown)


app/assets/javascripts/angular/controllers/pages/canvasCourseProvisionController.js
```diff
@@ -92,34 +92,35 @@
...
-          'termSlug': $scope.currentSemester,
-          'ccns': ccns
-        };
-        if ($scope.is_admin) {
-          if ($scope.adminMode !== 'by_ccn' && $scope.admin_acting_as) {
-            newCourse.admin_acting_as = $scope.admin_acting_as;
-          } else if ($scope.adminMode === 'by_ccn' && $scope.admin_by_ccns) {
-            newCourse.admin_by_ccns = $scope.admin_by_ccns.match(/\w+/g);
-            newCourse.admin_term_slug = $scope.currentAdminSemester;
+      if ($scope.createCourseSiteForm.$valid) {
```
christianv: @paulkerschen might be a bit cleaner to stop executing the function instead of wrapping everything in a big `if` statement.

```javascript
if (!$scope.createCourseSiteForm.$valid) {
  return
}
setErrorText();
// ...
```

app/assets/templates/canvas_embedded/course_provision.html
```diff
@@ -130,7 +130,11 @@
                     <label for="siteName" class="right">Site Name:</label>
                   </div>
                   <div class="medium-8 columns">
-                    <input type="text" id="siteName" data-ng-model="siteName" data-ng-required="true">
+                    <input type="text" name="siteName" id="siteName" data-ng-model="siteName" data-ng-required="true">
+                    <div ng-show="createCourseSiteForm.siteName.$error.required" class="cc-notice-box cc-notice-error">
```
christianv: @paulkerschen couple of small suggestions:

* Use `ng-if` instead of `ng-show` (and see whether it still works) - http://stackoverflow.com/questions/19177732/what-is-the-difference-between-ng-if-and-ng-show-ng-hide
* Apply the `cc-radius` class
* Show the `exclamation` icon before the text

The last 2 items are to make it consistent with
https://github.com/ets-berkeley-edu/calcentral/blob/63ac6489b06ab4e7578f281ae9d1e780f5aac33c/app/assets/templates/canvas_embedded/course_add_user.html#L56

paulkerschen:
Thanks @christianv ; looks like ng-if does the job.

raydavis:
Thanks, @christianv & @paulkerschen !

PR #3263, opened 2015-01-15
CLC-4715 - Remove typo equal sign character

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-4715


raydavis:
Looks like an improvement to me!

PR #3277, opened 2015-01-21
OEC-521, Update settings.yml to have current term, proper depts

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/OEC-521


johncrossman:
I'll need to update a test here. Closing

PR #3278, opened 2015-01-21
CLC-4676: UpNext Selenium test

pfarestveit:
https://jira.ets.berkeley.edu/jira/browse/CLC-4676

The new test creates a calendar event at Google and then verifies that the event appears in CalCentral, with all the expected detail. It also opens the event in Google and verifies that the event detail view loads. Since the test will run in the wee hours, it will ensure that the UpNext feed is updated immediately each day.

This also includes some clean-up of related Google mail and task tests.


johncrossman:
Looks good. Nice work. Travis passed 1 out of 2.

PR #3279, opened 2015-01-21
CLC-4745 - Roster Photos: filtering by section removes all photos

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-4745


app/assets/javascripts/angular/controllers/pages/rosterController.js
```diff
@@ -9,6 +9,20 @@
       apiService.util.setTitle('Roster Photos');
     }
 
+    $scope.studentInSectionFilter = function(student) {
+      if (!$scope.searchSection) {
+        return true;
+      }
+      var response = false;
+      for (var i = 0; i < student.sections.length; i++) {
```
christianv: @paulkerschen this code seems pretty inefficient, would it be possible to return a `sectionsIds` array from the back-end and run `indexOf` on it to compare?

app/assets/javascripts/angular/controllers/pages/rosterController.js
```diff
@@ -9,6 +9,20 @@
...
+        if (student.sections[i].id === parseInt($scope.searchSection)) {
```
christianv: @paulkerschen fyi - `parseInt` should always have a radix parameter. (here `10`)

app/assets/javascripts/angular/controllers/pages/rosterController.js
christianv: @paulkerschen looks like we already have `section_ccns` in the JSON

christianv:
lgtm as soon as travis is green

PR #3281, opened 2015-01-21
CLC-4639 Only show official sections in bCourses Roster Photos

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-4639

Per Ray's spec, the Roster Photos tool no longer checks Canvas course site memberships and relies solely on campus data when loading. This should substantially speed up loading, with some UX effects as detailed in the JIRA.

### DIFF
--- a/app/assets/javascripts/angular/controllers/pages/rosterController.js
+++ b/app/assets/javascripts/angular/controllers/pages/rosterController.js
@@ -5,7 +5,7 @@
   /**
    * Canvas roster photos LTI app controller
    */
-  angular.module('calcentral.controllers').controller('RosterController', function(apiService, rosterFactory, $routeParams, $scope) {
+  angular.module('calcentral.controllers').controller('RosterController', function(apiService, rosterFactory, $routeParams, $scope, $window) {
     if ($routeParams.canvasCourseId) {
       apiService.util.setTitle('Roster Photos');
     }
@@ -14,12 +14,14 @@
       if (!$scope.searchSection) {
         return true;
       }
-      return (student.section_ccns.indexOf($scope.searchSection) !== -1);
+      var section_ccn = parseInt($scope.searchSection, 10);
+      return (student.section_ccns.indexOf(section_ccn) !== -1);
     };
 
     var getRoster = function() {
       $scope.context = $scope.campusCourseId ? 'campus' : 'canvas';
       $scope.courseId = $scope.campusCourseId || $routeParams.canvasCourseId || 'embedded';
+      $scope.origin = $window.location.origin;
 
       rosterFactory.getRoster($scope.context, $scope.courseId).success(function(data) {
         angular.extend($scope, data);
--- a/app/assets/templates/canvas_embedded/roster.html
+++ b/app/assets/templates/canvas_embedded/roster.html
@@ -22,14 +22,14 @@
 
   <ul class="cc-page-roster-list">
     <li data-ng-repeat="student in students | filter:search | filter:studentInSectionFilter | orderBy:'last_name'">
-      <div data-ng-show="student.photo">
-        <a data-ng-href="{{student.profile_url}}">
-          <img data-ng-src="{{student.photo}}" width="72" height="96" />
+      <div data-ng-if="student.profile_url">
+        <a data-ng-href="student.profile_url" target="_top">
+          <div data-ng-include="'canvas_embedded/roster_photo.html'"></div>
         </a>
       </div>
-      <div data-ng-hide="student.photo">
-        <a data-ng-href="{{student.profile_url}}">
-          <div class="cc-image-responsive cc-page-roster-image-unavailable" title="Photo unavailable"></div>
+      <div data-ng-if="!student.profile_url">
+        <a data-ng-href="{{origin}}/{{context}}/{{courseId}}/profile/{{student.login_id}}" target="_top">
+          <div data-ng-include="'canvas_embedded/roster_photo.html'"></div>
         </a>
       </div>
       <div class="cc-page-roster-student-name" data-ng-bind="student.first_name"></div>
--- a/app/assets/templates/canvas_embedded/roster_photo.html
+++ b/app/assets/templates/canvas_embedded/roster_photo.html
@@ -1,0 +1,2 @@
+<img data-ng-if="student.photo" data-ng-src="{{student.photo}}" width="72" height="96" alt="Photo of {{student.first_name}} {{student.last_name}}"/>
+<div data-ng-if="!student.photo" class="cc-image-responsive cc-page-roster-image-unavailable" title="Photo unavailable"></div>
--- a/app/controllers/canvas_rosters_controller.rb
+++ b/app/controllers/canvas_rosters_controller.rb
@@ -37,4 +37,13 @@ class CanvasRostersController < RostersController
     serve_photo
   end
 
+  # GET /canvas/:canvas_course_id/profile/:person_id
+  def profile
+    user_id = Integer(params[:person_id], 10)
+    if (profile_url = Canvas::CanvasRosters.new(session[:user_id], course_id: canvas_course_id).profile_url_for_ldap_id(user_id))
+      redirect_to profile_url
+    else
+      redirect_to url_for_path '/404'
+    end
+  end
 end
--- a/app/models/canvas/canvas_rosters.rb
+++ b/app/models/canvas/canvas_rosters.rb
@@ -11,63 +11,60 @@ module Canvas
         students: []
       }
       campus_enrollment_map = {}
-      # Fill in the Canvas course sections (not to be confused with official campus sections).
-      response = Canvas::CourseSections.new(course_id: @canvas_course_id).sections_list
-      return feed unless (response && response.status == 200 && canvas_sections = safe_json(response.body))
-      canvas_sections.each do |canvas_section|
-        canvas_section_id = canvas_section['id']
-        sis_id = canvas_section['sis_section_id']
+      # Look up Canvas course sections associated with official campus sections.
+      official_sections = Canvas::CourseSections.new(course_id: @canvas_course_id).official_section_identifiers
+      return feed unless official_sections
+      official_sections.each do |official_section|
+        canvas_section_id = official_section['id']
+        sis_id = official_section['sis_section_id']
         feed[:sections] << {
           id: canvas_section_id,
-          name: canvas_section['name'],
+          name: official_section['name'],
           sis_id: sis_id
         }
-        # Get the official campus section enrollments (if any) which are associated with these Canvas course sections.
-        if (campus_section = Canvas::Proxy.sis_section_id_to_ccn_and_term(sis_id))
-          section_enrollments = CampusOracle::Queries.get_enrolled_students(campus_section[:ccn], campus_section[:term_yr], campus_section[:term_cd])
-          section_enrollments.each do |enr|
-            if (existing_entry = campus_enrollment_map[enr['ldap_uid']])
-              # We include waitlisted students in the roster. However, we do not show the official photo if the student
-              # is waitlisted in ALL sections.
-              if existing_entry[:enroll_status] == 'W' &&
-                enr['enroll_status'] == 'E'
-                existing_entry[:enroll_status] = 'E'
-              end
-            else
-              campus_enrollment_map[enr['ldap_uid']] = {
-                student_id: enr['student_id'],
-                first_name: enr['first_name'],
-                last_name: enr['last_name'],
-                email: enr['student_email_address'],
-                enroll_status: enr['enroll_status'],
-                photo_bytes: enr['photo_bytes']
-              }
+
+        section_enrollments = CampusOracle::Queries.get_enrolled_students(official_section[:ccn], official_section[:term_yr], official_section[:term_cd])
+        section_enrollments.each do |enr|
+          if (existing_entry = campus_enrollment_map[enr['ldap_uid']])
+            existing_entry[:sections] << {id: canvas_section_id}
+            existing_entry[:section_ccns] << canvas_section_id
+            # We include waitlisted students in the roster. However, we do not show the official photo if the student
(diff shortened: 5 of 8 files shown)


app/assets/javascripts/angular/controllers/pages/rosterController.js
```diff
@@ -5,22 +5,34 @@
   /**
    * Canvas roster photos LTI app controller
    */
-  angular.module('calcentral.controllers').controller('RosterController', function(apiService, rosterFactory, $routeParams, $scope) {
+  angular.module('calcentral.controllers').controller('RosterController', function(apiService, rosterFactory, canvasProfileUrlService, $routeParams, $scope, $window) {
```
christianv: @paulkerschen can we order `canvasProfileUrlService` alphabetically?

app/assets/javascripts/angular/controllers/pages/rosterController.js
```diff
@@ -5,22 +5,34 @@
...
     };
 
-    var getRoster = function() {
-      $scope.context = $scope.campusCourseId ? 'campus' : 'canvas';
-      $scope.courseId = $scope.campusCourseId || $routeParams.canvasCourseId || 'embedded';
+    $scope.redirectToProfile = function(student) {
+      if (student.profile_url) {
+        $window.top.location.href = student.profile_url;
+      } else if ($scope.context === 'canvas') {
+        canvasProfileUrlService.getProfileUrl($scope.courseId, student.login_id).success(function(data) {
```
christianv: @paulkerschen instead of doing an extra fetch for this URL, can we make it part of the back-end feed?

Right now there will be a delay when the user click something that looks like a link. It would be good to avoid that.

app/assets/javascripts/angular/controllers/pages/rosterController.js
```diff
@@ -5,22 +5,34 @@
...
+      var section_ccn = parseInt($scope.searchSection, 10);
+      return (student.section_ccns.indexOf(section_ccn) !== -1);
     };
 
-    var getRoster = function() {
-      $scope.context = $scope.campusCourseId ? 'campus' : 'canvas';
-      $scope.courseId = $scope.campusCourseId || $routeParams.canvasCourseId || 'embedded';
+    $scope.redirectToProfile = function(student) {
+      if (student.profile_url) {
+        $window.top.location.href = student.profile_url;
```
christianv: @paulkerschen instead of using `$window.top.location.href`, does it also work if we use `target="_blank"` on the anchor tag?

app/assets/templates/canvas_embedded/roster.html
```diff
@@ -23,14 +23,10 @@
   <ul class="cc-page-roster-list">
     <li data-ng-repeat="student in students | filter:search | filter:studentInSectionFilter | orderBy:'last_name'">
       <div data-ng-show="student.photo">
-        <a data-ng-href="{{student.profile_url}}">
-          <img data-ng-src="{{student.photo}}" width="72" height="96" />
-        </a>
+        <img class="cc-pointer" data-ng-src="{{student.photo}}" ng-click="redirectToProfile(student)" width="72" height="96" />
```
christianv: @paulkerschen don't make an image clickable, instead, use an anchor tag `<a>` around the image. The biggest reason for this is accessibility.

app/assets/templates/canvas_embedded/roster.html
```diff
@@ -23,14 +23,10 @@
...
       </div>
       <div data-ng-hide="student.photo">
-        <a data-ng-href="{{student.profile_url}}">
-          <div class="cc-image-responsive cc-page-roster-image-unavailable" title="Photo unavailable"></div>
-        </a>
+        <div class="cc-image-responsive cc-page-roster-image-unavailable cc-pointer" ng-click="redirectToProfile(student)" title="Photo unavailable"></div>
```
christianv: @paulkerschen same comment here, don't make a div clickable, instead wrap it in an anchor tag.

fixtures/pretty_vcr_recordings/Canvas_logins_for_ldap_id.json
```diff
@@ -0,0 +1,118 @@
+{
```
christianv: @paulkerschen if I'm not mistaken the recommended way is not to use VCR but plain JSON fixtures. Please check in with @ctweney and @raydavis to verify.

app/assets/stylesheets/_useful.scss
```diff
@@ -60,6 +60,9 @@
     margin-right: 5px;
   }
 }
+.cc-pointer {
```
christianv: @paulkerschen we can probably remove this class if we don't have a clickable `img` and `div`

app/assets/javascripts/angular/controllers/pages/rosterController.js
raydavis: Sorry, @christianv , but for now we decided to take the UX hit of the delay when the user clicks the picture. The goal here is to avoid delays on the display of the whole roster. Full displays of very large course sites (400+ students) are frequent; clicking on a particular picture is very infrequent.

fixtures/pretty_vcr_recordings/Canvas_logins_for_ldap_id.json
raydavis: @christianv , unfortunately the superclass of all the Canvas proxies has not yet been changed to use plain JSON fixtures, so we're locked into that for the moment. Later today I'll add a JIRA ticket to change it.

The good news, though, is that I don't think we need a new proxy.

app/models/canvas/logins.rb
```diff
@@ -27,5 +28,40 @@ def change_sis_user_id(login_id, new_sis_user_id)
       })
     end
 
+    def canvas_id_for_ldap_id(ldap_id, options={})
```
raydavis: Good news, @paulkerschen -- no need for all this; you should be able to re-use the existing SisUserProfile proxy. Even better news: the most-used Canvas API call in CalCentral, and so very likely to be in the cache.

raydavis:
Still catching up with this between meetings, @paulkerschen , but https://github.com/ets-berkeley-edu/calcentral/pull/3279 made me wonder if Canvas-Roster & Campus-Roster are now able to share section-selection code on the front end...

app/assets/templates/canvas_embedded/roster.html
paulkerschen: @christianv Do you mean wrap the div in an anchor tag and set the ng-click on the anchor? Or is it possible to call an Angular controller function from an href attribute?

app/models/canvas/logins.rb
paulkerschen: @raydavis - I see it; I should have looked harder!

paulkerschen:
@raydavis - They may well share that code already; I think the template and controller are common to both paths. I'll confirm though.

app/assets/javascripts/angular/controllers/pages/rosterController.js
paulkerschen: @christianv I don't know! The problem is that, for back-end reasons, the anchor tag won't come populated with the destination URL, so our choice is between one of various JavaScript tricks to call a function on click, which retrieves the URL and changes the window location. It would be nice if `target="_blank"` worked.

app/assets/templates/canvas_embedded/roster.html
redconfetti: As @nicolaasmatthijs noted, if you wrap this with an anchor tag that includes a href="#", the data-ng-click attribute will override the default behavior, thus stopping the browser from scrolling to the top of the page.

Nico also noted that the reasoning behind this is accessibility, such as when a person is tabbing through the page. An image would be skipped, but an anchor would be recognized.

app/models/canvas/canvas_rosters.rb
```diff
@@ -11,22 +11,25 @@ def get_feed_internal
         students: []
       }
       campus_enrollment_map = {}
-      # Fill in the Canvas course sections (not to be confused with official campus sections).
+      # Look up Canvas course sections (not to be confused with official campus sections).
       response = Canvas::CourseSections.new(course_id: @canvas_course_id).sections_list
```
redconfetti: During our quick and high level pair review of this we came across this and it looks like you're doing something similar to what I did in the EGrades Export tool, which is try to obtain a unique list of campus sections that are associated with a course site.

I pointed out that [Canvas::CourseSections#official_section_identifiers](https://github.com/ets-berkeley-edu/calcentral/blob/7da54757/app/models/canvas/course_sections.rb#L14) already performs this same functionality. It would be best if this could be refactored to use this method, instead of duplicating the functionality.

app/assets/templates/canvas_embedded/roster.html
nicolaasmatthijs: Note that the click handler will need to `return false;` to prevent the event from bubbling up. Probably also good to add an `alt` tag to the image.

app/assets/javascripts/angular/controllers/pages/rosterController.js
christianv: @raydavis would it be possible to look at the proposed solution of lazy loading the profile URLs (in batch or 1 by 1) instead of getting them `onclick`?

app/assets/javascripts/angular/controllers/pages/rosterController.js
christianv: @paulkerschen as suggested below, we could lazy load the profile URLs and convert the images to links as soon as they're loaded. Another alternative would be to have the back-end handle the redirect.

fixtures/pretty_vcr_recordings/Canvas_logins_for_ldap_id.json
christianv: @raydavis makes sense, thanks for clarifying!

app/assets/templates/canvas_embedded/roster.html
christianv: @paulkerschen I would suggest lazy loading the profile URLs on load or doing the redirect on the back-end so we always have a URL available.

app/assets/javascripts/angular/controllers/pages/rosterController.js
raydavis: @christianv , sorry, I just don't see the benefit on that one given the relative numbers. Maybe we could talk about it in chat or person?

app/assets/javascripts/angular/controllers/pages/rosterController.js
raydavis: @christianv , @paulkerschen , FWIW having the back-end handle the redirect would work for me.

app/assets/templates/canvas_embedded/roster.html
```diff
@@ -22,14 +22,16 @@
 
   <ul class="cc-page-roster-list">
     <li data-ng-repeat="student in students | filter:search | filter:studentInSectionFilter | orderBy:'last_name'">
-      <div data-ng-show="student.photo">
-        <a data-ng-href="{{student.profile_url}}">
-          <img data-ng-src="{{student.photo}}" width="72" height="96" />
+      <div data-ng-if="student.profile_url">
+        <a data-ng-href="student.profile_url" target="_top">
+          <img data-ng-if="student.photo" data-ng-src="{{student.photo}}" width="72" height="96" alt="Photo of {{student.first_name}} {{student.last_name}}"/>
```
christianv: @paulkerschen can we use `ng-attr-alt` instead?
https://docs.angularjs.org/guide/directive#-ngattr-attribute-bindings

app/assets/templates/canvas_embedded/roster.html
```diff
@@ -22,14 +22,16 @@
...
+          <div data-ng-if="!student.photo" class="cc-image-responsive cc-page-roster-image-unavailable" title="Photo unavailable"></div>
         </a>
       </div>
-      <div data-ng-hide="student.photo">
-        <a data-ng-href="{{student.profile_url}}">
-          <div class="cc-image-responsive cc-page-roster-image-unavailable" title="Photo unavailable"></div>
+      <div data-ng-if="!student.profile_url">
+        <a data-ng-href="{{origin}}/{{context}}/{{courseId}}/profile/{{student.login_id}}" target="_top">
+          <img data-ng-if="student.photo" data-ng-src="{{student.photo}}" width="72" height="96" alt="Photo of {{student.first_name}} {{student.last_name}}"/>
```
christianv: @paulkerschen can we make this a template on its own? `ng-include`

app/controllers/canvas_rosters_controller.rb
```diff
@@ -37,4 +37,13 @@ def photo
     serve_photo
   end
 
+  # GET /canvas/:canvas_course_id/profile/:person_id
+  def profile
+    user_id = Integer(params[:person_id], 10)
+    if (profile_url = Canvas::CanvasRosters.new(session[:user_id], course_id: canvas_course_id).profile_url_for_ldap_id(user_id))
+      redirect_to profile_url
+    else
+      redirect_to url_for_path '/404'
```
christianv: @paulkerschen +1 thanks!

christianv:
@paulkerschen thanks for making the code, really looks way cleaner now. 2 small follow up comments and we are good to merge (front-end side) after those.

paulkerschen:
@raydavis @redconfetti This new version is substantially slimmer on the back end.

raydavis:
The Travis is half full. Looks grand from here, @paulkerschen -- thanks!

app/assets/templates/canvas_embedded/roster_photo.html
```diff
@@ -0,0 +1,2 @@
+<img data-ng-if="student.photo" data-ng-src="{{student.photo}}" width="72" height="96" alt="Photo of {{student.first_name}} {{student.last_name}}"/>
```
christianv: @paulkerschen seems like a comment was missed, can we use `ng-attr-alt` instead?

PR #3285, opened 2015-01-22
CLC-4575: Selenium test for Cal 1 Card UI

pfarestveit:
https://jira.ets.berkeley.edu/jira/browse/CLC-4575

This test verifies different permutations of the Cal 1 Card UI on My Finances, using a set of real UIDs.  The test verifies that Cal 1 Card data exists for at least one of the test UIDs, so it will fail if the feed is down or if the test UIDs are not useful.  It also verifies the external links to Cal 1 Card and Caldining.  Finally, it generates a CSV containing information about the UIDs to facilitate manual testing and management of the pool of test UIDs themselves.

### DIFF
--- a/spec/ui_selenium/my_finances_data_cal1card_spec.rb
+++ b/spec/ui_selenium/my_finances_data_cal1card_spec.rb
@@ -1,0 +1,148 @@
+require 'spec_helper'
+require 'selenium-webdriver'
+require 'page-object'
+require 'csv'
+require 'json'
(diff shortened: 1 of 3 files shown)


johncrossman:
Looks good. Style-wise which do you prefer? 
  splash_page.load_page(driver)
or 
  splash_page.load_page driver

pfarestveit:
@johncrossman, I prefer total consistency for readability.  If parentheses must be used sometimes, then I prefer to use them all the time.

PR #3288, opened 2015-01-22
CLC-4639 Fix alt tag in roster photos

paulkerschen:
Thanks for the catch on https://github.com/ets-berkeley-edu/calcentral/pull/3281 @christianv - it must have regressed when I moved it to its own template.


christianv:
@paulkerschen thanks! Merged

PR #3289, opened 2015-01-23
CLC-4696 Resize large LTI frames to avoid dual scroll bars

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-4696

### DIFF
--- a/public/canvas/canvas-customization.js
+++ b/public/canvas/canvas-customization.js
@@ -341,6 +341,10 @@
    *    This will respond with a window event back to the LTI iFrame with the following message:
    *    `{scrollPosition: <currentScrollPosition>}`
    *
+   *  - If the iFrame is so enormous as to hit the 5000px limit in Instructure's code,
+   *    resize it ourselves.
+   *    `{subject: 'resizeLargeFrame', height: <height>}`
+   *
    * @param  {Object}    ev         Event that is sent over from the iframe
    * @param  {String}    ev.data    The message sent with the event. Note that this is expected to be a stringified JSON object
    */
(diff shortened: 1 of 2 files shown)


public/canvas/canvas-customization.js
```diff
@@ -375,6 +379,13 @@
           var response = {scrollPosition: scrollPosition};
           ev.source.postMessage(JSON.stringify(response), '*');
         }
+      // Resize frame if too large for Instructure's `public/javascripts/tool_inline.js`
+      } else if (message.subject === 'resizeLargeFrame' && message.height !== undefined) {
```
christianv: @paulkerschen you should be able to replace
`message.height !== undefined`
with
`message.height`

public/canvas/canvas-customization.js
```diff
@@ -375,6 +379,13 @@
...
+        document.getElementById('tool_content').style.height = message.height + 'px';
```
christianv: @paulkerschen although I'm ok with using native JS, in this case jQuery is loaded so you might be able to replace this with:

```javascript
$('#tool_content').height(message.height);
$('.tool_content_wrapper').height('auto');
```

public/canvas/canvas-customization.js
```diff
@@ -375,6 +379,13 @@
           var response = {scrollPosition: scrollPosition};
           ev.source.postMessage(JSON.stringify(response), '*');
         }
+      // Resize frame if too large for Instructure's `public/javascripts/tool_inline.js`
+      } else if (message.subject === 'resizeLargeFrame' && message.height) {
+        document.getElementById('tool_content').style.height = message.height + 'px';
```
christianv: @paulkerschen although I'm ok with using native JS, in this case jQuery is loaded so you might be able to replace this with:

```javascript
$('#tool_content').height(message.height);
$('.tool_content_wrapper').height('auto');
```

paulkerschen: @christianv Sure, I'll take the concision.

redconfetti: I actually like @paulkerschen's survivalist mentality. What if Canvas uses a jQuery library hosted by Google, and a terrorist attack occurs on Google servers... so jQuery is not present, but then the LTI iframe is still resized because of Paul's native JS ... Guess what the result would be?!? The terrorists don't win! Hell yeah! Murica!

PR #3293, opened 2015-01-23
CLC-4745 Fix type inconsistency in roster feeds

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-4745

Changed CanvasRosters to return section CCNs as strings, as CampusRosters does. The front-end rosterController may receive feeds from either endpoint and was understandably getting confused.


christianv:
@paulkerschen lgtm

PR #3300, opened 2015-01-27
CLC-4696 Ensure that LTI frame size updates do in fact update

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-4696

This one's a facepalm. #3289 added variable assignments outside the `updateHeight` function, so they were only set once and wouldn't respond to subsequent changes in the frame height -- or, in the case of slow-loading tools, the initial height of the fully loaded frame.

### DIFF
--- a/app/assets/javascripts/angular/services/utilService.js
+++ b/app/assets/javascripts/angular/services/utilService.js
@@ -102,9 +102,9 @@
      */
     var iframeUpdateHeight = function() {
       if (isInIframe) {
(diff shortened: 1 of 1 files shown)


raydavis:
Oh, ouch! Yeah, that would definitely do it...

raydavis:
@paulkerschen , I'd love to get this into QA, too. Let me know if you'd like help cherrypicking into a QA PR.

PR #3301, opened 2015-01-27
QA CLC-4696 Ensure that LTI frame size updates do in fact update

paulkerschen:


raydavis:
QA version of https://github.com/ets-berkeley-edu/calcentral/pull/3300

raydavis:
Travis is half full.

PR #3302, opened 2015-01-28
OEC-531, secondary x-listed are expected in CSVs of both participating depts

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/OEC-531

For example, secondary MCELLBI sections with cross-listings in PSYCH dept should show in MCELLBI.csv and PSYCH.csv. This allows both depts to vote on whether course should be evaluated. (Reqts confusion has been ironed out. See the OEC-531.)

Fix was to remove the dept constraint in SQL where clause of *get_secondary_cross_listings*

Bonus: *oec.rake courses* gets more logging.


paulkerschen:
Happy Travis.

PR #3303, opened 2015-01-28
CLC-4739 - Roster Export: Add section information to the roster csv export

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-4739

### DIFF
--- a/app/models/rosters/campus.rb
+++ b/app/models/rosters/campus.rb
@@ -61,6 +61,10 @@ module Rosters
           end
         end
       end
+
+      # Create sections hash indexed by CCN
+      sections_index = index_by_attribute(feed[:sections], :ccn)
+
       return feed if campus_enrollment_map.empty?
       campus_enrollment_map.keys.each do |id|
         campus_student = campus_enrollment_map[id]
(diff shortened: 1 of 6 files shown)


app/models/rosters/campus.rb
```diff
@@ -61,6 +61,10 @@ def get_feed_internal
           end
         end
       end
+
+      # Create sections hash indexed by CCN
+      sections_index = index_by_attribute(feed[:sections], :ccn)
```
raydavis: @redconfetti , I think your "index_by_attribute" might do the same thing as a Rubyism we've used elsewhere:

```ruby
sections_index = Hash[feed[:sections].map {|s| [s[:ccn], s}]
```

app/models/rosters/common.rb
```diff
@@ -60,5 +61,18 @@ def photo_data_or_file(student_id)
         end
       end
     end
+
+    def index_by_attribute(array_of_hashes, attribute)
```
raydavis: Probably a good idea to use the Ruby one-liner above as an implementation of (or replacement for) "index_by_attribute", since it's a more-or-less standard formula for doing this.

raydavis:
Aside from my implementation query, this looks good, @redconfetti . I'm guessing it's not urgent for me to merge as is while you're gone, so I'll give you a chance to look at my comment first.

app/models/rosters/campus.rb
redconfetti: Yep. I decided to keep the index_by_attribute method, to keep things as DRY as possible with these two models with similar logic.

PR #3304, opened 2015-01-28
OEC-538, BIO_post_processor should skip non-BIO depts in BIO csv (e.g., MCB cross-listing)

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/OEC-538

One example: The BIO.csv would legitimately get:
  2015-B-42154,INTEGBI 95 LEC 002 SPEC RES IN BIO 1B...

Next, BIO_post_processor would move that row to INTEGBI.csv based on its dept_name=INTEGBI.  Problem is that INTEGBI.csv, based on its own SQL query, already has that data.

Our fix is to have BIO_post_processor skip rows in BIO.csv where dept_name = INTEGBI or MCELLBI.

I've also cleaned up logging summary in *oec.rake*.

### DIFF
--- a/app/models/oec/biology_post_processor.rb
+++ b/app/models/oec/biology_post_processor.rb
@@ -19,10 +19,12 @@ module Oec
           course_name = row[1]
           if index == 0
             header_row = row
+          elsif dept_name.include?(mcellbi_dept) || dept_name.include?(integbi_dept)
+            Rails.logger.warn "#{row[0]} #{row[1]} #{biology.base_file_name} skipped. Course is listed #{dept_name} CSV file."
           else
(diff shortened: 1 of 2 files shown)


ctweney:
This LGTM, although as noted in chat crosslistings with other departments will  need their own fix, which John is going to JIRA separately.

johncrossman:
Thanks @ctweney  For the record, I added your observation on "other departments" in the steps-to-verify https://jira.ets.berkeley.edu/jira/browse/OEC-538  and created a new PR (https://github.com/ets-berkeley-edu/calcentral/pull/3305) with what I think is a fix

PR #3308, opened 2015-01-30
CLC-4735 - Investigate alternative to include & load CalCentral front-end assets

christianv:
https://jira.ets.berkeley.edu/jira/browse/CLC-4735

**NOTE**: Do not merge before the Friday meeting on Jan 30, 2015


christianv:
@raydavis @ctweney @johncrossman @nicolaasmatthijs @redconfetti @paulkerschen - just updated the PR with removing the `bamboo` check and removing `imagemin` / `pngquant`.

christianv:
Travis & Bamboo seem to be green:
https://bamboo.ets.berkeley.edu/bamboo/browse/MYB-CLCMVPCHRISTIAN211-4

raydavis:
@christianv & crew: Will people have time to start testing this today, or should we wait until Monday morning after the release?

christianv:
@raydavis definitely wait with the merge until after Monday morning. It would be great if people have time to start testing early.

raydavis:
Thanks, @christianv -- I have mixed feelings, of course, because one of the things we want to find out about are problems with the final build, copy, deploy, and running-on-the-shared-server steps, which we can't test locally. But on the other hand it's nice to keep the systems clear for emergency post-release fixes.

nicolaasmatthijs:
@raydavis : Can this just be deployed on the shared server from a branch to allow you to keep master clear for emergencies?

raydavis:
@nicolaasmatthijs , when we very quickly talked about this at the meeting, it became clear that a merge to master followed by a revert to master would actually be considerably easier than deployment of a branch, just due to the way we do our deploys from Bamboo.

christianv:
@raydavis totally agree with the revert approach. If there is an emergency that comes up, we can revert, add the patch, test and then re-commit the front end assets part.

app/assets/stylesheets/_foundation_overrides.scss
```diff
@@ -1,35 +0,0 @@
-// Foundation overrides
-$base-font-size: 13px;
```
redconfetti: Oh okay. So these overrides went to another SCSS file - src/assets/stylesheets/_foundation_overrides.scss

Did you drop the $base-font-size, and $block-grid-default-spacing ?

config/routes.rb
```diff
@@ -85,6 +85,7 @@
   get '/api/cache/clear' => 'cache#clear', :defaults => { :format => 'json' }
   get '/api/cache/delete/:key' => 'cache#delete', :defaults => { :format => 'json' }
   get '/api/cache/warm/:uid' => 'cache#warm', :defaults => { :format => 'json' }
+  get '/api/calcentral_config' => 'calcentral_config#get', :via => :get, :defaults => { :format => 'json' }
```
ctweney: This could change to just /api/config (and the controller similarly named "ConfigController") since we know we're CalCentral already.

app/assets/stylesheets/_foundation_overrides.scss
christianv: @redconfetti as mentioned in the presentation, we copied over the CSS (not SCSS) file for foundation. So we no longer need these variables.

Having said that, one of the goals will be to update the foundation stylesheet to remove unused CSS and update properties that we're overriding again in CalCentral.

app/controllers/calcentral_config_controller.rb
```diff
@@ -0,0 +1,12 @@
+class CalcentralConfigController < ApplicationController
```
ctweney: Please add a controller spec for this -- blog_feed_controller_spec.rb provides a nice simple model to work from .

config/routes.rb
christianv: @ctweney Agreed, updated

app/controllers/calcentral_config_controller.rb
```diff
@@ -0,0 +1,12 @@
...
+  before_filter :get_settings, :initialize_calcentral_config
+
+  def get
+    render json: @calcentral_config.merge({
+      # See http://git.io/rgw3Pg
+      csrfParam: request_forgery_protection_token,
```
redconfetti: Can you convert these keys to underscore case please. :smiling_imp:

app/controllers/calcentral_config_controller.rb
christianv: @ctweney should have been added

redconfetti:
LGTM :+1: :smiling_imp: :computer: :boat:

paulkerschen:
It looks like in at least one case (font-awesome) the old gem version is persisting alongside the new package because the gem is a dependency of rails-admin. Can we expect the two to stay out of each other's way?

christianv:
@paulkerschen: yes, the whole reason we still have the the asset pipeline is because of `rails_admin`. They shouldn't conflict since `rails_admin` will only include the files necessary to run `rails_admin`. And CalCentral will only include the files necessary for CalCentral

PR #3309, opened 2015-01-30
OEC-540, omit x-listings where x_listed_flag is null and dept is not participating in OEC

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/OEC-540

We have unofficial cross-listings (e.g., room shares where cross_listed_flag is null) showing up in our OEC data. We will include those CCNs as long as the dept is participating in OEC.  This change will cause 11 CCNs to be omitted from the set of OEC CSV files.


paulkerschen:
LGTM.

PR #3310, opened 2015-01-30
CLC-4663 Remove unnecessary response parsing in Finaid module

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-4663

The goal is to replace the current divergent state of our XML and JSON feed parsing, as described in the ticket, with a standard approach using the parsed_response Ruby object that HTTParty gives us for free. 

However, working directly with that parsed_response object brings in some of the headaches and pitfalls that come with traversing large data structures. My approach here is to wrap that object, and its children as extracted, in a new FeedWrapper class. The main benefits are two: 
- Soft failures on missing or unexpectedly formatted elements (without losing the ability to log surprises). Because we don't have to explicitly check for nils and types, we can keep the concision of the current CSS and XPath selectors without having to actually run CSS and XPath queries on everything.
- Reasonably formatted coercion to strings, dates, etc., in order to replace existing module methods such as `to_text`, `to_array`, etc. in MyAcademics::AcademicsModule.

This PR introduces the overhaul only in the Finaid module, which contained some of the gnarliest XML parsing. Once we have consensus on an approach, I can apply it to the rest of the proxy classes, a task which should involve deleting much more code than is added.

### DIFF
--- a/lib/proxies/feed_wrapper.rb
+++ b/lib/proxies/feed_wrapper.rb
@@ -1,0 +1,86 @@
+class FeedWrapper
+  include Enumerable
+
+  def initialize(obj)
+    @obj = obj
+  end
+
+  def [](key)
+    if @obj.is_a?(Hash) || (@obj.is_a?(Array) && key.is_a?(Integer))
+      self.class.new(@obj[key])
+    else
+      self.class.new(nil)
+    end
+  end
+
+  def ==(obj)
+    if obj.respond_to?(:unwrap)
+      self.unwrap == obj.unwrap
+    else
+      self.unwrap == obj
+    end
+  end
+
+  # This corrects for a quirk in XML-to-object parsing; usually an XML element
+  # is converted to a hash with the element name as key and a hash as a value,
+  # but sibling XML elements with the same name are converted to a hash with the
+  # element name as key and an array of hashes as value. In cases where the number
+  # of elements is unknown, this can ensure that single or blank results still come
+  # wrapped in an array.
+  def as_collection
+    @obj.is_a?(Array) ? self : self.class.new(self.to_a)
+  end
+
+  def blank?
+    @obj.blank?
+  end
+
+  def content(default='')
+    if @obj.is_a? String
+      self.to_text(default)
+    else
+      self['__content__'].to_text(default)
+    end
+  end
+
+  def each
+    if @obj.respond_to?(:each)
+      @obj.each { |e| yield self.class.new(e) }
+    end
+  end
+
+  def find_by(key, value)
+    self.find { |e| e[key].unwrap == value } || self.class.new(nil)
+  end
+
+  def to_a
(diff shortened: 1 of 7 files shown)


spec/lib/proxies/feed_wrapper_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+          <Flora>f</Flora>
+          <Gawain>g</Gawain>
+          <Hrothgar>h</Hrothgar>
+        </Subdocument>
+      </Document>
+    '))
+  end
+
+  context 'wrapping a collection' do
+    before { pp subject.unwrap }
```
raydavis: Looks like this slipped through...

lib/proxies/http_requester.rb
```diff
@@ -6,13 +6,27 @@ def verify_ssl?
...
         url,
         {
           timeout: Settings.application.outgoing_http_timeout,
           verify: verify_ssl?
         }.merge(additional_options)
       )
+      begin
+        if response.parsed_response.nil?
+          logger.error "Unable to parse response from URL (#{url})"
+          logger.error "Remote server status: #{response.code}, body: #{response.body}"
```
raydavis: Mind merging all these pairs into single messages? Makes it easier to match logged errors to causes.

lib/proxies/feed_wrapper.rb
```diff
@@ -0,0 +1,86 @@
...
+    else
+      [@obj]
+    end
+  end
+
+  def to_date(default='')
+    Date.parse(self.to_s).in_time_zone.to_datetime rescue default
+  end
+
+  def to_s(default='')
```
raydavis: I'm a bit worried about undesirable loss of information when overriding such a basic method (in debugging messages, for example). How would you feel about carrying forward the "to_text" brandname?

raydavis:
@paulkerschen, this is a big fat wrapper class, but FinAid certainly demonstrates its use. As you roll out the follow-up PRs, can you watch for clients who only need one little slice of functionality and might be better served by finer-grained mix-ins?

spec/lib/proxies/feed_wrapper_spec.rb
paulkerschen: D'oh!

lib/proxies/feed_wrapper.rb
paulkerschen: That's a good idea.

lib/proxies/http_requester.rb
paulkerschen: Right.

paulkerschen:
Yes indeed. A lot of the smaller XML and JSON clients should better lend themselves to _removing_ code, as per the title of the ticket.

raydavis:
Lookin' sharp, @paulkerschen -- I'll merge post-Travis & pre-assets-revolution.

PR #3311, opened 2015-01-31
NOJIRA - Modified Postgres instructions 

NolanChan:
When I ran rake db:schema:load, I experienced this error:

PG::Error: ERROR:  permission denied to create database CALCENTRAL_DEVELOPMENT

After granting 'createdb' privileges to the user CALCENTRAL_DEVELOPMENT, everything seemed to work ok

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,14 +59,17 @@ http://www.oracle.com/technetwork/java/javase/downloads/index.html
     pg_ctl -D /usr/local/var/postgres -l /usr/local/var/postgres/server.log start
     psql postgres
     create database calcentral_development;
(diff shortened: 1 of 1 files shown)


NolanChan:
Hi @paulkerschen , that was exactly my experience too. I was unsure if my 'ALTER DATABASE calcentral_development OWNER TO calcentral_development' was a necessary line, as I was just shooting in the dark with all the commands that I tried. Since both our experiences were the same, I'll take that as confirmation that we should update the readme with the ALTER command

paulkerschen:
Right on, this looks like how I got it working.

PR #3317, opened 2015-02-02
CLC-4815 - bCourses - AngularJS bootstrap not present

christianv:
https://jira.ets.berkeley.edu/jira/browse/CLC-4815

*Note* This is not the DRY'st solution out there but should unblock @redconfetti
We'll leave the jira open (or create a follow up one) to make this more DRY


christianv:
@raydavis updated the PR with the JSCS fix

PR #3320, opened 2015-02-03
CLC-4811 - Empty Academics "Profile" card for concurrent enrollments

christianv:
https://jira.ets.berkeley.edu/jira/browse/CLC-4811

### DIFF
--- a/src/assets/templates/academics_student_profile.html
+++ b/src/assets/templates/academics_student_profile.html
@@ -13,6 +13,9 @@
       <p data-ng-if="!api.user.profile.roles.student && api.user.profile.roles.exStudent">
         You are not currently considered an active student.
       </p>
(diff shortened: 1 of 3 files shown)


src/assets/templates/academics_student_profile.html
```diff
@@ -63,7 +66,7 @@
       </div>
     </div>
 
-    <div class="cc-table cc-table-top-border cc-clearfix">
+    <div class="cc-table cc-table-top-border cc-clearfix" ng-if="collegeAndLevel.standing || gpaUnits.totalUnits !== null">
```
raydavis: What happens if "standing" is null but we have a "totalUnits" value? This is a real-world possibility because we get standing from the very flaky Bear Facts API but we get "totalUnits" from the more reliable Oracle views.

christianv: @raydavis good point, added a follow up commit

PR #3325, opened 2015-02-03
OEC-531, exclude zero enrollment courses in the rb code, not SQL

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/OEC-531

In some cases of cross-listing, we have sections where there are zero enrollments in one department's cross-listing, but not in the other. The effect of this is that we see a few cases like 2015-B-11700 where it has enrollments, but its matched section 2015-B-53051 does not, so it fails to show up on the Chem spreadsheet.

If we exclude a zero enrollment course in the SQL then we never check the enrollment count of its cross-listings. Therefore, we move the exclusion logic out of SQL and specifically at the point of *output << row*.
 
(Apologies for four commits in one PR. The diff is not overwhelming.)

### DIFF
--- a/app/models/oec/courses.rb
+++ b/app/models/oec/courses.rb
@@ -49,7 +49,7 @@ module Oec
       unless visited_row_set.include? row_as_string
         enrollment_count = course['enrollment_count'].to_i
         catalog_id = course['catalog_id']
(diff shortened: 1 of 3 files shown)


raydavis:
LGTM

PR #3327, opened 2015-02-03
OEC-533, omit wait listed students from course_students.csv

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/OEC-533

Both student-related queries now use the clause: (r.enroll_status = 'E' OR r.enroll_status = 'C')

Also, we use two queries (instead of one per dept) to generate the global list of student and enrollment data.

### DIFF
--- a/lib/tasks/oec.rake
+++ b/lib/tasks/oec.rake
@@ -25,17 +25,15 @@ namespace :oec do
       dept_set.add 'INTEGBI'
       dept_set.add 'MCELLBI'
     end
-    students_csv_created = false
+    ccn_set = Set.new
+    gsi_ccn_set = Set.new
     dept_set.each do |dept_name|
       filename = "#{dept_name.gsub(/\s/, '_')}_courses.csv"
       csv_file = "#{src_dir}/#{filename}"
       if File.exists? csv_file
         reader = Oec::FileReader.new csv_file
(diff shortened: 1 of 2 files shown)


lib/tasks/oec.rake
```diff
@@ -25,17 +25,15 @@ namespace :oec do
...
       csv_file = "#{src_dir}/#{filename}"
       if File.exists? csv_file
         reader = Oec::FileReader.new csv_file
-        [Oec::Students, Oec::CourseStudents].each do |klass|
-          klass.new(reader.ccns, reader.gsi_ccns, dest_dir).export !students_csv_created
-        end
-        students_csv_created = true
-        Rails.logger.warn "#{hr}Files wrote to #{dest_dir}#{hr}"
+        ccn_set.merge reader.ccns.to_set
+        gsi_ccn_set.merge reader.gsi_ccns.to_set
```
paulkerschen: Do these need to be `merge!` for the changes to persist?

paulkerschen: Never mind, Set#merge is different from Hash#merge. I didn't know that!

johncrossman:
Thanks @paulkerschen.  I was doing my homework on merge!  just as you merged my change.

PR #3328, opened 2015-02-04
CLC-4821 - Remove unused views in /apps/views

christianv:
https://jira.ets.berkeley.edu/jira/browse/CLC-4821


paulkerschen:
So long, ActionView.

PR #3330, opened 2015-02-04
CLC-3515 Replace class Foo::Bar with module Foo; class Bar

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-3513

I'm really sorry that the differ isn't smarter about indentation changes. Apart from the first and last lines of the classes, the only changes I made were to namespacing: 
 - removing explicit module references where they weren't needed; 
 - and prepending a double colon to top-level modules (Sakai, Canvas) where the namespaced module would otherwise be assumed.

I also prepended a double colon to CalLink, which escapes confusion with MyGroups::Callink only by virtue of capitalization (and perhaps that should be standardized in a follow-up PR).

### DIFF
--- a/app/models/my_classes/campus.rb
+++ b/app/models/my_classes/campus.rb
@@ -1,55 +1,58 @@
-class MyClasses::Campus
-  include MyClasses::ClassesModule
+module MyClasses
+  class Campus
+    include ClassesModule
 
-  def fetch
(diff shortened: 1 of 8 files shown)


johncrossman:
1 of 2 Travis are good.

johncrossman:
All seems in order. Thanks @paulkerschen

PR #3331, opened 2015-02-04
CLC-4663 Weaken testext to avoid Bamboo failures

paulkerschen:


johncrossman:
The bar has been lowered. Travis says good.

PR #3333, opened 2015-02-04
OEC-539, rake oec:* with irregular 'src' / 'dest' args use wrong working dir as base

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/OEC-539

This feature is requested by Ops. Examples of the bug below. First example writes files to *~/calcentral* instead of working dir *~/calcentral/tmp/oec*
> cd ~/calcentral/tmp/oec
> rake oec:courses

Old code did not properly expand "..":
> cd ~/calcentral
> rake oec:courses dest=../foo


lib/tasks/oec.rake
```diff
@@ -56,9 +60,21 @@ namespace :oec do
...
   end
 
+  def get_path(arg_name)
+    path = ENV[arg_name].to_s
+    return working_dir if path == ''
+    path.start_with?('.') ? File.expand_path(path, working_dir) : Pathname.new(path).realpath
+  end
+
+  def working_dir
+    ENV['PWD'].to_s == '' ? Dir.pwd : ENV['PWD']
```
raydavis: ```ENV['PWD'].blank?``` would be the more common pattern, FWIW.

raydavis: Also, have you tested that ENV['PWD'] will work? I hadn't seen that before, but websearch + http://ruby-doc.org/stdlib-1.9.3/libdoc/rake/rdoc/Rake/Application.html suggests that "Rake.original_dir" might be the magic word.

johncrossman:
Thanks @raydavis   I will come back to this soon.  I'll submit new PR when ready.

PR #3336, opened 2015-02-05
CLC-4663 Remove unnecessary parsing in EtsBlog proxies

paulkerschen:
Continuing cleanup on https://jira.ets.berkeley.edu/jira/browse/CLC-4663.

EtsBlog::Alerts uses the new FeedWrapper class. EtsBlog::ReleaseNotes is managing without it.

The tests for EtsBlog::Alerts have been slimmed down, because the removal of parsing has left them less to do, and because certain conditions, like a missing fake file, didn't seem to require an explicit test separate from the tests for exception handling.


raydavis:
Totally agree on the test slimming, @paulkerschen -- thanks!

PR #3338, opened 2015-02-05
CLC-4825 - Create a disabledLinkDirective

christianv:
https://jira.ets.berkeley.edu/jira/browse/CLC-4825


redconfetti:
@christianv This still results in the browser seeing the element as disabled because it has a disabled attribute. Could we make this more like ng-disabled, but call it cc-disabled, and have it apply the disabled attribute if the cc-disabled value expression is true?

christianv:
@redconfetti good point, follow up at #3340

PR #3344, opened 2015-02-06
CLC-4823 - Site Provisioning: Review Site Accessibility

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-4823


raydavis:
I like the name switch from "canvas_embedded" to "canvas_shared".

raydavis:
And thank you for making it look more realistic outside the iframe!

src/assets/stylesheets/_canvas_create_course_site.scss
```diff
@@ -68,52 +64,47 @@
...
   }
 
-  .cc-page-provision-form-button {
+  .cc-page-create-course-site-form-button {
     background: linear-gradient($bc-color-form-button-gradient1, $bc-color-form-button-gradient2);
   }
 
-  .cc-page-course-provision-help-notice {
+  .cc-page-create-course-site-hidden {
+    display: none;
```
christianv: @redconfetti this should not be `display:none` since you're applying this to the `label`. Instead, we should be using the `cc-visuallyhidden` class which will make sure a screenreader still reads it.

PR #3345, opened 2015-02-06
OEC-560, logger statements changed from warn to info

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/OEC-560

Simple enough.


raydavis:
Can't argue with that.

PR #3346, opened 2015-02-06
CLC-4823 - Converted label to use cc-visuallyhidden instead of style with display:none

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-4823


christianv:
@redconfetti thanks! lgtm as soon as travis is green

PR #3347, opened 2015-02-06
OEC-561, diff task: log proper message when no diff found

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/OEC-561

Includes:
1. log level tweaks
1. def diff_found? on CoursesDiff
1. more better messaging in oec.rake


johncrossman:
Needs diff resolved. Checking

PR #3349, opened 2015-02-07
OEC-561, diff task: log proper message when no diff found

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/OEC-561

Includes:
1. log level tweaks
2. def diff_found? on CoursesDiff
3. mo' better messaging in oec.rake

### DIFF
--- a/app/models/oec/courses.rb
+++ b/app/models/oec/courses.rb
@@ -50,7 +50,7 @@ module Oec
           output << row
           visited_row_set << row_as_string
         else
-          Rails.logger.debug "#{@dept_name}.csv: Skipping #{course_id}, #{course['dept_name']} #{catalog_id} because enrollment_count=#{enrollment_count}"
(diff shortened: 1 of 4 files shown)


paulkerschen:
Not picking a nit so much as pure curiosity, @johncrossman : why the `diff_found?` method instead of setting `attr_reader` on `@difference_found`? Is it just to have the question mark make clear it's a boolean?

johncrossman:
@paulkerschen Thanks. I'll use attr_reader and add test in spec.

johncrossman:
@paulkerschen  Change is in. To answer your question, using method (instead of attr_reader) gave readability (ie, '?') and default boolean (nil is false). Readability now comes with new var name. Default is explicitly set in init. I thought of using alias_method (with attr_reader) to get '?' in method but that's over-engineering.

paulkerschen:
Makes sense to me! Half a Travis.

PR #3350, opened 2015-02-09
CLC-4830 - CFV - Improve the accessibility on the search details page

christianv:
https://jira.ets.berkeley.edu/jira/browse/CLC-4830

* Add table scopes
* Show `search` instead of `go` or `return` when you're in the search text box on iOS. Thanks @nicolaasmatthijs
* Add labels

### DIFF
--- a/src/assets/templates/cars_activity.html
+++ b/src/assets/templates/cars_activity.html
@@ -7,7 +7,7 @@
   </div>
 
   <div data-cc-spinner-directive>
-    <form name="cc_page_myfinances_account_filter" class="cc-page-myfinances-account-filter">
+    <form name="cc_page_myfinances_account_filter" class="cc-page-myfinances-account-filter" action="/">
       <div class="row">
         <div class="cc-page-myfinances-filter-label">
           <div class="cc-left cc-page-myfinances-filter-element">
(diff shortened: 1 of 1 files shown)


src/assets/templates/cars_activity.html
```diff
@@ -7,7 +7,7 @@ <h2 class="cc-left">Activity</h2>
   </div>
 
   <div data-cc-spinner-directive>
-    <form name="cc_page_myfinances_account_filter" class="cc-page-myfinances-account-filter">
+    <form name="cc_page_myfinances_account_filter" class="cc-page-myfinances-account-filter" action="/" role="search">
```
nicolaasmatthijs: @christianv : If I remember correctly, the form should have an `id` attribute that contains the string `search` somewhere for the `Search` button to be shown on all iOS versions

nicolaasmatthijs: Apparently, you have to have "search" in the name or id attribute. I thought it was required on the form, but maybe it works on the input field as well.

christianv: @nicolaasmatthijs I used http://stackoverflow.com/questions/4864167/show-search-button-in-iphone-ipad-safari-keyboard/26253512#26253512 and tried it on the iOS simulator were everything worked correctly. (see screenshot in jira - https://jira.ets.berkeley.edu/jira/browse/CLC-4830

src/assets/templates/cars_activity.html
```diff
@@ -54,7 +63,7 @@ <h2 class="cc-left">Activity</h2>
               <label class="cc-hidden cc-page-myfinances-print-label cc-print-show cc-left">
                 <strong>Search: </strong>
               </label>
-              <input type="text" placeholder="Search..." data-ng-model="search.$">
+              <input type="text" placeholder="Search..." data-ng-model="search.$" aria-label="Search" name="search">
```
nicolaasmatthijs: It looks like entering text here will immediately narrow down the list. Currently, there isn't anything announcing that that's happening though. Ideally, we would add an ARIA “polite” alert that announces this.

nicolaasmatthijs:
Looks good. I left one more comment that we can deal with as part of a follow-up ticket.

src/assets/templates/cars_activity.html
christianv: @nicolaasmatthijs totally agree - follow up at #3353

PR #3353, opened 2015-02-09
CLC-4830 - Follow up - add live region

christianv:
https://jira.ets.berkeley.edu/jira/browse/CLC-4830


raydavis:
We should all try to live politely.

christianv:
@raydavis Thank you sir!

PR #3355, opened 2015-02-09
CLC-4833 Disable E-Grades export during Canvas transition to new design

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-4833

The existing code seems nicely set-up for a one-liner.

### DIFF
--- a/public/canvas/canvas-customization.js
+++ b/public/canvas/canvas-customization.js
@@ -229,6 +229,7 @@
   /**
    * Adds E-Grades Export to Canvas Gradebook feature
    */
+  /* jshint ignore:start */
   var addEGradeExportOption = function() {
     // obtain course context id
     if (window.ENV && window.ENV.GRADEBOOK_OPTIONS && window.ENV.GRADEBOOK_OPTIONS.context_id) {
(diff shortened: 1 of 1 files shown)


raydavis:
… except that JSHint won't let it be a one-liner.

public/canvas/canvas-customization.js
```diff
@@ -229,6 +229,7 @@
   /**
    * Adds E-Grades Export to Canvas Gradebook feature
    */
+  /* jshint unused::false */
```
christianv: @raydavis shouldn't this be one `:`?

raydavis:
Indeed it should, @christianv , and it would have been wiser if I'd run jshint locally before pushing the change.

raydavis:
BTW, @christianv and @redconfetti , how do you feel about going with commented-out code (rather than a simple delete) in this case? Generally I'm against it, given versioned source control. For some reason this feels like an edge case, but I could be very, very easily talked out of it.

christianv:
@raydavis is fine by me, as long as there is an easy way to find what is currently `unused`, e.g. by searching `ignore:start`.

PR #3356, opened 2015-02-10
OEC-539, use Rake.original_dir as base dir when src/dest not specified

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/OEC-539

OEC rake tasks have optional src and dest args (default is Rake.original_dir).  Arg values are expected to be relative or absolute directory paths. Valid usage would be:
> rake oec:courses dest=~/oec
> rake oec:courses dest=tmp/oec
> rake oec:courses dest=/var/tmp/oec
> rake oec:courses dest=../tmp/oec

Paths should be determined relative to user's working dir and *not* base of calcentral repo (as Dir.pwd would have it).

### DIFF
--- a/lib/tasks/oec.rake
+++ b/lib/tasks/oec.rake
@@ -1,12 +1,10 @@
 namespace :oec do
 
   hr = "\n" + '-------------------------------------------------------------' + "\n"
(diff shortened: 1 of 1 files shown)


paulkerschen:
Args!

PR #3357, opened 2015-02-10
CLC-4455 Organize cross-listed courses more reasonably

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-4455
subsumes https://jira.ets.berkeley.edu/jira/browse/CLC-3355

This PR collapses crosslisted courses for instructors in My Academics and My Classes. Because campus data indicates crosslisting by adding a code to primary sections, this change will only affect instructors assigned to a primary section (no love for GSIs this round).

On the UX side, the the list of section codes in the My Academics semester summary is replaced with a new  column containing summary information on the number and type of sections. When the screen gets small, it collapses to a two-column view that lists only the total number of sections under the course code.

Rather than jam multiple course codes into the breadcrumbs on the Class Info page, a new "Crosslisted As" heading signals the existence of multiple codes.

Most of the Class Info widgets ignore CCNs and display only one section out of a crosslisted group. The exception is Roster Photos, which groups students by CCN and separates out crosslistings that are implicitly merged elsewhere.

On the data-structure side, the code for merging crosslisted courses has been moved upstream from Canvas course provisioning to My Academics. The `teachingSemesters` portion of the My Academics feed, as well as the My Classes feed, now collapses crosslisted courses into single objects. To preserve listing-specific information such as department and course code, each course now includes a `listings` array which may contain one or more objects. There is still only one slug per course, and the client side now uses slugs instead of course codes when a unique identifier is needed.

The `semesters` portion of the My Academics feed, which is only for students and doesn't include crosslistings, is unchanged.

### DIFF
--- a/src/assets/templates/widgets/my_classes.html
+++ b/src/assets/templates/widgets/my_classes.html
@@ -19,16 +19,15 @@
               <span class="cc-ellipsis" data-ng-bind="class.name"></span>
             </a>
             <div data-ng-switch-default>
-              <div class="cc-ellipsis">
+              <div class="cc-ellipsis cc-widget-classes-title" data-ng-repeat="listing in class.listings">
                 <a data-ng-click="api.analytics.trackExternalLink('My Classes', class.emitter, class.site_url)"
                    data-ng-href="{{class.site_url}}">
-                  <strong data-ng-bind="class.course_code" data-ng-class="{'cc-italic': class.waitlistPosition}"></strong>
-                  <strong data-ng-bind="class.courseCodeSection" data-ng-class="{'cc-italic': class.waitlistPosition}"></strong>
+                  <strong data-ng-bind="listing.course_code" data-ng-class="{'cc-italic': class.waitlistPosition}"></strong>
+                  <strong data-ng-bind="listing.courseCodeSection" data-ng-class="{'cc-italic': class.waitlistPosition}"></strong>
                 </a>
-                <span class="cc-text-light" data-ng-show="class.waitlistPosition" data-ng-bind-template="- #{{ class.waitlistPosition }} on the wait list"></span>
               </div>
-              <div class="cc-ellipsis cc-text-light cc-widget-classes-subtitle" data-ng-class="{'cc-italic': class.waitlistPosition}" data-ng-show="!!class.name" data-ng-bind="class.name">
-              </div>
+              <span class="cc-text-light" data-ng-show="class.waitlistPosition" data-ng-bind-template="- #{{ class.waitlistPosition }} on the wait list"></span>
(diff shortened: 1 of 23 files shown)


src/assets/stylesheets/_academics.scss
```diff
@@ -297,6 +297,22 @@
       display: none;
     }
   }
+  @media #{$medium-up} {
```
christianv: @paulkerschen  can we use the `hide-for-small` & `show-for-medium-up` classes instead?

raydavis:
Thanks for this heroically heavy lifting, @paulkerschen! I'm not done reviewing, but I think I see one minor issue already: the public/dummy/json/academics.json sample feed (used for standalone front-end development and as a cheesy API contract) doesn't look like it's been updated to match the new data structure.

If you've got some sample feed handy, you might even add a cross-listed example to both "academics.json" and "classes.json".

src/assets/templates/widgets/my_classes.html
```diff
@@ -20,11 +20,13 @@ <h3 data-ng-bind="category.categoryTitle"></h3>
             </a>
             <div data-ng-switch-default>
               <div class="cc-ellipsis">
-                <a data-ng-click="api.analytics.trackExternalLink('My Classes', class.emitter, class.site_url)"
-                   data-ng-href="{{class.site_url}}">
-                  <strong data-ng-bind="class.course_code" data-ng-class="{'cc-italic': class.waitlistPosition}"></strong>
-                  <strong data-ng-bind="class.courseCodeSection" data-ng-class="{'cc-italic': class.waitlistPosition}"></strong>
-                </a>
+                <div class="cc-widget-classes-title" data-ng-repeat="listing in class.listings">
```
christianv: @paulkerschen did you check whether the `ellipsis` still works? We basically use it to add 3 dots `...` when there's not enough space. I'm not sure whether it works on `div`s as well.

src/assets/stylesheets/_academics.scss
paulkerschen: Thanks @christianv, I should have guessed there was something of the sort in the toolbox.

src/assets/templates/widgets/my_classes.html
paulkerschen: Well caught @christianv - it doesn't work on `div`s.

paulkerschen:
Thanks for reminding me about the dummy JSON, @raydavis. I've updated it, and uncovered a couple of bugs in the process. It turns out the code in myClassesFactory.js which matches courses to sites really does need unique course IDs; my attempted fix to have it work with slugs wasn't sufficient. So I've gone back to using IDs, and just made it to go one loop deeper to get all IDs from all listings.

Mary Salinger has been made instructor of a primary section because, campus data being what it is, she wouldn't get the benefit of this change otherwise.

PR #3359, opened 2015-02-10
OEC-551, avoid duplicate rows in INTEGBI.csv caused by biology_post_process

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/OEC-551
Avoid duplicate rows in INTEGBI.csv caused by biology_post_process. BIO.csv entries were being moved to INTEGBI.csv causing, in the case of certain cross-listings, duplicates. The fix is to identify all unique CCN/UID entries in INTEGBI.csv before moving more over from BIO.csv.

This PR includes simple fix for https://jira.ets.berkeley.edu/jira/browse/OEC-567 in oec.rake:

>      if courses_diff.was_difference_found
>        Rails.logger.warn "#{hr}Find summary in #{courses_diff.output_filename}#{hr}"
>      else
>        File.delete courses_diff.output_filename
>        Rails.logger.warn "#{hr}No diff found in #{dept_name} csv.#{hr}"
>      end

### DIFF
--- a/app/models/oec/biology_post_processor.rb
+++ b/app/models/oec/biology_post_processor.rb
@@ -9,13 +9,14 @@ module Oec
     end
 
     def post_process
-      path_to_biology_csv = path_to_csv @biology_dept_name
+      biology = Oec::Courses.new(@biology_dept_name, @src_dir)
+      path_to_biology_csv = biology.output_filename
       if File.exist? path_to_biology_csv
         files_to_archive = { @biology_dept_name => path_to_biology_csv }
         uids_in_target_files = {}
         sorted_dept_rows = {}
         @biology_relationship_matchers.keys.each do |target_dept_name|
-          path_to_target_csv = path_to_csv target_dept_name
+          path_to_target_csv = Oec::Courses.new(target_dept_name, @src_dir).output_filename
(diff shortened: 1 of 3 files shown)


app/models/oec/biology_post_processor.rb
```diff
@@ -1,47 +1,85 @@
...
+            unless index == 0
+              uids_in_target_files[target_dept_name] << "#{get_row_uid row}"
+              put_row_per_dept(sorted_dept_rows, target_dept_name, row)
+            end
+          end
+        end
+        header_row = nil
+        CSV.read(path_to_biology_csv).each_with_index do |row, index|
+          unless preexisting_uid?(get_row_uid(row), uids_in_target_files, files_to_archive.keys)
+            course_name = row[1]
```
paulkerschen: @johncrossman Completely minor, but if this assignment isn't doing anything in the context of header rows, it probably makes sense to move it down under the `else`.

johncrossman: Thanks @paulkerschen. I removed need for header_row by making Courses#headers available in post_process method. Conveniently, this removes need for *if index == 0* in CSV.read loop.

paulkerschen:
Makes sense to me!

PR #3360, opened 2015-02-10
CLC-4773 - Store "View As" user data in the back-end

edwlook:
https://jira.ets.berkeley.edu/jira/browse/CLC-4773

### DIFF
--- a/app/controllers/stored_users_controller.rb
+++ b/app/controllers/stored_users_controller.rb
@@ -1,0 +1,49 @@
+class StoredUsersController < ApplicationController
+  before_filter :authenticate
+  before_filter :numeric_uid?, except: [:get]
+  respond_to :json
+
+  def get
+    authorize(current_user, :can_view_as?)
+    users_found = User::StoredUsers.get(current_user.real_user_id)
+    render json: { users: users_found }.to_json
+  end
+
+  def store_saved_uid
+    response = User::StoredUsers.store_saved_uid(current_user.real_user_id, params['uid'])
+    render json: response.to_json
+  end
+
+  def store_recent_uid
+    response = User::StoredUsers.store_recent_uid(current_user.real_user_id, params['uid'])
+    render json: response.to_json
+  end
+
+  def delete_saved_uid
+    response = User::StoredUsers.delete_saved_uid(current_user.real_user_id, params['uid'])
+    render json: response.to_json
+  end
+
+  def delete_recent_uid
+    response = User::StoredUsers.delete_recent_uid(current_user.real_user_id, params['uid'])
+    render json: response.to_json
+  end
+
+  private
+
+  def error_response
+   result = { success: false, message: 'Please provide a numeric UID.' }
+    respond_with(result) do |format|
+      format.json { render json: result.to_json, status: :bad_request }
+    end
+  end
+
+  def numeric_uid?
+    begin
+      Integer(params[:uid], 10)
+    rescue ArgumentError
+      error_response
+    end
+  end
+
+end
--- a/app/models/user/search_users_by_uid.rb
+++ b/app/models/user/search_users_by_uid.rb
@@ -4,11 +4,23 @@ module User
 
     def initialize(options={})
       @id = options[:id]
+      @ids = options[:ids]
     end
 
     def search_users_by_uid
-      self.class.fetch_from_cache "#{@id}" do
-        CampusOracle::Queries.find_people_by_uid(@id)
+      search_batch([@id])
+    end
+
+    def search_users_by_uid_batch
+      search_batch(@ids)
+    end
+
+    private
+
+    def search_batch(uids)
+      cache_key = uids.sort.join('-')
+      self.class.fetch_from_cache "#{cache_key}" do
+        CampusOracle::Queries.get_basic_people_attributes(uids)
       end
     end
 
--- a/app/models/user/stored_users.rb
+++ b/app/models/user/stored_users.rb
@@ -1,0 +1,125 @@
+module User
+  class StoredUsers
+    include ActiveRecordHelper
+
+    def self.get(uid)
+      users = {
+        :saved => [],
+        :recent => []
+      }
+
+      uid_entries = get_stored_uid_entries(uid)
+
+      all_uids = []
+
+      uid_entries[:saved].each do |entry|
+        all_uids << entry[:stored_uid]
+      end
+
+      uid_entries[:recent].each do |entry|
+        all_uids << entry[:stored_uid]
+      end
+
+      return users unless all_uids.present?
+
+      users_found = User::SearchUsersByUid.new(:ids => all_uids).search_users_by_uid_batch
+      uid_hash = {}
+
+      users_found.each do |user|
+        uid_hash[user['ldap_uid']] = user
+      end
+
+      uid_entries[:saved].each do |entry|
+        user = uid_hash[entry[:stored_uid]]
+        if user.present?
+          users[:saved] << user
+        end
+      end
+
+      uid_entries[:recent].each do |entry|
+        user = uid_hash[entry[:stored_uid]]
+        if user.present?
+          users[:recent] << user
+        end
+      end
+
+      users
+    end
+
+    def self.store_saved_uid(uid, uid_to_store)
+      user = get_user(uid)
+      return error_response("Could not find user #{uid}.") unless user
+      store(user.saved_uids, uid, uid_to_store)
+    end
+
+    def self.store_recent_uid(uid, uid_to_store)
+      user = get_user(uid)
+      return error_response("Could not find user #{uid}.") unless user
+      store(user.recent_uids, uid, uid_to_store)
+    end
+
+    def self.delete_saved_uid(uid, uid_to_delete)
+      user = get_user(uid)
+      return error_response("Could not find user #{uid}.") unless user
+      delete(user.saved_uids, uid, uid_to_delete)
+    end
+
+    def self.delete_recent_uid(uid, uid_to_delete)
+      user = get_user(uid)
+      return error_response("Could not find user #{uid}.") unless user
+      delete(user.recent_uids, uid, uid_to_delete)
+    end
+
+    private
+
+    def self.get_stored_uid_entries(uid)
+      stored_entries = {
+        :saved => [],
+        :recent => []
+      }
+      user = get_user(uid)
+      if user
+        stored_entries[:saved] = user.saved_uids.order(:created_at).reverse_order
+        stored_entries[:recent] = user.recent_uids.order(:created_at).reverse_order
+      end
+      stored_entries
+    end
+
+    def self.store(model, uid, uid_to_store)
+      if (model.where(:stored_uid => uid_to_store.to_s).size == 0)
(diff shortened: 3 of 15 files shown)


christianv:
@raydavis @ctweney could you both have a look at this?

app/models/user/stored_users.rb
```diff
@@ -0,0 +1,107 @@
...
+    def self.get(uid)
+      users = {
+        :saved => [],
+        :recent => []
+      }
+      uid_entries = get_stored_uid_entries(uid)
+
+      uid_entries[:saved].each do |entry|
+        id = entry[:stored_uid]
+        user = User::SearchUsersByUid.new(:id => id.to_s).search_users_by_uid[0]
```
ctweney: There's a classic N+1 query problem here. You have 1 SQL query for each stored UID. You should instead aggregate the stored and recent UIDs into an array and query User::SearchUsersByUid just once.

db/schema.rb
```diff
@@ -11,7 +11,10 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20140916213448) do
+ActiveRecord::Schema.define(version: 2014021814020707) do
```
ctweney: Something bad happened to schema.rb when you ran db:migrate. The version number here is way too old. And there are extraneous changes -- all those lines creating indexes with "using: btree" on them shouldn't be there since they're not part of this change.

spec/controllers/stored_users_controller_spec.rb
```diff
@@ -0,0 +1,139 @@
+require "spec_helper"
```
ctweney: Please change the double quotes to single everywhere (except where string interpolation is happening).

app/controllers/stored_users_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+  respond_to :json
+
+  def get
+    authorize(current_user, :can_view_as?)
+    users_found = User::StoredUsers.get(current_user.real_user_id)
+    render json: { users: users_found }.to_json
+  end
+
+  def store_saved_uid
+    return error_response unless valid_params?(params)
```
paulkerschen: I would use a before_filter for all these `valid_params?` checks (and probably give it a more descriptive name, like `ensure_numeric_uids`.)

edwlook: @paulkerschen Thanks for the tip!

app/models/user/stored_users.rb
edwlook: @ctweney I will update User::SearchUsersByUid to support batch queries.

db/schema.rb
ctweney: The "using: btree" change appears harmless to me, actually.  Edward's patch just makes all the add_index statements consistent in having it. I think it's an artifact of a change in Rails. @raydavis @paulkerschen What do you think about this?

paulkerschen: @ctweney I think you're probably right, and Rails has just decided to make index type explicit in the schema. I wouldn't expect it to cause trouble.

edwlook: Also, Postgres uses a btree by default for creating indexes (http://www.postgresql.org/docs/9.3/static/sql-createindex.html).

raydavis: The btree change seems OK to me, too, but the version change still worries me -- not just the embedded date but also the difference in number of digits.

edwlook: @raydavis Chris helped me resolve this earlier today. Let me push up the new changes.

edwlook:
@raydavis @ctweney new commits pushed.

app/models/user/search_users_by_uid.rb
```diff
@@ -4,11 +4,38 @@ class SearchUsersByUid
...
+      @ids = options[:ids]
     end
 
     def search_users_by_uid
-      self.class.fetch_from_cache "#{@id}" do
-        CampusOracle::Queries.find_people_by_uid(@id)
+      search(@id)
+    end
+
+    def search_users_by_uid_batch
```
ctweney: This batch query mode is still N+1 queries.  You should revise CampusOracle::Queries.find_people_by_uid so that it can take an array of uids. The cache key used to cache the results will have to change, too.

edwlook: Ok, I misunderstood what you said earlier and just changed it so User:: SearchUsersByUid would only receive 1 call.

raydavis: See CampusOracle::Queries.get_basic_people_attributes : https://github.com/ets-berkeley-edu/calcentral/blob/master/app/models/campus_oracle/queries.rb#L25

raydavis: Oops, wrong context for my last comment -- sorry for the noise.

redconfetti: So this is trying to get a bunch of users by UID. Is this preferred because it either uses the cache, or populates the cache.... over using something like [CampusOracle::Queries.get_basic_people_attributes](https://github.com/ets-berkeley-edu/calcentral/blob/8e61814d/app/models/campus_oracle/queries.rb#L25)?

spec/controllers/stored_users_controller_spec.rb
```diff
@@ -0,0 +1,139 @@
...
+    it 'should return error_response on invalid uid' do
+      post :store_recent_uid, { format: 'json', uid: 'not_numeric' }
+      expect(response.status).to eq(400)
+      json_response = JSON.parse(response.body)
+      expect(json_response['success']).to eq false
+      expect(json_response['message']).to eq 'Please provide a numeric UID.'
+    end
+
+    it 'should return success_response on valid uid' do
+      User::StoredUsers.should_receive(:store_recent_uid).with('1234', '100').and_return(success_response)
```
redconfetti: The syntax of Rspec you're using will eventually be deprecated. You can write expressions such as this one instead like:
```ruby
expect(User::StoredUsers).to receive(:store_recent_uid).with('1234','100').and_return(success_response)
```

It's basically making it so that Rspec doesn't add a #should_receive method to User::StoredUsers during testing. They realized that it's bad form and conflicts with some peoples code.

spec/controllers/stored_users_controller_spec.rb
```diff
@@ -0,0 +1,139 @@
+require 'spec_helper'
+
+describe StoredUsersController do
+
+  before do
+    session[:user_id] = '1234'
+    User::Auth.stub(:where).and_return([User::Auth.new(uid: '1234', is_superuser: true, active: true)])
```
redconfetti: The newer Rspec syntax for a stub like this is similar to the mock I just commented about.

```ruby
allow(User::Auth).to receive(:where).and_return([User::Auth.new(uid: '1234', is_superuser: true, active: true)])
```

When you're working with mocking or stubbing methods on an instance, instead of a class, you can then use:

```ruby
allow_any_instance_of(Some::ClassName).to receive(:instance_method).with(1234).and_return(true)
expect_any_instance_of(Some::ClassName).to receive(:self_destruct).with(self_destruct_code).and_return(true)
```

app/models/user/search_users_by_uid.rb
raydavis: @redconfetti , you re-made my deleted goof. :) This search is on our PostgreSQL DB, not on campus Oracle.

raydavis: Oh bloody dang, it *is* on campus Oracle. I think I'll go home now.

raydavis: OK, I went back and tried to orient myself in the current code.

The currently installed User::SearchUsersByUid search_users_by_uid method is solely for use by a browser-facing API, through the SearchUsersController, with some (unnecessary in this case) allowance for duplicate results so as to match returns from the other user-entered searches. I don't think you want to build on that for code that's cycling through storage of already-successfully-fetched UIDs. So if I'm following the intention successfully this time, I think you would be better off building on get_basic_people_attributes.

@redconfetti , @ctweney , does that match your understanding of the goal here, or am I again missing something?

ctweney: I think @raydavis  is right -- using get_basic_people_attributes (or a derivative of it) is probably the best bet here.

ctweney:
@edwlook Heads up, it looks like pull #3367 will get merged before this one. That means there'll be a merge conflict on schema.rb and you'll have to rebase. You will have to change your db migrations in this pull so that they come AFTER 20150211123814_create_summer_sub_term.rb in sequence. Sorry for the inconvenience -- let me know if you need help resolving the conflict/ordering issue.

spec/controllers/stored_users_controller_spec.rb
edwlook: @redconfetti Thanks for the heads up!

edwlook:
@ctweney Pushed changes addressing your comments + squash + rebase. How does it look now?

ctweney:
Looks good to me, @edwlook , thanks for making the changes!
@raydavis  @christianv can you give it another once-over? And, should we wait until Tuesday given the crowded last day of sprint?

christianv:
@ctweney it looks like this is only putting the back-end in place to allow for saving it there. So the front-end will need to change after that. And merging on Tuesday sounds fine to me @raydavis @edwlook

raydavis:
Yes, please; let's wait till after Bug Bash before merging this to master, since it includes another DB migration and it can't easily be QA-ed as is.

app/models/user/search_users_by_uid.rb
```diff
@@ -4,11 +4,23 @@ class SearchUsersByUid
...
+      search_batch(@ids)
+    end
+
+    private
+
+    def search_batch(uids)
+      cache_key = uids.sort.join('-')
+      self.class.fetch_from_cache "#{cache_key}" do
+        CampusOracle::Queries.get_basic_people_attributes(uids)
```
raydavis: If I'm following your logic correctly, the final JSON delivered to the browser consists of everything returned by the Oracle query. Is that the same as what's needed by the UX? If not, it would be good for this class to filter out unused properties.

By the way, when we add new API feeds, we generally also try to add a sample feed to the "public/dummy/json" directory. That helps speed front-end development by letting us point the Angular factory to the static on-disk file, and also serves as primitive documentation of the API contract. It would be nice to get one of those here.

christianv: @edwlook when you do add a dummy file, make sure the data is obfuscated (FERPA) so there is no way to link it to any real people on Campus

raydavis: Thanks for the reminder, @christianv !

edwlook: @raydavis Yes, the JSON is the same as what's needed on the front-end.

edwlook:
Added dummy JSON file named `stored_users.json`.

raydavis:
Apologies for the delay on these. We have a couple of urgent fixes to get into the QA branch and I'd like to be able to check them on Dev before making big changes to Master. Shouldn't be much longer...

raydavis:
After a much-too-compact QA week, we can finally start merging more ambitious PRs into the master branch. Thank you for your patience.

PR #3364, opened 2015-02-11
CLC-4811 - Empty Academics "Profile" card for concurrent enrollments

christianv:
https://jira.ets.berkeley.edu/jira/browse/CLC-4811


johncrossman:
All good.

PR #3365, opened 2015-02-11
CLC-4781 - Implement 'Create a Project Site' form

redconfetti:
Also included is code for CLC-4782 - Implement Project site creation processing

https://jira.ets.berkeley.edu/jira/browse/CLC-4781
https://jira.ets.berkeley.edu/jira/browse/CLC-4782

### DIFF
--- a/app/models/canvas/course.rb
+++ b/app/models/canvas/course.rb
@@ -19,6 +19,23 @@ module Canvas
       end
     end
 
+    def create(account_id, course_name, course_code, term_id, sis_course_id)
+      request_params = {
+        'account_id' => account_id,
+        'course' => {
+          'name' => course_name,
+          'course_code' => course_code,
+          'term_id' => term_id,
+          'sis_course_id' => sis_course_id
+        }
+      }
+      request_options = {
+        :method => :post,
+        :body => request_params,
+      }
+      response = request_uncached("accounts/#{account_id}/courses", "_course_creation", request_options)
+    end
+
     private
 
     def request_course
--- a/app/models/canvas/course_enrollments.rb
+++ b/app/models/canvas/course_enrollments.rb
@@ -3,7 +3,6 @@ module Canvas
 
     ENROLLMENT_STATES = ['active', 'invited']
     ENROLLMENT_TYPES = ['StudentEnrollment', 'TeacherEnrollment', 'TaEnrollment', 'ObserverEnrollment', 'DesignerEnrollment']
-    PROJECT_ENROLLMENT_TYPES = ['Owner', 'Maintainer', 'Member']
 
     def initialize(options = {})
       super(options)
--- a/app/models/canvas/course_enrollments.rb
+++ b/app/models/canvas/course_enrollments.rb
@@ -1,0 +1,40 @@
+module Canvas
+  class CourseEnrollments < Proxy
+
+    ENROLLMENT_STATES = ['active', 'invited']
+    ENROLLMENT_TYPES = ['StudentEnrollment', 'TeacherEnrollment', 'TaEnrollment', 'ObserverEnrollment', 'DesignerEnrollment']
+    PROJECT_ENROLLMENT_TYPES = ['Owner', 'Maintainer', 'Member']
+
+    def initialize(options = {})
+      super(options)
+      raise ArgumentError, 'Canvas Course ID option required' unless options.has_key?(:canvas_course_id)
+      @canvas_course_id = options[:canvas_course_id]
+    end
+
+    # Interface to Enroll a User in a Canvas Course
+    # See https://canvas.instructure.com/doc/api/enrollments.html#method.enrollments_api.create
+    # Include 'role_id' option to specify custom roles (i.e. Waitlist Student, Owner, Maintainer, Member)
+    def enroll_user(canvas_user_id, enrollment_type, enrollment_state, notify = false, options = {})
+      raise ArgumentError, 'Notification flag must be a Boolean' unless notify == true || notify == false
+      sentence_options = {:last_word_connector => ', or ', :two_words_connector => ' or '}
+      raise ArgumentError, "Enrollment type argument '#{enrollment_type}', must be #{ENROLLMENT_TYPES.to_sentence(sentence_options)}" unless ENROLLMENT_TYPES.include?(enrollment_type.to_str)
+      raise ArgumentError, "Enrollment state argument '#{enrollment_state}', must be #{ENROLLMENT_STATES.to_sentence(sentence_options)}" unless ENROLLMENT_STATES.include?(enrollment_state.to_str)
+      request_params = {
+        'enrollment' => {
+          'user_id' => canvas_user_id.to_int,
+          'type' => enrollment_type.to_str,
+          'enrollment_state' => enrollment_state.to_str,
+          'notify' => !!notify,
+        }
+      }
+      request_params['enrollment'].merge!({'role_id' => options[:role_id]}) if options[:role_id]
+      request_options = {
+        :method => :post,
+        :body => request_params,
+      }
+      response = request_uncached("courses/#{@canvas_course_id}/enrollments", '_course_enroll_user', request_options)
+      JSON.parse(response.body)
+    end
+
+  end
+end
--- a/spec/models/canvas/course_enrollments_spec.rb
+++ b/spec/models/canvas/course_enrollments_spec.rb
@@ -1,0 +1,95 @@
+require "spec_helper"
+
+describe Canvas::CourseEnrollments do
+
+  let(:uid)               { rand(99999).to_s }
+  let(:canvas_user_id)    { rand(99999) }
+  let(:canvas_course_id)  { rand(99999) }
+  let(:add_enrollment_response_body)  { '{
+      "id":20959,
+      "root_account_id":90242,
+      "course_id":11611,
+      "course_section_id":1311,
+      "user_id":1234567,
+      "associated_user_id":null,
+      "enrollment_state":"active",
+      "start_at":null,
+      "end_at":null,
+      "type":"TeacherEnrollment",
+      "role":"Owner",
+      "role_id":12,
+      "limit_privileges_to_course_section":false,
+      "last_activity_at":null,
+      "sis_import_id":null,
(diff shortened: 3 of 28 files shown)


app/models/canvas/course.rb
```diff
@@ -19,6 +19,23 @@ def course(options = {})
       end
     end
 
+    def create(account_id, course_name, course_code, term_id, sis_course_id)
+      request_params = {
+        'account_id' => account_id,
```
christianv: @redconfetti what is the reason here for not using `:account_id => account_id `?

app/models/canvas/course_enrollments.rb
```diff
@@ -0,0 +1,40 @@
+module Canvas
+  class CourseEnrollments < Proxy
+
+    ENROLLMENT_STATES = ['active', 'invited']
+    ENROLLMENT_TYPES = ["StudentEnrollment", "TeacherEnrollment", "TaEnrollment", "ObserverEnrollment", "DesignerEnrollment"]
```
christianv: @redconfetti shouldn't these be single quotes?

spec/models/canvas/course_enrollments_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+
+  context "when enrolling user into canvas course" do
+    before do
+      add_response = double
+      add_response.stub(:body).and_return(add_enrollment_response_body)
+      subject.stub(:request_uncached).and_return(add_response)
+    end
+
+    it "raises exception if canvas_user_id is not an integer" do
+      expect { subject.enroll_user('not_an_integer', 'TaEnrollment', 'active', false) }.to raise_error(NoMethodError, "undefined method `to_int' for \"not_an_integer\":String")
```
christianv: @redconfetti really trying to understand when we're using `'` or `"`

src/assets/javascripts/angular/controllers/pages/canvasCreateProjectSiteController.js
```diff
@@ -4,7 +4,49 @@
...
    */
-  angular.module('calcentral.controllers').controller('CanvasCreateProjectSiteController', function(apiService) {
+  angular.module('calcentral.controllers').controller('CanvasCreateProjectSiteController', function(apiService, canvasProjectProvisionFactory, canvasSiteCreationFactory, canvasSiteCreationService, $location, $route, $scope, $window) {
     apiService.util.setTitle('Create a Project Site');
+
+    $scope.accessDeniedError = 'This feature is only available to faculty and staff.';
+    $scope.linkToSiteOverview = canvasSiteCreationService.linkToSiteOverview($route.current.isEmbedded);
+
+    $scope.createProjectSite = function() {
+      $scope.isLoaded = false;
```
christianv: @redconfetti can we remove this? We shouldn't need it if we're returning a correct structure from the back-end & are doing the `angular.extend`

src/assets/javascripts/angular/controllers/pages/canvasCreateProjectSiteController.js
```diff
@@ -4,7 +4,49 @@
...
+      canvasProjectProvisionFactory.createProjectSite($scope.projectSiteName)
+        .success(function(data) {
+          angular.extend($scope, data);
+          if ($scope.projectSiteUrl && $scope.projectSiteUrl.length > 0) {
```
christianv: @redconfetti can we remove the `$scope.projectSiteUrl.length > 0`? An empty string, null is also falsy in JavaScript.

src/assets/javascripts/angular/factories/canvasProjectProvisionFactory.js
```diff
@@ -0,0 +1,22 @@
+/* jshint camelcase: false */
```
christianv: @redconfetti can we remove this?

src/assets/javascripts/angular/controllers/pages/canvasCreateCourseSiteController.js
```diff
@@ -5,9 +5,12 @@
   /**
    * Canvas course provisioning LTI app controller
    */
-  angular.module('calcentral.controllers').controller('CanvasCreateCourseSiteController', function(apiService, canvasCourseProvisionFactory, canvasCourseProvisionService, $scope, $timeout) {
+  angular.module('calcentral.controllers').controller('CanvasCreateCourseSiteController', function(apiService, canvasCourseProvisionFactory, canvasSiteCreationService, $route, $scope, $timeout) {
     apiService.util.setTitle('Create a Course Site');
 
+    $scope.accessDeniedError = 'This feature is currently only available to instructors with course sections scheduled in the current or upcoming terms.';
```
christianv: @redconfetti why do we need this in the JavaScript? Can't we show / hide it in the HTML?

app/models/canvas/course.rb
redconfetti: @christianv The keys of this hash are going to be converted to strings anyway when the request is sent to the Canvas API. I guess I could convert them to Ruby symbols and it would still work.

@raydavis You have any opinion on what format of hash we should be using in this context?

spec/models/canvas/course_enrollments_spec.rb
redconfetti: @christianv I really think we need to discuss this further. It should be that we use single quotes when there is not an interpolated value being inserted into the string like "blah #{inserted_value} blah".

But I don't see us making this a priority in pull requests that much. I'd rather we either agree to be consistent with it, or not focus on this minor detail.

http://stackoverflow.com/questions/20870375/single-vs-double-quotes

In this case, I opted to use the double quotes because it meant I didn't have to escape the single quotes around 'to_int'. We're also talking about an rspec test here, not live code.

src/assets/javascripts/angular/controllers/pages/canvasCreateProjectSiteController.js
redconfetti: @christianv Yeah. It doesn't look like the spinner is showing after i call this method with or without the $scope.isLoaded = false. I'm removing it, but could use some help. Any ideas why it isn't showing? I have it applied to the [primary DIV in the template](https://github.com/redconfetti/calcentral/blob/CLC-4781-create-project-site-form/src/assets/templates/canvas_embedded/create_project_site.html#L1).

src/assets/javascripts/angular/factories/canvasProjectProvisionFactory.js
redconfetti: @christianv  Oh yeah, I actually made the keys camelcase so that I could remove this. Thanks for reminding me.

src/assets/javascripts/angular/controllers/pages/canvasCreateCourseSiteController.js
redconfetti: @christianv The HTML template is shared across three controllers, each have their own message for situations where access is denied.

spec/models/canvas/course_enrollments_spec.rb
christianv: @redconfetti totally agree with making it consistent. And indeed, afaik that was the agreement. So would it be possible to change it to single quotes everywhere we're not doing interpolation?

In the 'to_int' case it actually looks like you used a back quote in the beginning and a single quote at the end ``to_int'`

src/assets/javascripts/angular/controllers/pages/canvasCreateCourseSiteController.js
christianv: @redconfetti I see, thanks

app/models/canvas/course.rb
raydavis: @redconfetti , generally our code uses symbol keys for new hashes below the controller. (I personally prefer ```{account_id: account_id}``` syntax in such cases, but we never decided to make that a project default so the code varies more on that one.)

raydavis: Ah, I see, this hash is specifically to fill in an API request. Since the hash is being built for someone else's code and not ours, I don't have much of an opinion.

app/models/canvas/course_enrollments.rb
```diff
@@ -0,0 +1,40 @@
+module Canvas
+  class CourseEnrollments < Proxy
+
+    ENROLLMENT_STATES = ['active', 'invited']
+    ENROLLMENT_TYPES = ['StudentEnrollment', 'TeacherEnrollment', 'TaEnrollment', 'ObserverEnrollment', 'DesignerEnrollment']
```
raydavis: I'm confused by a couple of things here, @redconfetti …

1. Should this be called ```COURSE_ENROLLMENT_TYPES``` to match "PROJECT"?
2. Should it exist at all? For "real" course sites, I think we'll only ever create *section* enrollments through our code, right?
3. But then I don't see ```PROJECT_ENROLLMENT_TYPES``` used anywhere in the code.

raydavis:
Other than those couple of lines in CourseEnrollments, this looks good to me.

app/models/canvas/course_enrollments.rb
redconfetti: @raydavis I called it 'PROJECT_ENROLLMENT_TYPES' because these are custom types only present under the 'Project' sub-account. I found out that the specification of a [custom 'role' string is being deprecated](https://canvas.instructure.com/doc/api/enrollments.html#method.enrollments_api.create), and thus opted to make this class support a 'role_id' option that is optionally included with the request.

I've configured a Canvas Proxy value that stores the ID for the 'Owner' role configured under the 'Projects' sub-account, and I'm specifying that in the Canvas::ProjectProvision class to make sure that it's used.

I'm removing this constant because it's not used. One thing I think we should perhaps think about however is if our 'Find a Person to Add' tool will continue to add Waitlist Student's after this option is deprecated.

redconfetti: @raydavis Nevermind, I see that we don't support adding 'Waitlist Students' via the Find a Person to Add tool. The 'Waitlist Student' role is used during official enrollment processing via the SIS Import API, not this API where the text based custom role field has been deprecated.

raydavis:
Test failures for Canvas::ProjectProvision, @redconfetti

PR #3366, opened 2015-02-11
OEC-566, rake diff task: Allow for multiple file inputs

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/OEC-556

See 'rake oec:diff' section of [OEC manual](https://docs.google.com/a/berkeley.edu/document/d/1MAZN5TA39fpRD5zRrUOqafHI1FEKcyqVHnCufpARblw/edit#heading=h.hkx6y2ahi39s) to understand how the optional 'departments' arg can be used.

### DIFF
--- a/lib/tasks/oec.rake
+++ b/lib/tasks/oec.rake
@@ -39,24 +39,50 @@ namespace :oec do
 
   desc 'Spreadsheet from dept is compared with campus data'
   task :diff => :environment do
-    dept_name = ENV['dept_name']
-    if dept_name.blank?
-      Rails.logger.warn "#{hr}Usage: rake oec:diff dept_name=BIOLOGY [src=/path/to/files] [dest=/export/path/]#{hr}"
-    else
-      src_dir = get_path_arg 'src'
-      dest_dir = get_path_arg 'dest'
-      courses_diff = Oec::CoursesDiff.new(dept_name.upcase.gsub(/_/, ' '), src_dir, dest_dir)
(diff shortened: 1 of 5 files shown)


lib/tasks/oec.rake
```diff
@@ -39,25 +39,51 @@ namespace :oec do
...
   desc 'Spreadsheet from dept is compared with campus data'
   task :diff => :environment do
-    dept_name = ENV['dept_name']
-    if dept_name.blank?
-      Rails.logger.warn "#{hr}Usage: rake oec:diff dept_name=BIOLOGY [src=/path/to/files] [dest=/export/path/]#{hr}"
-    else
-      src_dir = get_path_arg 'src'
-      dest_dir = get_path_arg 'dest'
-      courses_diff = Oec::CoursesDiff.new(dept_name.upcase.gsub(/_/, ' '), src_dir, dest_dir)
+    src_dir = get_path_arg 'src'
```
redconfetti: I don't want to block you on this PR specifically, but something that is usually recommended is to keep your rake tasks and controllers skinny, and your models fat. It might be a good idea to move all this logic into a model, and then simply make the call to the primary method for this task, that way you can implement an rspec test if needed on that method, and also call the method from any context (web or cronjob).

johncrossman: Thanks @redconfetti, good tip. In fact, this logic will soon become shared (ie, new model) and rakefile will be skinnier.

PR #3370, opened 2015-02-12
CLC-4842 Course Provisioning: make first listing the default for course data

paulkerschen:
This should fix https://jira.ets.berkeley.edu/jira/browse/CLC-4842, but needs to be verified as I'm not set up to test locally.


redconfetti:
LGTM! Ready to merge after Travis goes green.

PR #3371, opened 2015-02-13
CLC-4849 Provide course codes when provisioning by CCN

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-4849


johncrossman:
Yes I see no problem. Waiting on Mr.T

PR #3372, opened 2015-02-13
CLC-4837 Standardize logging of remote server errors

paulkerschen:
This PR addresses https://jira.ets.berkeley.edu/jira/browse/CLC-4837 by making a couple of changes to consolidate our error logging.

First, Errors::ProxyError now takes more options and hands them over to ResponseWrapper to format the log output. This should make it easier to keep formatting consistent and avoid problems like CLC-4837, where a multi-line response body was pushing the important information out of view. 

Second, since almost all of the classes using HTTPRequester raise the same error on unexpected status codes, I’ve consolidated that error handling into HTTPRequester itself. Classes that need to raise specialized errors or handle particular status codes themselves can pass HTTPRequester an `on_error` option hash.

The first line of the log shows the actual error message along with URL and status code. More detailed information like UID, cache key and response body appears on the next line, before the stack trace.

### DIFF
--- a/app/models/advising/my_advising.rb
+++ b/app/models/advising/my_advising.rb
@@ -41,34 +41,24 @@ module Advising
       else
         url = "#{@settings.base_url}/student/#{student_id}"
         logger.info "Internal_get: Fake = #@fake; Making request to #{url} on behalf of user #{@uid}; cache expiration #{self.class.expires_in}"
-        begin
-          response = get_response(
-            url,
-            basic_auth: {username: @settings.username, password: @settings.password}
-          )
-        rescue Timeout::Error
-          logger.error("Timeout error: url = #{url}, uid = #{@uid}")
-          raise Errors::ProxyError.new
-        end
+        response = get_response(
+          url,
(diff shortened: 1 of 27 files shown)


raydavis:
This looks like a big improvement in a couple of different ways, @paulkerschen -- really nice work. But I'm afraid it touches just too many places to fit comfortably into the end of a sprint which has already touched so many places. So unless @pfarestveit overrides for some reason, I'm afraid I'll have to ask for a merge to wait until Tuesday. Annoying, I know, especially given the likelihood of merge conflicts.

paulkerschen:
@raydavis Given that there's nothing user-facing here, I quite agree it can chill till next week.

paulkerschen:
Rebased and fixed conflicts with #3384.

A separate commit adds `check_for_errors.sh` to the repository.

raydavis:
Apologies for the delay on these. We have a couple of urgent fixes to get into the QA branch and I'd like to be able to check them on Dev before making big changes to Master. Shouldn't be much longer...

PR #3377, opened 2015-02-13
CLC-4857 - Travis: use Container-Based Infrastructure and set Git depth

christianv:
https://jira.ets.berkeley.edu/jira/browse/CLC-4857


johncrossman:
LGTM

PR #3378, opened 2015-02-13
CLC-4847 Fix NODEONE so it will work in calcentral-sis-dev; bring calcen...

ctweney:
...tral init.d script up to date with what's actually in use

The "sleep" command is present in the copy of the "calcentral" start script that Ops actually uses (which lives in /calcentral-dev/init.d). 

https://jira.ets.berkeley.edu/jira/browse/CLC-4847


raydavis:
Let this merge be on my head!

PR #3383, opened 2015-02-17
[QA] CLC-4859 - Project sites: non-admin site owner cannot use "find a person to add" tool

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-4859

For Master PR see #3381


christianv:
+1 on the front-end

PR #3326, opened 2015-02-03
CLC-4663 Remove unnecessary parsing from Bearfacts API clients

paulkerschen:
Part two of the unexpectedly expansive https://jira.ets.berkeley.edu/jira/browse/CLC-4663.

PR #3384, opened 2015-02-17
CLC-4865 Fix verbose ERROR logging on student ID lookup failure

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-4865

Fixes regressions from #3310 and #3326 that treated student ID lookup failure as an error condition rather than a fact of everyday life.


raydavis:
Thanks for the quick fix, @paulkerschen ! Code looks right -- just waiting on Travis...

paulkerschen:
@raydavis - the fix was quick because I remembered deliberately causing the problem. :)

PR #3399, opened 2015-02-18
CLC-4673 - Converted span to button with link styling for improved accessibility

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-4673

### DIFF
--- a/public/canvas/canvas-customization.js
+++ b/public/canvas/canvas-customization.js
@@ -12,10 +12,10 @@
       var altMediaPanel = [
         '<div class="cc-alt-media-alert-container">',
         '  <div class="alert alert-info cc-alt-media-alert">',
-        '    <span class="element_toggler" aria-controls="cc-alt-media-alert-content" aria-expanded="false" aria-label="View Instructions" role="button">',
+        '    <button class="btn-link element_toggler cc-alt-media-alert-button" aria-controls="cc-alt-media-alert-content" aria-expanded="false" aria-label="Notice to Instructors for Making Course Materials Accessible">',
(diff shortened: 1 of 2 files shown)


public/canvas/canvas-customization.js
```diff
@@ -12,10 +12,10 @@
       var altMediaPanel = [
         '<div class="cc-alt-media-alert-container">',
         '  <div class="alert alert-info cc-alt-media-alert">',
-        '    <span class="element_toggler" aria-controls="cc-alt-media-alert-content" aria-expanded="false" aria-label="View Instructions" role="button">',
+        '    <button class="btn-link element_toggler cc-alt-media-alert-button" aria-controls="cc-alt-media-alert-content" aria-expanded="false" aria-label="Notice to Instructors for Making Course Materials Accessible">',
```
christianv: @redconfetti can we remove the `aria-expanded` attribute since we're not updating it?

christianv: @redconfetti do we ever update the `aria-expanded` attribute?

redconfetti: Yes. The 'element_toggler' class is tied to Javascript (jQuery) logic implemented by Canvas which toggles the attributes value.

christianv: @redconfetti thanks - for future purposes:

https://github.com/instructure/canvas-lms/blob/d78d65cce6ff30f495ddf319e3b4e23ffa3fb481/app/stylesheets/components/_element-toggler.scss

redconfetti: @christianv Thanks for looking that. I'm confused now though because I'm no longer using a SPAN, but instead a button, because the span wasn't being recognized by the screenreader I was testing with.

PR #3402, opened 2015-02-19
CLC-4883 Translate additional instruction formats

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-4883


johncrossman:
Looks good.

PR #3406, opened 2015-02-19
OEC-571, Diff task not respecting 'departments' arg

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/OEC-571


paulkerschen:
Looks good.

PR #3407, opened 2015-02-19
CLC-4883 Add remaining instruction formats from registrar's list

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-4883

Rounding out the list. Source, page 9: http://registrar.berkeley.edu/DisplayMedia.aspx?ID=Class_Sched_Users_Guide.pdf


johncrossman:
LGTM

PR #3410, opened 2015-02-23
CLC-4884 - Add-Delete: Add Preferred Course Ordering in Feed

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-4884

### DIFF
--- a/spec/models/canvas/course_provision_spec.rb
+++ b/spec/models/canvas/course_provision_spec.rb
@@ -5,7 +5,7 @@ describe Canvas::CourseProvision do
   let(:user_id) { rand(99999).to_s }
   let(:canvas_admin_id) { rand(99999).to_s }
   let(:canvas_course_id) { rand(999999).to_s }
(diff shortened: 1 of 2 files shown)


spec/models/canvas/course_provision_spec.rb
```diff
@@ -85,6 +85,7 @@
       before do
         allow_any_instance_of(Canvas::CourseSections).to receive(:official_section_identifiers).and_return(official_sections)
         allow_any_instance_of(Canvas::Course).to receive(:course).and_return(course_hash)
+        allow(subject).to receive(:group_by_used!).and_return(true)
```
paulkerschen: I don't think the test is using this return value, but wouldn't you expect it to return the feed argument itself, as in the block on line 104?

redconfetti: Oh yeah, thanks for catching that. Doh!

PR #3411, opened 2015-02-24
CLC-4767 Script to configure all currently defined LTI tools on bCourses

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-4767

This gives us an easy way to add new LTI tools, to reset keys and secrets, and to configure apps on a Canvas server from the CalCentral environment associated with it (i.e., the script can configure bCourses-beta apps that CalCentral-production doesn't yet know about).

### DIFF
--- a/app/models/canvas/external_app_configurations.rb
+++ b/app/models/canvas/external_app_configurations.rb
@@ -1,0 +1,80 @@
+module Canvas
+  module ExternalAppConfigurations
+    extend self
+    include ClassLogger
+
+    def lti_app_definitions
+      {
+        'site_creation' => {
+          xml_name: 'lti_site_creation',
+          app_name: 'Create a Site',
+          account: main_account_id
+        },
+        'rosters' => {
+          xml_name: 'lti_roster_photos',
+          app_name: 'Roster Photos',
+          account: official_courses_account_id
+        },
+        'user_provision' => {
+          xml_name: 'lti_user_provision',
+          app_name: 'User Provisioning',
+          account: main_account_id
+        },
+        'course_add_user' => {
+          xml_name: 'lti_course_add_user',
+          app_name: 'Find a Person to Add',
+          account: main_account_id
+        },
+        'course_mediacasts' => {
+          xml_name: 'lti_course_mediacasts',
+          app_name: 'Webcasts',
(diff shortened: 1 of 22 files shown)


app/models/canvas/external_app_configurations.rb
```diff
@@ -0,0 +1,80 @@
...
+        },
+        'course_mediacasts' => {
+          xml_name: 'lti_course_mediacasts',
+          app_name: 'Webcasts',
+          account: main_account_id
+        },
+        'course_manage_official_sections' => {
+          xml_name: 'lti_course_manage_official_sections',
+          app_name: 'Official Sections',
+          account: nil
```
raydavis: To enable the "Official Sections" tool, set "account" to "official_courses_account_id".

lib/tasks/canvas.rake
```diff
@@ -58,11 +58,18 @@ namespace :canvas do
       canvas_hosts_to_calcentrals_string.split(/=|,/).each_slice(2) {|pair|
         canvas_hosts_to_calcentrals.push({host: pair[0], calcentral: pair[1]})
       }
-      Canvas::ReconfigureExternalApps.reconfigure_external_apps(reachable_xml_host, canvas_hosts_to_calcentrals)
+      Canvas::ReconfigureExternalApps.new.reset_external_app_hosts_by_url(reachable_xml_host, canvas_hosts_to_calcentrals)
       Rails.logger.info("Reconfigured external apps from #{reachable_xml_host} for #{canvas_hosts_to_calcentrals_string}")
     end
   end
 
+  desc 'Configure all default Canvas external apps provided by the current server'
```
paulkerschen: Needs a 'do'

raydavis: Actually I think it needed to lose an 'end' :)

paulkerschen: Right, so it did!

app/models/canvas/external_app_configurations.rb
redconfetti: Sweet. I like how we have the configuration of our apps so clearly defined in our code now, including which accounts they are configured under. :+1: :godmode:

PR #3417, opened 2015-02-25
CLC-4906 bCourses should use student ID for concurrent enrollment

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-4906

Also cleared out some unused and repeated code.


paulkerschen:
LGTM. I'm glad to see CampusOracle::Queries lose the method that's not actually a query.

PR #3418, opened 2015-02-25
CLC-4734 Show grades from enrollment data while transcripts unavailable

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-4734

The grading_in_progress period is slightly extended to 4 weeks from semester's end. Current and grading_in_progress terms attempt to fetch grades from enrollment data only. Past terms for which grading is no longer in progress fetch grades from transcript data as before.


raydavis:
The PR looks great, but @paulkerschen and I agree that maybe we'll wait to merge until we've got confirmation that the new column will land in the production view. @paulkerschen , mind closing this temporarily?

paulkerschen:
OK, closing for now.

PR #3420, opened 2015-02-26
CLC-4863 Do not show REMOVED or LAPSED transcript entries

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-4863

I went with filtering out the entries in the model rather than adding more lines to the database query; but I could easily be persuaded to do it the other way.


raydavis:
Thanks, @paulkerschen . I agree with your instinct so far as filtering in the model: we might someday have to get even fancier about interpreting that column.

PR #3425, opened 2015-03-02
CLC-4941 Properly log Canvas proxy errors

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-4941

### DIFF
--- a/lib/errors/proxy_error.rb
+++ b/lib/errors/proxy_error.rb
@@ -8,7 +8,11 @@ module Errors
       @uid = opts[:uid]
 
       if (response = opts[:response])
-        @status = response.code
+        @status = if response.respond_to? :code
+                    response.code
+                  elsif response.respond_to? :status
+                    response.status
+                  end
         @body = response.body
       end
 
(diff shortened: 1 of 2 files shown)


lib/errors/proxy_error.rb
```diff
@@ -8,7 +8,11 @@ def initialize(log_message = '', opts={})
       @uid = opts[:uid]
 
       if (response = opts[:response])
-        @status = response.code
+        @status = if response.respond_to? :code
+                    response.code
+                  elsif response.respond_to? :status
+                    response.status
+                  end
         @body = response.body
```
johncrossman: @paulkerschen For sake of my education, does status need a default when those two conditions are not met?

paulkerschen:
@johncrossman In the edge case where this method somehow gets sent a response without a status, leaving `@status` as an implicit nil should be fine. The code in ResponseWrapper.handle_exception will just skip the property when it comes time to log the error.

The weird conditional is needed because it turns out ProxyError can be initialized with response objects from either HTTParty or Faraday, and these libraries use `status` and `code`, respectively, for status codes. (Also educational! normal Rails idiom would be to compress this with `try`, but HTTParty's Response object inherits from `BasicObject`, not `Object`, so `try` doesn't work.)

johncrossman:
Thanks @paulkerschen  Looks good and seems to address issue described in da Jira.

PR #3426, opened 2015-03-02
CLC-4734 Show grades from enrollment data while transcripts unavailable

paulkerschen:
The return of https://jira.ets.berkeley.edu/jira/browse/CLC-4734, with database now in readiness.


raydavis:
Still looks good the second time around -- just waiting for Travis to nod...

PR #3434, opened 2015-03-03
CLC-4933 Do not return expired affiliations from CourseAddUser search

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-4933


johncrossman:
Looks good.

PR #3436, opened 2015-03-04
OEC-580, info useful to Ops is put to final output of script

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/OEC-580

This is a follow up to resolved sub-tasks of OEC-580. When rake oec:diff is done the summary contain, in addition to errors, where the diff results can be found. For example:
```ruby
Summary
  STAT
    Diff report
      Created /Users/john/Projects/github/johncrossman/calcentral/tmp/oec/diff_STAT_courses.csv
    ... 
```


paulkerschen:
Looks good from here.

PR #3440, opened 2015-03-04
CLC-4895 Add regression tests for Add-Delete Sections server-side

raydavis:
Follow-up on https://jira.ets.berkeley.edu/jira/browse/CLC-4895

I exposed a few more instance variables in ProvideCourseSite to make it a _little_ more test-friendly...


johncrossman:
Looks good.

PR #3442, opened 2015-03-05
CLC-4934 Account for daylight savings transitions in task due dates

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-4934


christianv:
@paulkerschen nice catch!

PR #3445, opened 2015-03-05
CLC-4887 Give the big bCourses refresh scripts more heap to fall back on

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-4887

The default MaxHeapSize is about 310M, so a boost for these two consistently-growing scripts seems reasonable.


johncrossman:
Looks good. I think CMSPermGenSweepingEnabled is deprecated but Oracle docs are not clear on that question.

PR #3446, opened 2015-03-05
CLC-4448 Remove duplicate site listings under repeated courses in My Classes

paulkerschen:
Fixes duplicate repeater error and hidden bCourses sites in https://jira.ets.berkeley.edu/jira/browse/CLC-4448.

The back-end change ensures that bCourses site data in the MyClasses feed does not include duplicate links to campus courses.

The front-end change ensures that even if a campus course appears multiple times in the feed, all entries will be correctly associated with course sites. This matching logic is getting a little unwieldy, and at some point it might be a good idea to move it to the back end.


johncrossman:
Works for me.

PR #3447, opened 2015-03-05
CLC-4899 - Add-Delete: Implement minimized maintenance notice official course tools

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-4899

### DIFF
--- a/src/assets/stylesheets/_canvas_shared.scss
+++ b/src/assets/stylesheets/_canvas_shared.scss
@@ -4,39 +4,6 @@
   padding: 0;
 }
 
-// Maintenance Notice
-.cc-maintenance-notice {
-  color: $bc-color-off-black;
-  font-size: 13px;
-  margin: 8px 6px;
-  min-height: 20px;
-}
-
-.cc-notice-text-container {
-  line-height: 18px;
-  margin-left: 45px;
-}
-
-.cc-notice-text-header {
-  color: #333;
-  font-size: 16px;
-  font-weight: 500;
-  line-height: 20px;
-  margin: 4px 0;
-}
-
-.cc-notice-icon {
-  color: $bc-color-gold;
-  float: left;
-  font-size: 36px;
-  line-height: 44px;
-}
-
-// Error / Notices
-.cc-page-error-notice {
-  margin: 10px 0;
-}
-
 // Site Provisioning Headers
 .cc-page-site-provisioning-header {
   color: $bc-color-off-black;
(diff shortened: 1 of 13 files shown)


redconfetti:
Notice: There may be one last change to the maintenance notice. Do not merge yet.

src/assets/stylesheets/_canvas_shared.scss
```diff
@@ -146,6 +146,76 @@
   }
 }
 
+// Notices
+
+.cc-notice-icon {
```
christianv: @redconfetti only a fyi - we need to start to be careful with this file since we can have overlaps with other styles.

Would it be possible to open a ticket to have a wrapper around all of these styles? e.g. `.bc-main`.

```
.bc-main {
  .cc-notice-box {
  }
}
```

src/assets/templates/canvas_embedded/_shared/maintenance_notice.html
```diff
@@ -0,0 +1,16 @@
...
+
+  <div data-ng-click="maintenanceCollapsed = !maintenanceCollapsed" data-ng-attr-aria-controls="maintenance-details">
+    <i class="fa cc-left cc-template-canvas-maintenance-notice-collapse-icon"
+      data-ng-class="{'fa-caret-right': maintenanceCollapsed, 'fa-caret-down': !maintenanceCollapsed}"></i>
+
+    <i class="fa fa-exclamation-triangle cc-icon-yellow"></i>
+    <span class="cc-template-canvas-maintenance-notice-text">You may experience delays before your <span data-ng-bind="courseActionVerb"></span> during our daily scheduled maintenance from 8-9 AM.</span>
+  </div>
+
+  <div id="maintenance-details" data-ng-show="!maintenanceCollapsed" data-ng-attr-aria-expanded="{{maintenanceCollapsed}}" role="region" class="cc-template-canvas-maintenance-notice-details cc-template-canvas-maintenance-notice-text">
```
christianv: @redconfetti does `ng-if` work here?

src/assets/templates/canvas_embedded/create_course_site.html
```diff
@@ -96,10 +89,10 @@ <h2 class="cc-page-site-provisioning-header cc-page-site-provisioning-header2">C
                 <div class="cc-page-help-notice cc-page-create-course-site-help-notice">
                   <i class="fa fa-question-circle cc-left cc-page-help-notice-icon"></i>
                   <div class="cc-page-help-notice-left-margin">
-                    <button class="cc-button-link cc-page-help-notice-toggle-button" data-ng-click="toggle.displayHelp = !toggle.displayHelp">
+                    <button class="cc-button-link cc-page-help-notice-toggle-button" data-ng-click="toggle.displayHelp = !toggle.displayHelp" aria-controls="section-selection-help">
                       Need help deciding which rosters to select?
                     </button>
-                    <div class="cc-page-help-notice-content" data-ng-if="toggle.displayHelp">
+                    <div id="section-selection-help" class="cc-page-help-notice-content" data-ng-show="toggle.displayHelp" data-ng-attr-aria-expanded="{{toggle.displayHelp}}" role="region">
```
christianv: @redconfetti interesting, what broke when you used `ng-if`?

src/assets/stylesheets/_canvas_shared.scss
redconfetti: @christianv It's actually redundant. I'm renaming it to .cc-canvas-notice-icon, and also removing the styles that are no longer relevant now that the original maintenance notice is no longer present.

src/assets/templates/canvas_embedded/_shared/maintenance_notice.html
redconfetti: Will ng-if still work with the aria-expanded tag in a screen reader? Something makes me suspect it won't.

src/assets/templates/canvas_embedded/create_course_site.html
redconfetti: @christianv I'm suspecting that the screen-reader won't be able to detect that an element of the DOM has been expanded/displayed if i'm using an ng-if.

christianv: @redconfetti that's why we have the `aria-expanded` attribute, so using an `ng-if` should be fine. I agree that we should test this with screenreaders though.

PR #3451, opened 2015-03-07
Adviser Picture now on Academics Page

NolanChan:
CLC-4641 - Integrate advisor picture into the L&S Advising card

https://jira.ets.berkeley.edu/jira/browse/CLC-4641

Advisers that don't have photos only have their names' displayed

### DIFF
--- a/src/assets/templates/academics_ls_advising.html
+++ b/src/assets/templates/academics_ls_advising.html
@@ -10,6 +10,9 @@
       </p>
       <div data-ng-if="caseloadAdviser">
         <h3 class="cc-widget-lsadvising-heading-first">College Advisor</h3>
+        <div data-ng-if="caseloadAdviser.photoUrl" class="cc-widget-profile-picture">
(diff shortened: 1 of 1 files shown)


src/assets/templates/academics_ls_advising.html
```diff
@@ -10,6 +10,9 @@ <h2 class="cc-left">L&amp;S Advising</h2>
       </p>
       <div data-ng-if="caseloadAdviser">
         <h3 class="cc-widget-lsadvising-heading-first">College Advisor</h3>
+        <div class="cc-widget-profile-picture">
```
christianv: @NolanChan can we wrap this in an `ng-if` so it's not loaded in the DOM when there is no picture?

NolanChan: Oh yeah, sorry about that. I was anticipating adding stuff for future and past appointments but that'll be for later

PR #3452, opened 2015-03-07
CLC-4950 Standardize CCN formatting in bCourses to fix Official Sections

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-4950

Some back-end code was not zero-padding, causing mismatches in front-end.

This must be fixed before QA testing on Monday, and so if the new branch has been cut, I'll prepare a QA PR.


paulkerschen:
Yes indeed.

PR #3454, opened 2015-03-09
CLC-4899 - Add-Delete: Implement minimized maintenance notice official course tools

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-4899

This one should do the job without breaking the 'Create a Course Site' tool.


christianv:
Lgtm

PR #3455, opened 2015-03-10
CLC-4819 - Analytics: Develop an analytics prototype for bCourses

nicolaasmatthijs:
Added PoC tool usage tracking using MixPanel.

https://jira.ets.berkeley.edu/jira/browse/CLC-4819


raydavis:
Many JSHint complaints from Travis, @nicolaasmatthijs.

public/canvas/canvas-customization.js
```diff
@@ -322,6 +322,68 @@
...
 
   /**
+   * Track the course site and course tool usage using MixPanel
+   */
+  (function() {
+    // Only add mixpanel tracking to the production, beta and test instances
+    // TODO: Once we have decided that we want to continue working with MixPanel,
+    // these tokens should be retrieved from a dedicated REST endpoint
+    var mixpanelToken = null;
+    if (window.location.hostname === 'bcourses.berkeley.edu') {
```
christianv: @nicolaasmatthijs please add a REST endpoint for these. I did read the comments but since this PR is for master it will go in production.

It's totally fine with me if we test a different branch of CalCentral for a while on QA that contains this code but we shouldn't merge it into master

public/canvas/canvas-customization.js
```diff
@@ -322,6 +322,68 @@
...
+    }
+
+    if (mixpanelToken) {
+      // Add the MixPanel tracking script
+      /* jshint ignore:start */
+     (function(f,b){if(!b.__SV){var a,e,i,g;window.mixpanel=b;b._i=[];b.init=function(a,e,d){function f(b,h){var a=h.split(".");2==a.length&&(b=b[a[0]],h=a[1]);b[h]=function(){b.push([h].concat(Array.prototype.slice.call(arguments,0)))}}var c=b;"undefined"!==typeof d?c=b[d]=[]:d="mixpanel";c.people=c.people||[];c.toString=function(b){var a="mixpanel";"mixpanel"!==d&&(a+="."+d);b||(a+=" (stub)");return a};c.people.toString=function(){return c.toString(1)+".people (stub)"};i="disable track track_pageview track_links track_forms register register_once alias unregister identify name_tag set_config people.set people.set_once people.increment people.append people.track_charge people.clear_charges people.delete_user".split(" ");for(g=0;g<i.length;g++)f(c,i[g]);b._i.push([a,e,d])};b.__SV=1.2;a=f.createElement("script");a.type="text/javascript";a.async=!0;a.src="//cdn.mxpnl.com/libs/mixpanel-2-latest.min.js";e=f.getElementsByTagName("script")[0];e.parentNode.insertBefore(a,e)}})(document,window.mixpanel||[]);mixpanel.init(mixpanelToken);
```
christianv: @nicolaasmatthijs instead of adding the minified code in here, can we load it asynchronously?

christianv:
As mentioned in one of the comments, I would be perfectly fine with running this branch for a while on DEV / QA for a while on a specific (non-master) branch as is. But we should find a solution for the following areas before merging it into master:

1. Tokens should not be embedded in the code, they should come from a REST endpoint
1. Hide it behind a feature flag
1. Find a better way to include the minified mixpanel code

public/canvas/canvas-customization.js
nicolaasmatthijs: @christianv : I had cleared with Rich that it is more important to get this out as a PoC than to make the investment in creating the dedicated REST endpoint for this.

nicolaasmatthijs: PS: Do you have an example of such a REST endpoint that allows for an incoming cross-domain request? I'll look at what I can do.

christianv: @nicolaasmatthijs as mentioned in the other comments, it's totally fine to put this in a separate branch but I don't feel comfortable merging this into master. I know we do have high standards but that's a decision we made as a team.

public/canvas/canvas-customization.js
nicolaasmatthijs: How ?

christianv: @nicolaasmatthijs for instance - http://api.jquery.com/jquery.getscript/. I'm more than willing to spend some time together in person to figure this out.

public/canvas/canvas-customization.js
christianv: @nicolaasmatthijs as for the REST endpoint example, I think @raydavis & @redconfetti might be able to help out here.

public/canvas/canvas-customization.js
nicolaasmatthijs: So just create a separate js file that contains the same mixpanel embed code? Wouldn't that be a bit redundant?

raydavis:
Thanks for the help, @christianv !

Regarding the particular question of "which branch": Given that this request comes from the bCourses project manager and other ETS leadership, and given the difficulties of testing bCourses customizations from anyplace _other_ than the shared servers, I'm comfortable with it landing on the current "master" branch and on Dev.

But I'd like to make sure that it doesn't get folded automatically into the next sprint's QA branch in two and a half weeks (I can add a comment to that effect at the JIRA ticket, and a calendar reminder). And I agree that a feature flag (or equivalent) is a good idea.

christianv:
@raydavis I'm fine with it landing in master as well when the 3 above mentioned areas are resolved.

public/canvas/canvas-customization.js
christianv: @nicolaasmatthijs indeed, but that's fine. That way we know that it's 3th party code, which makes it easier to exclude it from tools like our linting services and makes it easier to recognize / debug.

We're already taking a similar approach for our google analytics code:
https://github.com/ets-berkeley-edu/calcentral/blob/master/src/assets/javascripts/lib/analytics.js
In that case though it's bundled with the main CalCentral code JavaScript in order to avoid the extra async request.

raydavis:
The current PR was based on a misunderstanding. We've agreed to close it and use it as a starting point for a normally scheduled-and-released task in the next sprint. Thank you for the work thus far, @nicolaasmatthijs & @christianv !

PR #3457, opened 2015-03-10
CLC-4958 - Add-Delete: Implement current design for Sections Available to Add

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-4958

See QA PR in #3458


raydavis:
Thanks, @redconfetti -- let's see how this looks on Dev before merging the QA PR.

paulkerschen:
A couple of notes, @raydavis and @redconfetti: this will still show the colon when course title is blank, which may be undesirable; and it doesn't explicitly handle the crosslisting case. I think it will just default to showing whichever course code was listed first in the feed.

Also, is it being told to order the course listings in any particular way? I know it's alphabetical by title in Course Provisioning.

raydavis:
Thanks, @paulkerschen -- my guess is that these sorts of considerations could keep the QA PR on ice for now. Maybe we can move them into a comment at the JIRA ticket?

PR #3459, opened 2015-03-10
CLC-4960 - Advising: advisor photo could use a better layout

christianv:
https://jira.ets.berkeley.edu/jira/browse/CLC-4960


raydavis:
Works for me -- thanks, @christianv ! Just going to let Travis put its two cents in...

PR #3468, opened 2015-03-12
CLC-4969: misc Selenium test bug fixes

pfarestveit:
https://jira.ets.berkeley.edu/jira/browse/CLC-4969

See JIRA ticket for detail on the issues addressed in this PR.  In summary they are:
* Reducing false test failures caused either by recent changes in the expected application behavior or by bugs in the Selenium code itself.
* Avoiding intermittent errors instantiating the browser and passing it to the rest of a given test.  This involves separating the browser instantiation into its own "before" hook, apart from other test setup code.  I'm not certain this will fix the errors, but it's worth a try.  The bulk of the noise in this PR comes from rearranging these hooks and consequently shifting the rest of the (unaltered) test code.


johncrossman:
Nice work @pfarestveit. Looks good. Waiting on Travis.

PR #3475, opened 2015-03-14
GPA has at least 2 Significant Digits

NolanChan:
CLC-4868 - Include at least two significant digits in GPA

https://jira.ets.berkeley.edu/jira/browse/CLC-4868

There was multiple ways I wanted to approach the problem (filters in AngularJS, conversion to significant digits in js file) but the simplest way seemed to be to pass in a presentable value in Ruby. Feedback definitely needed for smarter ways

### DIFF
--- a/app/models/my_academics/gpa_units.rb
+++ b/app/models/my_academics/gpa_units.rb
@@ -7,7 +7,7 @@ module MyAcademics
       return data if student_info.nil?
 
       data[:gpaUnits] = {
-        cumulativeGpa: student_info['cum_gpa'].nil? ? nil: student_info['cum_gpa'].to_f,
+        cumulativeGpa: student_info['cum_gpa'].nil? ? nil: student_info['cum_gpa'].to_s,
         totalUnits: student_info['tot_units'].nil? ? nil : student_info['tot_units'].to_f,
         totalUnitsAttempted: student_info['lgr_tot_attempt_unit'].nil? ? nil : student_info['lgr_tot_attempt_unit'].to_f
       }
--- a/src/assets/templates/academics_student_profile.html
+++ b/src/assets/templates/academics_student_profile.html
@@ -64,7 +64,7 @@
       </div>
 
       <div class="cc-widget-profile-content-gpa" data-ng-if="gpaUnits.cumulativeGpa !== null && gpaUnits.cumulativeGpa > 0 && !studentInfo.isLawStudent">
-        <span data-ng-bind="gpaUnits.cumulativeGpa"></span>
+        <span data-ng-bind="gpaUnits.cumulativeGpaFloat"></span>
         GPA
       </div>
 
--- a/src/assets/javascripts/angular/controllers/pages/academicsController.js
+++ b/src/assets/javascripts/angular/controllers/pages/academicsController.js
@@ -298,6 +298,8 @@
           $scope.widgetSemesterName = $scope.selectedTeachingSemester.name;
         }
       }
+      $scope.gpaUnits.cumulativeGpaFloat = $scope.gpaUnits.cumulativeGpa; // cumulativeGpa is passed as a string to maintain two significant digits
+      $scope.gpaUnits.cumulativeGpa =  parseFloat($scope.gpaUnits.cumulativeGpa); // converted to Float to be processed regularly
     };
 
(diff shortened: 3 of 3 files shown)


src/assets/templates/academics_student_profile.html
```diff
@@ -64,7 +64,7 @@
       </div>
 
       <div class="cc-widget-profile-content-gpa" data-ng-if="gpaUnits.cumulativeGpa !== null && gpaUnits.cumulativeGpa > 0 && !studentInfo.isLawStudent">
-        <span data-ng-bind="gpaUnits.cumulativeGpa"></span>
+        <span data-ng-bind="gpaUnits.printableGpa"></span>
```
christianv: @NolanChan does it also work if you just use the number filter? https://docs.angularjs.org/api/ng/filter/number

gpaUnits.cumulativeGpa | number: 2

NolanChan: `gpaUnits.cumulativeGpa | number: 2`

would make 
4 => 4.00
3.3422 => 3.34
3.1 => 3.10

Is this the desired behavior? I just took it that we wanted 4 => 4.0 and not restrict GPA's with more trailing numbers

christianv: @paulkerschen or @raydavis could you chime in?

raydavis: Since @paulkerschen and @NolanChan have experience as UCB students, I bow to their instincts on this.

paulkerschen: What I see coming out of campus data for GPA is anywhere between one and three decimal places. I like @NolanChan 's approach of doing the string conversion on the back-end because it preserves whatever precision was in the database, and I think is in line with the philosophy of not hiding data.

That said, it does seem a little messy to be returning string and numeric values both in the JSON. It looks like the only place we're currently using the numeric value is in the GPA calculator. @christianv, would you be all right with having the calculator parse a string GPA value back into a numeric value, or would you prefer to keep a numeric value in the feed?

app/models/my_academics/gpa_units.rb
```diff
@@ -9,7 +9,8 @@ def merge(data)
       data[:gpaUnits] = {
         cumulativeGpa: student_info['cum_gpa'].nil? ? nil: student_info['cum_gpa'].to_f,
         totalUnits: student_info['tot_units'].nil? ? nil : student_info['tot_units'].to_f,
-        totalUnitsAttempted: student_info['lgr_tot_attempt_unit'].nil? ? nil : student_info['lgr_tot_attempt_unit'].to_f
+        totalUnitsAttempted: student_info['lgr_tot_attempt_unit'].nil? ? nil : student_info['lgr_tot_attempt_unit'].to_f,
+        printableGpa: student_info['cum_gpa'].nil? ? nil: student_info['cum_gpa'].to_f.to_s
```
paulkerschen: The intermediate .to_f shouldn't be necessary; just `student_info['cum_gpa'].to_s` ought to work.

src/assets/templates/academics_student_profile.html
christianv: @paulkerschen You're right, it would be better if we don't have a string & numeric value for the same property in the feed. @NolanChan can you see whether we can parse it in the front-end and only send the string over?

NolanChan: sure, I'll get on that

raydavis: Re-pinging @christianv in case he missed @NolanChan 's last question.

christianv: @NolanChan indeed, sorry for the delay, but I would suggest keeping the name `cumulativeGpa` and make it a `string`. Then in the `.js` file convert it with `parseFloat` to a float and call it something like `cumulativeGpaFloat`. We use very similar logic in the financials JavaScript. Does that sounds like a good approach to you?

Also make sure to ping me with `@christianv` so I'm faster in replying to your messages.

@raydavis thanks for the ping, never got a message in my inbox.

NolanChan:
@christianv sorry, I forgot to include your name, I defaulted to IM'ing mode. Anyways, I made the recommended changes

christianv:
@NolanChan thanks for working this one through.

PR #3476, opened 2015-03-16
CLC-4989 Set alternate tmp directory to prevent FFI error starting Torquebox

ctweney:
Longer explanation and background on the JIRA: https://jira.ets.berkeley.edu/jira/browse/CLC-4989


paulkerschen:
LGTM

PR #3477, opened 2015-03-16
CLC-4989 Fix error in reference to tmp dir

ctweney:
My fix to https://jira.ets.berkeley.edu/jira/browse/CLC-4989 broke the master branch deployment. This should fix it.


johncrossman:
Bueno.

PR #3478, opened 2015-03-16
CLC-4990 Fix Teacher/TA drops from sections with short CCNs

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-4990

This is a critical production problem, and so I'll also be submitting a PR against our current QA branch (AKA v52).


paulkerschen:
In it goes.

PR #3483, opened 2015-03-17
CLC-4537 Show EAP/AP/UC Extension transcripts in My Academics

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-4537

Transcript-only semesters (including education-abroad, UC Extension, and very old semesters) are now merged into the My Academics semester feed. Education-abroad and UC Extension semesters get notations indicating their special status.

Additional course credits such as AP units are included in a new feed object. An "Additional Course Credit" box appears at the bottom of the Semesters card and follows the same visibility rules as past semesters: if there are no other past semesters it shows by default, otherwise it's hidden behind a "Show More" button.

The PR looks bigger than it is because the work encouraged some cleanup along the way.
	- Hash keys in intermediate data structures are standardized to `:units` everywhere, instead of a mix of `:unit` and `:units`.
	- The dummy academics.json is updated to include a number of values that the MyAcademics feed now includes.
	- The ever-growing CampusOracle::UserCourses spec is split into several different specs matching the classes in the model directory.

The specs with actual changes are those for CampusOracle::UserCourses::Transcripts and MyAcademics::Semesters. The latter was updated to use randomized mock data structures instead of depending on the H2 sample data. If this looks good, it could be considered a preview of a more factory-based approach in our unit tests.


raydavis:
Thank you for this massive amount of work, and for finishing it so early in the sprint!

PR #3488, opened 2015-03-18
CLC-5000 - Optimize the maintenance image

christianv:
https://jira.ets.berkeley.edu/jira/browse/CLC-5000


raydavis:
Another 191 bytes gone!

PR #3489, opened 2015-03-18
CLC-5001 Include expected sections array in feed for transcript-only courses

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5001


raydavis:
Thanks for the fast fix!

PR #3490, opened 2015-03-18
CLC-4909 Add Regstatus API as a new proxy

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-4909

The structure of this new proxy is based on the Bearfacts proxies. It returns the same custom hashes on error conditions, so that tests can be conveniently bundled in with the other student data proxies.

This simple API endpoint also affords a opportunity to try out a different, VCR-less approach to mocking up web responses. (See https://jira.ets.berkeley.edu/jira/browse/CLC-3383 .) A mock proxy class uses WebMock to intercept HTTP requests and creates a fake response matching the query parameters.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -145,6 +145,8 @@ group :development, :test , :testext do
   # Spork can speed up multiple test runs.
   gem 'spork', :git => 'https://github.com/sporkrb/spork.git'
   gem 'spork-rails', '~> 4.0.0'
+
+  gem 'webmock', '~> 1.20.4'
 end
 
 group :development do
@@ -162,8 +164,6 @@ group :test, :testext do
   # RSpec results that Hudson + Bamboo + xml happy CI servers can read. See https://rubygems.org/gems/rspec_junit_formatter
   # TODO: Use gem 'rspec_junit_formatter', '~> 0.2.x' when deprecated concern of CLC-3565 is resolved.
   gem 'rspec_junit_formatter', :git => 'https://github.com/sj26/rspec_junit_formatter.git'
-
-  gem 'webmock', '~> 1.20.4'
 end
 
 group :shell_debug do
--- a/app/models/regstatus/proxy.rb
+++ b/app/models/regstatus/proxy.rb
@@ -1,0 +1,68 @@
+module Regstatus
+  class Proxy < BaseProxy
+
+    include ClassLogger
+    include Cache::UserCacheExpiry
+    include Proxies::Mockable
+    include User::Student
+
+    def initialize(options = {})
+      super(Settings.regstatus_proxy, options)
+      initialize_mocks if @fake
+    end
+
+    def get(term=nil)
+      term ||= Berkeley::Terms.fetch.current
+      internal_response = self.class.smart_fetch_from_cache(id: "#{@uid}/#{term.slug}") do
+        get_internal term
+      end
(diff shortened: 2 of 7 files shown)


raydavis:
The proxy code itself is nice and straightforward, @paulkerschen. And thank you for thinking about the fake-proxy question. If possible, it would be nice for @ctweney to weigh in on this, too.

The first thing that strikes me in Proxies::Mocks::Regstatus is its exclusive use of hard-coded singleton methods. Does that lose the ability (which both VCR and our later read-the-body-from-a-file have) to set up different mock responses for different initial conditions? (For example, to set up tests in which ``regStatus[:isRegistered]`` is ``true``?) Or do you have a factory-style scheme in mind which hasn't surfaced in this PR?

paulkerschen:
@raydavis, good questions all round. I was myself surprised to find that the mock proxy had no need to maintain state, at least in this implementation; the reason being that state is handled by WebMock, and that once the mock proxy has used its singleton methods to stub WebMock with the appropriate request matchers and responses, that behavior will hang around until WebMock is reset. What's nice about this is that the Proxies::Mocks::Regstatus singleton methods can be called from anywhere (either proxy code or test code) without having to worry about instantiating or passing around objects.

As far as modifying initial conditions, there's a half-considered factory-style scheme which may not be clear from this code. My thinking is that `self.set_stubs` serves as the public interface and can be made to accept options, which in turn override the defaults that are hard-coded within the mock proxy. I didn't want to go wild enforcing any particular convention for this, since overly finicky conditions are part of what's frustrating about VCR, but you're right that this should be made more apparent. I'll push up an amendment that shows a more flexible mock.

raydavis:
FWIW, our usual weirdo-Ruby idiom for providing methods which might either be called as a convenient singleton OR be mixed in and potentially overridden is to drop the explicit "self." prefixes on each method name, and instead put this line at the top:

```
extend self
```

This lets the methods be used either as a utility function OR to to be treated as instance mix-ins. Just plain mix-ins are often preferable, though, and of course you can have them add class-level methods to the including class as well.

paulkerschen:
@raydavis, ah, that's a good un.

paulkerschen:
@raydavis, thanks for talking this out with me. I've pushed an amended commit with more of a road map.

Regstatus::Proxy now mixes in a Proxies::Mockable module, which expects the proxy to define its own mock_request and mock_response defaults for stubbing. These defaults can be overriden from within tests in various ways, as shown in the spec for Proxies::Mockable.

Look forward to feedback from you and @ctweney on this.

raydavis:
Thank you so much for working through this, @paulkerschen . It's looking good.

Could you move your "sample" tests into regstatus/proxy_spec.rb as "real" tests, since they are tied to that proxy and do test aspects of the logic? And just to zip up the last of the code coverage on this proof-of-concept, how about adding the "noStudentId: true" case?

paulkerschen:
@raydavis: The tests are moved. Testing `noStudentId: true` in this context is actually knottier because the API endpoint is never hit: we check campus Oracle for an SID and return the error if not found. (That behavior is tested in the 'a student data proxy' example group.) It would be great if the API returned an error on bad IDs, but it seems to just return false on anything it doesn't understand.

I also made a key change to the Gemfile, which is to include WebMock in the development group.

raydavis:
> (That behavior is tested in the 'a student data proxy' example group.)

Ah, good -- thanks for the reminder.

PR #3491, opened 2015-03-19
CLC-5004 Standardize "params"/"session" hash access

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-5004

The HTTP request-and-session variables' underlying Hash data structures (which use string keys) can leak out of Rails's HashWithIndifferentAccess abstraction, unexpectedly clashing with symbol-key-using code.

In terms of lines of code, this is my biggest PR in ages. But as you'll see, it gets a little samey...

### DIFF
--- a/app/controllers/act_as_controller.rb
+++ b/app/controllers/act_as_controller.rb
@@ -6,22 +6,22 @@ class ActAsController < ApplicationController
 
   def start
     authorize current_user, :can_view_as?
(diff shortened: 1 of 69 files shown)


redconfetti:
If there are any conflicts, it will be easy to resolve them as this mostly is just a pattern of converting symbol keys to string versions. Looks good to me.

PR #3492, opened 2015-03-19
CLC-4913 - Course Provisioning: As an Instructor, I want to only be added to appropriate sections when creating a course site

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-4913

Establishes API based method for creation of a course section. Updates Canvas::ProvideCourseSite to use this method to automatically create a default course section and enroll the site creator as a 'Teacher' in that section. The exception to this is when the course site is created by CCN.

### DIFF
--- a/app/models/canvas/provide_course_site.rb
+++ b/app/models/canvas/provide_course_site.rb
@@ -55,7 +55,8 @@ module Canvas
       import_course_site(@import_data['course_site_definition'])
       retrieve_course_site_details
       import_sections(section_definitions)
-      add_instructor_to_sections(section_definitions) unless is_admin_by_ccns
+      enroll_instructor unless is_admin_by_ccns
+      # add_instructor_to_sections(section_definitions) unless is_admin_by_ccns
 
       expire_instructor_sites_cache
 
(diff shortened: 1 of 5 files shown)


app/models/canvas/provide_course_site.rb
```diff
@@ -207,22 +207,21 @@ def import_sections(canvas_section_rows)
...
-    def add_instructor_to_sections(canvas_section_rows)
-      canvas_section_rows.each do |canvas_section|
-        response = Canvas::CourseAddUser.add_user_to_course_section(@uid, 'TeacherEnrollment',
-          "sis_section_id:#{canvas_section['section_id']}")
-        if response.blank?
-          logger.error("Imported course from #{@import_data['courses_csv_file']} but could not add #{@uid} as teacher to section #{canvas_section['section_id']}")
-          raise RuntimeError, 'Course site was created but the teacher could not be added!'
-        end
+    def enroll_instructor
+      default_section = Canvas::CourseSections.new(:course_id => course_details['id']).create(@import_data['site_name'], '')
```
raydavis: Our original idea was to give the default section a sis_section_id (so that we could get more access routes to it), just not an sis_section_id that matched our official section naming convention. Instead, it looks like the default section will be created with a nil SIS ID.

Although that's not quite what I expected, lack of an SIS ID doesn't actually cause any functional issues, so I'm fine with changing our minds on that requirement. But in that case, since no code ever calls the "create" method with a real sis_section_id, why not just leave the argument out?

raydavis:
Looks great, @redconfetti -- I just had the one question about the new ``create`` method.

app/models/canvas/provide_course_site.rb
redconfetti: As we discussed, lets implement an SIS Section ID that is formatted like "DEFSEC:#{canvas_course_id}".

raydavis:
Thanks for the SIS ID, @redconfetti !

PR #3493, opened 2015-03-19
CLC-4949 CLC-4996 Modify bCourses Official Sections authorization rules

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-4949
https://jira.ets.berkeley.edu/jira/browse/CLC-4996

Changing authorization touched enough places that it seemed best to submit this PR separately. Next step is to fill in more course section data in the feed, but that part of the task should have fewer chances of merge conflicts.

* Authorization rules have been changed, made explicit, & moved to the controller.
* A "canEdit" property has been added to the feed to let the front-end know whether to show the Edit button.
* Two API paths were changed to make the authz checks easier and follow existing conventions.
* The Official Sections app can now be run outside an embedded iframe.

TO DO: Add not-an-instructor site section data.

### DIFF
--- a/app/controllers/canvas_course_provision_controller.rb
+++ b/app/controllers/canvas_course_provision_controller.rb
@@ -1,11 +1,30 @@
 class CanvasCourseProvisionController < ApplicationController
   include ClassLogger
+  include SpecificToCourseSite
 
-  before_filter :api_authenticate
-  before_filter :validate_admin_mode, :only => [:get_feed, :create_course_site]
+  before_action :api_authenticate
+  before_action :validate_admin_mode, :only => [:get_feed, :create_course_site]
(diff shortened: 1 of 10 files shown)


app/controllers/canvas_course_provision_controller.rb
```diff
@@ -17,6 +36,15 @@ def get_feed
     end
   end
 
+  # GET /api/academics/canvas/course_provision/sections_feed/:canvas_course_id.json
+  def get_sections_feed
+    if (feed = Canvas::CourseProvision.new(session['user_id'], options_from_params).get_feed)
```
redconfetti: instead of duplicating the code here, does this work if you just put 'get_feed' ?

PR #3496, opened 2015-03-20
CLC-4910 Merge Regstatus into My Academics feed during term transitions

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-4910

Merges in registration data only during transitions (as determined by whether the term in the Bearfacts profile matches the current term in CalCentral).

Dummy JSON feeds for different scenarios are also included. In addition to registration-related changes, the `collegeAndLevel` data in the dummies has been made consistent with our current feed structure.

### DIFF
--- a/app/models/my_academics/college_and_level.rb
+++ b/app/models/my_academics/college_and_level.rb
@@ -87,12 +87,15 @@ module MyAcademics
         end
       end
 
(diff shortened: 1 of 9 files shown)


raydavis:
Thank you for the extra fake feeds, @paulkerschen !

PR #3487, opened 2015-03-18
CLC-5003 Remove large size for GPA

szhu:
Most students visit the Academics tab to check things other than GPA, so there's no particular reason for GPA to be larger.

In addition, many students visit CalCentral in public settings and would prefer to have their GPA less visible to people around them.

This is a temporary solution, but a "show GPA" checkbox might be more appropriate in the future.

Before:
screen shot 2015-03-18 at 1 56 34 am

After:
screen shot 2015-03-18 at 1 56 42 am

PR #3497, opened 2015-03-20
CLC-5003 Remove large size for GPA

raydavis:
Revived from https://github.com/ets-berkeley-edu/calcentral/pull/3487 for @szhu -- please see the earlier PR for comments.

Most students visit the Academics tab to check things other than GPA,
so there's no particular reason for GPA to be larger.

In addition, many students visit CalCentral in public settings and
would prefer to have their GPA less visible to people around them.

This is a temporary solution, but a "show GPA" checkbox might be more
appropriate in the future.

### DIFF
--- a/src/assets/stylesheets/_profile.scss
+++ b/src/assets/stylesheets/_profile.scss
@@ -6,9 +6,11 @@
     margin-left: 87px;
     padding-bottom: 10px;
   }
+  .cc-widget-profile-content-fullname {
+    font-size: 16px;
+  }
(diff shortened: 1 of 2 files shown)


raydavis:
As you can see at the old PR, this is not the self-merge it appears to be...

@szhu , thank you for donating time and attention to this fix. Your change should hit production on or around April 6. If you happen to be near Dwinelle sometime, drop us a line and pick up a T-shirt!

szhu:
Thanks!

PR #3499, opened 2015-03-20
CLC-4987 Show better registration data during term transitions

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-4882
https://jira.ets.berkeley.edu/jira/browse/CLC-4987

Front-end counterpart to #3490 and #3496.

### DIFF
--- a/src/assets/stylesheets/_profile.scss
+++ b/src/assets/stylesheets/_profile.scss
@@ -1,4 +1,11 @@
 .cc-widget-profile {
+  .cc-widget-profile-message {
+    padding: 15px 15px 0;
+  }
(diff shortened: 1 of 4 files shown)


src/assets/stylesheets/_profile.scss
```diff
@@ -24,4 +27,7 @@
       color: $cc-color-white;
     }
   }
+  p {
```
christianv: @paulkerschen would it be possible to turn this into a `CSS` class?

paulkerschen:
@christianv, yes, a class on a `div` seems to work just as well, as long as it gets a `line-height` property.

PR #3502, opened 2015-03-24
CLC-4911 - Accessibility: As a screen reader user, I want the Create a Site tool functional with my screen reader

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-4911


redconfetti:
@christianv I've reviewed this with Ray. Leaving it open for you to review. Keep in mind that we will do an actual functional review with a screen reader in a separate Jira task. I tried to do that myself with NVDA, and I'm aware that some things do not work as I expected, however this task was about implementing things we documented in the Accessibility Guide.

raydavis:
@christianv, to avoid merge conflicts I'm going to merge this now. @redconfetti is planning some follow-up work, and so any comments you can make will still be very welcome!

src/assets/javascripts/angular/controllers/pages/canvasCreateCourseSiteController.js
```diff
@@ -229,6 +229,13 @@
     $scope.selectedSections = canvasSiteCreationService.selectedSections;
     $scope.toggleCheckboxes = canvasSiteCreationService.toggleCheckboxes;
 
+    /*
+     * Used with bc-official-sections-table directive to ensure every section is displayed
+     */
+    $scope.displayRow = function() {
```
christianv: @redconfetti where are we updating this variable? When is it being updated to false.

And instead of using a function, does it work when it's just a boolean? `$scope.displayRow = true`

src/assets/stylesheets/_canvas_create_course_site.scss
```diff
@@ -124,4 +123,51 @@
     width: 140px;
   }
 
+  //////////////////////////////////////////////////////
```
christianv: @redconfetti can we remove this line? Just use the regular comment:

```
// Form with Expandable Courses & Sections
```

src/assets/templates/canvas_embedded/create_course_site.html
```diff
@@ -5,6 +5,14 @@
   <div data-ng-show="feedFetched && !displayError">
     <div data-ng-show="is_admin">
 
+      <p id="cc-page-reader-alert-ccn-mode" class="cc-visuallyhidden" role="alert" aria-live="polite" aria-relevant="text" data-ng-if="adminMode === 'by_ccn'">
```
christianv: @redconfetti where are these `id`s being used?

src/assets/javascripts/angular/controllers/pages/canvasCreateCourseSiteController.js
redconfetti: @christianv The directive that I introduced to provide a shared template between the 'Create a Course Site' and 'Official Sections' tools requires a function to determine if a row is displayed or not. In the Official Sections tool this is dynamic, but in the 'Create a Course Site' context it is always true.

This is needed because the template relies on an ng-if [that calls](https://github.com/ets-berkeley-edu/calcentral/blob/8a1e60a9/src/assets/templates/canvas_embedded/_shared/sections_table.html#L18) a [rowDisplayLogic method](https://github.com/ets-berkeley-edu/calcentral/blob/8a1e60a9/src/assets/javascripts/angular/directives/bcSectionsTableDirective.js#L16). In the 'Create a Course Site' template I [pass in this displayRow method](https://github.com/ets-berkeley-edu/calcentral/blob/8a1e60a9/src/assets/templates/canvas_embedded/create_course_site.html#L140) as that rowDisplayLogic method.

Curiously, I don't define two arguments on this function in the the 'Create a Course Site' controlller, but I do in the 'Official Sections' controller for the [rowDisplayLogic method](https://github.com/ets-berkeley-edu/calcentral/blob/8a1e60a9/src/assets/javascripts/angular/controllers/pages/canvasCourseManageOfficialSectionsController.js#L321). I'm curious why I don't end up receiving an invalid argument number error with the Javascript.

Anyway, in this case I'm trying to emulate this other function call and always return true.

redconfetti: @christianv I'm actually going to make a PR now to make this more consistent and avoid an invalid number of method arguments error. Still, it would cause an error for the template to try to pass arguments to a variable that has a value right?

```javascript
var valueVar = true;
valueVar('listMode', 'section');
// Uncaught TypeError: boolean is not a function
```

redconfetti: I see why it worked without an error now.

```javascript
>> var twoArgMethod = function(arg1, arg2) {
  console.log(arg1 + ' ' + arg2);
};
=> undefined
>> twoArgMethod();
=> undefined undefined
```

src/assets/stylesheets/_canvas_create_course_site.scss
redconfetti: @christianv But it's a nice visual separator, easily caught by the eye of the developer.

You're no fun! :tongue:

src/assets/templates/canvas_embedded/create_course_site.html
redconfetti: @christianv At one point there were associated with aria-control attributes, but those must have been removed when they were seen to be unnecessary.

PR #3505, opened 2015-03-24
CLC-4992 Zero-pad CCN strings from Oracle queries

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-4992

Removes workarounds from #3478.


johncrossman:
Mighty fine.

PR #3506, opened 2015-03-24
CLC-4793 Allow cache deletion API to accept keys in query strings

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-4793


johncrossman:
All good.

PR #3513, opened 2015-03-26
CLC-4916  - Find a Person to Add: Refactor authorization to respect role type rather than role

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-4916

Updates the Course User Roles API end-point to include the users role types for authorization in addition to the actual roles. Removes custom roles from the roles hash that were added to temporarily authorize access to the 'Find a Person to Add' tool for the Owner, Maintainer, and Member roles that were introduced to the 'Projects' account.

### DIFF
--- a/app/models/canvas/course_user.rb
+++ b/app/models/canvas/course_user.rb
@@ -67,6 +67,12 @@ module Canvas
       roles_hash
     end
 
+    def role_types
+      profile = course_user
+      return [] if profile.nil? || profile['enrollments'].nil? || profile['enrollments'].empty?
+      role_types = profile['enrollments'].collect {|enrollment| enrollment['type'] }
+    end
+
     # Do not need to log a stack trace when the user is not a course site member.
     def existence_check
       true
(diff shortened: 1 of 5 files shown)


app/models/canvas/course_user.rb
```diff
@@ -61,12 +61,15 @@ def roles
       roles_hash['ta'] = true if roles.include?('TaEnrollment')
       roles_hash['designer'] = true if roles.include?('DesignerEnrollment')
       roles_hash['waitlistStudent'] = true if roles.include?('Waitlist Student')
-      roles_hash['owner'] = true if roles.include?('Owner')
```
raydavis: Why would we want to lose this information, @redconfetti ?

raydavis:
I get the utility of better access to the built-in role-types. But I'm worried about losing access to our customizable role-names. Let's talk about it?

redconfetti:
@raydavis This appears to have failed, but if you look at both TravisCI builds, the Rspec pass: `1744 examples, 0 failures`, and the Javascript errors have nothing to do with my code.

raydavis:
Yes, I see that, @redconfetti. Even knowing that the JavaScript errors weren't caused by you, would you mind adding a fix for them to this PR to unblock us? It's annoying but I expect we'll all get our turns...

christianv:
@raydavis @redconfetti please see #3516 for the `jscs` errors fix. Then just restart travis for this PR after the merge.

PR #3517, opened 2015-03-26
CLC-4911 - Misc Front-end fixes

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-4911

Misc fixes related to comments by @christianv in #3502

### DIFF
--- a/src/assets/javascripts/angular/controllers/pages/canvasCreateCourseSiteController.js
+++ b/src/assets/javascripts/angular/controllers/pages/canvasCreateCourseSiteController.js
@@ -232,7 +232,7 @@
     /*
      * Used with bc-official-sections-table directive to ensure every section is displayed
      */
-    $scope.displayRow = function() {
+    $scope.rowDisplayLogic = function() {
       return true;
     };
 
--- a/src/assets/templates/canvas_embedded/create_course_site.html
+++ b/src/assets/templates/canvas_embedded/create_course_site.html
@@ -137,7 +137,7 @@
 
                         <div class="row">
                           <div class="medium-12 columns">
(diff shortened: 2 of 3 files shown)


redconfetti:
@christianv I defined arguments in the function because I knew that it would have arguments passed to the function, but the function doesn't use the arguments and thus I'm getting these JSHint errors:

```shell
src/assets/javascripts/angular/controllers/pages/canvasCreateCourseSiteController.js: line 235, col 49, 'section' is defined but never used.
src/assets/javascripts/angular/controllers/pages/canvasCreateCourseSiteController.js: line 235, col 39, 'listMode' is defined but never used.
```

Is it valid to pass arguments to a function even if it isn't defined to receive them?

src/assets/javascripts/angular/controllers/pages/canvasCreateCourseSiteController.js
```diff
@@ -232,7 +232,7 @@
     /*
      * Used with bc-official-sections-table directive to ensure every section is displayed
      */
-    $scope.displayRow = function() {
+    $scope.rowDisplayLogic = function(listMode, section) {
```
christianv: @redconfetti indeed, we should remove those when they're never being used inside the function. Passing in arguments and never using them in the function is something we should avoid.

src/assets/templates/canvas_embedded/create_course_site.html
```diff
@@ -5,11 +5,11 @@
   <div data-ng-show="feedFetched && !displayError">
     <div data-ng-show="is_admin">
 
-      <p id="cc-page-reader-alert-ccn-mode" class="cc-visuallyhidden" role="alert" aria-live="polite" aria-relevant="text" data-ng-if="adminMode === 'by_ccn'">
```
christianv: @redconfetti thanks for removing the ID's, since we didn't use them

PR #3518, opened 2015-03-26
CLC-5023 Parse tasks without times as Date instead of DateTime

ctweney:
https://jira.ets.berkeley.edu/jira/browse/CLC-5023


ctweney:
@raydavis I pushed a second commit that applies the change to Google as well .

raydavis:
Thank you for noticing and fixing this, @ctweney ! If it tests out on Dev next week, I think it's worth adding to the QA branch, too.

PR #3519, opened 2015-03-30
CLC-5026 Update path for Google Drive badge icons

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5026


raydavis:
Thanks, @paulkerschen ! If this tests OK in Dev, I think a QA PR would be nice, too.

PR #3523, opened 2015-03-30
CLC-4873 - Create a Project Site: Exceed max char limit and boom: "Unexpected Error"

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-4873


raydavis:
Looks like a nice improvement, @redconfetti -- thanks!

PR #3535, opened 2015-04-01
CLC-5043 Revise card display for current transcript-only semesters

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5043


raydavis:
Looks right from here!

PR #3536, opened 2015-04-01
OEC-585, do not allow diff on BIOLOGY_courses_confirmed.csv

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/OEC-585

BIOLOGY data is managed within:
* MCELLBI_courses_confirmed.csv
* INTEGBI_courses_confirmed.csv

This change causes an alert when the always-unexpected _BIOLOGY_courses_confirmed.csv_ is passed in.


paulkerschen:
BIOLOGY(.*)is not allowed! Go home!

PR #3538, opened 2015-04-02
OEC-581, allow for CCNs with those damn leading zeroes

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/OEC-581

Let's be tolerant of CCNs such as 09876


paulkerschen:
It takes all kinds.

PR #3540, opened 2015-04-02
CLC-4955 - Update E-Grade Export tool to run in CalCentral Context

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-4955

### DIFF
--- a/app/controllers/canvas_course_add_user_controller.rb
+++ b/app/controllers/canvas_course_add_user_controller.rb
@@ -18,7 +18,13 @@ class CanvasCourseAddUserController < ApplicationController
   # GET /api/academics/canvas/course_user_roles
   def course_user_roles
     profile = user_profile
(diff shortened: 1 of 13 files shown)


raydavis:
Loads of RSpec errors, @redconfetti , but they look all of a type.

christianv:
@redconfetti front-end lgtm

raydavis:
Back-end changes look good too -- just need to kick the tests back into line, and we can wrap it up.

redconfetti:
@raydavis - It's odd that the [default setting for url_root](https://github.com/ets-berkeley-edu/calcentral/blob/0a0f65cb/config/settings.yml#L66), which isn't being overridden in the [test config](https://github.com/ets-berkeley-edu/calcentral/blob/0a0f65cb/config/settings/test.yml#L8), is 'http://localhost:12345', but instead the test is getting 'https://ucb.beta.example.com' ?!?

I guess I'll just have to mock the response from Settings.canvas_proxy for url_root in this test so that it passes in any environment.

redconfetti:
Oh nevermind. I'm getting the same errors locally. I expected them to only happen within Travis environment.

PR #3542, opened 2015-04-03
CLC-1484 Symbolize keys in tasks feed

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-1484


raydavis:
Everything I can think to say about colons sounds odd out of context.

PR #3543, opened 2015-04-03
CLC-4748 Run our rake tasks from shell script

ctweney:
https://jira.ets.berkeley.edu/jira/browse/CLC-4748

Step 1 in getting rid of our archaic Ruby Bamboo plugin dependency. 

This job shows the simplification in Bamboo structure -- we get to drop 2 tasks from the rake stage: https://bamboo.ets.berkeley.edu/bamboo/build/admin/edit/editBuildTasks.action?buildKey=MYB-CLCMVPCHRIS-JOB1

### DIFF
--- a/script/bamboo-tests.sh
+++ b/script/bamboo-tests.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+# Script to install necessary dependencies and then run testext tests, for use on Bamboo CI
+
+cd $( dirname "${BASH_SOURCE[0]}" )/..
+
+# Enable rvm and use the correct Ruby version and gem set.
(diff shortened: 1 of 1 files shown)


ctweney:
If this PR is accepted I'd recommend also merging it into QA so that we can get rid of the plugin sooner. This script only interacts with Bamboo and won't affect any production code.

raydavis:
Thanks, @ctweney! As hoped, looks like rolling our own shell script beats the plugin regardless. I'm just waiting on Travis now.

If Ops can wait that long, I'd vote for changing the developer-specific Bamboo job definitions first and dealing with QA a bit later.

PR #3546, opened 2015-04-06
CLC-5063 Do not query an empty list of UIDs from Canvas Refresh

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-5063

Multiples of a thousand led to an empty SQL query clause and sudden script termination.


johncrossman:
Looks good

PR #3547, opened 2015-04-06
CLC-4991 Detangle LTI-embedded and shared templates

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-4991

This PR attempts the minimum needed to cleanly separate bCourses-specific styling from LTI-iframe-embedded state.

The next steps include supporting different styles in the three templates (roster, roster-photo, and webcasts) which are shared by CalCentral page and bCourses tools.


src/assets/javascripts/angular/configuration/routeConfiguration.js
```diff
@@ -129,8 +130,7 @@
     }).
     when('/canvas/course_grade_export/:canvasCourseId', {
       templateUrl: 'canvas_embedded/course_grade_export.html',
-      controller: 'CanvasCourseGradeExportController',
-      isEmbedded: true
```
redconfetti: @raydavis Good catch. I think I missed that `isEmbedded: true` with a copy+paste action without realizing that it wasn't relevant.

PR #3551, opened 2015-04-07

raydavis:
_Possible_ next step for: https://jira.ets.berkeley.edu/jira/browse/CLC-4991

@redconfetti , @christianv , I feel fairly confident that the isBcourses and SCSS moves match our agreement but considerably less sure about ```_canvas_webcast.scss```. 

The best way to mix distinct styles into the shared Roster and Webcasts logic will probably be with directives and ngClass. On the other hand I wanted to make a visible change ASAP so as to make sure the code reorganization was OK. Thus the hackish button override. Just let me know if you'd like me to take that bit of code out before the PR is merged.


raydavis:
Whoops, forgot to scss-lint locally. I'll close this for now, but @christianv & @redconfetti , if you get a chance please take a look.

src/assets/stylesheets/_canvas_webcast.scss
```diff
@@ -0,0 +1,7 @@
+.bc-main {
+  .cc-button-group {
+    .cc-button, .cc-button-selected {
+      @extend .bc-canvas-button;
```
christianv: @raydavis would there be a way to avoid the `@extend` here? Would it be ok to just add the `bc-canvas-button` class for now? In general, using `@extend` should be avoided as much as possible.

redconfetti: @christianv Really, that seems to make the code drier. What's the reasoning behind not using the `@extend` except for in special cases?

src/assets/javascripts/angular/controllers/pages/canvasCreateProjectSiteController.js
```diff
@@ -2,7 +2,7 @@
   'use strict';
 
   /**
-   * Canvas 'Create a Site' overview index controller
+   * Canvas 'Create a Project Site' overview index controller
```
redconfetti: image

src/assets/templates/canvas_embedded/course_mediacasts.html
```diff
@@ -1,3 +1,3 @@
-<div class="cc-widget cc-widget-webcast">
+<div class="bc-main cc-widget-webcast">
```
christianv: @raydavis I think a bug might be lurking here, would it be ok to keep the `cc-widget` class here? See https://github.com/ets-berkeley-edu/calcentral/commit/ec9d302f39a03bc29f69acdac25e524de2a824d0

So we can have both `bc-main` and `cc-widget` on there.

src/assets/stylesheets/_canvas_webcast.scss
raydavis: @christianv , interesting suggestion. I'll try it and see.

My amended branch extends a placeholder version of ```%bc-canvas-button'```, by the way.

PR #3553, opened 2015-04-07

raydavis:
Possible next step for: https://jira.ets.berkeley.edu/jira/browse/CLC-4991
Amended version of: https://github.com/ets-berkeley-edu/calcentral/pull/3551
Same notes apply.

@redconfetti , @christianv , I feel fairly confident that the isBcourses and SCSS moves match our agreement but considerably less sure about _canvas_webcast.scss. The best way to mix distinct styles into the shared Roster and Webcasts logic will probably be with directives and ngClass. On the other hand I wanted to make a visible change ASAP so as to make sure the code reorganization was OK. Thus the hackish button override.

I need to get clarification from @christianv on one of his notes at the first PR. In the meantime, this adds basic Canvas styling for selected button.


redconfetti:
I'm not necessarily understanding how directives would be used to accomplish dynamic styling into the tools, unless you're suggesting that custom directives be defined if/when ngClass doesn't meet your needs through something like `data-ng-class="{'bc-specific-class-example':(isBcourses)}"`.

@christianv Do you agree with such an approach? Or do you see this as less preferable to the use of a special wrapper DIV with a bCourses related class name (e.g. 'bc-wrapper' or 'cc-canvas-wrapper') for templates presented in the bCourses LTI contexts?

raydavis:
Good questions, @redconfetti -- I'm interested in @christianv's response too.

My first thought was directives mostly because they seem to be the recommended way to share bits of front-end logic across multiple contexts. So far as styling goes, I remembered you passing class-logic as an attribute to bcSectionsTable. But I don't yet know if that level of flexibility is needed by the two shared templates.

src/assets/stylesheets/_canvas_webcast.scss
```diff
@@ -0,0 +1,10 @@
+.bc-main {
+  .cc-button-group {
+    .cc-button {
+      @extend %bc-canvas-button;
```
christianv: @raydavis as mentioned before, we should try to avoid using `@extend` as much as possible.

It's just a different way of extending the CSS, you can use multiple classes on an element:

Something like this:
```html
<div class="cc-button-group">
  <button class="cc-button bc-canvas-button">
```
Or, in the CSS, you would also add the button group:
```css
.bc-canvas-button, .bc-main .cc-button-group .cc-button {
...
}
```

src/assets/stylesheets/_canvas_main.scss
```diff
@@ -0,0 +1,4 @@
+// Wrapper class for all bCourses apps, in place of CalCentral's cc-widget
+.bc-main {
+  background: #fff;
```
christianv: @raydavis Can we make sure this becomes a color SASS variable?

src/assets/stylesheets/_canvas_emulated.scss
```diff
@@ -7,9 +10,12 @@ $bc-canvas-button-primary-hover-border: #525252;
 $bc-canvas-button-hover-background: #e8e8e8;
 $bc-canvas-button-hover-border: #d1d1d1;
 $bc-canvas-off-black: #555;
+$bc-canvas-button-selected-background: #6b6b6b;
+$bc-canvas-button-selected-border: #525252;
+$bc-canvas-button-selected-color: #fff;
 
 // Canvas Button Styles
-.bc-canvas-button {
+.bc-canvas-button, %bc-canvas-button {
```
christianv: @raydavis Can we avoid the use of `%...`? The main reason for this is that we'll want to move away from SASS eventually, as soon as CSS variables become supported in most browsers.

src/assets/stylesheets/_canvas_webcast.scss
raydavis: @christianv , I can't insert the bCourses-specific bc-canvas-button class into a non-bCourses-specific shared template. That would call for use of ngClass.

The second approach might work, though. I'll give it a try.

FWIW, Canvas's own SCSS definitions use ```@extend``` extensively.

raydavis:
I'd like to get the less controversial aspects of this branch merged, so I'll close again and issue a new PR without any visible changes whatsoever.

christianv:
@raydavis @redconfetti I would suggest doing the adding of the specific bCourses class on the highest level possible, which is our current `<body>` tag in the `base.html` file.

`data-ng-class="{'bc-wrapper':(isBcourses)}"`.

This means we'll need to pass the `isBcourses` variable on the highest level, which is our [CalcentralController](https://github.com/ets-berkeley-edu/calcentral/blob/f1298db3a4b91d2f6c4979a89ec67085777845be/src/assets/javascripts/angular/controllers/calcentralController.js#L18).


```javascript
apiService.util.checkIsBcourses();
```

and then add the `checkIsBcourses` method to the utilService:
```
var checkIsBcourses = function() {
  $rootscope.isBcourses = $route.current.isBcourses;
}
```

PR #3554, opened 2015-04-07
CLC-4939 Show short descriptions for course sites in My Classes

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-4939

### DIFF
--- a/src/assets/stylesheets/_classes.scss
+++ b/src/assets/stylesheets/_classes.scss
@@ -9,6 +9,21 @@
   .cc-widget-classes-title {
     margin-bottom: -4px;
   }
+  .cc-widget-course-site-margin {
+    margin: 0 0 4px 22px;
+  }
+  .cc-widget-course-site-subtitle {
+    font-size: 12px;
+    margin-bottom: -4px;
+  }
+  .cc-widget-icon {
+    margin-top: -8px;
+    position: absolute;
+    top: 50%;
+  }
+  .cc-widget-icon-container {
+    position: relative;
+  }
   .cc-widget-list {
     font-size: 13px;
     > li {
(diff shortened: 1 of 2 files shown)


raydavis:
Looks like a big improvement, @paulkerschen. Thanks especially for coming up with the only-display-distinct-site-names wrinkle.

src/assets/stylesheets/_classes.scss
```diff
@@ -9,6 +9,21 @@
   .cc-widget-classes-title {
     margin-bottom: -4px;
   }
+  .cc-widget-course-site-margin {
+    margin: 0 0 4px 22px;
+  }
+  .cc-widget-course-site-subtitle {
+    font-size: 12px;
+    margin-bottom: -4px;
```
christianv: @paulkerschen as a fyi, we should try to avoid negative `margin` and absolute positioning as much as possible. Instead we should try to use things like `line-height`. However, if there is no other way at all, then yes, that's what we should go for.

paulkerschen: @christianv - ok, good to know as a general point. These particular negative margins just follow the lead of their neighbor classes.

src/assets/stylesheets/_classes.scss
```diff
@@ -9,6 +9,21 @@
   .cc-widget-classes-title {
     margin-bottom: -4px;
   }
+  .cc-widget-course-site-margin {
+    margin: 0 0 4px 22px;
+  }
+  .cc-widget-course-site-subtitle {
+    font-size: 12px;
```
christianv: @paulkerschen do we need the `font-size` here? Doesn't seem to change when I remove it.

paulkerschen: @christianv yes, that was carried over from a sister class. Removed in #3559 .

PR #3556, opened 2015-04-07

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-4991

Amended version of: https://github.com/ets-berkeley-edu/calcentral/pull/3553#issuecomment-90729136


src/assets/stylesheets/_canvas_buttons.scss
```diff
@@ -0,0 +1,79 @@
...
+$bc-canvas-button-primary-hover-border: #525252;
+$bc-canvas-button-hover-background: #e8e8e8;
+$bc-canvas-button-hover-border: #d1d1d1;
+$bc-canvas-off-black: #555;
+$bc-canvas-button-selected-background: #6b6b6b;
+$bc-canvas-button-selected-border: #525252;
+$bc-canvas-button-selected-color: #fff;
+
+// Canvas Button Styles
+.bc-canvas-button, .bc-main .cc-widget-webcast .cc-button {
```
raydavis: We could override ```cc-button``` styling for _all_ bCourses apps by simply removing ```.cc-widget-webcast``` from this line. However, that would introduce a massive UX change at one go (for example, in Create a Course Site, which currently shows a mix of cc-button and bc-button styles), and we said we'd like to standardize styling one app at a time. This more focused change should be enough to demonstrate the concept.

raydavis:
My proof-of-concept goal was to support both a ngClass approach and a pure SCSS approach for styling those layouts which were needed by both CalCentral and bCourses contexts. This shows the SCSS wrapper approach at the widget level and shows the ngClass approach at the base.html level.

I believe we're in agreement that unshared, bCourses-only templates should just explicitly use "bc-"-specific classes.

raydavis:
As usual, pinging @redconfetti & @christianv .

src/assets/stylesheets/_canvas_buttons.scss
christianv: @raydavis +1

christianv:
front-end lgtm - passing over to @redconfetti for final review.

redconfetti:
Looks good to me.

src/assets/javascripts/angular/services/utilService.js
```diff
@@ -28,6 +28,13 @@
     };
 
     /**
+     * Pass whether this is a bCourses app context or not.
+     */
+    var checkIsBcourses = function() {
+      $rootScope.isBcourses = $route.current.isBcourses;
```
redconfetti: @christianv Is it safe now to expect the last statement in a function to be the return with Javascript now, or must this and other functions explicitly use `return` ?

christianv: @redconfetti should be an explicit return.

PR #3559, opened 2015-04-08
CLC-5087 Vertically center icons in My Classes

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5087


johncrossman:
LGTM. Waiting on Trav

PR #3562, opened 2015-04-08
CLC-5050, add rooms.json URL to settings; rename property to webcast_proxy

johncrossman:

rooms.json feed lists capture-enabled rooms. URL scheme is /warehouse/2015/B/rooms.json

Note: I've created https://jira.ets.berkeley.edu/jira/browse/WO-2449 in order to get JSON file in proper location.


config/settings.yml
```diff
@@ -147,9 +147,10 @@ textbooks_proxy:
   base_url: "https://calstudentstore.berkeley.edu"
   token: secret set per deployment layer
 
-playlists_warehouse_proxy:
+webcast_proxy:
   fake: false
   base_url: 'https://video-dev.ets.berkeley.edu/warehouse/webcast.json'
+  rooms_url: 'https://video-dev.ets.berkeley.edu/warehouse/2015/B/rooms.json'
```
paulkerschen: @johncrossman , the way we tend to do settings for proxies with multiple URLs is to have one `base_url` value set to the longest common element, and then build the full URL for each endpoint in code. CalLink is a good example:

https://github.com/ets-berkeley-edu/calcentral/blob/master/config/settings.yml#L106
https://github.com/ets-berkeley-edu/calcentral/blob/master/app/models/cal_link/memberships.rb#L14
https://github.com/ets-berkeley-edu/calcentral/blob/master/app/models/cal_link/organization.rb#L19

Does that seem doable here?

johncrossman: Thanks @paulkerschen. I'll investigate the doability.

paulkerschen: Sorry @johncrossman - I should have mentioned this above, but is the `2015/B` portion of the URL expected to change reliably from term to term? If so, and if it seems possible to build the URL in code, it might be better to call `Berkeley::Terms.fetch.current` and get the current term year and code dynamically.

christianv:
@johncrossman would it be possible to add a dummy file / fixture somewhere so we can easily see the JSON structure?

config/settings.yml
johncrossman: @paulkerschen  I don't think that's appropriate here. For example, during late spring and summer of 2015 we'll want that URL to be 2015/D.  Like OEC, the necessary term is not always the 'current'.

paulkerschen: @johncrossman Makes sense, thanks. Do you think there's any hope of reflecting the term requirements in code, or are we going to be stuck swapping in hard-coded values?

raydavis: I second the plea for programmatic transitions if at all possible. Time-dependent manual configurations are a near-guarantee of production emergencies. If nothing else, it would be good to choose a time to start nagging in the log files.

johncrossman:
@christianv : see fixtures/json/rooms.json
@paulkerschen : now using base_url convention. See sub-tasks of https://jira.ets.berkeley.edu/jira/browse/CLC-4925 for related work in progress.

johncrossman:
@raydavis @paulkerschen  Re: programmatic term transition for webcast integration, I created https://jira.ets.berkeley.edu/jira/browse/CLC-5102. I'll follow up.

paulkerschen:
Looks good to me. Thank you for replacing that weird `playlists_warehouse` name!

PR #3563, opened 2015-04-08
CLC-5090 - Instructor class page: class info/roster buttons push webcast & textbook widgets down

christianv:
https://jira.ets.berkeley.edu/jira/browse/CLC-5090


paulkerschen:
Thanks!

PR #3564, opened 2015-04-08
NOJIRA Fix variable interpolation in logs

paulkerschen:


raydavis:
Thanks for the clean-up!

PR #3565, opened 2015-04-08
CLC-5040 Change transition dates for legacy Financial Aid system

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-5040

This shouldn't ever be needed, but just in case...


paulkerschen:
The future that wasn't!

PR #3566, opened 2015-04-08
CLC-5050, bug fix in webcast.json feed gives us more exact start times

johncrossman:

Expected recordingStart date needed change according to a bug fix on the Webcast side. 

Bamboo build against this branch (CLC-5050) was a success: 
https://bamboo.ets.berkeley.edu/bamboo/browse/MYB-CJT41-1


johncrossman:
Note: Bamboo server now has proper webcast configs. Once this PR is merged I will unleash and verify Mediacasts specs on master build.

PR #3568, opened 2015-04-09
CLC-5101, base proxy for webcast in support of rooms and all_playlists feeds

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5101
Passed my Bamboo branch plan: https://bamboo.ets.berkeley.edu/bamboo/browse/MYB-CJT42-1

Webcast system now provides two feeds: courses and webcastEnabledRooms. This PR has a new base proxy class,  and corresponding specs.


app/models/mediacasts/proxy.rb
```diff
@@ -0,0 +1,56 @@
...
+      :proxy_error_message => 'There was a problem fetching the Webcast-related data'
+    }
+
+    ERRORS = {
+      :video_error_message => 'No Webcast data found.'
+    }
+
+    def initialize(options = {})
+      super(Settings.webcast_proxy, options)
+      @json_url = "#{@settings.base_url}/#{get_json_path}"
```
raydavis: Just curious, @johncrossman -- what led you to use an instance variable rather than a method arg?

app/models/mediacasts/rooms.rb
```diff
@@ -0,0 +1,26 @@
...
+  class Rooms < Proxy
+
+    def initialize(term_yr, term_cd, options = {})
+      super(options)
+      @term_yr = term_yr
+      @term_cd = term_cd.upcase
+    end
+
+    def get_json_path
+      "#{@term_yr}/#{@term_cd}/rooms.json"
```
raydavis: The superclass's ```get``` assumes a global cache key. Do we have some ironclad guarantee that only one Rooms feed will ever be instantiated in a running server (that is, it's a singleton)? One easy way to ensure that would be for Mediacasts::Rooms to figure out the one-and-only term_yr and term_cd itself rather than taking them as initialization arguments.

app/models/mediacasts/proxy.rb
johncrossman: @raydavis Mostly for readability. I'll go the prudent route and move it to get_json_data.

app/models/mediacasts/rooms.rb
johncrossman: @raydavis : I've added your concern to CLC-5102. I'm sure we can cook up a programmatic solution and then we can get rid of term args of Mediacasts::Rooms init. Does that work for you?

raydavis: Sure does, @johncrossman -- thanks!

PR #3571, opened 2015-04-09
CLC-5114 Modify dummy JSON for multiple primaries

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5114


raydavis:
Travis is half-full, and I'm always happy to see more useful fake-JSON.

PR #3574, opened 2015-04-10
CLC-5102, webcast-integration in Junction programmatic term transition scheme

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5102
Bamboo build: https://bamboo.ets.berkeley.edu/bamboo/browse/MYB-CJT43-1

Webcast provides /webcast-system-status.json which has property 'isSignUpActive'=[true | false]. This allows us to take term_yr and term_cd args out of Room class. This is a programmatic way of enabling webcast alerts without adding hard-to-maintain configs.

### DIFF
--- a/app/models/mediacasts/rooms.rb
+++ b/app/models/mediacasts/rooms.rb
@@ -1,21 +1,18 @@
 module Mediacasts
   class Rooms < Proxy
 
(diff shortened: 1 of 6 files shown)


raydavis:
Thanks, @johncrossman, that's a perfect solution!

PR #3577, opened 2015-04-10
SISRP-2515 Investigatory code - convert to UTF-8 at point in code where ...

ctweney:
...we get UndefinedConversionError

This is kind of a stab in the dark but might help us get past the "UndefinedConversion" errors we've been seeing in the Slate feed sometimes. 

https://jira.berkeley.edu/browse/SISRP-2515


christianv:
let's try this out

PR #3583, opened 2015-04-11
CLC-5068 - E-Grades Export: Implement background job process for course grade preparation

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-5068

### DIFF
--- a/app/controllers/canvas_course_grade_export_controller.rb
+++ b/app/controllers/canvas_course_grade_export_controller.rb
@@ -25,7 +25,7 @@ class CanvasCourseGradeExportController < ApplicationController
   # GET /api/academics/canvas/egrade_export/status/:canvas_course_id.json?jobId=Canvas::BackgroundJob.1383330151057-67f4b934525501cb
   def job_status
     background_job = Canvas::BackgroundJob.find(params['jobId'])
-    render json: background_job.background_job_report and return if background_job.class == Canvas::Egrades
+    render json: background_job.background_job_report.to_json and return if background_job.class == Canvas::Egrades
     render json: { jobId: params['jobId'], jobStatus: 'Error', errors: ['Unable to find Canvas::EGrades background job'] }.to_json
   end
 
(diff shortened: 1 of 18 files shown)


app/controllers/canvas_course_grade_export_controller.rb
```diff
@@ -14,6 +14,21 @@ def authorize_exporting_grades
     authorize canvas_course, :can_export_grades?
   end
 
+  # POST /api/academics/canvas/egrade_export/prepare/:canvas_course_id.csv
```
redconfetti: This shouldn't end with .csv, but without extension or '.json'

app/models/canvas/background_job.rb
```diff
@@ -0,0 +1,122 @@
...
+      Hash.new
+    end
+
+    def background_job_set_total_steps(total_steps)
+      @background_job_total_steps = total_steps.to_i
+      background_job_save
+    end
+
+    def background_job_add_error(error)
+      @background_job_errors ||= []
```
paulkerschen: @redconfetti - Is there a reason to repeat the initialization code instead of using your `background_job_errors` method above?

app/models/canvas/background_job.rb
```diff
@@ -0,0 +1,122 @@
...
+        completedSteps: background_job_completed_steps,
+        percentComplete: (background_job_completed_steps.count.to_f / background_job_total_steps.to_f).round(2),
+      }
+      json_hash[:errors] = background_job_errors if background_job_errors.count > 0
+      json_hash.reverse_merge!(background_job_report_custom)
+      json_hash.to_json
+    end
+
+    def background_job_complete_step(step_text)
+      @background_job_completed_steps ||= []
```
paulkerschen: @redconfetti - See above comment. I like this module, but I'm a little thrown off by having all the different instance variables lazy-initialized in different methods; it makes the logic harder to follow, especially since the expected usage seems to be that the including class will call everything in an `initialize` method anyway. Do you think it might work to just consolidate all the initialization in a `background_job_initialize` method that the including class could call with an options hash?

johncrossman:
@paulkerschen concerns addressed. Ready for merge

PR #3588, opened 2015-04-13
CLC-5121 - Update front-end dependencies: gulp-angular-templatecache / gulp-rename / gulp-rev-all / gulp-uglify

christianv:
https://jira.ets.berkeley.edu/jira/browse/CLC-5121


johncrossman:
Looks good.

PR #3591, opened 2015-04-13
CLC-4940, Webcast widget: No Video / Audio tabs when term >= Fall 2015

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-4940

Prepare course_media data for front end consumption. The name of the var includes "rss" because the RSS feeds are the underlying requirement of audio player.


johncrossman:
Investigating Travis fail.

christianv:
@johncrossman didn't we decide on a more general solution?

From Judy:
> hide the audio & video tabs (and show the video part) when there is no audio

So make it no depended on the year?

johncrossman:
@christianv You're right. :)  A new PR will be in soon.

PR #3593, opened 2015-04-13
CLC-5035 - Update preventBubble for outbound links

christianv:
https://jira.ets.berkeley.edu/jira/browse/CLC-5035


paulkerschen:
Thanks for the update, @christianv !

christianv:
@paulkerschen thank you for digging into it, would be awesome if you can do an early test on `calcentral-dev` and comment on the jira.

PR #3595, opened 2015-04-13
CLC-5123 - Student Lookup: "Recent" tab is confusing

edwlook:
https://jira.ets.berkeley.edu/jira/browse/CLC-5123
Screencast: https://www.youtube.com/watch?v=xLK8bA_0Lyw

### DIFF
--- a/src/assets/templates/widgets/student_lookup.html
+++ b/src/assets/templates/widgets/student_lookup.html
@@ -18,9 +18,9 @@
       </ul>
     </div>
 
-    <!-- Search Users -->
-    <div data-ng-show="admin.currentSelection === 'Search'">
-      <form data-ng-submit="admin.lookupUser()">
+    <!-- Search/Saved/Recent Users -->
+    <div data-ng-show="admin.currentSelection === tab.name" data-ng-repeat="tab in admin.tabs" class="cc-widget-student-lookup-tab">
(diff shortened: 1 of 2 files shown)


src/assets/templates/widgets/student_lookup.html
```diff
@@ -18,9 +18,9 @@
       </ul>
     </div>
 
-    <!-- Search Users -->
-    <div data-ng-show="admin.currentSelection === 'Search'">
-      <form data-ng-submit="admin.lookupUser()">
+    <!-- Search/Saved/Recent Users -->
+    <div data-ng-show="admin.currentSelection === tab.name" data-ng-repeat="tab in admin.tabs" data-ng-init="showLimit = tab.showLimit" class="cc-widget-student-lookup-tab">
```
christianv: @edwlook can we try a different solution than the `ng-init`?

See the angular docs on this:
> The only appropriate use of ngInit is for aliasing special properties of ngRepeat, as seen in the demo below. Besides this case, you should use controllers rather than ngInit to initialize values on a scope.

christianv:
@edwlook thanks! Lgtm as soon as we change the `ng-init` solution with something else.

edwlook:
@christianv Done, `ng-init` is no longer used. Since each ng-repeat creates a new child scope for every item it iterates over, I can simply initialize `showLimit` in the controller.

PR #3596, opened 2015-04-13
CLC-4789 Separate multiple primary section enrollments in Class Info

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-4789

A few notes:

The `course_option` database column, which was only being queried for instructor assignments, is now queried for student enrollments as well because we need it for the edge case of assigning secondary sections among multiple primaries. The value is not included in the final My Academics or My Classes feeds.

Our policy of not exposing section associations for Canvas sites had to be very slightly relaxed in order to point Canvas sites to the correct primary section in the multiple-primaries case. These associations were already exposed in My Classes, and are now exposed in My Academics as well.

The front-end component looks much bigger than it actually is because our collection of utility functions for parsing the academics feed was moved from academicsController.js to the new academicsService.js, so that the booklist controller can use them as well.

### DIFF
--- a/src/assets/javascripts/angular/services/academicsService.js
+++ b/src/assets/javascripts/angular/services/academicsService.js
@@ -1,0 +1,227 @@
+/* jshint camelcase: false */
+(function(angular) {
+  'use strict';
+
+  angular.module('calcentral.services').service('academicsService', function() {
+    // Selects the semester of most pressing interest. Choose the current semester if available.
+    // Otherwise choose the next semester in the future, if available.
+    // Otherwise,choose the most recent semester.
+    var chooseDefaultSemester = function(semesters) {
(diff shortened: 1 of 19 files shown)


src/assets/javascripts/angular/services/academicsService.js
```diff
@@ -0,0 +1,227 @@
+/* jshint camelcase: false */
+(function(angular) {
+  'use strict';
```
christianv: @paulkerschen thanks for splitting this up!

christianv:
front-end lgtm

johncrossman:
Thanks for the readable specs @paulkerschen. Logic looks right. Let's see what the QA dept has to say.

PR #3597, opened 2015-04-13
CLC-4940, show tabs if videos.length AND audio.length

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-4940


christianv:
@johncrossman will this also automatically switch to audio when there are no videos?

johncrossman:
@christianv Yes, that's the behavior. For example, https://calcentral-qa.berkeley.edu/academics/semester/spring-2014/class/asamst-132ac

christianv:
@johncrossman thanks!

PR #3599, opened 2015-04-14
CLC-5071 - E-Grades Export: Update error display as per UX design recommendations

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-5071


paulkerschen:
Looks good to me.

PR #3602, opened 2015-04-14
CLC-5073 - E-Grades Export: Add muted assignments table with notice and link to error display

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-5073

### DIFF
--- a/src/assets/javascripts/angular/controllers/pages/canvasCourseGradeExportController.js
+++ b/src/assets/javascripts/angular/controllers/pages/canvasCourseGradeExportController.js
@@ -140,6 +140,7 @@
 
     /* Load and initialize application based on muted Assignments */
     var handleMutedAssignments = function(mutedAssignments) {
+      $scope.mutedAssignments = mutedAssignments;
       if (mutedAssignments.length > 0) {
         $scope.appState = 'error';
         $scope.mutedAssignmentsPresent = true;
--- a/src/assets/javascripts/angular/controllers/pages/canvasCourseGradeExportController.js
+++ b/src/assets/javascripts/angular/controllers/pages/canvasCourseGradeExportController.js
@@ -138,6 +138,14 @@
       }
     };
 
+    /* Load and initialize application based on muted Assignments */
+    var handleMutedAssignments = function(mutedAssignments) {
+      if (mutedAssignments.length > 0) {
+        $scope.appState = 'error';
+        $scope.mutedAssignmentsPresent = true;
+      }
+    };
+
     /* Load and initialize application based on official sections */
     var loadOfficialSections = function(officialSections) {
       if (officialSections && officialSections.length > 0) {
@@ -157,6 +165,7 @@
         }
         if ($scope.appState !== 'error') {
           handleGradingStandardState(data.gradingStandardEnabled);
+          handleMutedAssignments(data.mutedAssignments);
         }
         if ($scope.appState !== 'error') {
           loadOfficialSections(data.officialSections);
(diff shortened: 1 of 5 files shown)


src/assets/javascripts/angular/controllers/pages/canvasCourseGradeExportController.js
```diff
@@ -138,6 +138,15 @@
       }
     };
 
+    /* Load and initialize application based on muted Assignments */
+    var handleMutedAssignments = function(mutedAssignments) {
+      $scope.mutedAssignments = mutedAssignments;
+      if (mutedAssignments.length > 0) {
+        $scope.appState = 'error';
+        $scope.mutedAssignmentsPresent = true;
```
christianv: @redconfetti so we don't need an `else` clause where we set this to `false`?

src/assets/stylesheets/_canvas_course_grade_export.scss
```diff
@@ -38,8 +38,16 @@
   }
 
   .cc-page-course-grading-export-error-notice-container {
+    display: table;
```
christianv: @redconfetti interesting, why do you have to use this for the text? Would it be possible to provide an image on the jira with the before / after result?

src/assets/templates/canvas_embedded/course_grade_export.html
```diff
@@ -106,33 +106,100 @@ <h2 class="cc-page-course-grade-export-download-header">Final Grades Download</h
...
+            </p>
+
+            <h3 class="cc-page-course-grading-export-muted-assignments-table-header">
+              <span data-ng-pluralize count="mutedAssignments.length" when="{'one': '1 Muted Assignment', 'other': '{{mutedAssignments.length}} Muted Assignments'}"></span>
+            </h3>
+
+            <table role="grid">
+              <thead>
+                <tr>
+                  <th>Name</th>
```
christianv: @redconfetti can we add the `scope` parameter?

src/assets/stylesheets/_canvas_course_grade_export.scss
redconfetti: I searched online for a solution to make the text align in the middle vertically so that it would look nice next to the icon. This was the only one I found that worked in Chrome, and I hope it works in others. `vertical-align: middle` didn't work in a normal DIV.

image

src/assets/javascripts/angular/controllers/pages/canvasCourseGradeExportController.js
redconfetti: @christianv If you're using a variable in an ng-if that isn't defined, doesn't that just evaluate as not-true?

src/assets/templates/canvas_embedded/course_grade_export.html
redconfetti: @christianv Oh. I didn't even know that existed. Is that needed for accessibility?

src/assets/javascripts/angular/controllers/pages/canvasCourseGradeExportController.js
christianv: @redconfetti it does evaluate as [`falsy`](http://www.sitepoint.com/javascript-truthy-falsy/) indeed. Was just wondering whether `handleMutedAssignments` is only executed once or multiple times on the same page load.

src/assets/stylesheets/_canvas_course_grade_export.scss
christianv: @redconfetti would have done it with a `float:left` and setting a `margin-top` on the text but this is ok for now. I really hope soon we'll be moving to a flexbox way of doing things: http://caniuse.com/#search=flexbox

src/assets/templates/canvas_embedded/course_grade_export.html
christianv: @redconfetti indeed

PR #3604, opened 2015-04-15
CLC-5074 - E-Grades Export: Add option to enable grading scheme from error page

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-5074


christianv:
@redconfetti looks like we have a merge conflict now, would it be possible to rebase?

src/assets/templates/canvas_embedded/course_grade_export.html
```diff
@@ -123,16 +123,39 @@ <h1 class="cc-page-course-grade-export-header">Provide some extra information be
...
+            <button class="cc-button-link" data-ng-click="goToCourseDetails()">course details</button> page for this course site.
+            Once this step is completed, return here and try again.
           </p>
           <p>
             For more information see <a href="http://guides.instructure.com/m/4152/l/57089-how-do-i-enable-a-grading-scheme-for-my-course">How do I enable a grading scheme for my course?</a>
           </p>
 
+          <div>
+            <input type="checkbox"
+            data-ng-model="enableDefaultGradingScheme" aria-label="Checkbox for Enable Default Letter Grades +/- Grading Scheme"
```
christianv: @redconfetti can we just use `Enable and use default Letter Grades +/- grading scheme with E-Grades Download` as the label for this checkbox?

It's also not best practice to use the `type` of the `input` in the `aria-label`: `Checkbox for `

src/assets/templates/canvas_embedded/course_grade_export.html
```diff
@@ -123,16 +123,39 @@ <h1 class="cc-page-course-grade-export-header">Provide some extra information be
...
+            aria-checked="{{enableDefaultGradingScheme}}">
```
christianv: @redconfetti since this is a regular `checkbox` with only `true`/`false` as options, can we remove the `aria-checked` attribute? We should only do that for elements that aren't really a `checkbox`. E.g. when we put `role="checkbox"` on them.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_checkbox_role

src/assets/javascripts/angular/factories/canvasCourseGradeExportFactory.js
```diff
@@ -9,8 +9,11 @@
       return $http.get('/api/academics/canvas/egrade_export/options/' + canvasCourseId);
     };
 
-    var prepareGradesCacheJob = function(canvasCourseId) {
-      return $http.post('/api/academics/canvas/egrade_export/prepare/' + canvasCourseId);
+    var prepareGradesCacheJob = function(canvasCourseId, enableGradingScheme) {
+      enableGradingScheme = (typeof(enableGradingScheme) !== 'undefined') ? enableGradingScheme : false;
```
christianv: @redconfetti what are all the possible values of `enableGradingScheme`?

Does this work instead:

```javascript
enableGradingScheme = !!enableGradingScheme
```

src/assets/stylesheets/_canvas_course_grade_export.scss
```diff
@@ -78,4 +78,8 @@
   .cc-page-course-grade-export-notice-pending-request {
     margin: 15px auto;
   }
+
+  .cc-page-course-grade-export-error-continue-container {
```
christianv: @redconfetti can we just use `cc-text-right` instead and remove the class?

src/assets/templates/canvas_embedded/course_grade_export.html
redconfetti: Sure. I did that because I wasn't sure how it presents it to the user, but now that I think of it, it would make sense that the screen reader would represent it as a checkbox properly.

src/assets/javascripts/angular/factories/canvasCourseGradeExportFactory.js
redconfetti: @christianv It's a boolean value. Thanks for the tip. I did that in Ruby, but didn't think of doing that here. It was like 3 or 4 in the morning. :wink:

src/assets/templates/canvas_embedded/course_grade_export.html
```diff
@@ -125,19 +125,28 @@ <h2 class="cc-page-course-grade-export-sub-header">Set a Grading Scheme</h2>
...
+              <button class="cc-button-link" data-ng-click="goToCourseDetails()">course details</button> page for this course site.
+              Once this step is completed, return here and try again.
             </p>
             <p>
               For more information see <a href="http://guides.instructure.com/m/4152/l/57089-how-do-i-enable-a-grading-scheme-for-my-course">How do I enable a grading scheme for my course?</a>
             </p>
 
+            <div>
+              <input type="checkbox"
+              data-ng-model="enableDefaultGradingScheme" aria-label="Enable Default Letter Grades +/- Grading Scheme">
```
christianv: @redconfetti instead of using `aria-label`, can you use a regular `label`? We should only use `aria-` if there is no better way of doing it.

So create a `<label for="...`

src/assets/javascripts/angular/factories/canvasCourseGradeExportFactory.js
```diff
@@ -9,8 +9,11 @@
       return $http.get('/api/academics/canvas/egrade_export/options/' + canvasCourseId);
     };
 
-    var prepareGradesCacheJob = function(canvasCourseId) {
-      return $http.post('/api/academics/canvas/egrade_export/prepare/' + canvasCourseId);
+    var prepareGradesCacheJob = function(canvasCourseId, enableGradingScheme) {
+      enableGradingScheme = !!enableGradingScheme;
+      return $http.post('/api/academics/canvas/egrade_export/prepare/' + canvasCourseId, {
+        enableGradingScheme: enableGradingScheme
```
christianv: @redconfetti sorry, seems like we don't even need line 13 anymore, can we just use:

```javascript
enableGradingScheme: !!enableGradingScheme
```

christianv:
@redconfetti 2 more small comments, should be good to merge after they're resolved

src/assets/javascripts/angular/factories/canvasCourseGradeExportFactory.js
redconfetti: :+1:

PR #3609, opened 2015-04-16
CLC-5129, calcentral db: create table webcast_preferences

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5129

The opt_out column will support "no thank you" checkbox on Webcast widget. Down the road, this table might support the Webcast sign-up form, hosted by bCourses.  I will update 'release instructions' if / when this is merged.

### DIFF
--- a/db/migrate/20150415232131_create_webcast_preferences_table.rb
+++ b/db/migrate/20150415232131_create_webcast_preferences_table.rb
@@ -2,11 +2,11 @@ class CreateWebcastPreferencesTable < ActiveRecord::Migration
   def change
 
     create_table :webcast_preferences do |t|
-      t.integer :year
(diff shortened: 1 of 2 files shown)


ctweney:
@johncrossman DB migration is automagic so you shouldn't need to modify release instructions for this (unless there's a feature flag or something coming down the road?). 

One question: What about summer sub-terms? Is there a need for webcast opt-in on a summer-subterm basis? If so, the summer sub_term code needs to be part of the composite key.

johncrossman:
@ctweney Thanks for educating. Webcast does not (and does not plan to) do summer terms. Nonetheless, a good question.

paulkerschen:
Looks good from here.

PR #3612, opened 2015-04-16
CLC-5127 Reorganize instructor Class Info page for multiple primaries

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5127

Multiple-primary backend code is moved from MyAcademics::Semesters to MyAcademics::AcademicsModule so that MyAcademics::Teaching can use it as well.

textbookController.js is reworked along the lines of booklistController.js to handle concurrent textbook queries for multiple primary sections when necessary. Some common code is extracted to academicsService.js.

The Class Info template is modified to fit the redesign, and the Class Sites card extracted to its own template.

### DIFF
--- a/app/models/my_academics/academics_module.rb
+++ b/app/models/my_academics/academics_module.rb
@@ -153,6 +153,27 @@ module MyAcademics
       term << course_info
     end
 
+    def merge_multiple_primaries(course, course_option)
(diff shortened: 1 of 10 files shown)


johncrossman:
I see no problems. Over to QA dept...

PR #3613, opened 2015-04-16
CLC-5124 Let Timeshifter code work on the new Mockable proxy code

ctweney:
https://jira.ets.berkeley.edu/jira/browse/CLC-5124

NOTE: No Mockable proxy currently uses the timeshifting in ets/master, but there are 2 in the SIS branch that will need it.

### DIFF
--- a/lib/timeshifter.rb
+++ b/lib/timeshifter.rb
@@ -8,8 +8,6 @@ class Timeshifter
     str
   end
 
-  private
-
   def self.substitutions
     @substitutions ||= build_substitutions
   end
--- a/lib/timeshifter.rb
+++ b/lib/timeshifter.rb
@@ -1,0 +1,79 @@
+class Timeshifter
+
+  include ClassLogger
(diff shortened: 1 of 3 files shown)


lib/timeshifter.rb
```diff
@@ -0,0 +1,79 @@
...
+
+  include ClassLogger
+
+  # substitute tokens in string with formatted date values from #Timeshifter.substitutions
+  def self.process(str)
+    Timeshifter.substitutions.each { |k, v| str.gsub!(k, v) }
+    str
+  end
+
+  private
```
paulkerschen: @ctweney, any particular reason for `private` here? As far as I know it doesn't work on class methods.

ctweney: @paulkerschen Good catch, it was just there out of habit. Removed!

paulkerschen:
Thanks, @ctweney !

PR #3614, opened 2015-04-16
CLC-5066 - Misc bug fixes

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-5066

### DIFF
--- a/src/assets/templates/canvas_embedded/course_grade_export.html
+++ b/src/assets/templates/canvas_embedded/course_grade_export.html
@@ -82,7 +82,7 @@
     </div>
   </div>
 
-  <div data-ng-if="appState === 'error'">
+  <div data-ng-show="appState === 'error'">
 
     <!-- Standard Error -->
     <div data-ng-if="errorStatus">
@@ -95,7 +95,7 @@
     </div>
 
     <!-- No Grading Standard Enabled -->
-    <div data-ng-if="!errorStatus && (noGradingStandardEnabled || mutedAssignmentsPresent)">
+    <div data-ng-show="!errorStatus && (noGradingStandardEnabled || mutedAssignmentsPresent)">
 
       <div class="row collapse">
         <div class="medium-12 columns">
--- a/src/assets/templates/canvas_embedded/course_grade_export.html
+++ b/src/assets/templates/canvas_embedded/course_grade_export.html
@@ -215,7 +215,7 @@
       <div class="row collapse">
         <div class="medium-12 columns end">
           <div class="cc-text-right">
-            <button class="cc-button-link cc-page-button-grey" type="button" data-ng-click="goToGradebook()" type="button" aria-label="Go Back to Gradebook">Back</button>
+            <button class="cc-button-link cc-page-button-grey cc-page-course-grading-export-back-link" type="button" data-ng-click="goToGradebook()" type="button" aria-label="Go Back to Gradebook">Back</button>
             <button data-ng-click="preloadGrades()" data-ng-disabled="!enableDefaultGradingScheme" class="bc-canvas-button bc-canvas-button-primary" type="submit">Continue</button>
           </div>
         </div>
--- a/src/assets/templates/canvas_embedded/course_grade_export.html
+++ b/src/assets/templates/canvas_embedded/course_grade_export.html
@@ -110,7 +110,7 @@
         </div>
       </div>
 
-      <div data-ng-if="noGradingStandardEnabled" class="cc-page-course-grading-export-error">
+      <div data-ng-show="noGradingStandardEnabled" class="cc-page-course-grading-export-error">
         <div class="row collapse">
           <div class="medium-12 columns">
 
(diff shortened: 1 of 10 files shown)


johncrossman:
@redconfetti and I reviewed this together. What came up:
1. why must background id be created outside of job instance?
2. is ng-show used properly? Is there an accessibility concern?
3. code styles are not standard (e.g., no parens when single arg)
4. function naming should align with biz rules

Jason has agreed to investigate #1 and talk to Christian about accessibility concern. Other items will be fixed as we go.

redconfetti:
I added some logger statements in to recall what happened before when I would call #background_job_initialize from within the Canvas::Egrades#canvas_course_student_grades method.

These were the calls from the Controller method:
```ruby
    egrades_worker = Canvas::Egrades.new(:canvas_course_id => canvas_course_id, :enable_grading_scheme => !!params['enableGradingScheme'])
    # egrades_worker.background_job_initialize
    egrades_worker.background.canvas_course_student_grades(true)
    Rails.logger.warn("Post Background Job Call background_job_id: #{egrades_worker.background_job_id}")
    Rails.logger.warn("Post Background Job Call background_job_report: #{egrades_worker.background_job_report}")
```

I added the call back in the Canvas::Egrades#canvas_course_student_grades method that is run as a background job like so:

```ruby
    def canvas_course_student_grades(force = false)
      background_job_initialize
      Rails.logger.warn("Background Job background_job_id: #{background_job_id}")
      course_settings = Canvas::CourseSettings.new(:course_id => @canvas_course_id)
      # ...
    end
```

This is what I got in the log:
```
13:48:00,681 INFO  [stdout] (http-localhost/127.0.0.1:3000-1) [2015-04-16 13:48:00] [WARN] Post Background Job Call background_job_id:
13:48:00] [WARN] Background Job background_job_id: Canvas::Egrades.1429217280717-27a16ca525f2d518
```

So what's happening is that the background job might not be called, and thus the #background_job_initialize isn't called to set the background_job_id, and thus the controller returns no background_job_id because it hasn't been set. So it's essentially a race condition between the call to retrieve the background_job_id and return it to the front-end, and the background job itself initializing the background job state (which includes setting a unique background job ID).

redconfetti:
@johncrossman So it turns out that anything inside of an element inserts into the DOM via an [ng-if](https://docs.angularjs.org/api/ng/directive/ngIf) exists within a child scope. 

See [Angularjs ng-model doesn't work inside ng-if](http://stackoverflow.com/questions/18342917/angularjs-ng-model-doesnt-work-inside-ng-if)

I didn't know this, and that's why I converted the ng-if directives into ng-show directives in the template. I thought I tried to update the ng-model directive to use $parent.enableDefaultGradingScheme, but that didn't work.

```html
<input id="cc-page-course-grading-export-enable-grading-scheme" type="checkbox"
data-ng-model="$parent.enableDefaultGradingScheme" aria-label="Enable Default Letter Grades +/- Grading Scheme">
```

Since there are essentially 3 nested ng-if directives at use, I'll need to use $parent.$parent.$parent.enableDefaultGradingScheme. I tested this out and it worked. I'll create another PR now to fix this.

@christian, do you know if ng-show properly disables a DOM element so that it's not seen by screen readers or text based browsers? Do they register `display:none` as the element not shown?

redconfetti:
Follow up PR #3616

src/assets/templates/canvas_embedded/course_grade_export.html
```diff
@@ -119,37 +119,44 @@ <h1 class="cc-page-course-grade-export-header">Required adjustments before downl
...
+            <h2 class="cc-page-course-grade-export-sub-header">Set Grading Scheme</h2>
 
-            <p>
-              Select to enable and use the default 'Letter Grades +/-' grading scheme below and then click on 'Continue'.
+            <p>A grading scheme must be enabled to calculate student letter grades. You can choose to enable the
+            default grading scheme, or manually enable a custom grading scheme.</p>
+
+            <h3 class="cc-page-course-grade-export-error-grading-scheme-header">Default Grading Scheme</h3>
+
+            <p style="margin: 10px 0;">
```
christianv: @redconfetti can we please remove the inline styling here?

src/assets/templates/canvas_embedded/course_grade_export.html
```diff
@@ -119,37 +119,44 @@ <h1 class="cc-page-course-grade-export-header">Required adjustments before downl
...
+              Select to enable and use the default 'Letter Grades +/-' grading scheme below and then click on 'Continue' below.
             </p>
+            <div>
+              <input id="cc-page-course-grading-export-enable-grading-scheme" type="checkbox"
+              data-ng-model="enableDefaultGradingScheme" aria-label="Enable Default Letter Grades +/- Grading Scheme">
```
christianv: @redconfetti the `aria-label` is redundant, since we already have an associated `label`, can we remove it?

redconfetti:
@johncrossman I've resolved the other two updates that we discussed in #3618
- update the Javascript function name from goToCourseDetails to goToCourseSettings
- Add more verbose comments regarding the need to fake 30 total steps before passing the background job processing to the Canvas::CourseUsers#course_users method to retrieve grades

PR #3615, opened 2015-04-16
CLC-5130, active-record class for webcast_preferences table

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5130

Preferences class bound to db table 'webcast_preferences' with spec

### DIFF
--- a/app/models/webcast/preferences.rb
+++ b/app/models/webcast/preferences.rb
@@ -5,16 +5,5 @@ module Webcast
 
     attr_accessible :year, :term_cd, :ccn, :opt_out
 
-    def self.lookup(year, term_cd, ccn)
-      result_set = Preferences.limit(1).where(
-        'year = :year AND term_cd = :term_cd AND ccn = :ccn',
-        {
-          year: year,
-          term_cd: term_cd,
-          ccn: ccn
-        })
-      result_set[0]
-    end
-
   end
 end
--- a/app/models/webcast/preferences.rb
+++ b/app/models/webcast/preferences.rb
@@ -1,0 +1,20 @@
+module Webcast
+  class Preferences < ActiveRecord::Base
+
+    self.table_name = 'webcast_preferences'
+
+    attr_accessible :year, :term_cd, :ccn, :opt_out
+
+    def self.lookup(year, term_cd, ccn)
+      result_set = Preferences.limit(1).where(
+        'year = :year AND term_cd = :term_cd AND ccn = :ccn',
+        {
+          year: year,
+          term_cd: term_cd,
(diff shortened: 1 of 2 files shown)


app/models/webcast/preferences.rb
```diff
@@ -0,0 +1,20 @@
+module Webcast
+  class Preferences < ActiveRecord::Base
+
+    self.table_name = 'webcast_preferences'
+
+    attr_accessible :year, :term_cd, :ccn, :opt_out
```
paulkerschen: @johncrossman, `attr_accessible` shouldn't be needed here (and should probably be removed from our other models if it's still there); as of Rails 4, controllers and not models have the job of whitelisting parameters for assignment.

app/models/webcast/preferences.rb
```diff
@@ -0,0 +1,20 @@
...
+    def self.lookup(year, term_cd, ccn)
+      result_set = Preferences.limit(1).where(
+        'year = :year AND term_cd = :term_cd AND ccn = :ccn',
+        {
+          year: year,
+          term_cd: term_cd,
+          ccn: ccn
+        })
+      result_set[0]
+    end
```
paulkerschen: The usual way to do this would be a one-liner `Preferences.find_by` with your options hash; that will just return the first matching record without having to interpolate SQL, get an array, and extract the first element.

spec/models/webcast/preferences_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+
+    it 'should not find matching record' do
+      expect(Webcast::Preferences.lookup(2015, 'B', 1234)).to be_nil
+    end
+
+    it 'should return record with opt_out equal false' do
+      record = Webcast::Preferences.lookup(2015, 'D', 1234)
+      expect(record).to_not be_nil
+      expect(record.opt_out).to be true
+    end
```
paulkerschen: Likewise here, if you just use ActiveRecord's `find_by` instead of the `lookup` method, the tests should still pass.

app/models/webcast/preferences.rb
paulkerschen: @johncrossman , https://jira.ets.berkeley.edu/jira/browse/CLC-5152 has been created to investigate whether we still need `attr_accessible`. In the meantime I withdraw the comment above; if it turns out not to be needed, we can clean everything up at one go.

johncrossman:
@paulkerschen Thanks for the review. As you say, 'strong parameters' is preferred but only if compatible with RailsAdmin. TBD.  Latest commit swaps in find_by.

PR #3616, opened 2015-04-16
CLC-5066 - Added header to progress bar display. Applied fix to ng-if scope issue.

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-5066

### DIFF
--- a/src/assets/templates/canvas_embedded/course_grade_export.html
+++ b/src/assets/templates/canvas_embedded/course_grade_export.html
@@ -140,7 +140,7 @@
 
             <h3 class="cc-page-course-grade-export-error-grading-scheme-header">Default Grading Scheme</h3>
 
-            <p style="margin: 10px 0;">
+            <p class="cc-page-course-grading-export-error-notice">
               Select to enable and use the default 'Letter Grades +/-' grading scheme below and then click on 'Continue' below.
             </p>
             <div>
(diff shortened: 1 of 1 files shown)


src/assets/templates/canvas_embedded/course_grade_export.html
```diff
@@ -136,7 +145,7 @@ <h3 class="cc-page-course-grade-export-error-grading-scheme-header">Default Grad
             </p>
             <div>
               <input id="cc-page-course-grading-export-enable-grading-scheme" type="checkbox"
-              data-ng-model="enableDefaultGradingScheme" aria-label="Enable Default Letter Grades +/- Grading Scheme">
+              data-ng-model="$parent.$parent.$parent.enableDefaultGradingScheme" aria-label="Enable Default Letter Grades +/- Grading Scheme">
```
christianv: @redconfetti this is some very interesting code... Is there another way to solve this without having to go to the great-grandparents?

christianv:
As mentioned to @redconfetti in IM, we should use `ng-if` as much as possible, but if it doesn't work, use `ng-show`. We shouldn't use ugly hacks like trying to get the great-grandparents `$scope`. /cc @ctweney @raydavis @paulkerschen

christianv:
Please don't merge until after another push from @redconfetti

PR #3617, opened 2015-04-16
CLC-5126 - View-as doesn't work on calcentral-dev

christianv:
https://jira.ets.berkeley.edu/jira/browse/CLC-5126


ctweney:
The fix looks right to me: "Do this AND THEN do that" as opposed to "Do this and that". Just waiting on T.

PR #3619, opened 2015-04-16
CLC-5151: fix webcast Selenium test to expect new UI

pfarestveit:
https://jira.ets.berkeley.edu/jira/browse/CLC-5151

Test expectations updated per https://jira.ets.berkeley.edu/jira/browse/CLC-4940.


johncrossman:
Service reduction in effect.

PR #3621, opened 2015-04-17
CLC-5078 - E-Grades Export: Add refresh grades option to download display

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-5078

### DIFF
--- a/src/assets/templates/canvas_embedded/course_grade_export.html
+++ b/src/assets/templates/canvas_embedded/course_grade_export.html
@@ -81,6 +81,13 @@
         <p class="cc-page-course-grade-export-more-info-container">
           For more information, see <a href="http://ets.berkeley.edu/bcourses/how-do-i-submit-grades-e-grades-bcourses">How do I Submit Grades to E-Grades from bCourses?</a>
         </p>
+        <p class="cc-page-course-grade-export-more-info-container">
+          Updated Gradebook?
+          <button class="bc-canvas-button cc-page-course-grade-export-refresh-button" data-ng-click="$parent.preloadGrades()">
+            <i class="fa fa-refresh"></i>
+            Refresh Prepared Grades
+          </button>
+        </p>
       </div>
     </div>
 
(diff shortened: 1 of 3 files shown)


src/assets/templates/canvas_embedded/course_grade_export.html
```diff
@@ -81,6 +81,13 @@ <h2 class="cc-page-course-grade-export-download-header">Final Grades Download</h
         <p class="cc-page-course-grade-export-more-info-container">
           For more information, see <a href="http://ets.berkeley.edu/bcourses/how-do-i-submit-grades-e-grades-bcourses">How do I Submit Grades to E-Grades from bCourses?</a>
         </p>
+        <p class="cc-page-course-grade-export-more-info-container">
+          Updated Gradebook?
+          <button class="bc-canvas-button cc-page-course-grade-export-refresh-button" data-ng-click="$parent.preloadGrades()">
```
christianv: @redconfetti Does it still work without the $parent?

redconfetti: @christianv Nope. It looks like I did that just because I expected the ng-if to block again. So does the ng-if simply introduce some sort of child scope for variables, but not for functions?

christianv: @redconfetti can't really explain it better than the current angular docs:

> Note that when an element is removed using ngIf its scope is destroyed and a new scope is created when the element is restored. The scope created within ngIf inherits from its parent scope using prototypal inheritance. An important implication of this is if ngModel is used within ngIf to bind to a javascript primitive defined in the parent scope. In this case any modifications made to the variable within the child scope will override (hide) the value in the parent scope.

https://docs.angularjs.org/api/ng/directive/ngIf

christianv: Also, try to avoid `$parent` as much as possible.

PR #3626, opened 2015-04-17
CLC-5079 - E-Grades Export: Force refresh after 35 minute timeout

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-5079


johncrossman:
Looks ok to me. @christianv - no concerns?

redconfetti:
Mecha-Christian (jshint, jscs, scsshint) approves....

src/assets/javascripts/angular/controllers/pages/canvasCourseGradeExportController.js
```diff
@@ -187,6 +196,16 @@
...
     };
 
+    /*
+     * Forces refresh of prepared grades 34 minutes after called
+     */
+    var cacheTimeoutPromise;
+    var refreshExpiredCache = function() {
+      cacheTimeoutPromise = $timeout(function() {
+        $scope.preloadGrades();
+      }, 2040000);
```
christianv: @redconfetti this will automatically reload the information without letting the user know. Even though it's only after a long time, we should let them know before changing the data.

christianv: @redconfetti A solution might be to show a timer on the page that something will be happening in `xx` amount of minutes.

PR #3629, opened 2015-04-17
CLC-5104 Serve root from bootstrap.html to force processing by BootstrapController

ctweney:
https://jira.ets.berkeley.edu/jira/browse/CLC-5104

@christianv should review this for sure. 

Please wait for green Bamboo & Travis, and don't merge until after QA branches.


christianv:
@ctweney is there a way to call the `BootstrapController` on `index.html` instead? Or perhaps renaming it to `IndexController`?

ctweney:
No, the index.html filename in /public is magic for Rails. When it sees that, it serves it through the static file server, bypassing the Rails controller stack entirely.

christianv:
@ctweney very interesting, and it doesn't seem like there is a good way to change that Rails magic...

http://stackoverflow.com/questions/1080747/how-do-you-get-a-rails-project-to-start-with-index-html-erb-instead-of-index-htm
http://stackoverflow.com/questions/18485429/rails-use-ruby-in-public-index
http://stackoverflow.com/questions/5174160/routing-around-static-index-html-in-rails

The main reason for asking this is because `index.html` is usually the default, main HTML file for any front-end project. Was also thinking about using something like `index-main.html` to get around this.

Would you be ok with `index-main.html` and explaining why we can't use `index.html` somewhere?

ctweney:
Fair enough, @christianv , I renamed the file to index-main.html and put in explanatory comments.

christianv:
@ctweney really appreciate it, thank you. Have a great weekend.

PR #3639, opened 2015-04-21
CLC-5061, webcast-signup.html wiring prior to adding logic

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5061
Bamboo, too: https://bamboo.ets.berkeley.edu/bamboo/browse/MYB-CJT49-1

The logic will come next. For example:
1. visibility def in erb file
2. render sign_up widget according to instructor type, ccn(s)
3. etc.

### DIFF
--- a/src/assets/javascripts/angular/controllers/widgets/webcastSignUpController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/webcastSignUpController.js
@@ -1,0 +1,11 @@
+(function(angular) {
+  'use strict';
+
+  /**
+   * Webcast sign-up controller
+   */
+  angular.module('calcentral.controllers').controller('WebcastSignUpController', function(apiService, webcastFactory, $route, $routeParams, $scope) {
+
+
+  });
+})(window.angular);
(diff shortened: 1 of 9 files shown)


src/assets/javascripts/angular/controllers/widgets/webcastSignUpController.js
```diff
@@ -0,0 +1,11 @@
+(function(angular) {
+  'use strict';
+
+  /**
+   * Webcast sign-up controller
+   */
+  angular.module('calcentral.controllers').controller('WebcastSignUpController', function(apiService, webcastFactory, $route, $routeParams, $scope) {
+
```
christianv: @johncrossman can we remove the arguments `apiService, webcastFactory, $route, $routeParams, $scope` for now since we're not using them?

johncrossman: done

johncrossman:
Fixing jscs problem.

PR #3644, opened 2015-04-21
CLC-5177 Mock request matcher can now handle ports or no ports

ctweney:
https://jira.ets.berkeley.edu/jira/browse/CLC-5177

The use case for this only exists in the sis-master branch at the moment, but I think it's good to get into master.


raydavis:
Agreed, @ctweney -- thanks!

PR #3647, opened 2015-04-22
CLC-5184 Move Advising::MyAdvising to Mockable framework

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5184

### DIFF
--- a/app/models/advising/my_advising.rb
+++ b/app/models/advising/my_advising.rb
@@ -3,6 +3,7 @@ module Advising
     include Cache::LiveUpdatesEnabled
     include Cache::FeedExceptionsHandled
     include HttpRequester
+    include Proxies::Mockable
     include SafeJsonParser
     include User::Student
     include ClassLogger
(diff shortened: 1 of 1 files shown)


app/models/advising/my_advising.rb
```diff
@@ -33,32 +35,24 @@ def get_parsed_response
...
-        )
-        status_code = response.code
-        if status_code == 404
-          logger.debug "404 response from advising API for user #{@uid}"
-          parsed_json = {'body' => 'No advising data could be found for your account.'}
-        else
-          unless (parsed_json = safe_json(response.body))
-            raise Errors::ProxyError.new('Empty response', response: response, url: url, uid: uid)
-          end
+        unless (parsed_json = safe_json(response.body))
```
raydavis: Although this may be scope creep, this sort of "safe_json(response.body)" call should (eventually) be replaced by "response.parsed_response" (as you already did for XML response bodies).

paulkerschen: @raydavis Good point; how about I open up another issue for this and other needless JSON parsings around the code base?

raydavis: Sounds fine, @paulkerschen . I'll wait to merge this for a bit, though, so as to prioritize testing of the potential last fix for the release.

PR #3653, opened 2015-04-22
CLC-5186 Move BearFacts proxies to Mockable framework

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5186

VCR recordings are replaced with human-readable XML fixtures.

Because file paths are set dynamically based on student ID, the Mockable read_file method gets an existence check.

Note that variables affecting WebMock matching (here, student ID and Telebears term ID), along with any associated stubbing in the specs, have to be set before the mocks are initialized. It's also possible to override the mocking behavior after initialization, as in the Regstatus::Proxy spec, but that involves setting up a different sort of test.


raydavis:
Nice to get those XML payloads unpacked, @paulkerschen !

PR #3655, opened 2015-04-23
CLC-5190 Fix production UI test

ctweney:
https://jira.ets.berkeley.edu/jira/browse/CLC-5190

https://bamboo.ets.berkeley.edu/bamboo/browse/MYB-CPUT-JOB1-213/log failed, I believe because "bundle install" ran before the RAILS_ENV was set, and because the spec:xml rake task isn't defined when RAILS_ENV=production. This fixes both issues.


johncrossman:
Looks good. Thanks, Chris

ctweney:
This does not fix the issue. I didn't close it in time before John merged -- I'll issue a revert PR shortly.

PR #3659, opened 2015-04-23
CLC-5195 Move CalLink proxies to Mockable; refine Mockable's grain

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5195 and ongoing tweaks to Proxies::Mockable.

URI parsing has been moved out of the Mockable module and into the MockHttpInteraction class. Proxies that need to specify their own URIs can now add a `uri_matching` parameter to `mock_request` and get it automatically parsed the way WebMock wants it. (The change to Bearfacts::Proxy in this PR is an example.)

Along the same lines, a `query_including` parameter now takes advantage of WebMock's convenient `hash_including` to do partial matches on request parameters. In this PR it's used to ensure that a fake CalLink::Organization proxy will only stub out requests that include its own org ID.


ctweney:
I foresaw the need for query param matching when I tweaked that code, but my use case didn't need it. I'm glad someone else added the functionality. LGTM!

PR #3664, opened 2015-04-23
CLC-5206 - Use data-ng everywhere

christianv:
https://jira.ets.berkeley.edu/jira/browse/CLC-5206


paulkerschen:
Travis the First is happy with this, and so am I.

PR #3665, opened 2015-04-23
CLC-5196 Move CFV Financials to Mockable framework

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5196

Tricks in the mock response: a custom header with the API version, and a dynamic 404 response using the same substitution convention as Timeshifter. 

financials_not_found.json is not valid JSON before the student ID is substituted in, but I am hopeful that it's not too great a sin.


christianv:
Note that 1 travis is already happy.

raydavis:
Thanks for teaching the new dog old tricks, @paulkerschen

PR #3666, opened 2015-04-23
CLC-5120: fix view-as Selenium tests

pfarestveit:
https://jira.ets.berkeley.edu/jira/browse/CLC-5120

These fixes adapt the existing view-as spec to the new UI.  There are also a few syntax changes in the contexts to improve readability of the examples.


johncrossman:
Looks good. Waiting on Trav

PR #3667, opened 2015-04-23
CLC-5205 - Create the cc-table-finances CSS class

christianv:
https://jira.ets.berkeley.edu/jira/browse/CLC-5205


christianv:
Follow up from #3663

PR #3669, opened 2015-04-24
CLC-5198 Follow-up: fix failing test

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5198


raydavis:
Thanks for the quick fix, @paulkerschen

PR #3671, opened 2015-04-27
CLC-5170 First two CalMail API proxies

raydavis:

I've updated Bamboo and Dev configurations to enable testing.

### DIFF
--- a/app/models/calmail/check_namespace.rb
+++ b/app/models/calmail/check_namespace.rb
@@ -1,0 +1,45 @@
+module Calmail
+  class CheckNamespace < Proxy
+    include ClassLogger
+    include ResponseWrapper
+
+    MAILING_LIST_EXISTS = 'Invalid: localpart: A mailing list by that name already exists'
+
+    def name_available?(list_name)
+      handling_exceptions(list_name) do
+        response = check_namespace(list_name)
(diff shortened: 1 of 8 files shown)


paulkerschen:
It all looks right to me, and I'm glad to learn about HTTParty's parser option.

app/models/calmail/check_namespace.rb
```diff
@@ -0,0 +1,45 @@
...
+    include ResponseWrapper
+
+    MAILING_LIST_EXISTS = 'Invalid: localpart: A mailing list by that name already exists'
+
+    def name_available?(list_name)
+      handling_exceptions(list_name) do
+        response = check_namespace(list_name)
+        if response.code == 200 && response.parsed_response['available']
+          true
+        elsif response.code == 500 && response.parsed_response['message'] == MAILING_LIST_EXISTS
```
johncrossman: @raydavis should code=503 get same treatment as 500? A lesson learned from Google api responses. In same spirit, might we get valid 3xx code (redirection)?

raydavis: @johncrossman , this is an exceptional (i.e., no exceptions) case, since the CalMail API uses 500 to signal success.

PR #3673, opened 2015-04-27
CLC-4014 Remove Research Hub support

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-4014

This should just do what it says, but I wouldn't mind a quick walkthrough with @johncrossman or @raydavis, if available, on our workflow for release instructions and config changes.


src/assets/stylesheets/_icons.scss
```diff
@@ -48,9 +48,6 @@
 .cc-icon-oskicat {
   background-image: url('../images/icon_oskicat_16x16.png');
 }
-.cc-icon-researchhub {
-  background-image: url('../images/icon_research_16x16.png');
```
christianv: @paulkerschen can we also remove the actual icon (if it's not used anywhere else)

paulkerschen: @christianv Oh, yes, thanks for catching that.

raydavis:
Thanks for thinking of that, @paulkerschen . I generally edit the configurations for Bamboo and for Dev, and then edit the Release-Specific Instructions to match.

Of course, my attempt from this morning didn't work so well! :)

raydavis:
Oh, and in the specific case of _removing_ properties, we sometimes delay removal for a release so as to enable rollbacks, builds of unmerged branches, etc. That may be most appropriate in this case.

PR #3679, opened 2015-04-28
CLC-5170 Lock down mocked tests; disable integration tests

raydavis:

We're going to live unhappily with some mystery for a while...


johncrossman:
I see no problems here. Congrats on finally getting a happy Bamboo.

PR #3682, opened 2015-04-29
CLC-5202 First two CalGroups proxies

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5202
https://jira.ets.berkeley.edu/jira/browse/CLC-5180

NB: New cal_groups_proxy local settings have been added to calcentral-dev. These should be copied over to ets-dev prior to merge, or Bamboo will fail.

Otherwise, trying to hew close to the logic in #3671 for consistency's sake. Like those proxies, these don't currently contemplate caching, but the calls could take a cache wrapper straightforwardly enough.

### DIFF
--- a/app/models/cal_groups/find_groups.rb
+++ b/app/models/cal_groups/find_groups.rb
@@ -1,0 +1,44 @@
+module CalGroups
+  class FindGroups < Proxy
+    include ClassLogger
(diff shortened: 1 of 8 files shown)


raydavis:
Works for me!

PR #3683, opened 2015-04-29
CLC-5170 Add all CalMail API calls needed for Mailman mailing lists

raydavis:

I believe these are all the APIs of use to the Mailman integration as currently scoped out.


johncrossman:
@raydavis Travis fails with "torquebox-server-3.1.2-java.gem may be corrupted". Transient problem or must we refresh the gem somehow?

raydavis:
Looks transient, @johncrossman.

paulkerschen:
I'll take it!

PR #3687, opened 2015-04-30
CLC-5216, Webcast::Merged used by canvas and calCentral controllers (backwards compat)

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5216
Bamboo, too: https://bamboo.ets.berkeley.edu/bamboo/browse/MYB-CJT50-8

Previously, the underlying Webcast code would expect dept_name and catalog_id then return the matching set of recordings from the feed. Now, we (1) query for all course sections that relate to dept_name and catalog_id, (2) pass the list of CCNs to webcast code and (3) return multiple sets of recordings. In the case of Canvas integration, we use the set of sis_section_ids (associated with a course site) to generate the necessary list of CCNs.

For backwards compatibility, we still support :videos, :audio and :itunes in the feed. They will be removed once the front-end is using :media property exclusively.

More test coverage, too. Next, we focus on front-end consumer code.


spec/models/webcast/merged_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+        expect(feed[:rooms]['VALLEY LSB']).to contain_exactly('2040', '2050', '2060')
+        course = feed[:media]['2014-B-1']
+        expect(course).to eq Webcast::Recordings::ERRORS
+        expect(course[:videos]).to be_nil
+        expect(course[:audio]).to be_nil
+        # Verify backwards compatibility
+        expect(feed[:videos]).to be_empty
+        expect(feed[:audio]).to be_empty
+        expect(feed[:itunes]['audio']).to be_nil
+        expect(feed[:itunes]['audio']).to be_nil
```
paulkerschen: @johncrossman Dupe expectation.

app/models/webcast/course_media.rb
```diff
@@ -1,19 +1,18 @@
 module Webcast
   class CourseMedia
 
-    def self.course_id(year, term, dept, catalog_id)
-      # allow lookups by either term_cd or term name
+    def self.id_per_ccn(year, term, ccn)
+      term.to_s.strip!
```
paulkerschen: @johncrossman I don't think the stripping would persist if `term` is not a string (`term.to_s` would create a new string object, and that string would be stripped, but it's not assigned to a variable.) Are you expecting non-string term argument to come in, such that `to_s` is needed?

app/controllers/mediacasts_controller.rb
```diff
@@ -2,11 +2,20 @@ class MediacastsController < ApplicationController
...
-    render :json => Webcast::CourseMedia.new(
-      params['year'], params['term_code'], params['dept'], params['catalog_id']
-    ).get_feed
+    term_yr = params['year']
+    term_cd = params['term_code']
+    dept_name = params['dept']
+    catalog_id = params['catalog_id']
+    ccn_list = []
+    sections = CampusOracle::Queries.get_all_course_sections(term_yr, term_cd, dept_name, catalog_id)
+    sections.each { |section| ccn_list << section['course_cntl_num'].to_i } if sections.any?
```
paulkerschen: @johncrossman FYI you can call `map` instead of `each` as a convenient idiom for building a new array with a common transformation.

spec/controllers/mediacasts_controller_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+
+      context 'when no Webcast recordings found' do
+        it 'should pay attention to term code' do
+          json = post_course chem_101
+          expect(json['media']).to have(3).items
+          %w(1 2 3).each do |ccn|
+            expect(json['media']["2014-B-#{ccn}"]['video_error_message']).to eq 'No Webcast data found.'
+          end
+        end
+      end
```
paulkerschen: @johncrossman I think it would be usual to wrap these two examples in one context instead of having two identical contexts?

paulkerschen:
I picked a few nits but this looks like good handling of knotty logic to me.

spec/controllers/mediacasts_controller_spec.rb
```diff
@@ -0,0 +1,76 @@
+describe MediacastsController do
+
+  before { session['user_id'] = rand(99999).to_s }
+
+  # Single digit CCNs have no webcast recordings
```
raydavis: That's an interesting quirk I didn't know about. Is it documented anywhere else? If not, it might be worth a bit more explanation for future devs.

app/models/webcast/course_media.rb
johncrossman: @paulkerschen I believe it does work, although counter-intuitive for reasons you give. Try with irb and tell me what you think. This does not jive with my Java-based sensibilities. Nonetheless, I'll change the line to erase all doubt.

paulkerschen: ` % irb
 :001 > term = 666
 => 666 
 :002 > term.to_s.strip!
 => nil 
 :003 > term
 => 666`

So `term` remains an integer.

spec/controllers/mediacasts_controller_spec.rb
johncrossman: Doh!

spec/controllers/mediacasts_controller_spec.rb
johncrossman: Thanks @raydavis  A reference to test data. I reworded.  @paulkerschen Fixes are in. Thx for the review.

PR #3688, opened 2015-04-30
CLC-5238 Remove call on sometimes-undefined WebMock

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5238


ctweney:
This was my mistake. As long as Travis & Bamboo are green it looks good to me!

PR #3691, opened 2015-04-30
NOJIRA Update google-api-client gem

raydavis:
We have a history of frustrations around Google token handling (see CLC-4277). We're a couple of versions behind on the gem which provides it, so we might well try an upgrade.


johncrossman:
Travis happy. Looks good.

christianv:
@raydavis shouldn't we also update the actual `Gemfile` and not just the `Gemfile.lock`?

raydavis:
Good point, @christianv . I'll do a follow-up PR -- assuming we don't have to revert this one!

PR #3692, opened 2015-04-30
CLC-5202 Add proxies for group save and delete

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5202
https://jira.ets.berkeley.edu/jira/browse/CLC-5239

Also factoring some of the response checks out of CalGroups::Proxy and into subclasses for finer grain.

### DIFF
--- a/app/models/cal_groups/find_groups.rb
+++ b/app/models/cal_groups/find_groups.rb
@@ -12,17 +12,17 @@ module CalGroups
 
     def find_group_by_name(qualified_name)
       response = request({
-          body: {
-            wsLiteObjectType: 'WsRestFindGroupsLiteRequest',
-            queryFilterType: 'FIND_BY_GROUP_NAME_EXACT',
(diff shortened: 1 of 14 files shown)


app/models/cal_groups/find_groups.rb
```diff
@@ -12,17 +12,17 @@ def name_available?(name)
 
     def find_group_by_name(qualified_name)
       response = request({
-          body: {
-            wsLiteObjectType: 'WsRestFindGroupsLiteRequest',
```
raydavis: Looks familiar. :) I've reported the RubyMine indentation bug at https://youtrack.jetbrains.com/issue/RUBY-16639 if anyone wants to upvote it.

raydavis:
Lookin' good!

raydavis:
(Well, as good as that API can look, anyway.)

PR #3693, opened 2015-04-30
CLC-5046 Update developer-seed-data based on production dump from April ...

ctweney:
...30 2015

https://jira.ets.berkeley.edu/jira/browse/CLC-5046


raydavis:
I plan to do another one of these later in the sprint, but nothing wrong with getting an early update in!

PR #3695, opened 2015-04-30
CLC-5136, rake task and shell scrpt in support of Webcast cron job

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5136

The rake task will be implemented per https://jira.ets.berkeley.edu/jira/browse/CLC-5242.

### DIFF
--- a/script/webcast.sh
+++ b/script/webcast.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+# Script to manage Webcast LTI tool settings in Canvas. This is intended to be run from cron.
+
+# Make sure the normal shell environment is in place, since it may not be when running as a cron job.
+source "${HOME}/.bash_profile"
+
+cd $( dirname "${BASH_SOURCE[0]}" )/..
+
+LOG=`date +"${PWD}/log/webcast_%Y-%m-%d.log"`
+LOGIT="tee -a ${LOG}"
+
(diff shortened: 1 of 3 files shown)


raydavis:
Travis is half full.

script/webcast.sh
```diff
@@ -0,0 +1,27 @@
...
+[[ -s "${HOME}/.rvm/scripts/rvm" ]] && . "${HOME}/.rvm/scripts/rvm"
+source .rvmrc
+
+export RAILS_ENV=production
+export LOGGER_STDOUT=only
+export LOGGER_LEVEL=INFO
+
+echo | $LOGIT
+echo "------------------------------------------" | $LOGIT
+echo "`date`: Webcast cron job started on app node: `hostname -s`..." | $LOGIT
```
raydavis: "Webcast cron job" doesn't describe what the script is doing. How about "Webcast LTI app visibility checks" or something like that?

raydavis:
Another naming request, @johncrossman -- "webcasts.sh" seems like too general a filename to convey at a glance what the script is for. We have a lot of script files, so it's probably worth being more specific so that we can quickly spot the one which checks to see if bCourses webcast apps should be unhidden.

johncrossman:
Thanks @raydavis  I've done some renaming

raydavis:
Thank you, @johncrossman !

PR #3696, opened 2015-04-30
CLC-5046 Remove lock_timeout setting, which only works in Postgres > 9.3

ctweney:
https://jira.ets.berkeley.edu/jira/browse/CLC-5046

This followup should fix the build on Bamboo.


raydavis:
Thanks, @ctweney . I'll create a ticket to update ets-dev PostgreSQL.

raydavis:
Or not. :)  Just saw CS-4515.

PR #3700, opened 2015-05-01
CLC-5135, revise calcentral db table: webcast_preferences

johncrossman:

Drop webcast_preferences table in favor of the new. Update spec, etc.


paulkerschen:
As frequently cranky as Travis looks these days, it is half happy.

raydavis:
Much more communicative, @johncrossman , thanks!

ctweney:
There's a problem with this migration. I ran rake db:migrate, then tried to revert it with "rake db:migrate VERSION=20150416164519" and got:

== 20150430222956 AlterWebcastPreferencesTable: reverting =====================
rake aborted!
An error has occurred, this and all later migrations canceled:

remove_index is only reversible if given a :column option.

PR #3702, opened 2015-05-01
CLC-5135, remove_index only reversible if given a :column option

johncrossman:

Index was on a composite key so remove_index per three columns. No new migrate file here.


ctweney:
looks right to me, thanks for fixing!

PR #3703, opened 2015-05-01
CLC-5244 - E-Grades Export 2: Apply error resolution actions from Error view when clicking 'Continue'

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-5244


redconfetti:
Do not merge yet. I forgot to implement the front-end animations required.

redconfetti:
@christianv Would you like to review the front-end ?

app/models/canvas/egrades.rb
```diff
@@ -11,16 +11,9 @@ class Egrades
...
     def initialize(options = {})
-      default_options = {
-        :enable_grading_scheme => false,
-        :unmute_assignments => false
-      }
-      options.reverse_merge!(default_options)
-
       raise RuntimeError, "canvas_course_id required" unless options.include?(:canvas_course_id)
       @canvas_course_id = options[:canvas_course_id]
       @enable_grading_scheme = options[:enable_grading_scheme]
```
redconfetti: I just noticed that this is left over but not used.

src/assets/templates/canvas_embedded/course_grade_export.html
```diff
@@ -214,7 +214,10 @@ <h2 class="cc-page-course-grade-export-sub-header">Set Grading Scheme</h2>
         <div class="medium-12 columns">
           <div class="cc-text-right">
             <button class="bc-canvas-button cc-page-course-grading-export-back-link" type="button" data-ng-click="goToGradebook()" type="button" aria-label="Go Back to Gradebook">Cancel</button>
-            <button data-ng-click="switchToSelection()" data-ng-disabled="notReadyForPreparation()" class="bc-canvas-button bc-canvas-button-primary" type="button">Continue</button>
+            <button data-ng-disabled="notReadyForPreparation() || resolvingCourseState" data-ng-click="resolveIssues()" class="bc-canvas-button bc-canvas-button-primary" type="submit">
```
christianv: @redconfetti since this doesn't look like an actual `form` `submit`, would it be possible to remove `type="submit"`?

src/assets/templates/canvas_embedded/course_grade_export.html
```diff
@@ -214,7 +214,10 @@ <h2 class="cc-page-course-grade-export-sub-header">Set Grading Scheme</h2>
...
+              <span data-ng-if="!resolvingCourseState">Continue</span>
+              <span data-ng-if="resolvingCourseState"><i class="fa fa-spinner fa-spin"></i> Saving</span>
```
christianv: @redconfetti nice that we're showing progress to the user this way.

christianv:
@redconfetti 1 small comment on the front-end. Other front-end parts look good to me.

src/assets/templates/canvas_embedded/course_grade_export.html
redconfetti: @christianv I actually changed it to a submit on purpose, thinking that this would help register what looks like a form (two checkboxes and a submit button) regarding web accessibility. Do we not use submit because technically it's not allowing the browsers typical submission actions, but instead is being overridden by AngularJS actions?

raydavis:
Server-side's looking nice, @redconfetti, but I'll wait on @christianv 's answer to your question.

src/assets/templates/canvas_embedded/course_grade_export.html
christianv: @redconfetti since it's not a real form submit, we shouldn't use `type="submit"`. If we really want to make this more accessible we should

1. Wrap this functionality in a form
2. Use `data-ng-submit` on the form
3. Remove the `data-ng-click` event from the button
4. Use `type="submit"` on the button

Not sure whether this is in scope for the ticket though.

redconfetti: @christianv :+1:

PR #3706, opened 2015-05-01
CLC-5242, db migrate - drop table, etc. - must be reversible

johncrossman:

Painful exercise in understanding reversible db migrations.

### DIFF
--- a/db/migrate/20150430222956_alter_webcast_preferences_table.rb
+++ b/db/migrate/20150430222956_alter_webcast_preferences_table.rb
@@ -1,19 +1,29 @@
 class AlterWebcastPreferencesTable < ActiveRecord::Migration
-  def change
 
-    remove_index(:webcast_preferences, {name: 'webcast_preferences_unique_index'})
-    remove_index(:webcast_preferences, :year)
-    remove_index(:webcast_preferences, :term_cd)
-    remove_index(:webcast_preferences, :ccn)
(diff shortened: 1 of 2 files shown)


ctweney:
I tested by checking out ets/qa and doing "rake db:schema:load db:seed" against a brand-new database. Then, I checked out this branch and did "rake db:migrate" followed by a downgrade "rake db:migrate VERSION=20150416164519". Both migrations up and down worked, but it still wants to modify schema.rb like so, which is fishy: 

```

   create_table "webcast_course_site_log", force: true do |t|
-    t.integer  "canvas_course_site_id",     null: false
-    t.datetime "webcast_tool_unhidden_at",  null: false
-    t.datetime "created_at",                null: false
-    t.datetime "updated_at",                null: false
+    t.integer  "canvas_course_site_id",    null: false
+    t.datetime "webcast_tool_unhidden_at", null: false
+    t.datetime "created_at",               null: false
+    t.datetime "updated_at",               null: false
   end
```

That looks like just a whitespace change.

johncrossman:
@ctweney  I removed the space that would otherwise be removed by 'rake db:migrate'. Expect no more diffs as above.

ctweney:
@johncrossman Confirmed that the whitespace change doesn't happen anymore. The migrate up-down-up sequence works! Thanks for fixing this!

PR #3710, opened 2015-05-04
CLC-5234 Remove obsolete 'token' from MyFinAid proxy

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5234

Finaid tests pass on Bamboo with token removed from code.


johncrossman:
Looks good. Verified no other instance of myfinaid_proxy.token in ymls. Waiting on T

PR #3713, opened 2015-05-04
CLC-5158: fix enrollments Selenium spec

pfarestveit:
Test updated to include coverage for the following:
* Multiple primary sections:
  * On semester cards, appending section codes to course codes under the right circumstances
  * Offering multiple class pages for multiple primary sections, also under the right circumstances
* EAP and additional credit
  * Verifying the content of cards with transcript data but no enrollment data

The test changes have been verified with the full set of ~180 test user accounts.


johncrossman:
Looks berry nice.

PR #3714, opened 2015-05-04
CLC-5253: misc fixes for Selenium tests

pfarestveit:
https://jira.ets.berkeley.edu/jira/browse/CLC-5253

Includes the following changes:
* New util method for quitting the browser, which tolerates errors:  this is to overcome situations where the browser hasn't instantiated in the first place.  The subsequent NoMethodError kills the remaining test suite.  Also renamed the command to launch the browser, to better distinguish it from the quit method.
* A hardcoded pause after quitting the browser: this is another experiment to avoid the RuntimeError occasionally encountered when launching the browser.  When one test quits and another launches, the browser instances might be tripping over one another.  In any case, these sporadic errors cause regular build failures.
* Fixed a broken XPath for the finances details test.

### DIFF
--- a/spec/ui_selenium/my_academics_data_advising_appts_spec.rb
+++ b/spec/ui_selenium/my_academics_data_advising_appts_spec.rb
@@ -19,7 +19,7 @@ describe 'My Academics L&S Advising card', :testui => true do
     include ClassLogger
 
     begin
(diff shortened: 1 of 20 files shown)


paulkerschen:
Ladies and gents, the Automator.

PR #3716, opened 2015-05-05
CLC-5211 New DB table for bCourses site mailing lists

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-5211
https://jira.ets.berkeley.edu/jira/browse/CLC-5212

This is a dependency for both the admin app side and the membership refresh script side of bCourses site mailing lists. Once it's in place, either side should be doable semi-independently. CCAdmin can be used to create mailing-list-associations by hand for early testing of refresh script logic.

### DIFF
--- a/app/models/mailing_lists/site_mailing_list.rb
+++ b/app/models/mailing_lists/site_mailing_list.rb
@@ -1,0 +1,10 @@
+module MailingLists
+  class SiteMailingList < ActiveRecord::Base
+    include ActiveRecordHelper
+
+    self.table_name = 'canvas_site_mailing_lists'
+
+    attr_accessible :canvas_site_id, :list_name, :state, :populated_at
+
+  end
+end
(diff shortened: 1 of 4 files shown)


paulkerschen:
Looks good! Glad we're committing the migration on a Tuesday so it has plenty of time to decide if it wants to cause trouble.

ctweney:
@raydavis Should the new table be added to the developer-seed-data recipe in README? i.e., would developers want a production-like copy on their own machines? That's a common occurrence where CCAdmin is used to manage production data sets.

raydavis:
Thank you for asking, @ctweney. I'll try to remember to be explicit about the developer-seed implications of new tables.

In this case, the table should remain empty by default, since the Canvas sites they refer to will be different in every environment. When CLC-5159 is complete, new rows will always and only be added by admins (or in the case of localhost, developers) using a bCourses app, and CCAdmin will only be used for rarer maintenance tasks. I mentioned creating new rows through CCAdmin just as a way to let refresh-script development proceed without being delayed by app development. (It's even possible that Jon H. will want to start maintenance of a few mailing lists early in production.)

PR #3717, opened 2015-05-05
CLC-5224 - Web Accessibility: Apply Accessibility Guidelines to 'Roster Photos' Tool

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-5224


raydavis:
Looks good to me, but like @redconfetti I'm curious about why @christianv didn't use an "img" element to begin with. I'll leave this PR open for a bit longer in case he happens to have time to look in.

src/assets/templates/widgets/roster_photo.html
```diff
@@ -1,2 +1,2 @@
 <img data-ng-if="student.photo" data-ng-src="{{student.photo}}" width="72" height="96" data-ng-attr-alt="Photo of {{student.first_name}} {{student.last_name}}"/>
-<div data-ng-if="!student.photo" class="cc-image-responsive cc-page-roster-image-unavailable" title="Photo unavailable"></div>
+<img data-ng-if="!student.photo" src="/assets/images/svg/photo_unavailable_official_72x96.svg" width="72" height="96" alt="No Official Photo is Available for {{student.first_name}} {{student.last_name}}"/>
```
christianv: @redconfetti can you verify that the img url hashed when we run the gulp production build? Afaik this only happens with images in the CSS/SCSS. We would need this for cache busting reasons.

redconfetti: @christianv I don't see it hashing in templates, only via the SCSS. Thanks for letting us know that we're doing this for cache busting reasons.

I'll switch it back to the DIV solution for now.

christianv:
@redconfetti thanks for working this through!

PR #3719, opened 2015-05-05
CLC-5225 - Web Accessibility: Apply Accessibility Guidelines to 'User Provisioning' Tool

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-5225


christianv:
@redconfetti thanks for making these accessibility changes!

PR #3722, opened 2015-05-05
CLC-5223 - Added hidden label for CCN list textarea

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-5223

### DIFF
--- a/src/assets/templates/canvas_embedded/create_course_site.html
+++ b/src/assets/templates/canvas_embedded/create_course_site.html
@@ -46,7 +46,8 @@
           </div>
         </div>
         <form data-ng-submit="fetchFeed()" class="cc-page-form">
-          <textarea data-ng-model="admin_by_ccns"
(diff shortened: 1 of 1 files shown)


src/assets/templates/canvas_embedded/create_course_site.html
```diff
@@ -46,7 +46,8 @@
           </div>
         </div>
         <form data-ng-submit="fetchFeed()" class="cc-page-form">
-          <textarea data-ng-model="admin_by_ccns"
+          <label for="ccn-list" class="cc-visuallyhidden">CCN List Separated by Commas or Spaces</label>
+          <textarea id="ccn-list" data-ng-model="admin_by_ccns"
```
christianv: @redconfetti can we make sure to prefix the ids with cc- ? Since they are part of the global namespace.

redconfetti: @christianv sure - fixed in #3727

PR #3723, opened 2015-05-05
CLC-5256, webcast cron must iterate across all official_sections

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5256
Our new rake task will use Canvas::Courses class to pull info on all official sections.


app/models/canvas/courses.rb
```diff
@@ -0,0 +1,14 @@
+module Canvas
+  class Courses < Proxy
+
+    def official_courses
+      self.class.fetch_from_cache(@uid) do
+        account_id = Settings.canvas_proxy.official_courses_account_id
+        params = 'published=true&include=[sections]'
+        response = request_uncached("accounts/#{account_id}/courses?#{params}", '_all_official_course_sites', {:method => :get})
```
raydavis: This will probably only give you the first 10 course sites. You'd need to page through the API to get them all. Here's an example: https://github.com/ets-berkeley-edu/calcentral/blob/master/app/models/canvas/course_users.rb#L29

raydavis:
Unless you need to restrict this to published sites, it would probably be much more efficient to re-use the course sections CSV downloaded for the daily site membership refresh: https://github.com/ets-berkeley-edu/calcentral/blob/master/app/models/canvas/refresh_all_campus_data.rb#L53

@redconfetti or I would need to help with that, since the code's pretty gnarly.

app/models/canvas/courses.rb
johncrossman: Thanks @raydavis  I thought that might be the case. I'll add in the pagination today.

johncrossman:
@raydavis I'm now using Canvas::SectionsReport (per conversation with @redconfetti ). Also new is WebcastEligibleCourses which filters out course-sections that are being recorded and/or eligible for webcast.  The piece that is still missing - and intended for future PR - is identifying ccns in webcast enabled rooms. For now, I have a hard coded list.  In fact, after today's talk with Judy I realize we'll need a modified, and perhaps simpler, approach to identifying sections in webcast-equipped rooms. More on that topic later.

app/models/canvas/webcast_lti_refresh.rb
```diff
@@ -1,64 +1,62 @@
...
-    def update_hidden_on_webcast_tab(show_tab, tab)
+    def update_hidden_on_webcast_tab(show_tab, canvas_course_id)
       modified_tab = nil
+      external_tools = Canvas::ExternalTools.new @options.merge(canvas_course_id: canvas_course_id)
+      tab = external_tools.find_canvas_course_tab @canvas_webcast_tool_id
       if tab
         is_tab_showing = !tab.has_key?('hidden') || !tab['hidden']
         if show_tab == is_tab_showing
-          Rails.logger.warn "Webcast tab on course site #{@canvas_course_id} is already hidden=#{show_tab}. Do nothing."
+          Rails.logger.warn "Webcast tab on course site #{canvas_course_id} is already hidden=#{show_tab}. Do nothing."
```
raydavis: @johncrossman , the next time you're in this code, could you try ``include ClassLogger`` for your logging needs? Gives you less typing & more info in the logs. I think we still need to take the "Rails.logger." approach in Rake task definitions, but I don't remember many other places.

PR #3724, opened 2015-05-05
CLC-5226 - Web Accessibility: Apply Accessibility Guidelines to 'Find a Person to Add' Tool

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-5226

### DIFF
--- a/src/assets/stylesheets/_canvas_course_add_user.scss
+++ b/src/assets/stylesheets/_canvas_course_add_user.scss
@@ -19,6 +19,10 @@
     margin: 0 0 10px;
   }
 
+  .bc-page-course-add-user-fieldset {
+    border: 0;
+  }
+
   .bc-fa-black {
     color: $bc-color-off-black !important;
   }
--- a/src/assets/stylesheets/_canvas_course_add_user.scss
+++ b/src/assets/stylesheets/_canvas_course_add_user.scss
@@ -1,13 +1,14 @@
 .bc-page-course-add-user {
-
   background: $cc-color-white;
-  font-family: $header-font-family;
-  font-weight: normal;
+  font-family: $bc-header-font-family;
+  font-size: 14px;
+  font-weight: 300;
   padding: 10px;
 
-  h2 {
+  .bc-page-course-add-user-header {
     color: $bc-color-off-black;
     font-size: 24px;
+    font-weight: normal;
     line-height: 40px;
     margin: 8px 0;
   }
--- a/src/assets/stylesheets/_canvas_course_add_user.scss
+++ b/src/assets/stylesheets/_canvas_course_add_user.scss
@@ -19,10 +19,6 @@
     margin: 0 0 10px;
   }
 
-  .bc-page-course-add-user-fieldset {
-    border: 0;
-  }
-
   .bc-fa-black {
     color: $bc-color-off-black !important;
   }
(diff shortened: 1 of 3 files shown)


src/assets/stylesheets/_canvas_course_add_user.scss
```diff
@@ -18,6 +19,10 @@
     margin: 0 0 10px;
   }
 
+  .bc-page-course-add-user-fieldset {
```
christianv: @redconfetti I'm imagining that we'll have to do this for every `fieldset`. Would you be ok with adding it to `_base.scss`?

```css
fieldset {
  border: 0;
}
```

src/assets/templates/canvas_embedded/course_add_user.html
```diff
@@ -12,14 +12,18 @@
         <div data-ng-show="noSearchTextAlert" class="cc-notice-box cc-notice-error cc-radius">
           You did not enter any search terms. Please try again.
           <div class="cc-notice-close-button">
```
christianv: @redconfetti can we put `cc-notice-close-button` in it's own template to dry it up?

src/assets/templates/canvas_embedded/course_add_user.html
```diff
@@ -130,26 +140,33 @@
...
-                  <td data-ng-bind="user.emailAddress"></td>
-                </tr>
-              </tbody>
-            </table>
+            <fieldset class="bc-page-course-add-user-fieldset">
+              <legend class="cc-visuallyhidden">Select the user you wish to add to the course site:</legend>
+              <table class="bc-user-search-table">
+                <thead>
+                  <tr>
+                    <th><span class="cc-visuallyhidden">Actions</span></th>
```
christianv: @redconfetti since this is an accessibility commit, can you also add `scope="col"` to each `th`?

christianv:
@redconfetti really good accessibility changes, +1!

src/assets/stylesheets/_canvas_course_add_user.scss
redconfetti: @christianv Yep. I'd rather do that than have to do it again and again.

src/assets/templates/canvas_embedded/course_add_user.html
redconfetti: @christianv Each of these has it's own logic concerning what it does to close out the notice (see ngClick property). I would have to develop a directive with template to accomplish this. Doesn't that seem like overkill to dry up what seems like a grouped repetition of a pattern in 4 places?

src/assets/templates/canvas_embedded/course_add_user.html
redconfetti: @christianv sure. I've added this in our [Web Accessibility Guidelines](https://confluence.ets.berkeley.edu/confluence/display/CANVAS/Web+Accessibility+Guidelines#WebAccessibilityGuidelines-TabularData) for tabular data as well.

src/assets/templates/canvas_embedded/course_add_user.html
christianv: @redconfetti let's leave this for now. I agree that it might be overkill

redconfetti:
@christianv So are we good to go with merging this then?

PR #3544, opened 2015-04-04
CLC-4905 - Front-end: Store "View As" user data in the back-end

edwlook:
https://jira.ets.berkeley.edu/jira/browse/CLC-4905

PR #3725, opened 2015-05-06
CLC-5203 Revert to old View As UX, keeping back-end storage

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5203

This PR attempts the minimum to get the old View As UX working with the new database approach.
  - The template is almost entirely reverted to the pre-#3544 settings.html. (One change: the duplicate "Stop Acting As" button first noted in CLC-3506 is no longer present.)
  - The pre-#3544 adminController.js and the post-#3544 studentLookupController.js are stitched together into a single admin controller.
  - "Clear all users" API endpoints are added, as the old UX requires.
  - Unused front-end code is removed.
  - Saving a user now removes them from the Recent Users list. Duplication between two lists was no big deal in the tabbed UI, but it looks distinctly odd when both lists are visible.
  - Verbose logging in the stored-user models is removed.

There are couple of issues that should be addressed as follow-up tasks.
  - Code duplication in the server-side controller, model and tests.
  - While users can now manually clear the Recent Users list, the issue noted in CLC-5157 (no automatic limit to list size) remains.

I've removed large chunks of code committed in #3544, but of course that implies no disrespect for the work done there. I needed to revert as comprehensively as possible to be sure that I wasn't relying on any hidden dependencies, but I expect that whoever next works on View As will want to revert most of this commit in turn, and take it from there.


src/assets/templates/settings.html
```diff
@@ -26,12 +26,97 @@ <h1 class="cc-heading-page-title">Settings for {{api.user.profile.fullName}}</h1
...
+      <div class="row">
+        <div class="small-9 small-offset-3 columns">
+          <span data-ng-bind="admin.actAsErrorStatus"></span>
+        </div>
+      </div>
+      <div class="row" data-ng-if="admin.userPool">
+        <div class="small-9 small-offset-3 columns">
+          <ul data-ng-repeat="user in admin.userPool">
+            <li>
+              UID: <button class="cc-button-link" data-ng-bind="user.ldap_uid" data-ng-click="admin.actAsUser(user)"></button> (SID: {{user.student_id || "none"}})
```
christianv: @paulkerschen can we wrap the `{{` in a `<span>` with `ng-bind` everywhere in this page?

http://stackoverflow.com/questions/16125872/why-ng-bind-is-better-than-in-angular

paulkerschen: @christianv Thanks for the tip!

christianv:
front-end lgtm - would be good to also have a back-end review
/cc @ctweney @redconfetti @raydavis

raydavis:
Thanks, @paulkerschen (and @christianv for the review)!

PR #3727, opened 2015-05-06
CLC-5223 - Updated ID for form input to avoid conflict in global namespace

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-5223


christianv:
@redconfetti thanks!

PR #3728, opened 2015-05-06
CLC-5120: fix view-as tests again

pfarestveit:
https://jira.ets.berkeley.edu/jira/browse/CLC-5120

These changes restore the view-as and look-up portions of the User Authorizations spec to the former UI (see CLC-5203), also removing supporting code that pertained to the formerly-new UI.


johncrossman:
I see no problems.

PR #3732, opened 2015-05-08
CLC-5134, ExternalTools : update_canvas_course_site_tab needs proper method PUT

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5242
Bamboo build: https://bamboo.ets.berkeley.edu/bamboo/browse/MYB-CJT51-19

Few related changes in this PR:
* class that pulls from Webcast audio RSS must tolerate 404s
* optional options hash to Faraday connection so you can config in dev (e.g., ssl verify: false)
* ExternalTools needs proper method-PUT syntax in update_canvas_course_site_tab

### DIFF
--- a/app/models/canvas/sections_report.rb
+++ b/app/models/canvas/sections_report.rb
@@ -1,6 +1,10 @@
 module Canvas
   class SectionsReport < Canvas::Report
 
+    def initialize(options = {})
+      super options
+    end
+
     def get_csv(term_id)
       get_provisioning_csv('sections', term_id)
     end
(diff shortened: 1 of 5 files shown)


app/models/canvas/sections_report.rb
```diff
@@ -1,6 +1,10 @@
 module Canvas
   class SectionsReport < Canvas::Report
 
+    def initialize(options = {})
+      super options
+    end
+
```
paulkerschen: @johncrossman Just curious, does this method do anything, or is it a placeholder for future initialization code?

johncrossman: @paulkerschen It's used by WebcastEligibleCourses in this PR. It allows, for example, to pass options to Faraday connection.

PR #3734, opened 2015-05-08
CLC-5283 View As: update ID field when recent/saved user clicked

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5283

Restores a function that #3725 inexplicably left out.


christianv:
lgtm

PR #3735, opened 2015-05-08
CLC-5272 LTI app visibility "members" does not mean members

raydavis:
It means teachers and students. Only "public" links can be seen by other types of site members, such as observers.

https://jira.ets.berkeley.edu/jira/browse/CLC-5272


johncrossman:
Looks good

PR #3737, opened 2015-05-08
CLC-5256, restrict options sent to Faraday; bug fix on recordings.has_key?

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5256
Changes / fixes:
* Only allow ssl override to Faraday; otherwise nil
* check for 'has Webcast recordings' needs explicit [:videos] check
* clean up logger statements

Waiting on Bamboo, too: https://bamboo.ets.berkeley.edu/bamboo/browse/MYB-CJT51-22

### DIFF
--- a/app/models/canvas/report.rb
+++ b/app/models/canvas/report.rb
@@ -54,7 +54,9 @@ module Canvas
       if @fake
         csv = CSV.read("fixtures/pretty_vcr_recordings/Canvas_#{report_type}_report_#{object_type}_csv.csv", {headers: true})
       else
-        conn = Faraday.new(file_info["url"], @options) do |c|
+        # Faraday requires nil when options hash is empty. For now, we only allow ssl options.
+        conn_options = @options.any? && @options[:ssl].present? ? @options : nil
+        conn = Faraday.new(file_info["url"], conn_options) do |c|
           c.use FaradayMiddleware::FollowRedirects
           c.use Faraday::Adapter::NetHttp
         end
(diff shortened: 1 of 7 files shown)


johncrossman:
Bamboo is happy

app/models/canvas/report.rb
```diff
@@ -54,7 +54,9 @@ def get_account_csv(report_type, object_type, term_id)
       if @fake
         csv = CSV.read("fixtures/pretty_vcr_recordings/Canvas_#{report_type}_report_#{object_type}_csv.csv", {headers: true})
       else
-        conn = Faraday.new(file_info["url"], @options) do |c|
+        # Faraday requires nil when options hash is empty. For now, we only allow ssl options.
+        conn_options = @options.any? && @options[:ssl].present? ? @options : nil
```
raydavis: @johncrossman , why are the options only viable if they include ``:ssl``?

raydavis:
LGTM, just saw that one puzzler...

app/models/canvas/report.rb
johncrossman: I wanted to be cautious. Faraday does a lot of inspection on 'options' and I didn't want to open a can worms. I noticed fragility when I passed Faraday an empty hash and it blew up. This current use of options seems to work. E.g., importance of passing nil instead of empty hash.

johncrossman: @raydavis  After some testing I realize that sending empty hash to Faraday is fine. Previously, I inadvertently passed an instance of OpenStruct and, as you would expect, Faraday choked. I will commit one more change to this PR and remove the hack-ish restriction on @options

raydavis:
Thank you for the extra cleanup, @johncrossman -- that code is hard enough to follow as it is!

PR #3740, opened 2015-05-08
CLC-5134, remove use of SKIP_CANVAS_SSL_VERIFY option in rake task

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5134

The rake task webcast_lti_refresh code does not need to know about our SSL verification problems.

PR #3741, opened 2015-05-08
CLC-5134, remove use of SKIP_CANVAS_SSL_VERIFY option in rake task

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5134

The rake task webcast_lti_refresh code does not need to know about our SSL verification problems.


johncrossman:
@redconfetti  Same change as #3740 but only one commit.

PR #3742, opened 2015-05-08
CLC-5269: Selenium test for academics teaching UI

pfarestveit:
https://jira.ets.berkeley.edu/jira/browse/CLC-5269

This is a new test to cover what instructors see in the Academics UI.  The test iterates through UIDs, checking the teaching semester information that appears in the Academics landing page, semester pages, and class pages.

There is also some minor refactoring and reorganizing of code already used by the student academics test, since there is much overlap with this new test.


johncrossman:
A whole lot of goodness there... even tho I don't understand half of it.  Merging.

PR #3745, opened 2015-05-12
CLC-5242, fix raise error when tab update fails; test coverage; logging

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5242

Bug fix on verify tab update. Test coverage, too.


johncrossman:
ExternalTools class is odd in how it (1) looks for canvas_course_id and then (2) falls back to canvas_account_id if former is not found. I removed the test which expects exception in case #2.

PR #3747, opened 2015-05-12
CLC-5299, Webcasts: recordings duplicated on cross-listed course

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5299

This bug has to do with the deprecated :video and :audio elements of Webcast merged feed. In the case of cross-listings we were merging all recordings into an array and thus the duplicates. My fix is to use a Set object instead. I added a test to verify.


paulkerschen:
Definitely enough planets in that feed already without including them all twice.

PR #3758, opened 2015-05-13
CLC-5301, use eligible-for-webcast.json feed to determine if CCN is eligible for sign-up

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5301
For reasons explained in the Jira, we now use Webcast::SignUpEligible instead of Webcast::Rooms. The related work on Webcast has been done: https://jira.ets.berkeley.edu/jira/browse/WCT-5443
Other details:
* use slug convention: #{term.downcase}-#{year} 
* 2015 test data for webcast_eligible_courses_spec
* rename *in_webcast_enabled_room* hash key to *is_webcast_eligible*


paulkerschen:
Alles gut.

PR #3759, opened 2015-05-13
CLC-5292 Allow saved users to appear under Recent Users without save link

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5292

Note that this reverts the change in #3725 to StoredUsersController#store_saved_uid, which removed a user from the Recent Users list after saving.


johncrossman:
I agree with your decision described in CLC-5292. This looks good.

PR #3760, opened 2015-05-13
CLC-5228 - Converted aria-hidden attribute to use ng-attr directive

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-5228


johncrossman:
Good by me. Follows pattern of data-ng-attr-aria-expanded, etc

christianv:
@redconfetti thanks

PR #3762, opened 2015-05-14
NOJIRA Update activerecord-related gems

raydavis:
CHANGELOG FOR ACTIVERECORD-JDBC-ADAPTER FROM 1.3.13 TO 1.3.16
CHANGELOG FOR ACTIVERECORD-JDBCH2-ADAPTER FROM 1.3.13 TO 1.3.16
CHANGELOG FOR ACTIVERECORD-JDBCPOSTGRESQL-ADAPTER FROM 1.3.13 TO 1.3.16
CHANGELOG FOR ACTIVERECORD-JDBCSQLITE3-ADAPTER FROM 1.3.13 TO 1.3.16
 VERSION 1.3.16
[h2] allow jdbch2 adapter to use jdbc-h2 1.4 (#639)
[postgres] quote and type_cast should not be overriden on 4.2
[postgres] fix table column resolution (slightly align it to our needs) on 4.2
[postgres] re-arrange Column internals as some helpers end-up on Adapter in 4.2
[mysql] change_column table is redefined on AR 4.2 so that it uses _sql version
[mysql] more tuning/compatibility backports for AR 4.2 (working booleans #626)
[mysql] type-map support copy-pasta from AR 4.2 (closer to full compat) (#617)
[postgres] improved support for Amazon Redshift (based on user feedback #403)
fixed problem with create table statements with references in MySQL (#629)
fix RecordNotUnique reference in ArJdbc::SQlite3 (#634)
[postgres] add support for jsonb column type (#633 #635 #637)
 VERSION 1.3.15
[informix] adapter undefined method 'column_types' with Rails 4.x (#622)
improve connection factory Ruby interface impls to match Java method-name (#621)
[sqlite] truncate_fake - as due compatibility with do not want a truncate method
truncate support 4 all (introduced in AR 4.2 but we let it work in all versions)
[postgres] do not double load array_parser (on 4.2)
[postgres] connection.truncate impl (AR 4.2 compat)
[mysql] back-port missing truncate connection method (since AR 4.2)
auto setup type-casting compat (pre 4.2 emulation) 4 all
[postgres] backport bigserial key and bigint native type support
 VERSION 1.3.14
[mysql] Fixed an undefined exception error in typetosql for MySQL (#616)
[derby] support for empty insert statements (e.g. DbModel.create!) on AR 4.x
[db2] revert limit/offset OVER( ORDER BY ) matching as it was not complete (#606)
[db2] handle empty INSERT statement (default) VALUES in a way that works on AR 4.x
[db2] select return value should be the same as core AR (Array on 3.2) (#607)
[mysql] backport bulk change & schema-creation to be compatible with AR 4.x
handle possible null values while extracting from XML columns (#615)
DateTime columns are returned always as nil values (on Rails 4.2 #613)
[oracle] do not override primary= within column methods on AR 4.2
handle removed AR Column#primary= method on 4.2 (#609)
[oracle] fix missing sequencenamelength method when doing a rename_table
[mysql] do not force DEFAULT CHARSET=utf8 when creating new tables
[postgres] rename the primary key index when renaming a table in pg
[db2] improve create_table on Z/OS (now supporting AR 4.x TableDefinition) (#581)

CHANGELOG FOR ACTIVERECORD-ORACLE_ENHANCED-ADAPTER FROM 1.5.5 TO 1.5.6
 VERSION 1.5.6
Enhancements
Support Rails 4.1.10 [#530]
Remove warning message when JDK 8 is used [#525]
Support RAW column types [#471]
Properly quote database links [#556]
Grant create view privilege to db user [#528]
Read SYSTEM password from ENV ORACLESYSTEMPASSWORD optionally [#529]
Show original error message when loading ruby-oci8 library fails [#532]
Update README that OracleEnhancedProcedures is not auto loaded [#474]
Fix legacy schema support syntax [#507]
Peform all unit test when tested with Oracle 12c [#465]
Add :if_exists option to drop_table [#541]
Extract OracleEnhancedDatabaseStatements [#449]
Removed self.visitor_for(pool) method [#501]
Bug Fix
Fix serialized readonly lobs [#515]
Do not dump schema information during structure dump [#558]
Structure dump generates correct create or replace synonym [#453]
Procedures and functions are created correctly by removing semi-colon [#456]
Show support matrix of Java and JDBC Driver only when java_version >= '1.8' [#455]
Update Gemfile dependencies so specs can run [#472]


johncrossman:
Some nice enhancements. Scrutinized to the best of my ability. Merging...

PR #3766, opened 2015-05-15
CLC-5092 - Summarize with proper SASS

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-5092


paulkerschen:
I'm glad to know about that idiom too.

PR #3768, opened 2015-05-15
CLC-5313: add "testable users" check to all UID-driven Selenium tests

pfarestveit:
https://jira.ets.berkeley.edu/jira/browse/CLC-5313

Adding more checks for the health of various feeds, plus minor cleanup.


johncrossman:
All good

PR #3769, opened 2015-05-15
CLC-5093 - bCourses CSS: Course Site Roster Photos styling

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-5093

### DIFF
--- a/src/assets/images/canvas/ic-icon-arrow-down.svg
+++ b/src/assets/images/canvas/ic-icon-arrow-down.svg
@@ -1,0 +1,4 @@
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
+<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px" viewBox="0 0 300 170" enable-background="new 0 0 300 170" xml:space="preserve">
+  <polygon fill="#333333" points="150,170 300,0 0,0 "/>
+</svg>
--- a/src/assets/stylesheets/_canvas_base.scss
+++ b/src/assets/stylesheets/_canvas_base.scss
@@ -1,11 +1,15 @@
 // Canvas Fonts
 $bc-base-font-family: 'Helvetica Neue', Helvetica, Arial, sans-serif;
 
-.bc-canvas-application {
-  background: $cc-color-white;
+.bc-canvas-base-typography {
   font-family: $bc-base-font-family;
   font-size: 14px;
   font-weight: 300;
+}
+
+.bc-canvas-application {
+  background: $cc-color-white;
+  @extend .bc-canvas-base-typography;
 
   input, input[type="text"], button {
     &, &:hover, &:focus {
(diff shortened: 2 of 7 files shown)


raydavis:
scss-lint failed, @redconfetti .

src/assets/images/canvas/ic-icon-arrow-down.svg
```diff
@@ -0,0 +1,4 @@
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
+<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px" viewBox="0 0 300 170" enable-background="new 0 0 300 170" xml:space="preserve">
+  <polygon fill="#333333" points="150,170 300,0 0,0 "/>
```
christianv: @redconfetti can we:

1) Use [svgo](https://github.com/svg/svgo) to optimize the `svg`? Also see our [styleguide](https://github.com/ets-berkeley-edu/calcentral/blob/2928469e5e0ebb52b4535d335db78585a99c54b5/docs/styleguide.md)
2) Use the following naming convention: `icon_youtube_red_85x60.svg`
3) Make sure we have an extra line at the end of the file so GitHub doesn't complain about it.

src/assets/stylesheets/_canvas_base.scss
```diff
@@ -1,11 +1,30 @@
...
-  background: $cc-color-white;
+// Placeholder Selectors
+%bc-canvas-base-typography {
   font-family: $bc-base-font-family;
   font-size: 14px;
   font-weight: 300;
+}
+
+%bc-form-input-text {
+  @extend %bc-canvas-base-typography;
```
christianv: @redconfetti can we please avoid any use of `extend`? We don't want to create unmaintainable CSS

src/assets/stylesheets/_canvas_buttons.scss
```diff
@@ -64,6 +53,26 @@ $bc-canvas-button-selected-color: #fff;
   }
 }
 
+// Canvas Button Styles
+.bc-canvas-button, .bc-main .cc-widget-webcast .cc-button {
+  @extend %bc-canvas-button;
```
christianv: @redconfetti please avoid using `extend`

src/assets/stylesheets/_canvas_base.scss
redconfetti: @christianv I can understand eliminating the use of @extend in the _canvas_base.scss, _canvas_buttons.scss, and any other file with globally used styles.

But what about in files such as _canvas_roster.scss where I'm declaring them to apply to certain classes inside of the `cc-canvas-embedded` class (the wrapper used to apply different styles to the Canvas embedded Roster Photos tool than that of the CalCentral version)?

Do you foresee any better approach other than applying those styles to the elements as needed?

src/assets/images/canvas/ic-icon-arrow-down.svg
redconfetti: @christianv I've optimized this image, but it's not an icon. It's a down arrow used to replace the down arrow indicator in [Canvas styled select drop-down menus](https://canvas.instructure.com/styleguide#forms). I've made the file name more descriptive. Since it's an SVG, I don't think the dimensions in the filename apply.

See #3771

christianv: @redconfetti thanks!

src/assets/stylesheets/_canvas_base.scss
christianv: @redconfetti I do see you point and think the best way to resolve it would be to have a short chat in person when I get back, would you be ok with that?

redconfetti: @christianv See #3774 for refactor

PR #3780, opened 2015-05-20
CLC-5209, remove webcast_sign_up lti configs, template

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5209

We will have a single widget (canvas_embedded/course_mediacasts.html) to handle webcast scenarios, including sign-up.


paulkerschen:
Who could argue with all those minus signs?

PR #3782, opened 2015-05-21
CLC-5296 - Include the ngAria module

christianv:
https://jira.ets.berkeley.edu/jira/browse/CLC-5296


christianv:
@redconfetti since you've been adding `aria` roles, it would be great if you can take a look at this one.

PR #3784, opened 2015-05-21
CLC-5211 Mailing lists: add members count and recent errors to database

paulkerschen:
Back-end work under https://jira.ets.berkeley.edu/jira/browse/CLC-5211, in order to implement the designs at https://jira.ets.berkeley.edu/jira/browse/CLC-5321.

Adjustments to the SiteMailingList model logic will follow in another PR.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20150507215402) do
+ActiveRecord::Schema.define(version: 20150521163404) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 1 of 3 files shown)


db/schema.rb
```diff
@@ -23,6 +23,9 @@
     t.datetime "populated_at"
     t.datetime "created_at"
     t.datetime "updated_at"
+    t.integer  "members_count"
+    t.integer  "populate_add_errors"
+    t.integer  "populate_remove_errors"
   end
```
johncrossman: @paulkerschen Are these all nullable?

paulkerschen: For reference on @johncrossman 's question: these columns, like `populated_at`, will have null values after a record is first created and before any population jobs have run. After the first population job, they will have perhaps-zero integer values.

johncrossman: Thanks @paulkerschen  Makes sense

PR #3790, opened 2015-05-22
CLC-4967 - Add-delete: various styling issues with footer, buttons, borders... styling

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-4967


raydavis:
Thank you for centralizing the iframe resize, @redconfetti

PR #3793, opened 2015-05-22
NOJIRA - Merge master in SIS master

christianv:


christianv:
Seems like the textbook tests were causing some errors so I restarted travis. If it doesn't work, @ctweney would you mind having a look?

christianv:
1 green travis, merging

PR #3795, opened 2015-05-25
CLC-5336 - Campus page doesn't show on mobile

christianv:
https://jira.ets.berkeley.edu/jira/browse/CLC-5336


johncrossman:
Looks good

PR #3797, opened 2015-05-26
CLC-5217, instructor_func added to instructors_feed; update course_sections_spec

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5217
Webcast front-end needs to know about instructor_func. PR includes refactor of course_sections_spec.rb

Note: In an upcoming PR the Webcast merged feed will leverage this change in order to put "webcast authorized" instructors in its front-end api


paulkerschen:
Thanks for the refactor!

PR #3802, opened 2015-05-26
SISRP-4274 Connect to real CS feeds; reduce duplication by establishi…

ctweney:
…ng DirectProxy; update dummy JSON feeds and move them to correct directory

https://jira.berkeley.edu/browse/SISRP-4274

This will allow a real live connection to the real CS feed if you have the URL and basic auth information.


christianv:
lgtm if tests pass

raydavis:
And the tests did pass.

PR #3806, opened 2015-05-27
CLC-5174, Webcast widget needs a layout tweak

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5174

* Top margin in webcast widget removed.
* Unneeded sign-up assets removed

### DIFF
--- a/src/assets/stylesheets/_webcast.scss
+++ b/src/assets/stylesheets/_webcast.scss
@@ -21,7 +21,6 @@
     margin-top: 10px;
   }
   .cc-widget-webcast-select {
-    margin-top: 15px;
     max-width: 300px;
   }
 }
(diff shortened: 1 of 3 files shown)


src/assets/stylesheets/_webcast.scss
```diff
@@ -21,7 +21,6 @@
     margin-top: 10px;
   }
   .cc-widget-webcast-select {
-    margin-top: 15px;
```
christianv: @johncrossman does this mean we won't have webcasts with both audio and video feeds anymore?

johncrossman: @christianv That is true for future webcasts. Old webcasts *will* have audio and video feeds.

johncrossman: @christianv  And now that I look at the two-tab scenario (Video | Audio) I see the problem. I'll reopen the Jira.

PR #3809, opened 2015-05-27
CLC-5091 - Misc Tweaks #2

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-5091


src/assets/javascripts/angular/controllers/pages/canvasCourseAddUserController.js
```diff
@@ -142,6 +142,14 @@
       });
     };
 
+    $scope.noUserSelected = function() {
```
christianv: @redconfetti can you rewrite this as:

```javascript
$scope.noUserSelected = function() {
  return !!$scope.selectedUser;
}
```

or 

```javascript
$scope.noUserSelected = !!scope.selectedUser;
```

Not sure whether the last one will update correctly.

christianv:
Only 1 comment, rest lgtm

src/assets/javascripts/angular/controllers/pages/canvasCourseAddUserController.js
redconfetti: Will do. For some reason that approach slipped my mind.

PR #3814, opened 2015-05-28
CLC-4579 Generate report on bCourses TurnItIn usage as CSV

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-4579

This lets us produce a one-off report from the console. If it meets requirements, a cron job script can land in the next sprint.

### DIFF
--- a/app/models/canvas/turnitin_reporter.rb
+++ b/app/models/canvas/turnitin_reporter.rb
@@ -1,0 +1,62 @@
+module Canvas
+  class TurnitinReporter < Csv
+    include ClassLogger
+
+    def self.print_current_term
+      current_term_id = Canvas::Proxy.current_sis_term_ids.first
+      worker = TurnitinReporter.new(current_term_id)
+      csv_filename = worker.generate_csv
+      parsed_csv = CSV.read(csv_filename, {headers: true})
+      print parsed_csv.to_csv
+    end
+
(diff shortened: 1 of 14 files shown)


app/models/canvas/turnitin_reporter.rb
```diff
@@ -0,0 +1,62 @@
...
+    def self.print_current_term
+      current_term_id = Canvas::Proxy.current_sis_term_ids.first
+      worker = TurnitinReporter.new(current_term_id)
+      csv_filename = worker.generate_csv
+      parsed_csv = CSV.read(csv_filename, {headers: true})
+      print parsed_csv.to_csv
+    end
+
+    def initialize(sis_term_id)
+      super()
```
paulkerschen: @raydavis My only nit to pick is that the empty parentheses don't seem very Ruby.

raydavis: @paulkerschen , fair cop. (I think I had an argument there in an earlier draft.) I'll amend.

paulkerschen: @raydavis, ack, I misled you. `super` is magic and, without empty parentheses, will try to pass the same arguments it was called with. My bad!

raydavis: Ruby and its magic!

PR #3816, opened 2015-05-28
NOJIRA, fix spec w.r.t recordingStartUTC feature

johncrossman:
We refreshed Webcast's QA env with production webcast.json which does not have the recordingStartUTC property because that feature is not yet live. I've commented out the line in spec and created a Jira: https://jira.ets.berkeley.edu/jira/browse/WCT-5456


paulkerschen:
Makes sense. Waiting on T

PR #3821, opened 2015-05-28
CLC-5145 Improve My Classes messaging for EAP students

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-5145

Also took Bernie's suggestion on making the default no-classes message less scary.


paulkerschen:
Much more polite!

PR #3824, opened 2015-05-29
CLC-4579 Rake task to generate TurnItIn report for a term

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-4579

This should make QA a little easier.

PR #3827, opened 2015-05-29
NOJIRA Uncomment cd in mailing list update script

paulkerschen:
Review of #3824 alerted me to the log in my own eye.


johncrossman:
Iz good

PR #3828, opened 2015-05-29
CLC-5350 Add UID to roster photos cache key

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5350


raydavis:
Thanks for the quick fix, @paulkerschen

PR #3830, opened 2015-05-29
CLC-5308 Remove obsolete Bearfacts token

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5308

And a little extra cleanup.


johncrossman:
Nice bit of cleanup

PR #3832, opened 2015-05-29
CLC-5022, do not show 'report problem' unless video OR audio

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5022

'Report problem' only if webcasts exist.


paulkerschen:
Makes sense. Waiting on T

PR #3837, opened 2015-06-01
CLC-5358 Replace lost LDAP UID with existing LDAP UID

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-5358

Wait for Bamboo!


paulkerschen:
Happy Bamboo.

PR #3840, opened 2015-06-01
CLC-5346 Set explicit user agent on Calmail proxy requests

paulkerschen:
A workaround for https://jira.ets.berkeley.edu/jira/browse/CLC-5346.


paulkerschen:
It passes on Bamboo....

johncrossman:
We [heart] workarounds

PR #3843, opened 2015-06-01
[QA] CLC-5346 Set explicit user agent on Calmail proxy requests

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5346
QA version of #3840.


johncrossman:
As before, looks good. Waiting on T

PR #3844, opened 2015-06-01
CLC-5364 - Front-end assets are not being hashed

christianv:
https://jira.ets.berkeley.edu/jira/browse/CLC-5364

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -307,30 +307,35 @@
     }
 
     var path = require('path');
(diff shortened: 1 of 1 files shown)


ctweney:
LGTM as long as Travis is happy.

christianv:
We should also check for a happy bamboo

raydavis:
I've grabbed a snapshot of requests from calcentral-dev, and will check for differences after this deploys.

PR #3845, opened 2015-06-02
CLC-5364 QA - Front-end assets are not being hashed

christianv:
https://jira.ets.berkeley.edu/jira/browse/CLC-5364

Follow up from #3844 - we should test that one first before merging this PR


raydavis:
We have cache-breaking hash on calcentral-dev. Thanks for the fix, @christianv !

PR #3851, opened 2015-06-02
SISRP-4458 -  Removed 'act_as' from the features list

winstonkim:
https://jira.berkeley.edu/browse/SISRP-4458


johncrossman:
I found no need to update specs. Good to go, @winstonkim

PR #3854, opened 2015-06-03
CLC-5218, sign-up eligible CCNs in merged feed; messaging in widget; specs

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5215
https://jira.ets.berkeley.edu/jira/browse/CLC-5218

Highlights:
* Bamboo build: https://bamboo.ets.berkeley.edu/bamboo/browse/MYB-CJT55-15
* sign-up eligible CCNs to feed based on what is relevant to the course 
* re-org of webcast/merged.rb for readability
* specs and test data to verify
* new widget message: Webcasts for {{...}} will appear here after classes start


spec/models/webcast/merged_spec.rb
```diff
@@ -60,6 +64,22 @@
...
+      end
+      before do
+        courses_list = [{
+            :classes=>[{ :sections => [{ :ccn=>'9688', :section_number=>'101', :instruction_format=>'LEC' }] }]
+          }
+        ]
+        expect_any_instance_of(MyAcademics::Teaching).to receive(:courses_list_from_ccns).once.and_return courses_list
+      end
+      it 'pads ccn with zeroes' do
+        expect(feed[:media][2008]['D']['09688']).to be_nil
```
paulkerschen: @johncrossman, I might be missing something here, but are you checking for the presence of the '09688' key? This expectation should pass either if the key is present with a nil value, or if the key is absent.

src/assets/templates/widgets/webcast.html
```diff
@@ -5,7 +5,21 @@ <h1 class="cc-visuallyhidden" data-ng-if="isEmbedded">Webcasts</h1>
 
     <div data-ng-bind="proxyErrorMessage"></div>
     <div data-ng-if="!proxyErrorMessage">
-      <div data-ng-if="!videos.length && !audio.length">There are no webcasts available.</div>
+      <div data-ng-if="!videos && !audio">
+        There are no Webcasts scheduled.
+      </div>
+      <div data-ng-if="media">
+        <div data-ng-repeat="(year, term_hash) in media">
```
christianv: @johncrossman Can we use camelCase everywhere? e.g. `termHash`?

src/assets/templates/widgets/webcast.html
```diff
@@ -5,7 +5,21 @@ <h1 class="cc-visuallyhidden" data-ng-if="isEmbedded">Webcasts</h1>
...
+          <div data-ng-repeat="(term, ccn_hash) in term_hash">
+            <div data-ng-repeat="(ccn, course) in ccn_hash">
```
christianv: @johncrossman interesting that we have to use so many `ngRepeats`, is there a way in the back-end to restructure the format so we can avoid some of those?

spec/models/webcast/merged_spec.rb
johncrossman: Good catch, @paulkerschen  There was a bug in the associated mock query results. And in my refactoring of test data I made the mistake of expecting nil. Fixed.

src/assets/templates/widgets/webcast.html
johncrossman: Thanks @christianv  Change made.

src/assets/templates/widgets/webcast.html
johncrossman: Yes, @christianv. I'm going to create separate Jira and reduce it to strictly a ccnHash.

paulkerschen:
LGTM, and https://jira.ets.berkeley.edu/jira/browse/CLC-5376 is created for the nested repeats.

PR #3856, opened 2015-06-03
[QA] CLC-5374 Make model callbacks play more nicely with RailsAdmin

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5374

QA version of #3855.


johncrossman:
Once again, me like.

PR #3859, opened 2015-06-03
CLC-5379: Selenium test fixes

pfarestveit:
https://jira.ets.berkeley.edu/jira/browse/CLC-5379


johncrossman:
Travis is happy and, after a quick glance, I see no prob.

PR #3870, opened 2015-06-08
CLC-5409-improve sanity check in reg status test

pfarestveit:
https://jira.ets.berkeley.edu/jira/browse/CLC-5409


johncrossman:
Waiting on T

PR #3871, opened 2015-06-08
CLC-5394, spec to verify iTunes audio when iTunes video is nil

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5394

Bug is already fixed on master. This spec hopes to prevent future breakage.


raydavis:
Thanks for the safety net, @johncrossman

PR #3874, opened 2015-06-08
CLC-5408 Quell registration alerts during term transitions

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5408

The most expedient approach to a quieter /my/badges feed:
1. Check if the CalCentral current term is not considered current in campus data;
2. If so, return nil registration data from CampusOracle::UserAttributes;
3. Have MyBadges::StudentInfo attempt to fill in that nil data by calling out to MyAcademics::TransitionRegStatus;
3a. If the student is registered for the CalCentral current term, say so;
3b. If the student is not registered for the CalCentral current term, pass along null, no-action-required data that won't trigger front-end alerts.

### DIFF
--- a/app/models/campus_oracle/user_attributes.rb
+++ b/app/models/campus_oracle/user_attributes.rb
@@ -19,16 +19,13 @@ module CampusOracle
     def get_feed_internal
       result = CampusOracle::Queries.get_person_attributes(@uid)
       if result
-        result[:reg_status] = {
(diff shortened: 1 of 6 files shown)


johncrossman:
This looks good to me. But @raydavis should peruse, too.

raydavis:
Perused!

PR #3876, opened 2015-06-08
CLC-5403, show sign-up text in CalCentral if user.viewing_as?

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5403

User in view-as mode should see sign-up info (e.g., when debugging).


raydavis:
Unhappy Travis.

PR #3886, opened 2015-06-09
CLC-5419 Surface residency status even if registration status is missing

paulkerschen:


johncrossman:
No probs. Waiting on T

PR #3890, opened 2015-06-10
README: Copyedit

mc10:
Should be uncontroversial changes.


raydavis:
You are quite right, @mc10. Thank you for the edits.

PR #3894, opened 2015-06-10
OEC-636, Update OEC's default yml configs (e.g., departments)

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/OEC-636

Summer term with dept list approved by Justin


johncrossman:
Fixing...

PR #3896, opened 2015-06-10
CLC-5366 Work around AngularJS Boolean bug in My Tasks

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-5366

Trigger bug: https://github.com/angular/angular.js/issues/11959


paulkerschen:
Not ineffective.

PR #3897, opened 2015-06-10
OEC-635, oec diff feature needs to know about new SIS_ID column

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/OEC-635

Notes:
* the historically hacky row_converter needs to know about new SIS_ID column
* to verify this fix I had to update lots o test data


paulkerschen:
@johncrossman Thanks for inserting comments about those skipped columns!

PR #3898, opened 2015-06-11
CLC-5414 Clean up transition term messaging in My Academics

paulkerschen:
Highlights:

`MyAcademics::TransitionRegStatus` gets the more succinct and accurate name `MyAcademics::TransitionTerm`.

We now know that Bearfacts profile information may or may not be out of date during term transitions. MyAcademics::TransitionTerm checks the reported term on the Bearfacts profile and sends a `profileUpdated` boolean to the front end. Transition-specific messaging will show at the top of the Profile card only if this value is false.

The quick fix to the Status and Blocks card in #3886 is mostly reverted, since the blunt instrument of a null `regStatus.code` is no longer used to communicate lack of registration during term transitions. Instead we use the usual blank-space code for "not registered," but set `regStatus.needsAction` to `false` to keep alarms from going off. Status and Blocks will tell students that they are not registered for summer, but will not include an alert icon.

A whole lot of dummy JSON is added, covering a plurality of cases.


app/models/berkeley/terms.rb
```diff
@@ -37,6 +37,8 @@ class Terms
     # during Spring terms, when most instructors and most enrolled students will have more
     # interest in the Fall schedule than in Summer classes.
     attr_reader :future
+    # The previous term (before the current one).
+    attr_reader :previous
```
raydavis: I was wondering when this shoe was finally going to drop!

raydavis:
Thank you!

PR #3899, opened 2015-06-11
CLC-5407 Show faculty-appropriate messages in My Classes

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5407


raydavis:
Wow, that's embarrassing... thanks for the fix.

PR #3900, opened 2015-06-11
CLC-5425 Update Textbooks integration tests for summer 2015

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5425


raydavis:
Thank you for fixing the tests, @paulkerschen. We'll also need to contact the vendor, since they previously told us that they would keep the Spring 2015 textbooks feeds around at least through Fall 2015.

PR #3902, opened 2015-06-12
CLC-5428, Point script/webcast_lti_refresh.sh at proper rake task

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5428


paulkerschen:
YES.

PR #3904, opened 2015-06-12
CLC-5389 - Web Accessibility 3: Webcasts

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-5389

* Removes automatic reading of loaded content (resolves issues with reading relevant alerts)
* Adds announcements when switching between 'Video' and 'Audio' tab.
* Adds announcements when selecting a specific video or audio recording.


johncrossman:
I don't see probs.

PR #3909, opened 2015-06-16
CLC-5450: fix Selenium tests for reg status

pfarestveit:
https://jira.ets.berkeley.edu/jira/browse/CLC-5450

Updates the logic in the existing "status and blocks" spec to incorporate the new handling of term transition periods in the status popover and in the My Academics UI.


johncrossman:
Berry nice.

PR #3910, opened 2015-06-16
CLC-5451 - Update Angular to 1.4.1

christianv:
https://jira.ets.berkeley.edu/jira/browse/CLC-5451


johncrossman:
@christianv verified that 1.4.1 fixes problem with evaluating the template.

PR #3911, opened 2015-06-16
CLC-5108  - Have Course Site Provisioning use new background job module

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-5108

### DIFF
--- a/app/controllers/canvas_course_provision_controller.rb
+++ b/app/controllers/canvas_course_provision_controller.rb
@@ -57,9 +57,9 @@ class CanvasCourseProvisionController < ApplicationController
 
   # GET /api/academics/canvas/course_provision/status.json
   def job_status
-    course_provision_job = Canvas::BackgroundJob.find(params['job_id'])
-    render json: course_provision_job.to_json and return if course_provision_job.class == Canvas::ProvideCourseSite
-    render json: { job_id: params['job_id'], jobStatus: "jobNotFoundError", error: "Unable to find course management job" }.to_json
+    background_job = Canvas::BackgroundJob.find(params['jobId'])
+    render json: background_job.background_job_report.to_json and return if background_job.class == Canvas::ProvideCourseSite
+    render json: { jobId: params['jobId'], jobStatus: "jobNotFoundError", error: "Unable to find course management job" }.to_json
   end
 
   def create_options_from_params
--- a/app/controllers/canvas_course_provision_controller.rb
+++ b/app/controllers/canvas_course_provision_controller.rb
@@ -58,7 +58,7 @@ class CanvasCourseProvisionController < ApplicationController
   # GET /api/academics/canvas/course_provision/status.json
   def job_status
     background_job = Canvas::BackgroundJob.find(params['jobId'])
-    render json: background_job.background_job_report.to_json and return if background_job.class == Canvas::ProvideCourseSite
+    render json: background_job.background_job_report.to_json and return if background_job.is_a? Canvas::ProvideCourseSite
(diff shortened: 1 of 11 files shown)


redconfetti:
Note: Still need to fix front-end for this. Don't merge yet

app/controllers/canvas_course_provision_controller.rb
```diff
@@ -57,9 +57,9 @@ def edit_sections
 
   # GET /api/academics/canvas/course_provision/status.json
   def job_status
-    course_provision_job = Canvas::BackgroundJob.find(params['job_id'])
-    render json: course_provision_job.to_json and return if course_provision_job.class == Canvas::ProvideCourseSite
-    render json: { job_id: params['job_id'], jobStatus: "jobNotFoundError", error: "Unable to find course management job" }.to_json
+    background_job = Canvas::BackgroundJob.find(params['jobId'])
+    render json: background_job.background_job_report.to_json and return if background_job.class == Canvas::ProvideCourseSite
```
johncrossman: @redconfetti  Instead of obj.class == Canvas::ProvideCourseSite, how about kind_of? (or is_a?) which is true if object is an instance of that class or a subclass.

app/models/canvas/provide_course_site.rb
```diff
@@ -3,26 +3,21 @@ class ProvideCourseSite < Csv
...
       @section_definitions = []
       background_job_initialize
     end
 
     def create_course_site(site_name, site_course_code, term_slug, ccns, is_admin_by_ccns = false)
-      @total_steps = 12.0
-      @jobStatus = 'Processing'
-      background_job_save
+      background_job_set_type('course_creation')
+      background_job_set_total_steps(10)
```
johncrossman: @redconfetti  Minor style issue I've been reminded of: Don't use parens when not necessary. E.g. background_job_set_type('course_creation') as background_job_set_type 'course_creation'

app/models/canvas/provide_course_site.rb
```diff
@@ -135,7 +120,7 @@ def prepare_users_courses_list
         courses_list = filter_courses_by_ccns(candidate_courses_list, @import_data['term_slug'], @import_data['ccns'])
       end
       @import_data['courses'] = courses_list
-      complete_step('Prepared courses list')
+      background_job_complete_step('Prepared courses list')
     end
```
johncrossman: @redconfetti  Should these log statements include canvas_course_info or similar?

app/controllers/canvas_course_provision_controller.rb
redconfetti: Yeah. Good call. I don't expect any sort of superclass to be received ever so I'll use is_a?, which is cleaner and easier to read. Thanks!

app/models/canvas/provide_course_site.rb
redconfetti: @johncrossman What do you mean by log statements? background_job_completed_step doesn't perform logging, it adds a new string to an array that is stored as an instance variable in the Canvas::ProvideCourseSite object. The entire object is written to a Rails cache storage container each time there is an update, including a new step being completed, and all other details revealed via the background_job_report method. See [Canvas::BackgroundJob module](https://github.com/redconfetti/calcentral/blob/CLC-5108-background-job/app/models/canvas/background_job.rb#L111).

johncrossman: I think I misunderstood the purpose of those statements - not realizing they are for the UI. Nevermind. Thanks @redconfetti

PR #3913, opened 2015-06-16
CLC-5457, readable list in '...will appear here after classes start' section of webcast widget

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5457

Plus, some logging.


ctweney:
Looks good!

PR #3921, opened 2015-06-17
CLC-5462 - Synchronous XMLHttpRequest warning

christianv:
https://jira.ets.berkeley.edu/jira/browse/CLC-5462
https://jira.berkeley.edu/browse/SISRP-4328

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -62,8 +62,6 @@
         external: [
           // Date parsing
           'node_modules/moment/moment.js',
-          // Libraries (google analytics)
-          'src/assets/javascripts/lib/**/*.js',
           // Human Sorting in JavaScript
           'node_modules/js-natural-sort/dist/naturalSort.js',
           // Remote JavaScript error logging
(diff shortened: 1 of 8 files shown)


gulpfile.js
```diff
@@ -86,7 +84,8 @@
         // Our own files, we put this in a separate array to make sure we run
         // ng-annotate on it
         internal: [
-          'src/assets/javascripts/**/*.js'
+          'src/assets/javascripts/**/*.js',
+          '!src/assets/javascripts/{angularlib}/**/*.js'
```
redconfetti: Is this how you exclude a certain directory from the wildcard?

christianv: @redconfetti indeed

src/assets/javascripts/angular/services/analyticsService.js
```diff
@@ -13,9 +10,7 @@
      * More info on https://developers.google.com/analytics/devguides/collection/analyticsjs/events
      */
     var sendEvent = function(category, action, label) {
-      if (isGaAvailable) {
-        $window.ga('send', 'event', category, action, label);
-      }
+      window.ga('send', 'event', category, action, label);
```
redconfetti: Why are we using `window` instead of `$window` now?

christianv: @redconfetti actually didn't work with `$window` in this case. Probably because the Google Analytics code (which we don't have control over) was setting variables on the `window` object after `$window` was instantiated.

PR #3925, opened 2015-06-17
CLC-5108 - Corrected total steps for course creation by CCN.

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-5108

* Also added tests for proper step calculation


johncrossman:
Makes sense.

PR #3926, opened 2015-06-17
CLC-5395 - Modify workflow of Create a course site tool

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-5395


johncrossman:
Looks good, @redconfetti

PR #3928, opened 2015-06-18
CLC-5106 Show webcast recordings in Recent Activity

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5106

- Webcast recordings from current-term courses dated within the last ten days will appear in Recent Activity.
- A new 'linkText' property overrides the 'More info' text with 'View webcast.'
- Clicking 'View webcast' passes the recording's YouTube ID to the Class Info page and pre-selects the proper recording in the Webcasts widget.

### DIFF
--- a/app/models/my_academics/academics_module.rb
+++ b/app/models/my_academics/academics_module.rb
@@ -1,5 +1,6 @@
 module MyAcademics
   module AcademicsModule
+    extend self
 
     def initialize(uid)
       @uid = uid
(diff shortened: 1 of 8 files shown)


paulkerschen:
Note for @raydavis and @johncrossman in particular: adding this feature means that My Activities now hits Webcast::Recordings and CampusOracle::UserCourses in order to build its feed. My hope is that caching will take care of any performance issues in production, but it's possible that uncached load tests will take a hit.

johncrossman:
This looks good to me, @paulkerschen  Thanks for explaining the workings of cutoff_date. I think caching will work in our favor here - I don't see a sign of performance hit.

PR #3931, opened 2015-06-18
CLC-5453, extra args for tab api; cleanup in find_canvas_course_tab()

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5453

This does not resolve the Jira but I would like to run webcast-cron with 'id' passed in body of the HTTP PUT. I'm also passing 'position' and 'visibility' to rule out their necessity.


paulkerschen:
Reasonable enough!

PR #3933, opened 2015-06-18
CLC-5472, New feature flag: webcast_sign_up_on_calcentral

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5472

Modified logic in can_view_webcast_sign_up?. Feature flag will be false in upcoming release.

### DIFF
--- a/spec/policies/authentication_state_policy_spec.rb
+++ b/spec/policies/authentication_state_policy_spec.rb
@@ -1,5 +1,3 @@
-require "spec_helper"
-
 describe AuthenticationStatePolicy do
   let(:session_state) do
     {'user_id' => user_id, 'original_user_id' => original_user_id, 'lti_authenticated_only' => lti_authenticated_only}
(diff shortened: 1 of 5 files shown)


spec/policies/authentication_state_policy_spec.rb
```diff
@@ -188,4 +186,36 @@
     end
   end
 
+  describe '#can_view_webcast_sign_up?' do
+    subject { AuthenticationState.new(session_state).policy.can_view_webcast_sign_up? }
+    context 'when feature flag is true' do
+      before { Settings.features.webcast_sign_up_on_calcentral = true }
```
raydavis: @johncrossman , assigning the static variable adds unnecessary chanciness to the tests. I'd suggest configuring the flag as ``true`` in ``test.yml``, and using ``allow(Settings.features).to receive(:webcast_sign_up_on_calentral).and_return(false)`` to test that the flag works.

raydavis:
RSpec failure.

spec/policies/authentication_state_policy_spec.rb
johncrossman: Thanks @raydavis  Fixed. I'll follow up on spec failure

johncrossman:
Mr. T is happy

PR #3934, opened 2015-06-19
CLC-5396 - Remove SIS import on Add a Person tool

redconfetti:
https://jira.ets.berkeley.edu/jira/browse/CLC-5396

Actually only performs SIS User Import if the user is not already present within Canvas. I had to make an executive decision... like a baws

like a baws


raydavis:
You're the bawss!

PR #3936, opened 2015-06-19
CLC-5476: fix two WebDriver tests

pfarestveit:
https://jira.ets.berkeley.edu/jira/browse/CLC-5476

Fixes two tests: one that sorts and filters Oski's financial transactions and another that tests Up Next.


johncrossman:
I see no evil.

PR #3943, opened 2015-06-22
CLC-5478 - Bad spacing in "Status and Blocks" header

christianv:
https://jira.ets.berkeley.edu/jira/browse/CLC-5478

Different approach to fix the bug described in CLC-5478 after reading @raydavis' comment


raydavis:
Thank you for the revision, @christianv !

PR #3944, opened 2015-06-22
CLC-5488: fix QA webcast test

pfarestveit:
https://jira.ets.berkeley.edu/jira/browse/CLC-5488


pfarestveit:
Added a further check for the "report a problem" link and slightly more reliable interaction with the iframe content.

johncrossman:
I see goodness.

PR #3945, opened 2015-06-23
CLC-5491 - IE doesn't like webcasts anymore

christianv:
https://jira.ets.berkeley.edu/jira/browse/CLC-5491


raydavis:
Thank you for jumping on this, @christianv

PR #3946, opened 2015-06-23
CLC-5419 Do not surface residency status during term transitions

paulkerschen:

Back to the blunt hammer! No one, resident or not, sees any residency status during term transitions including summer.


raydavis:
https://www.youtube.com/watch?v=vEHwy3quc14

PR #3952, opened 2015-06-23
[QA] CLC-5419 Do not surface residency status during term transitions

paulkerschen:

QA version of #3946.


johncrossman:
Looks good. Matches master PR.

PR #3956, opened 2015-06-24
CLC-5419: fix Selenium residency test

pfarestveit:

Suppressing residency breaks QA's test, which expects it to be visible regardless of term.  The test now checks for a transition term before assuming residency is present.


johncrossman:
Lots o' goodness

PR #3966, opened 2015-06-25
CLC-5488: Selenium test for webcast fix

pfarestveit:
https://jira.ets.berkeley.edu/jira/browse/CLC-5488

The previous test fix for the webcast UI changes didn't handle the UI well enough.  This should provide better timing while elements load and better control of elements in the iframe.


johncrossman:
One out of two Travis builds ain't bad.

PR #3968, opened 2015-06-25
CLC-2107 CLC-5482 Set source of course-site activities to campus course code

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5482 has incorporated much of https://jira.ets.berkeley.edu/jira/browse/CLC-2107.

See CLC-5482 on the motivation for the `activityFactory.js` changes, which cover the rare case of a course site associated with multiple campus courses.


raydavis:
Thank you so much for finally implementing the original Dashboard design, @paulkerschen . I look forward to seeing it in action.

PR #3970, opened 2015-06-26
CLC-5501 - Extend scrolling support in canvas-customization.js

nicolaasmatthijs:
In order to provide better scrolling support between an LTI iFrame and the parent container, I've added a `getScrollInformation` event that will respond with the scroll information such as:

- LTI iFrame height
- Parent container height
- Current scroll position
- Current distance to bottom of screen

All other scrolling support events will now also respond with the same scroll information.

This will make it possible to add infinite scrolling capabilities to tools such as Collabosphere.

### DIFF
--- a/public/canvas/canvas-customization.js
+++ b/public/canvas/canvas-customization.js
@@ -348,25 +348,60 @@
    * Resizing the iFrame based on its content is handled by Instructure's `public/javascripts/tool_inline.js`
    * file, and it determines the message format we use.
    *
-   * We include the following custom event types:
-   *
-   *  - Scroll the parent container to a specified position:
-   *    `{subject: 'changeParent', scrollTo: <scrollPosition>}`
-   *
-   *  - Scroll the parent container to the top of the screen:
-   *    `{subject: 'changeParent', scrollToTop: true}`
+   * The following custom event is provided for modifying the URL of the parent container:
    *
    *  - Change the location of the parent container:
-   *    `{subject: 'changeParent', parentLocation: <newLocation>}`
+   *    ```
+   *     {
+   *       subject: 'changeParent',
+   *       parentLocation: <newLocation>
+   *     }
+   *    ```
    *
-   *  - Get the scroll position of the parent container:
-   *    `{subject: 'getScrollPosition'}`
-   *    This will respond with a window event back to the LTI iFrame with the following message:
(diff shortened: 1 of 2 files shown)


public/canvas/canvas-customization.js
```diff
@@ -348,25 +348,60 @@
...
+   *       subject: 'getScrollInformation'
+   *     }
+   *    ```
    *
-   *  - Get the scroll position of the parent container:
-   *    `{subject: 'getScrollPosition'}`
-   *    This will respond with a window event back to the LTI iFrame with the following message:
-   *    `{scrollPosition: <currentScrollPosition>}`
-   *
-   *  - If the iFrame is so enormous as to hit the 5000px limit in Instructure's code,
```
raydavis: Is the ``resizeLargeFrame`` workaround no longer needed?

nicolaasmatthijs: You can use 

```
{
  subject: 'changeParent',
  height: <height>
}
```

instead, which is a little more in line with the other `changeParent` messages. I've modified the one place I could find where `resizeLargeFrame` was being used.

paulkerschen: @nicolaasmatthijs - if I'm reading right, it looks like the `changeParent` message sets the height attribute on a different element than the existing `resizeLargeFrame` message. It might still work to defeat the 5000px limit, but we should test with roster photos from a large course (the original case that required us to put in this workaround).

raydavis: Thanks for the follow-up, @paulkerschen . I definitely agree that we should test ASAP.

nicolaasmatthijs: Setting the height on the wrapper container and setting `height_overriden` on it appears to be the desired approach when reading the Canvas resizing code. The Asset Library list in the Collabosphere tools also often surpasses the 5000px limit and the approach in this PR has been extensively tested on that (it was also part of the Datacultures production code), but we should definitely also verify this with the roster photos tool.

paulkerschen: @raydavis @nicolaasmatthijs Unfortunately, it looks like resizing is indeed broken for the roster photos tool. I'll create a new issue for it.

paulkerschen: https://jira.ets.berkeley.edu/jira/browse/CLC-5502

PR #3973, opened 2015-06-29
CLC-5503 - application.js is no longer hashed

christianv:
https://jira.ets.berkeley.edu/jira/browse/CLC-5503


christianv:
@winstonkim just as a fyi - we should make sure we always have a `return` at the end of each task.

johncrossman:
Looks good.

PR #3981, opened 2015-07-02
CLC-5504 Fix goof in deploy.rb: Only run update-build.sh once per host

ctweney:
https://jira.ets.berkeley.edu/jira/browse/CLC-5504

Followup PR that fixes the broken build from #3974 . My mistake was omitting the "hosts" param in that run call.


raydavis:
Thanks, @ctweney . Given the low priority on this, and our current short-handedness, is it OK if we wait until Monday for a merge?

ctweney:
Fine by me @raydavis !

PR #3987, opened 2015-07-06
CLC-5520 Fix course associations for Canvas group sites

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5520

Fixes an association that was broken by integer/string mismatch, and extends the course-site improvement in #3968 to cover group sites as well.


raydavis:
Good catch & good fix!

PR #3988, opened 2015-07-07
CLC-5453, webcast LTI app default=enabled for the moment

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5453

Canvas admin console does not give me access to the Webcast LTI app's 'default' property so I'm going to temporarily modify the source config. I've created https://jira.ets.berkeley.edu/jira/browse/CLC-5521 to revert this change prior to end of sprint 59.


raydavis:
Left a comment at the JIRA, but I'm guessing this might be overkill.

PR #3990, opened 2015-07-07
CLC-5371 Add grading-in-progress term information to My Classes feed

paulkerschen:
See https://jira.ets.berkeley.edu/jira/browse/CLC-5371 for motivation.

A `gradingInProgressClasses` key is added to My Classes for the four-week period after end of term, when grading information is likely to show up. This surfaces no new information in My Classes, but lets the Recent Activity widget associate announcements with recently completed courses.

It might have been more elegant to arrange the My Classes semesters in an array, as with My Academics, but it's clearly more pragmatic to leave existing feed elements alone.

### DIFF
--- a/app/models/my_activities/dashboard_sites.rb
+++ b/app/models/my_activities/dashboard_sites.rb
@@ -1,9 +1,9 @@
 module MyActivities
   class DashboardSites
     def self.fetch(uid, options={})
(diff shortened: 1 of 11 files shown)


raydavis:
I agree with your pragmatism, @paulkerschen , but am glad that this helps open some possibility of CLC-206 being revisited (in some form) later.

PR #3991, opened 2015-07-07
SISRP-4446 - Write spec to cover ping_controller

Shuotong:
https://jira.berkeley.edu/browse/SISRP-4446

### DIFF
--- a/app/controllers/ping_controller.rb
+++ b/app/controllers/ping_controller.rb
@@ -4,7 +4,7 @@ class PingController < ApplicationController
     # IST's nagios and our check-alive.sh script use this endpoint to tell whether the server's up.
     # Don't modify its content unless you have general agreement that it's necessary to do so.
     begin
-    ping_state = ping
+      ping_state = ping
     rescue StandardError => e
       Rails.logger.fatal e
     end
--- a/app/controllers/ping_controller.rb
+++ b/app/controllers/ping_controller.rb
@@ -3,7 +3,11 @@ class PingController < ApplicationController
   def do
     # IST's nagios and our check-alive.sh script use this endpoint to tell whether the server's up.
     # Don't modify its content unless you have general agreement that it's necessary to do so.
+    begin
     ping_state = ping
+    rescue StandardError => e
+      Rails.logger.fatal e
+    end
     if ping_state
       feed = {
         server_alive: true
--- a/spec/controllers/ping_controller_spec.rb
+++ b/spec/controllers/ping_controller_spec.rb
@@ -47,26 +47,26 @@ describe PingController do
       end
 
       context 'do not do background jobs check' do
+        let (:expected) { { 'server_alive' => true }.to_json }
         before do
           Settings.features.stub(:background_jobs_check).and_return(false)
-          @expected = { :server_alive => true }.to_json
         end
         it 'renders a json file with server status' do
           get :do
-          response.body.should == @expected
+          expect(response.body).to eq expected
         end
       end
 
       context 'do background jobs check' do
+        let (:example_background_jobs_check) { { 'ets-calcentral-prod-01' => 'OK', 'ets-calcentral-prod-02' => 'OK', 'ets-calcentral-prod-03' => 'OK', 'status' => 'OK', 'last_ping' => '2015-07-06T16:31:50.161-07:00'}.to_json }
+        let (:expected) { { 'server_alive' => true, 'background_jobs_check' => example_background_jobs_check }.to_json }
         before do
           Settings.features.stub(:background_jobs_check).and_return(true)
-          @example_backgroundjobscheck = { :"ets-calcentral-prod-01" => "OK", :"ets-calcentral-prod-02" => "OK",:"ets-calcentral-prod-03" => "OK",:"status" => "OK","last_ping" => "2015-07-06T16:31:50.161-07:00"}.to_json
-          BackgroundJobsCheck.any_instance.stub(:get_feed => @example_backgroundjobscheck)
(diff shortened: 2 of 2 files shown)


app/controllers/ping_controller.rb
```diff
@@ -3,7 +3,11 @@ class PingController < ApplicationController
   def do
     # IST's nagios and our check-alive.sh script use this endpoint to tell whether the server's up.
     # Don't modify its content unless you have general agreement that it's necessary to do so.
+    begin
```
ctweney: Please fix indentation in this block.

spec/controllers/ping_controller_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+    context 'Both databases available' do
+      before do
+        User::Data.stub(:database_alive?).and_return(true)
+        CampusOracle::Queries.stub(:database_alive?).and_return(true)
+      end
+
+      context 'do not do background jobs check' do
+        before do
+          Settings.features.stub(:background_jobs_check).and_return(false)
+          @expected = { :server_alive => true }.to_json
```
ctweney: Please inline the expected variable instead of declaring it here.

spec/controllers/ping_controller_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+        it 'renders a json file with server status' do
+          get :do
+          response.body.should == @expected
+        end
+      end
+
+      context 'do background jobs check' do
+        before do
+          Settings.features.stub(:background_jobs_check).and_return(true)
+          @example_backgroundjobscheck = { :"ets-calcentral-prod-01" => "OK", :"ets-calcentral-prod-02" => "OK",:"ets-calcentral-prod-03" => "OK",:"status" => "OK","last_ping" => "2015-07-06T16:31:50.161-07:00"}.to_json
```
ctweney: Please declar example_backgroundjobscheck using "let" like so (outside of the before block but inside of the context block): 

```
let(:example_background_jobs_check) { { :"ets-calcentral-prod-01" => "OK", :"ets-calcentral-prod-02" => "OK",:"ets-calcentral-prod-03" => "OK",:"status" => "OK","last_ping" => "2015-07-06T16:31:50.161-07:00"}.to_json } 
```

Then you can refer to example_background_jobs_check in the code without an @ or a :.

spec/controllers/ping_controller_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+          BackgroundJobsCheck.any_instance.stub(:get_feed => @example_backgroundjobscheck)
+          @expected = { :server_alive => true, :background_jobs_check => @example_backgroundjobscheck }.to_json
```
ctweney: Please inline expected as above.

spec/controllers/ping_controller_spec.rb
ctweney: Also, you can drop the : in front of the hash keys, and use single quotes, like so:

```
'ets-calcentral-prod-01' => 'OK' ... etc .
```

spec/controllers/ping_controller_spec.rb
```diff
@@ -0,0 +1,73 @@
+require "spec_helper"
+
+describe PingController do
+
+
+    context 'User database unavailable' do
```
christianv: @Shuotong shouldn't we always use 2 spaces for indentation?

spec/controllers/ping_controller_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+describe PingController do
+
+  context 'User database unavailable' do
+    before do
+      User::Data.stub(:database_alive?).and_return(false)
+      CampusOracle::Queries.stub(:database_alive?).and_return(true)
+    end
+    it 'raises error' do
+      get :do
+      expect {raise 'CalCentral database is currently unavailable'}.to raise_error('CalCentral database is currently unavailable')
```
ctweney: I think this line is a tautology -- it isn't testing anything except the block you provide right here. In PingController, the raised error is caught and logged and never rethrown, so you can't test for it directly. Instead of expecting the error to be raised, check for the logging of the fatal error, something like this: 

```
expect(Rails.logger).to receive(:fatal)
```

That would cover the controller code nicely (every line of the raise & response to raised error would be covered).

spec/controllers/ping_controller_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+  end
+
+  context 'Campus database unavailable' do
+    before do
+      User::Data.stub(:database_alive?).and_return(true)
+      CampusOracle::Queries.stub(:database_alive?).and_return(false)
+    end
+    it 'raises error' do
+      get :do
+      expect {raise 'Campus database is currently unavailable'}.to raise_error('Campus database is currently unavailable')
```
ctweney: Same as above expect-raise block.

spec/controllers/ping_controller_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+  end
+
+  context 'Both databases unavailable' do
+    before do
+      User::Data.stub(:database_alive?).and_return(false)
+      CampusOracle::Queries.stub(:database_alive?).and_return(false)
+    end
+    it 'raises error' do
+      get :do
+      expect {raise 'CalCentral database is currently unavailable'}.to raise_error('CalCentral database is currently unavailable')
```
ctweney: Same as above expect-raise block.

ctweney:
Looks good - thanks for fixing the way it logs database unavailable problems, too!

PR #3993, opened 2015-07-07
CLC-5515 Do not request textbooks if no primary sections are available

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5515


raydavis:
Travis is half-full.

PR #3994, opened 2015-07-08
SISRP-5127 - Dev/Implement Front-End T&C page with dummy data

christianv:
https://jira.berkeley.edu/browse/SISRP-5127


public/dummy/json/finaid_summary.json
```diff
@@ -0,0 +1,35 @@
+{
+  "finaidSummary": {
+    "finaidYears": [
```
ctweney: This structure is so different from the way we represented aid years in finaid_award.json and finaid_budget.json. Those 2 had an array of terms each of which had a startTerm and endTerm property. I know that's not quite the right model for an aid year, which consists of up to 3 terms, and has a name for the whole aid year e.g. "Fall 2014 - Spring 2015 - Summer 2015" (maybe Summer isn't part of the name). 

I'd like to get a standard representation of aid year that we can port to the other feeds. The one here looks more correct than the one in finaid_awards.json, but it's not complete either. Won't we need the detail of terms that belong to the aid year?

src/assets/templates/widgets/finaid_summary.html
```diff
@@ -0,0 +1,49 @@
...
+        ></div>
+        <div
+          data-ng-if="finaidYear.termsAndConditions.approved && finaidSummary.title4.approved === null"
+          data-cc-finaid-summary-message-directive
+          data-item="finaidSummary.title4"
+          data-button-text="'Complete Title IV'"
+          data-finaid-year-id="finaidYear.id"
+        ></div>
+        <div data-ng-if="finaidYear.termsAndConditions.approved && finaidSummary.title4.approved !== null">
+          T&C approved - Title IV has been answered
```
ctweney: I think you want &amp; here.

ctweney:
2 comments, but neither of them is a merge blocker. LGTM otherwise!

public/dummy/json/finaid_summary.json
christianv: @ctweney totally agree that we should be more consistent, right now they're actually still figuring it out and hopefully we'll get a better answer this Thursday.

ctweney: Yes, let's defer a decision on the proper structure until the functional team & Bernie agree on the text. But I expect we'll need an outer aidyear object that contains a ID, a label (TBD), and an array of terms each with their own label and ID. It seems the functional disagreement comes from confusing the two levels.

PR #3996, opened 2015-07-08
CLC-5526, Canvas Tab API bug + Webcast cron = workaround B

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5526

When we use the Canvas Tab API we must avoid collateral damage. I.e., changes to 'hidden' property on non-targeted tabs. Therefore, the extraneous tabs are self-healing.


raydavis:
Sweet!

PR #4006, opened 2015-07-14
SISRP-4447 - Write spec to cover PhotoController

winstonkim:
https://jira.berkeley.edu/browse/SISRP-4447

### DIFF
--- a/spec/controllers/photo_controller_spec.rb
+++ b/spec/controllers/photo_controller_spec.rb
@@ -3,31 +3,30 @@ require 'spec_helper'
 describe PhotoController do
   let(:user_id) { random_id }
 
-  context "when serving a users photo" do
-    context "when user is logged in" do
+  context 'when serving a users photo' do
+    context 'when user is logged in' do
         before do
           session['user_id'] = random_id
         end
-      context "when user has photo" do
+      context 'when user has photo' do
         before do
           test_photo_object = {'photo' => 'photo_binary_content'}
-          CampusOracle::Queries.stub(:get_photo).and_return(test_photo_object)
+          allow(CampusOracle::Queries).to receive(:get_photo).and_return(test_photo_object)
         end
-        it "renders users raw image" do
-          expect(controller).to receive(:send_data).with('photo_binary_content', type: 'image/jpeg', disposition: 'inline'){
-            controller.render nothing: true
-          }
+        it 'renders users raw image' do
           get :my_photo
+          expect(response.status).to eq 200
+          expect(response.body).to eq 'photo_binary_content'
         end
       end
-      context "when user has no photo" do
+      context 'when user has no photo' do
         before do
-          test_photo_object = nil
-          CampusOracle::Queries.stub(:get_photo).and_return(test_photo_object)
+          allow(CampusOracle::Queries).to receive(:get_photo).and_return(nil)
         end
-        it "renders users raw image" do
-          allow(controller).to receive(:send_data).with(nil, type: 'image/jpeg', disposition: 'inline').and_return(true)
+        it 'renders users raw image' do
           get :my_photo
+          expect(response.status).to eq 200
+          expect(response.body).to eq ' '
         end
       end
     end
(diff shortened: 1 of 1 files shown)


spec/controllers/photo_controller_spec.rb
```diff
@@ -1,29 +1,34 @@
 require 'spec_helper'
 
 describe PhotoController do
+  let(:user_id) { random_id }
 
   context "when serving a users photo" do
```
ctweney: Please change double quotes to single quotes throughout the file (except where double quotes are needed for string interpolation).

spec/controllers/photo_controller_spec.rb
```diff
@@ -1,29 +1,34 @@
...
-          test_photo_object = {:photo => 'photo_binary_content'}
+          test_photo_object = {'photo' => 'photo_binary_content'}
           CampusOracle::Queries.stub(:get_photo).and_return(test_photo_object)
         end
         it "renders users raw image" do
-          allow(controller).to receive(:send_data).with('photo_binary_content', type: 'image/jpeg', disposition: 'inline').and_return(true)
+          expect(controller).to receive(:send_data).with('photo_binary_content', type: 'image/jpeg', disposition: 'inline'){
+            controller.render nothing: true
+          }
           get :my_photo
```
ctweney: This test needs some expectations!
Something along the lines of 
```
expect(response.status).to eq 200
```
And maybe some checks on the response.body content as well.

spec/controllers/photo_controller_spec.rb
```diff
@@ -1,29 +1,34 @@
...
+        end
       context "when user has photo" do
         before do
-          test_photo_object = {:photo => 'photo_binary_content'}
+          test_photo_object = {'photo' => 'photo_binary_content'}
           CampusOracle::Queries.stub(:get_photo).and_return(test_photo_object)
         end
         it "renders users raw image" do
-          allow(controller).to receive(:send_data).with('photo_binary_content', type: 'image/jpeg', disposition: 'inline').and_return(true)
+          expect(controller).to receive(:send_data).with('photo_binary_content', type: 'image/jpeg', disposition: 'inline'){
```
ctweney: I think this block is not necessary. Set expectations on the response instead.

spec/controllers/photo_controller_spec.rb
```diff
@@ -1,29 +1,34 @@
...
-
-    context "when user is not logged in" do
-      before { session['user_id'] = nil }
-      it "returns 401 error with no body" do
-        get :my_photo
-        expect(response.status).to eq 401
-        expect(response.body).to eq " "
+      context "when user has no photo" do
+        before do
+          test_photo_object = nil
```
ctweney: You can collapse this to a single line: 
```
CampusOracle::Queries.stub(:get_photo).and_return(nil)
```

spec/controllers/photo_controller_spec.rb
```diff
@@ -1,29 +1,34 @@
...
+          CampusOracle::Queries.stub(:get_photo).and_return(test_photo_object)
+        end
+        it "renders users raw image" do
+          allow(controller).to receive(:send_data).with(nil, type: 'image/jpeg', disposition: 'inline').and_return(true)
+          get :my_photo
```
ctweney: Again, the allow() shouldn't be necessary; instead, set expectations about what the response will be.

spec/controllers/photo_controller_spec.rb
paulkerschen: @winstonkim I believe that the `.stub` is also deprecated these days; the recommended way to do it now would be `allow(CampusOracle::Queries).to receive(:get_photo).and_return(nil)`.

winstonkim:
@ctweney @paulkerschen Thanks for the comments! I made the changes. Let me know if there is anything else that should be changed.

ctweney:
Travis is half-satisfied. I am completely satisfied.

PR #4011, opened 2015-07-17
Merge remote-tracking branch 'ets/master' into merge-sis

christianv:
+ do updates to JS files


christianv:
Self merging - definitely let me know if there are further changes we should make

PR #4014, opened 2015-07-17
CLC-3229 Bring Canvas proxy responses in line with other proxies

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-3229

This is the last large Canvas-cleanup bomb to be dropped. Reworking Canvas proxy error responses to include status codes and messages entailed doing the same thing to successful responses for symmetry's sake.

Canvas proxy requests now come wrapped in a hash object with `statusCode` and `body` for successful responses, and `statusCode` and `error` for error responses. This wasn't the only possible approach - the other proxies do vary somewhat in their behavior - but this brings them much more in line with their neighbors. 

The interesting parts of the PR are in `/app/models/canvas`. In addition to the response wrapping:
- Proxy subclasses are slimmed down by extracting a good deal of shared logic into `Canvas::Proxy`, including request paging, JSON parsing and optional caching. 
- In turn `Canvas::Proxy`, which was starting to look crowded, cedes to `Canvas::Terms` a large set of class methods dealing with Canvas term codes.
- Some tangled logic in `Canvas::SisImport` is combed out.

The boring parts of the PR are the specs and higher-order classes that receive Canvas proxy responses and had to be tweaked to account for the new wrapping.

Given the scope of this and other recent Canvas changes, I hope to spend some time checking for regressions on the dev servers before they land in QA's lap.


raydavis:
Slick refactoring!

PR #4016, opened 2015-07-20
CLC-5564 Have paged Canvas proxies return empty array on failure

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5564

`Canvas::Proxy#paged_get` is changed to patch leaky exception handling and guarantee the response body that other classes are expecting, by default an empty array.

Tests are tightened up, and the expectation is made clearer that on request failure, a paged proxy will return an empty array under `:body`, but an unpaged proxy will not return `:body`. 

For ease in testing, some short Canvas proxy methods are split into two even shorter methods: one returning a full response with headers, and another returning the body alone.


raydavis:
Thanks for centralizing this, @paulkerschen

PR #4018, opened 2015-07-21
CLC-5558 Change default mailing list settings: moderated, less verbose

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5558

@raydavis should sign off before merge.

- New members default to moderated.
- Production Calmail seems to evaluate Boolean query parameters as true, no matter what the value. Only if they are omitted entirely from the query will they default to false.


raydavis:
Signing off!

PR #4019, opened 2015-07-21
SISRP-5145 - Dev/Implement Front-End: Cost of Attendance Card

christianv:
https://jira.berkeley.edu/browse/SISRP-5145


ctweney:
This looks right to me as a start. I think there are some slight differences between your dummy JSON and the JSON that the back-end will produce from the feeds that we currently have. Let's resolve those slight differences after my PR for the back-end goes in (which will be soon, today probably).

PR #4021, opened 2015-07-21
CLC-5571 Use correct PUT endpoint for Canvas::ExternalTools updates

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5571

Continuing to save #4014 from itself.


johncrossman:
Thanks @paulkerschen  So simple

PR #4022, opened 2015-07-21
CLC-5567 - Do not show empty "Phase I" display for Tele-BEARS

christianv:
https://jira.ets.berkeley.edu/jira/browse/CLC-5567


raydavis:
Thanks, @christianv !

PR #4024, opened 2015-07-21
CLC-5572 Expect unwrapped response from Canvas::SisImport POST

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5572


johncrossman:
Makes sense to me.

PR #4028, opened 2015-07-22
CLC-5553, rename Webcast service to 'Course Capture'

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5553
https://jira.ets.berkeley.edu/jira/browse/WCT-5478

Includes name change in:
* LTI app
* My Activities
* Widget

The term "webcast" is still pervasive in the code (e.g., 'type' attribute in my_activities feed). My changes only apply to user-facing text.


paulkerschen:
"Recordings of" does read better than "Webcasts for."

PR #4031, opened 2015-07-23
CLC-5540 Treat Canvas::SisCourse requests as existence checks

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5540


johncrossman:
Looks good.

PR #4035, opened 2015-07-24
OEC-657 Informatively report bad characters in CSV files

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/OEC-657


johncrossman:
Thanks for coding an Ops friendly rescue block @paulkerschen

PR #4039, opened 2015-07-27
CLC-5576 Add course site creator to existing section rather than new one

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5576


raydavis:
Hasta la vista, DEFSEC!

PR #4041, opened 2015-07-29
CLC-5579 - CamelCase the keys on the 'features' property on the my status api

christianv:
https://jira.ets.berkeley.edu/jira/browse/CLC-5579


raydavis:
@christianv , can you update https://github.com/ets-berkeley-edu/calcentral/blob/master/public/dummy/json/status.json and https://github.com/ets-berkeley-edu/calcentral/blob/master/public/dummy/json/status_loggedout.json to match?

christianv:
@raydavis thanks for spotting, should have been updated.

raydavis:
And it is, thanks!

PR #4044, opened 2015-07-30
CLC-4449 Handle masquerade state in LTI authentication

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-4449


raydavis:
Look forward to reading the log files!

PR #4047, opened 2015-07-30
CLC-5545 - Check whether watchify would be able to look for newly added files

winstonkim:
https://jira.ets.berkeley.edu/jira/browse/CLC-5545


christianv:
lgtm

PR #4050, opened 2015-08-03
SISRP-5164 Rename feed to financial_aid_data

ctweney:
https://jira.berkeley.edu/browse/SISRP-5164

Note: Settings property change will require tweaks on sis-dev after merge.


christianv:
@ctweney thanks!

PR #4051, opened 2015-08-03
CLC-5591 Up Next: fall back to email address when display name is not available

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5591


raydavis:
Thanks, @paulkerschen ! I think we've already agreed that a QA PR is OK, as well.

PR #4056, opened 2015-08-04
CLC-5598 - Selenium test for My Classes UI

pfarestveit:
https://jira.ets.berkeley.edu/jira/browse/CLC-5598

This test covers the following:
* Expected display of enrolled classes, teaching classes, and site memberships
* Messaging for users with no classes or memberships
* Heading display for users with various mixes of classes and/or memberships

The spec also produces a CSV that lists the My Classes card content for each UID, which will be used for managing the pool of test UIDs and for manual testing.


paulkerschen:
Looks good! That's a complicated widget these days, isn't it.

PR #4057, opened 2015-08-05
SISRP-5145 - Dev/Implement Front-End: Cost of Attendance Card

christianv:
https://jira.berkeley.edu/browse/SISRP-5145


ctweney:
The test failure in campus_solutions_controller_spec.rb is caused by the changes in the mock aid_years.xml feed. Otherwise, LGTM.

christianv:
@ctweney pushed up a commit that hopefully fixes the test breakages.

PR #4060, opened 2015-08-05
CLC-3181 Canvas CSV imports: use full names instead of first and last

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-3181


raydavis:
V for Victory.

PR #4062, opened 2015-08-06
SISRP-5147 and SISRP-5148 - T&C and Title4 POST endpoints

ctweney:
Addresses https://jira.berkeley.edu/browse/SISRP-5147 and https://jira.berkeley.edu/browse/SISRP-5148 .

The front end does not yet call these endpoints. Here's how to do that: 
POST /api/campus_solutions/terms_and_conditions?uc_response=Y&aid_year=2016
POST /api/campus_solutions/title4?uc_response=N

Or put params in the form body and Rails will figure it out. 

Both T&C and Title4 go thru the hub. Params are sent to CS via the hub as an XML document. We use Rails to_xml to convert. It needs a couple of specific options because CS will choke on whitespace/newline.

### DIFF
--- a/app/controllers/campus_solutions_controller.rb
+++ b/app/controllers/campus_solutions_controller.rb
@@ -33,4 +33,9 @@ class CampusSolutionsController < ApplicationController
     render json: model.update(params)
   end
 
+  def title4
+    model = CampusSolutions::MyTitle4.from_session(session)
+    render json: model.update(params)
+  end
+
 end
--- a/app/controllers/campus_solutions_controller.rb
+++ b/app/controllers/campus_solutions_controller.rb
@@ -28,4 +28,9 @@ class CampusSolutionsController < ApplicationController
     render json: model.get_feed_as_json
   end
 
+  def terms_and_conditions
+    model = CampusSolutions::MyTermsAndConditions.from_session(session)
+    render json: model.update(params)
+  end
+
 end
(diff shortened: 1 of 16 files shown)


christianv:
@ctweney would it be possible to change it to this:

Doing a POST to: `POST /api/campus_solutions/terms_and_conditions` with the following POST params:
* "response": true
* "aidYear`: "2016"

We would need to do some conversion on the CalCentral back-end side.
And it might be good to look at the POST in the google tasks piece to see how we're sending the POSTs over from the front-end.

ctweney:
Sure, I'll revise when I'm back later this afternoon.

christianv:
@ctweney that would be great, thanks!

ctweney:
@christianv I pushed some new commits that fix the test failure and give you a camelCased JSON API surface. Here is an example post: 

```
POST /api/campus_solutions/terms_and_conditions HTTP/1.1
Host: localhost:3000
Authorization: XXXXXX
Content-Type: application/json
Cache-Control: no-cache
Postman-Token: dd85b234-0749-4865-bb67-c42c867640ac

{"response":"Y","aidYear":"2016"}
```

This is consistent with the tasks post. 

The camelCasing was not too hard, but unfortunately the value for "response" is still "Y" or "N". To handle your pure boolean, I'd have to build a value-mapping-and-transformation abstraction in addition to the fieldname-mapping I already have. I think that's a complexity too far.

PR #4071, opened 2015-08-11
CLC-5613 Ugly temporary fix for our double-alerts-feed marquee

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-5613

If this tests out on calcentral-dev, it will have to go into the interim release as well.


johncrossman:
Acceptable given the follow up in next release. @paulkerschen can chime in, too.

paulkerschen:
Chiming.

PR #4074, opened 2015-08-11
CLC-5649 - Add an option to dynamically disable the ccAccessibleFocusDirective

christianv:
https://jira.ets.berkeley.edu/jira/browse/CLC-5649


johncrossman:
Thanks for the inline docs @christianv  LGTM.

PR #4075, opened 2015-08-12
CLC-5639, spec as reference impl: get files within directory

johncrossman:

Specs to demonstrate 'get files in directory'.


paulkerschen:
Alles gut.

PR #4077, opened 2015-08-12
SISRP-6245 Expire cache for finaid endpoints when doing a POST

ctweney:
https://jira.berkeley.edu/browse/SISRP-6245

Whenever a POST to CS happens for T&C or Title4, the AidYears and FinancialAidData feeds get cache cleared as well. Also, other FinAid classes now participate correctly in the UserCacheExpiry scheme. 

Motivation for the RelatedCacheKeyTracker module: FinancialAidData feed needs two keys to work: uid and aid_year. This is because it's one-year-at-a-time. Since the feed needs two keys, so does its associated cache. Unfortunately, the UserCacheExpiry system (upon which HotPlate and LiveUpdates rely) can only expire based on uid. The set of uid-year keys cached by FinancialAidData remain unknown to UserCacheExpiry. So, we use RelatedCacheKeyTracker to record linked keys by uid as they go into the cache. That way we can expire them when UserCacheExpiry calls `klass.expire uid`. 

This malarkey would not be necessary if memcached allowed expiration based on key hierarchy or wildcard or similar. But then, it wouldn't be a key-value store if it did. 

FinancialAidExpiry is intended to invalidate cache of several feeds, but not all the user's feeds (which is what UserCacheExpiry does). We will have use of this pattern again and again as CS self-service endpoints get created. 

@raydavis and @paulkerschen , I'm curious to hear your input on the RelatedCacheKeyTracker solution.

### DIFF
--- a/lib/cache/related_cache_key_tracker.rb
+++ b/lib/cache/related_cache_key_tracker.rb
@@ -6,6 +6,10 @@ module Cache
     end
 
     module ClassMethods
+      def user_key(uid)
+        "related-cache-keys-#{uid}"
+      end
+
       def expire(uid=nil)
         super uid
         keys = related_cache_keys uid
@@ -13,11 +17,11 @@ module Cache
         keys.keys.each do |key|
           Rails.cache.delete key
         end
-        Rails.cache.delete "additional-keys-#{uid}"
+        Rails.cache.delete(user_key(uid))
       end
 
       def related_cache_keys(uid=nil)
-        Rails.cache.read("related-cache-keys-#{uid}") || {}
+        Rails.cache.read(user_key(uid)) || {}
       end
 
       def save_related_cache_key(uid=nil, related_key=nil)
@@ -25,7 +29,7 @@ module Cache
         return if keys[related_key].present?
         keys[related_key] = 1
         logger.debug "Writing related keys for uid #{uid}: #{keys}"
-        Rails.cache.write("related-cache-keys-#{uid}",
+        Rails.cache.write(user_key(uid),
                           keys,
                           :expires_in => Settings.maximum_expires_in,
                           :force => true)
--- a/lib/cache/related_cache_key_tracker.rb
+++ b/lib/cache/related_cache_key_tracker.rb
@@ -12,25 +12,25 @@ module Cache
 
       def expire(uid=nil)
         super uid
-        keys = related_cache_keys uid
-        logger.debug "Will now expire these associated keys for uid #{uid}: #{keys}"
-        keys.keys.each do |key|
-          Rails.cache.delete key
+        related_keys = related_cache_keys uid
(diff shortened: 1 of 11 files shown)


lib/cache/related_cache_key_tracker.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    module ClassMethods
+      def expire(uid=nil)
+        super uid
+        keys = related_cache_keys uid
+        logger.debug "Will now expire these associated keys for uid #{uid}: #{keys}"
+        keys.keys.each do |key|
+          Rails.cache.delete key
+        end
+        Rails.cache.delete "additional-keys-#{uid}"
```
paulkerschen: @ctweney Did you mean this to be `related-cache-keys`, or does `additional-keys` come up elsewhere?

ctweney: Oops! Copy-paste error. I'll fix by extracting a method.

lib/cache/related_cache_key_tracker.rb
```diff
@@ -0,0 +1,35 @@
...
+    def self.included base
+      base.extend ClassMethods
+    end
+
+    module ClassMethods
+      def expire(uid=nil)
+        super uid
+        keys = related_cache_keys uid
+        logger.debug "Will now expire these associated keys for uid #{uid}: #{keys}"
+        keys.keys.each do |key|
```
paulkerschen: @ctweney This made me scratch my head for a moment; then I worked out that it's using Hash instead of Set for performance reasons.

ctweney: Actually Ruby's Set uses Hash under the covers. The performance characteristics should be identical, and the semantics of Set would be prettier. I'll change to use Set -- assuming it's serializable in memcached -- I'll check to be sure.

paulkerschen:
@ctweney To me, this makes a lot of sense as a pattern for dealing with an issue that CalCentral's pure-dashboard functionality would never have encountered.

lib/cache/related_cache_key_tracker.rb
paulkerschen: @ctweney Since Set does use Hash, I'm not sure why Hash consistently benchmarks faster, but I've seen results in a couple of places implying that it does:

http://blog.x-aeon.com/2012/02/07/using-data-sets-in-ruby-performance-of-hash-vs-set/

raydavis:
@ctweney , I actually remember us and Yu-Hung discussing at some point whether we needed to implement this, but decided we could get by for a while with the UserCacheExpiry approach (a special case of the same idea). I'm for adding it to the toolkit.

lib/cache/related_cache_key_tracker.rb
ctweney: Wow, that benchmark code has Hash about 3x faster than Set on my machine. I'll refactor back to hash!

PR #4080, opened 2015-08-14
CLC-5651 Move current OEC code into OecLegacy module

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5651

Module is renamed, folders are moved, but YAML settings, Rake namespace and command-line interface are unchanged.


johncrossman:
Looks good, @paulkerschen. Includes changes to oec.rake

PR #4081, opened 2015-08-17
CLC-5639, DriveManager class can create dir and upload file; plus YAML scheme

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5646

Changes include:
* YAML scheme to support custom google api credentials
* CredentialStore class which aligns with Google API convention and follows our convention
* DriveManager provides CRUD operations. I don't see a need for a class per CRUD type.

### DIFF
--- a/app/models/google_apps/drive_manager.rb
+++ b/app/models/google_apps/drive_manager.rb
@@ -1,6 +1,8 @@
 module GoogleApps
   class DriveManager
 
+    include ClassLogger
+
     def initialize(credential_store)
       @credential_store = credential_store
     end
@@ -13,7 +15,7 @@ module GoogleApps
       dir.parents = [{ :id => parent_id }] if parent_id
       result = client.execute(:api_method => drive_api.files.insert, :body_object => dir)
       success = result.status == 200
-      Rails.logger.error "An error occurred: #{result.data['error']['message']}" unless success
+      logger.error "An error occurred: #{result.data['error']['message']}" unless success
       success ? result.data : nil
     end
 
@@ -31,7 +33,7 @@ module GoogleApps
         :media => media,
         :parameters => { :uploadType => 'multipart', :alt => 'json'})
       success = result.status == 200
-      Rails.logger.error "An error occurred: #{result.data['error']['message']}" unless success
+      logger.error "An error occurred: #{result.data['error']['message']}" unless success
       success ? result.data : nil
     end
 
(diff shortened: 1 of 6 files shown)


johncrossman:
Bamboo is happy: https://bamboo.ets.berkeley.edu/bamboo/browse/MYB-CJT-235

One more commit. The access_token override is now encapsulated in CredentialStore class. Notice that CredentialStore implements same as Google::APIClient::FileStore but does not cache credentials to disk.

paulkerschen:
NB Travis is half happy; the Google failures in the other half are intermittent failures that we've had for a long time, and for which we should probably open a ticket.

app/models/google_apps/drive_manager.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    def create_folder(title, parent_id)
+      client = get_google_api
+      drive_api = client.discovered_api('drive', 'v2')
+      metadata = { :title => title, :mimeType => 'application/vnd.google-apps.folder' }
+      dir = drive_api.files.insert.request_schema.new metadata
+      dir.parents = [{ :id => parent_id }] if parent_id
+      result = client.execute(:api_method => drive_api.files.insert, :body_object => dir)
+      success = result.status == 200
+      Rails.logger.error "An error occurred: #{result.data['error']['message']}" unless success
```
paulkerschen: @johncrossman If you include Junction's `ClassLogger` module in this class, calling `logger.error` instead of `Rails.logger.error` will give a more informative log message.

app/models/google_apps/drive_manager.rb
```diff
@@ -0,0 +1,47 @@
...
+      # Target directory is optional
+      file.parents = [{ :id => parent_id }] if parent_id
+      media = Google::APIClient::UploadIO.new(file_absolute_path, mime_type)
+      result = client.execute(
+        :api_method => drive_api.files.insert,
+        :body_object => file,
+        :media => media,
+        :parameters => { :uploadType => 'multipart', :alt => 'json'})
+      success = result.status == 200
+      Rails.logger.error "An error occurred: #{result.data['error']['message']}" unless success
```
paulkerschen: @johncrossman as above

paulkerschen:
@johncrossman This looks great and I can't wait to give it a test drive. One minor comment about logging.

johncrossman:
Thanks @paulkerschen  ClassLogger included.

PR #4083, opened 2015-08-17
CLC-5656 - JS error when you go to /campus unauthenticated

christianv:
https://jira.ets.berkeley.edu/jira/browse/CLC-5656


johncrossman:
Makes sense.

PR #4084, opened 2015-08-17
CLC-5639, find_folders_by_title and find_items_by_title (more generic)

johncrossman:

Google Drive is strange. A folder can have multiple items with the same name. Furthermore, an item can belong to multiple folders (i.e., parents).  The new methods in this PR try to account for such rules.


paulkerschen:
Up is down, black is white...

PR #4086, opened 2015-08-17
CLC-5620 Create and populate Oec::CourseCode table and model

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5651

After some dithering, I decided to use the opaque but machine-friendly five-letter L4 code to identify departments in the database. A `Berkeley::Departments` translation module follows the existing `Berkeley::Colleges` in using a simple in-memory map to deliver human-readable official department names without more mucking around in the DB.

`Berkeley::Colleges` gets a bit of cleanup on the side.


app/models/berkeley/colleges.rb
```diff
@@ -2,38 +2,31 @@ module Berkeley
...
+        'ENGR' =>       'College of Engineering',
+        'ENV DSGN' =>   'College of Environmental Design',
+        'GRAD DIV' =>   'College of Letters & Science',
+        'JOURN' =>      'Graduate School of Journalism',
+        'L & S' =>      'College of Letters & Science',
+        'LAW' =>        'School of Law',
+        'NAT RES' =>    'College of Natural Resources',
+        'OPTMETRY' =>   'School of Optometry',
+        'PUB HLTH' =>   'School of Public Health',
+        'PUB POL' =>    'Richard & Rhonda Goldman School of Public Policy',
```
johncrossman: I believe it is 'Richard & Rhoda Goldman School of Public Policy'. Lord knows, we want neither Rhoda nor Rhonda contacting us.

paulkerschen: Thank you, @johncrossman, amended. I assume they regularly scour Github for their own names.

db/schema.rb
```diff
@@ -153,6 +153,18 @@
...
+    t.string   "dept_name",      null: false
+    t.string   "catalog_id"
+    t.string   "dept_code",      null: false
+    t.boolean  "include_in_oec", null: false
+    t.datetime "created_at"
+    t.datetime "updated_at"
+  end
+
+  add_index "oec_course_codes", ["dept_code"], name: "index_oec_course_codes_on_dept_code", using: :btree
+  add_index "oec_course_codes", ["dept_name", "catalog_id"], name: "index_oec_course_codes_on_dept_name_and_catalog_id", unique: true, using: :btree
```
johncrossman: catalog_id is nullable and yet part of 'unique: true' index. Might that cause a problem?

spec/models/oec/course_code_spec.rb
```diff
@@ -0,0 +1,44 @@
+describe Oec::CourseCode do
+
```
johncrossman: Perhaps a test where dept_name contains a single quote?

db/schema.rb
paulkerschen: @johncrossman Excellent catch. Indeed, Postgres doesn't enforce unique constraints on null columns, and the intended constraint requires two partial indexes: 

http://stackoverflow.com/questions/8289100/create-unique-constraint-with-null-columns/8289253#8289253

I'll revise.

johncrossman: After a closer look I think the answer is "no problem".

spec/models/oec/course_code_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    Oec::CourseCode.create(dept_name: 'BIOLOGY', catalog_id: '1AL', dept_code: 'IMMCB', include_in_oec: true)
+    Oec::CourseCode.create(dept_name: 'BIOLOGY', catalog_id: '1B', dept_code: 'IBIBI', include_in_oec: true)
+    Oec::CourseCode.create(dept_name: 'BIOLOGY', catalog_id: '1BL', dept_code: 'IBIBI', include_in_oec: true)
+    Oec::CourseCode.create(dept_name: 'CATALAN', catalog_id: '1', dept_code: 'LPSPP', include_in_oec: true)
+    Oec::CourseCode.create(dept_name: 'CHEM', catalog_id: '3A', dept_code: 'CCHEM', include_in_oec: true)
+    Oec::CourseCode.create(dept_name: 'CHEM', catalog_id: '3B', dept_code: 'CCHEM', include_in_oec: true)
+    Oec::CourseCode.create(dept_name: 'INTEGBI', catalog_id: '99B', dept_code: 'IBIBI', include_in_oec: true)
+    Oec::CourseCode.create(dept_name: 'MCELLBI', catalog_id: '101', dept_code: 'IMMCB', include_in_oec: true)
+    Oec::CourseCode.create(dept_name: 'PORTUG', catalog_id: '10', dept_code: 'LPSPP', include_in_oec: true)
+    Oec::CourseCode.create(dept_name: 'SPANISH', catalog_id: '200', dept_code: 'LPSPP', include_in_oec: false)
```
johncrossman: Why include_in_oec: false for SPANISH?

db/migrate/20150813225141_create_oec_course_codes.rb
```diff
@@ -0,0 +1,221 @@
...
+    add_index :oec_course_codes, :dept_code
+
+    reversible do |dir|
+      dir.up do
+        [
+          'A,RESEC', nil, 'MBARC', true,
+          'AEROSPC', nil, 'QLROT', false,
+          'AFRICAM', nil, 'SAAMS', false,
+          'AFRKANS', nil, 'HZGER', false,
+          'AGR CHM', nil, 'MEPMB', false,
```
johncrossman: Should AGR CHM be 'true' since it's under MEPMB? It is a sibling of PLANTBI.

johncrossman:
@paulkerschen  I'm done reviewing. Looks really good. These dept mappings make me say...
 
971002bb-acac-4c7f-bc80-61eb11954ec8

paulkerschen:
@johncrossman Thanks for the considered review. I've pushed an amendment with changes:
- Partial indexes caused unnecessary problems switching between dev/prod Postgres and test sqlite, so wildcard `catalog_id` is now indicated by a blank string rather than `NULL`.
- `AGR CHEM` comes on board with its siblings.
- The `Oec::CourseCode` spec is fleshed out a bit more, including a department name with a real character, a hypothetical single course excluded from Spanish rather than the entire department, and examples illustrating the uniqueness constraints.

johncrossman:
Thank you @paulkerschen. Elegant work on an inelegant set of requirements.

PR #4091, opened 2015-08-18
CLC-5635, 'scope' value used by google_auth_controller is now set in yml 

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5635

In google_auth_controller, scope was set as an array. Now we're passing in a set of space-separated values. The underlying OAuth2::Client code is flexible and can handle both:

    case new_scope
    when Array
      ...
    when String
      @scope = new_scope.split(' ')


paulkerschen:
Muy conveniente.

PR #4092, opened 2015-08-18
CLC-5620 Really honestly fix Bamboo failures

paulkerschen:
Please do not merge until Bamboo checks out: https://bamboo.ets.berkeley.edu/bamboo/browse/MYB-CPT139-14

### DIFF
--- a/db/developer-seed-data.sql
+++ b/db/developer-seed-data.sql
@@ -383,6 +383,13 @@ ALTER TABLE ONLY links ALTER COLUMN id SET DEFAULT nextval('links_id_seq'::regcl
 -- Name: id; Type: DEFAULT; Schema: public; Owner: -
 --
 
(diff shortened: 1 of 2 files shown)


johncrossman:
Bamboo is no happy. It seems that course_code_spec sets up test data which duplicates seed-data.

paulkerschen:
@johncrossman Indeed. I've pushed a workaround and Bamboo will try here: https://bamboo.ets.berkeley.edu/bamboo/browse/MYB-CPT139-15

But we certainly have too much duplication between seed, migration and test data, and at some point it would be good to work out a better pattern for consistent database population across environments.

johncrossman:
I agree @paulkerschen With less test data mayhem, I wouldn't wince at schema changes. A good topic for sprint retrospective.

paulkerschen:
@johncrossman Looks like that last amendment finally unclogged the pipes.

johncrossman:
All clear on the Bamboo front.

PR #4094, opened 2015-08-18
Merge master into sis-master

christianv:


ctweney:
Ah, too long since I did one of these. Thanks!

PR #4096, opened 2015-08-18
SISRP-5405 Establish person EDO endpoint with fake data behind it

ctweney:
https://jira.berkeley.edu/browse/SISRP-5405

Using the most recently available example data from bmeta.berkeley.edu to set up a fake person EDO, so that @christianv can do front-end work against it. The real API is not yet working.


config/settings.yml
```diff
@@ -408,3 +408,8 @@ slate_checklist_proxy:
   username: secret
   password: secret
 
+hub_edos_proxy:
+  fake: true
+  base_url: 'https://apis.berkeley.edu/something'
```
christianv: @ctweney http://www.urbandictionary.com/define.php?term=something+something

PR #4097, opened 2015-08-18
SISRP-6502 - COA - Show budget item totals

christianv:
https://jira.berkeley.edu/browse/SISRP-6502


ctweney:
Looks like, totally good to me.

PR #4099, opened 2015-08-18
CLC-5654 Add Google Drive methods to move files between folders

paulkerschen:
In setting up Drive tasks for https://jira.ets.berkeley.edu/jira/browse/CLC-5654, I realized that one requirement I forgot to specify was copying a file to a new folder without having to download and re-upload.

The Drive API takes three steps to implement this action (duplicate file, add new parent, remove old parent). Fortunately, the new DriveManager class is nicely built and gives clear patterns to follow.

`find_folders` convenience method also thrown in.


johncrossman:
Es bueno.

PR #4100, opened 2015-08-19
CLC-5582 - Use theming option in pikaday instead of overwriting the CSS ourselves

winstonkim:
https://jira.ets.berkeley.edu/jira/browse/CLC-5582

### DIFF
--- a/src/assets/stylesheets/_datepicker.scss
+++ b/src/assets/stylesheets/_datepicker.scss
@@ -1,8 +1,8 @@
 // CalCentral overrides
-.pika-single {
+.pika-single.cc-datepicker-theme {
   position: absolute;
   width: 260px;
-  .pika-next, .pika-prev {
+  .cc-datepicker-theme .pika-next, .cc-datepicker-theme .pika-prev {
     padding: 15px;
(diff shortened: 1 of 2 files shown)


src/assets/stylesheets/_datepicker.scss
```diff
@@ -1,8 +1,8 @@
 // CalCentral overrides
-.pika-single {
+.pika-single.cc-datepicker-theme {
```
christianv: @winstonkim does it also work if we just use `.cc-datepicker-theme` instead of `.pika-single.cc-datepicker-theme`?

winstonkim: Yeah, it works. Let me make the changes.

johncrossman:
Before we merge this PR I'd like @raydavis and @christianv to decide on fixVersion for https://jira.ets.berkeley.edu/jira/browse/CLC-5582  Nonetheless, thanks for doing the work @winstonkim

christianv:
@johncrossman discussed with @raydavis and this should be fixVersion 64

PR #4102, opened 2015-08-19
CLC-5654 Implement term setup task

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5654

- Oec::TermSetupTask inherits from Oec::Task, which holds a lot of lower-level logic expected to be shared between tasks.
- Different CSV formats are defined as skeletal subclasses of Oec::CsvExport.
- DriveManager gets some logging and exception handling. Since it's quite a different animal from our usual proxies, there was unfortunately no good way to test the term setup task other than a lot of fiddly expectation-setting on a mock object.


johncrossman:
This looks good to me, @paulkerschen   I like logging improvement and 404 consideration in DriveManager. I'm ready to merge. And Task code is nice 'n readable.

PR #4104, opened 2015-08-20
SISRP-6542 Set statusCode to 400 and errored to true on user errors

ctweney:
https://jira.berkeley.edu/browse/SISRP-6542


christianv:
@ctweney thanks!

PR #4105, opened 2015-08-20
CLC-5604, new proxy_spec and better fake data handling in CredentialStore

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5604
Bamboo: https://bamboo.ets.berkeley.edu/bamboo/browse/MYB-CJT60-2

Includes:
* More test coverage
* fake data: *refresh_token* override to go with existing *access_token* override


paulkerschen:
More tests is good tests.

PR #4106, opened 2015-08-20
CLC-5103 Adieux, bSpace

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-5103


paulkerschen:
The incoming class will never know what they missed.

PR #4112, opened 2015-08-21
CLC-5674, spreadsheet api with gem update

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5674
Bamboo: https://bamboo.ets.berkeley.edu/bamboo/browse/MYB-CJT61-1

Includes:
* gem google_drive: https://github.com/gimite/google-drive-ruby 
* wrapper class SheetsManager provides Sheets API integration (phase one) and specs
* required tweak to 'scope'


app/models/google_apps/sheets_manager.rb
```diff
@@ -0,0 +1,36 @@
...
+      drive_api = client.discovered_api('drive', 'v2')
+      media = Google::APIClient::UploadIO.new(file_path, 'text/csv')
+      metadata = { :title => title, :description => description }
+      file = drive_api.files.insert.request_schema.new metadata
+      file.parents = [{ :id => parent_id }]
+      api_result = @session.execute!(
+        :api_method => drive_api.files.insert,
+        :body_object => file,
+        :media => media,
+        :parameters => { :uploadType => 'multipart', :convert => true })
```
paulkerschen: @johncrossman Would it be possible to have this API log its result and raise `ProxyError` on failures like the other DriveManager methods; or does the involvement of `GoogleDrive::Session` change things?

paulkerschen:
@johncrossman Looks really good; I just had the one question about logging and exceptions.

app/models/google_apps/sheets_manager.rb
johncrossman: Very good question, @paulkerschen.  I'll add that to upload_csv_to_spreadsheet(). But not so easy on the getter methods because, in those cases, I rely entirely on the wrapped logic. I had to code the upload() method with more direct access to underlying stuff.  Stay tuned for one more commit.

johncrossman:
Couple improvements based on comments so far:
* we now have explicit access to 'q' (query) string used by actual Google code
* spreadsheet_by_id() and upload() now have a handle on api_response and log accordingly
* error messages have more info

paulkerschen:
Thanks for the extra work, @johncrossman - I bet at some point we'll be glad to have those Google methods unpacked a bit.

PR #4115, opened 2015-08-21
CLC-5645 OEC scripts: Courses import task

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5603
Bamboo https://bamboo.ets.berkeley.edu/bamboo/browse/MYB-CPT140-7

A lot of this code will look eerily familiar to those versed in the existing OEC scripts.

Aside from the new Drive-upload functionality, the biggest changes to the logic are:
- Oec::CsvExport now wraps a Hash of rows; having lookup keys makes it easy to avoid duplicate insertions.
- Fixture data for tests has been broken out of CSV into JSON files matching individual queries, which should make it easier to write follow-up specs more fully describing the expected output structure.
- Queries and generated CSVs are no longer grouped by the "dept_name" portion of the course code, but by actual campus departments. The Oec::Queries class now takes as input the course-code-to-real-department mappings defined in Oec::CourseCode, and transforms them to SQL in its `depts_clause` method.
- Term and department information no longer comes from YAML settings. Terms are specified in arguments and the departments list comes out of Postgres (with department-filtering arguments to be enabled in a follow-up).


app/models/oec/courses.rb
```diff
@@ -1,6 +1,15 @@
 module Oec
   class Courses < Oec::CsvExport
 
+    attr_reader :dept_code
+
+    def initialize(export_dir, opts={})
+      if (@dept_code = opts.delete :dept_code)
+        opts[:filename] = "#{@dept_code}.csv"
```
johncrossman: What is the use case of no :dept_code arg in opts?

paulkerschen: @johncrossman There are cases where this same file structure is used for a non-specific `courses.csv` file; for example, in the `supplemental_sources` folder where the service lead might keep hand-edited course information to be merged in with SIS data.

app/models/oec/courses_import_task.rb
```diff
@@ -0,0 +1,73 @@
...
+        export_to_folder(courses, imports_today)
+      end
+    end
+
+    def import_courses(courses, course_codes)
+      Oec::Queries.courses_for_codes(@term_code, course_codes).each do |course|
+        # No practical way to combine these fields in SQL, so we'll do it here in Ruby.
+        if course['cross_listed_name'].present?
+          # get all the cross listings of this course, even if they're in departments not part of our filter.
+          Oec::Queries.courses_for_cntl_nums(@term_code, course['cross_listed_name']).each do |cross_listing|
```
johncrossman: This looks good. Couple notes:
* https://jira.ets.berkeley.edu/jira/browse/OEC-664
* room shares (e.g., https://jira.ets.berkeley.edu/jira/browse/OEC-568)
Does OEC intend to solve such cases on this go round?

paulkerschen: @johncrossman I think https://jira.ets.berkeley.edu/jira/browse/CLC-5628 may end up being the ticket for as much crosslisting/room-share improvement as we can fit into this go-round. Thanks for those two OEC links; I'll link them to the CLC issue.

app/models/oec/csv_export.rb
```diff
@@ -5,35 +5,48 @@ def self.base_filename
...
+      @filename || self.class.base_filename
     end
 
-    def record_to_csv_row(record)
-      row = {}
-      record.keys.each do |key|
-        row[key.upcase] = record[key]
+    def export
+      if @rows.any?
+        output = CSV.open(output_filename, 'wb', headers: headers, write_headers: true)
```
johncrossman: Minor question: I didn't write this CSV.open statement but I've wondered why we pass 'wb'. I understand that to mean 'Write-only, binary mode'. Docs say, "[...sets external encoding to ASCII-8BIT unless explicitly specified](http://ruby-doc.org/core-2.2.3/IO.html#method-c-new)" but I've seen no problems in past.

johncrossman:
This looks nice 'n clean to me. I posted a couple minor questions.

app/models/oec/csv_export.rb
paulkerschen: That's a good question. I wonder if people get into the habit for fear of cross-platform linebreak problems. It's not _supposed_ to do anything on a Mac...

johncrossman: I'll let you decide if that needs to change in future, @paulkerschen    I think I'm ready to merge this mofo.

johncrossman:
image

PR #4121, opened 2015-08-25
CLC-5625 Oec: skip never-evaluated instruction formats

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5625


paulkerschen:
Closing this for the moment in case more not-to-evaluate instruction formats appear.

PR #4122, opened 2015-08-25
CLC-5681 - Hovering a cc-button-link element with an icon doesn't underline the icon

christianv:
https://jira.ets.berkeley.edu/jira/browse/CLC-5681


raydavis:
Makes sense to me.

PR #4133, opened 2015-08-26
SISRP-6732 ignore another failing CS testext

ctweney:
https://jira.berkeley.edu/browse/SISRP-6732

Should fix the Bamboo thrashing.


christianv:
@ctweney lgtm - feel free to self merge when bamboo is green.

PR #4137, opened 2015-08-27
SISRP-6809 Always send required fields to CS

ctweney:
https://jira.berkeley.edu/browse/SISRP-6809?filter=-1


christianv:
@ctweney awesome & so neat!

PR #4139, opened 2015-08-28
CLC-5702 CLC-5705 Add Lead TA and Reader roles to Canvas course policy

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-5702
https://jira.ets.berkeley.edu/jira/browse/CLC-5705

### DIFF
--- a/app/models/canvas/course_user.rb
+++ b/app/models/canvas/course_user.rb
@@ -1,7 +1,8 @@
 module Canvas
   class CourseUser < Proxy
 
-    ADMIN_ROLES = %w(TeacherEnrollment TaEnrollment DesignerEnrollment Owner Maintainer)
+    ADMIN_ROLES = ['TeacherEnrollment', 'TaEnrollment', 'DesignerEnrollment',
+      'Owner', 'Maintainer', 'Lead TA']
 
     def initialize(options = {})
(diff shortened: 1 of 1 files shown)


app/models/canvas/course_user.rb
```diff
@@ -1,7 +1,8 @@
 module Canvas
   class CourseUser < Proxy
 
-    ADMIN_ROLES = %w(TeacherEnrollment TaEnrollment DesignerEnrollment Owner Maintainer)
+    ADMIN_ROLES = ['TeacherEnrollment', 'TaEnrollment', 'DesignerEnrollment',
+      'Owner', 'Maintainer', 'Lead TA']
```
paulkerschen: @raydavis Not at all a merge blocker, but I'm curious what led to this inconsistent scheme where only half the roles include `Enrollment`. Canvas-defined versus custom?

raydavis: Right, @paulkerschen -- Canvas's API has a bug/quirk which causes the names of Canvas's built-in roles to get replaced by their underlying enrollment type. See the "Enrollment object" description at: https://canvas.instructure.com/doc/api/enrollments.html

PR #4142, opened 2015-08-31
CLC-5628 Oec: include crosslistings for secondary sections

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5628

The CROSS_LISTED_NAME column is now populated for both primary and secondary sections, and is no longer a list of CSVs but a somewhat  human-friendlier guaranteed unique name. (The requirements for this are wonky, and are most clearly laid out in the 'cross-listed name generation' section of the CoursesImportTask spec.)

The CROSS_LISTED_FLAG column is now a little more informative, reporting 'Y' for official crosslistings and 'RM SHARE' for room shares. As before, the `import_courses` task pulls official crosslistings from any department, but room shares only from OEC-participating departments.

The Oracle queries are rewritten to hit the database only twice per department. On my local machine this cuts the time for a complete `import_courses` task from about 5:30 to about 1:45, most of that being spent uploading CSVs. 

Under the hood, Oec::CsvExport gets an `each` method and `Enumerable` mixin for easier data juggling. This class is slowly changing from a CSV exporter to an all-purpose data structure wrapper, and may want a rename at some point.


johncrossman:
This looks really good, @paulkerschen  Nice work. The performance improvement is Wowee Zowee. Notice my strategic capitalization.

PR #4144, opened 2015-08-31
CLC-5628 Compensate for disparate test data on Bamboo

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5628

At present Travis and Bamboo don't have the same test data for course code mappings, and I was not vigilant enough on this point.

One catch was a real catch: with the new `RM SHARE` value for the crosslisting flag, `row['CROSS_LISTED_FLAG'] == 'Y'` is no longer equivalent to `row['CROSS_LISTED_FLAG'].present?`

Otherwise, course-code mappings are just patched in with stubs to clean up the disparity.

### DIFF
--- a/spec/models/oec/courses_import_task_spec.rb
+++ b/spec/models/oec/courses_import_task_spec.rb
@@ -52,7 +52,7 @@ describe Oec::CoursesImportTask do
       it 'should include dept_form only for non-crosslisted courses' do
         subject.each do |row|
           courses.headers.each { |header| expect(row).to have_key header }
-          if row['CROSS_LISTED_FLAG'].present?
+          if row['CROSS_LISTED_FLAG'] == 'Y'
             expect(row['DEPT_FORM']).to be_nil
           else
             expect(row['DEPT_FORM']).to be_present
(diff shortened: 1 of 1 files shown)


paulkerschen:
Happier Bamboo at https://bamboo.ets.berkeley.edu/bamboo/browse/MYB-CPT146-6.

spec/models/oec/courses_import_task_spec.rb
```diff
@@ -52,7 +52,7 @@ def load_fixture_courses
       it 'should include dept_form only for non-crosslisted courses' do
         subject.each do |row|
           courses.headers.each { |header| expect(row).to have_key header }
-          if row['CROSS_LISTED_FLAG'].present?
+          if row['CROSS_LISTED_FLAG'] == 'Y'
```
johncrossman: case-insensitivity not needed here?

paulkerschen: @johncrossman That value comes out of Oracle and is always Y or null.

johncrossman:
We must appease the Bamboo gods. Nice work @paulkerschen

PR #4148, opened 2015-09-01
CLC-5724, upload_worksheet() brings abstraction to GoogleApps::SheetsManager

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5724

A worksheet object might wrap a CSV file read from disk or a hash in memory. Our GoogleApps::SheetsManager doesn't care and thus the abstraction.

### DIFF
--- a/app/models/google_apps/sheets_manager.rb
+++ b/app/models/google_apps/sheets_manager.rb
@@ -33,10 +33,19 @@ module GoogleApps
       spreadsheets
     end
 
-    def upload_csv_to_spreadsheet(title, description, file_path, parent_id = 'root')
+    def upload_worksheet(title, description, worksheet, parent_id = 'root')
+      content = CSV.generate do |csv|
+        headers = worksheet.headers
+        csv << headers
+        worksheet.each { |row| csv << row.values_at(*headers) }
+      end
+      upload_to_spreadsheet(title, description, StringIO.new(content), parent_id)
+    end
+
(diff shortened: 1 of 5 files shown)


johncrossman:
For the record, Bamboo passed: https://bamboo.ets.berkeley.edu/bamboo/browse/MYB-CJT65-1

paulkerschen:
@johncrossman, this looks sweet. Two questions:

1) Were you politely postponing switchover of Oec::ImportCoursesTask so as not to collide with my work in there? If so, I tip my hat to you and can easily switch it over as part of that tasks.
2) I see that the header-only uploads are still writing to disk and then uploading files. I found that Ruby's CSV library has annoying quirky behavior with header-only files, and I had to hack in some roundabout logic to make them output correctly. Did you run into similar weirdness?

johncrossman:
Answers to @paulkerschen questions:
1. My work to switchover the core OEC code was in Oec::Task. Which parts of Oec::CoursesImportTask still require refactoring? Regardless, please go for it! :)
2. Yes, I did not remove *worksheet.export* from header-only uploads method. Would you like me to remove that line from upload_worksheet_headers() method?

paulkerschen:
@johncrossman 
1. I was not looking closely enough and assumed that if Oec::TermSetupTask required changes, Oec::CoursesImportTask would as well. You're right, it doesn't!
2. Let's leave _worksheet.export_ in their for now. Later on it would be nice to have a toggle option for local-write-only or Google-upload.

I say good to merge.

PR #4149, opened 2015-09-01
CLC-5731 Allow OEC tasks to take department-filter and local-write arguments

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5731

Oec::CoursesImportTask can now filter by either friendly course-code "department names" (SPANISH, CATALAN) or unfriendly department codes (IBIBI, LPPSP). 

On all tasks, a `local_write` option can be set to any truthy value in order to skip Google Drive upload and instead output local CSVs and logfiles for debugging purposes. Otherwise local output-and-delete is skipped entirely in favor of the Better Way from #4148.

Task logging is standardized a bit.

### DIFF
--- a/app/models/oec/course_code.rb
+++ b/app/models/oec/course_code.rb
@@ -4,8 +4,8 @@ module Oec
     self.table_name = 'oec_course_codes'
     attr_accessible :dept_name, :catalog_id, :dept_code, :include_in_oec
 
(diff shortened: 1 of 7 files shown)


johncrossman:
Very nice. I'll make sure diffTask follows the same rules.

PR #4152, opened 2015-09-02
CLC-5719, pull dept spreadsheet from Drive with naming convention

johncrossman:
Bamboo: https://bamboo.ets.berkeley.edu/bamboo/browse/MYB-CJT66-1

Another step towards ReportDiffTask:
* option to pull concise department title per [requirements](https://confluence.ets.berkeley.edu/confluence/pages/viewpage.action?pageId=46142744)
* reusable find_or_create_today_subfolder() code moves to Oec::Task
* specs

### DIFF
--- a/app/models/berkeley/departments.rb
+++ b/app/models/berkeley/departments.rb
@@ -1,8 +1,13 @@
 module Berkeley
   class Departments
 
-    def self.get(dept_code)
-      department_map[dept_code] || dept_code
+    def self.get(dept_code, concise=false)
+      name = department_map[dept_code]
+      if name
+        concise ? name.sub('Department of ', '') : name
+      else
+        dept_code
+      end
     end
 
     #L4 codes from http://www.bai.berkeley.edu/BFS/BudgetGL/treeReports/UCBDTREE.HTM
--- a/app/models/berkeley/departments.rb
+++ b/app/models/berkeley/departments.rb
@@ -1,13 +1,15 @@
 module Berkeley
   class Departments
 
-    def self.get(dept_code, concise=false)
+    def self.get(dept_code, opts={})
       name = department_map[dept_code]
-      if name
-        concise ? name.sub('Department of ', '') : name
-      else
-        dept_code
-      end
+      name = shortened(name) if name && opts[:concise]
+      name || dept_code
+    end
+
+    def self.shortened(name)
+      name.sub!(/^[OCDSP][a-z]+ (of|for|in) /, '')
+      name.sub(/ (Department|Institute|Academic Program|Programs?)$/, '')
     end
 
     #L4 codes from http://www.bai.berkeley.edu/BFS/BudgetGL/treeReports/UCBDTREE.HTM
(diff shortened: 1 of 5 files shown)


app/models/berkeley/departments.rb
```diff
@@ -1,8 +1,13 @@
...
 
-    def self.get(dept_code)
-      department_map[dept_code] || dept_code
+    def self.get(dept_code, concise=false)
+      name = department_map[dept_code]
+      if name
+        concise ? name.sub('Department of ', '') : name
+      else
+        dept_code
+      end
```
paulkerschen: @johncrossman This is a most excellent option to have. Do you think that using a couple of regular expressions might work to get us a concise name across the board? Glancing through the list of official names, it seems like we might get something short and identifiable with the rules:
- Kill ' of ', ' for ', ' in ' and anything to the left;
- Kill rightmost 'Department', 'Institute', 'Program', 'Programs', 'Academic Program'.
Does that seem workable to you, or too awkward to put in code?

app/models/oec/report_diff_task.rb
```diff
@@ -2,12 +2,18 @@ module Oec
   class ReportDiffTask < Task
 
     def run_internal
-      unless (reports_today = find_or_create_today_reports_folder)
+      unless (reports_today = find_or_create_today_subfolder('reports'))
         raise RuntimeError, 'Failed to retrieve today\'s reports folder from remote drive'
       end
       Oec::CourseCode.by_dept_code(@course_code_filter).each do |dept_code, course_codes|
-        log :info, "Diff #{@term_code} #{dept_code} spreadsheet against latest SIS data. Summaries are in remote drive: #{reports_today}"
+        dept_title = Berkeley::Departments.get(dept_code, true)
```
paulkerschen: @johncrossman I might go with an options hash instead of a positional parameter here; a little more work for the receiving method, but if the `true` has a key, it's easier to see what the method call is doing.

spec/models/berkeley/departments_spec.rb
```diff
@@ -4,8 +4,16 @@
     expect(Berkeley::Departments.get 'LTSLL').to eq 'Department of Slavic Languages and Literatures'
   end
 
+  it 'should return a concise department name' do
+    expect(Berkeley::Departments.get('LTSLL', true)).to eq 'Slavic Languages and Literatures'
+    expect(Berkeley::Departments.get('DACED', true)).to eq 'College of Environmental Design'
+    expect(Berkeley::Departments.get('EAEDU', true)).to eq 'School of Education'
```
paulkerschen: @johncrossman Following on the comment above, I think 'Environmental Design' and 'Education' would be OK concise names for the latter two entities.

paulkerschen:
Looks good; left a couple of thoughts on the concise department naming.

app/models/berkeley/departments.rb
johncrossman: +1

johncrossman:
Thanks for the review, @paulkerschen. The code is more aggressive in making dept names concise. In the cast of 'History of Art Department' we do nothing. Are there similar exceptions? And thanks for the tip on options-hash.

paulkerschen:
@johncrossman Good catches on Haas and History of Art. This looks good for me; our set of department names is small and static enough that we can always tweak later if needed.

PR #4154, opened 2015-09-02
CLC-5627 Oec: Flag course IDs with joint faculty and GSI instructors

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5627

Also adds the duplicate COURSE_ID_2 column to course worksheets, and streamlines behavior in local write mode, and fixes a tricky intermittent test failure in Berkeley::Departments. The full department name in `@department_map` was getting overwritten by the shortened department name, so Travis would pass or fail depending on test order.


johncrossman:
+1

PR #4155, opened 2015-09-02
SISRP-5405 Adding Delete APIs and fixing test failures

ctweney:
https://jira.berkeley.edu/browse/SISRP-5405

Lots of new delete APIs, and some improvements/fixes in failing tests. Bamboo is green!


fixtures/xml/campus_solutions/phone_delete.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0"?>
+<PHONE_DELETE_RESPONSE xmlns="http://bcs.is.berkeley.edu">
+  <STATUS>Successfully deleted the contact details for individual 25738808 with the phone type CELL</STATUS>
```
christianv: @ctweney do we also have fixtures for when the DELETE fails?

christianv:
@ctweney lgtm, feel free to self merge when travis / bamboo are green.

fixtures/xml/campus_solutions/phone_delete.xml
ctweney: No, but it's not necessary; the failure message structure is exactly the same.

christianv: @ctweney ok, do we get an `errored = true` back when it fails?

ctweney: Yes, we should.

PR #4158, opened 2015-09-03
CLC-5720, diff_report class is Oec::Worksheet

johncrossman:

* Diff spreadsheet will be built from DiffReport instances. 
* Introducing Oec::RemoteDrive which will get helper methods out of Oec::Task
* Related Task gets a touch of work, too. Spec coming soon.


app/models/oec/report_diff_task.rb
```diff
@@ -6,14 +6,19 @@ def run_internal
         raise RuntimeError, 'Failed to retrieve today\'s reports folder from remote drive'
       end
       Oec::CourseCode.by_dept_code(@course_code_filter).each do |dept_code, course_codes|
-        dept_title = Berkeley::Departments.get(dept_code, concise: true)
+        spreadsheet = find_dept_courses_spreadsheet dept_code
```
paulkerschen: @johncrossman Has `find_dept_courses_spreadsheet` been defined yet? If not, it might be good to get a definition in the same commit.

johncrossman: Good catch, @paulkerschen. I just noticed the same. Stand by...

PR #4159, opened 2015-09-03
CLC-5735 Keep jshint at 2.8.0 until latedef regression is worked out

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-5735

Obviously want to wait for Travis-CI's verdict before merging!


paulkerschen:
I fear we may be waiting a while for big T...

PR #4163, opened 2015-09-03
CLC-5737 Oec: distinguish normalized and denormalized Courses worksheets

paulkerschen:
The previous `Oec::Courses`, which is a format we only expect to use for SIS imports, becomes `Oec::SisImportSheet`. 

`Oec::CoursesImportTask` becomes `Oec::SisImportTask` to signal the close connection between task and sheet format.

The new vanilla `Oec::Courses` includes no instructor-data columns and doesn't need the `dept_code` attribute.

Other cleanup:
  - Remove Oec::Worksheet's inheritance from CsvExport, since that inheritance was giving us nothing but the `export_directory` instance variable;
  - Rename Oec::Worksheet's `export` method to `write_csv`, since CSV output is no longer the usual means of export.

### DIFF
--- a/app/models/oec/courses.rb
+++ b/app/models/oec/courses.rb
@@ -1,15 +1,6 @@
 module Oec
   class Courses < Worksheet
 
-    attr_reader :dept_code
-
-    def initialize(export_dir, opts={})
-      if (@dept_code = opts.delete :dept_code)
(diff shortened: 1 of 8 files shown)


johncrossman:
LGTM.  Waiting on T

PR #4164, opened 2015-09-03
CLC-5720, diff_report worksheet; remote_drive extends generic Google operations

johncrossman:

* Spreadsheets will be built from DiffReport instances.
* Oec::RemoteDrive wraps SheetsManager so we can move helper methods out of Oec::Task
* Specs


paulkerschen:
Good! Waiting on once-again-helpful T -

PR #4168, opened 2015-09-04
CLC-5719, remote_drive gets find_dept_courses_spreadsheet and spec

johncrossman:

Added spec to verify find_dept_courses_spreadsheet(). For the moment, we verify 'should return nil'.


paulkerschen:
Bueno! In it goes.

PR #4173, opened 2015-09-04
SISRP-4445 Revise ets_blog specs to current style: Fix testext failures

mDibyo:
https://jira.berkeley.edu/browse/SISRP-4445


ctweney:
Thanks for the fix!

PR #4174, opened 2015-09-04
CLC-5726 Move drive methods from Oec::Task to Oec::RemoteDrive

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5726

Testing complex interactions with Google Drive is tricky. One the one hand, using a fake Google Drive involves fiddling with mock objects to the point of neuralgia. On the other, real Google Drive interactions require a ton of network calls, slowing down our testext runs.

I originally thought that I might split the difference by using a real drive for some tasks only, but executing and verifying a single OEC task adds about a minute to each Bamboo run, and that seemed Just Wrong.

What I've done here is to encapsulate a lot of repeated-interaction logic such as conflict checks inside Oec::RemoteDrive. This cuts down the number of total calls from Oec::Task to Oec::RemoteDrive, simplifying task code and making it much easier to write tests against a fake Oec::RemoteDrive object.

OecSpecHelper was not helping, and is gone.


johncrossman:
Let's get this merged before the neuralgia kicks in. Thanks for helping us avoid the pain... of complexity. Nice work, @paulkerschen.

PR #4175, opened 2015-09-08
CLC-5719, remote_drive_spec with #real Google API calls

johncrossman:

Verify remote_drive with real Google API calls. Bamboo: https://bamboo.ets.berkeley.edu/bamboo/browse/MYB-CJT68-1


paulkerschen:
Großartig.

PR #4179, opened 2015-09-08
CLC-5720, empty_trash() added to drive_manager

johncrossman:

The OEC specs will use empty_trash() to leave no trace of test data.


paulkerschen:
It's not just a good idea; it's the law.

PR #4180, opened 2015-09-08
SISRP-7300 and SISRP-7325 - Fixing test failures 

ctweney:
https://jira.berkeley.edu/browse/SISRP-7300
https://jira.berkeley.edu/browse/SISRP-7325

This PR should make Bamboo green again (apart from unrelated Google test failures). 

This removes the last of our thru-the-hub proxies, but I'm not going to drop that code yet, because the Person EDO will make use of it when we have the real thing available.


christianv:
lgtm

PR #4182, opened 2015-09-09
CLC-5718, fix empty_trash; option to permanently delete in trash_item

johncrossman:
* option to permanently delete in trash_item
* fix empty_trash
* simplify error handling

Bamboo is chewing on this change: https://bamboo.ets.berkeley.edu/bamboo/browse/MYB-CJT69-1

### DIFF
--- a/app/models/google_apps/drive_manager.rb
+++ b/app/models/google_apps/drive_manager.rb
@@ -153,10 +153,8 @@ module GoogleApps
         }
       )
       log_response result
-      # This call may return an empty 204 on success. TODO: Can we use result.error? instead?
-      success = result.status <= 204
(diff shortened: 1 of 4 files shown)


app/models/google_apps/drive_manager.rb
```diff
@@ -158,7 +153,7 @@ def remove_parent(id, parent_id)
         }
       )
       log_response result
-      # This call may return an empty 204 on success.
+      # This call may return an empty 204 on success. TODO: Can we use result.error? instead?
```
paulkerschen: @johncrossman Yes, `result.error?` should work here as in the other methods. Feel free to change as part of this PR.

johncrossman: 'Tis updated, @paulkerschen

PR #4185, opened 2015-09-09
CLC-5719, escape Google API query args

johncrossman:
Bamboo: https://bamboo.ets.berkeley.edu/bamboo/browse/MYB-CJT70-1

[Google API docs](https://developers.google.com/drive/web/search-parameters) say, "Surround with single quotes '. Escape single quotes in queries." I know of no other necessary escaping.

I was able to reproduce the problem using SWOME (Gender and Women's Studies) in oec_remote_drive_spec #real test. That spec is now passing.

### DIFF
--- a/app/models/google_apps/drive_manager.rb
+++ b/app/models/google_apps/drive_manager.rb
@@ -32,24 +32,24 @@ module GoogleApps
       items = []
       parent_id = options[:parent_id] || 'root'
       query = "'#{parent_id}' in parents and trashed=false"
-      query.concat " and title='#{options[:title]}'" if options.has_key? :title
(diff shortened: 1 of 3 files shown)


app/models/google_apps/drive_manager.rb
```diff
@@ -197,5 +204,9 @@ def log_response(api_response)
       logger.debug "Google Drive API request #{request_description} returned status #{api_response.status}"
     end
 
+    def escape(arg)
+      arg.gsub('\'', %q(\\\'))
```
paulkerschen: There it is! I can never, ever, ever remember how many total slashes it takes to do that.

paulkerschen:
Thanks @johncrossman.

PR #4187, opened 2015-09-09
CLC-5727 OEC: Merge in supplemental data as part of SisImportTask

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5727

This allows the service lead to set data such as modular dates and department-specific evaluation types for particular courses or course groups, filtering by department, course and section codes.


johncrossman:
Nice. Travis goodness on one out of two. Merging...

PR #4188, opened 2015-09-09
CLC-5728 OEC: Populate dates for non-modular courses

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5728

PR #4189, opened 2015-09-10
CLC-5728 OEC: Allow SisImportTask to handle nil terms

paulkerschen:
Check Bamboo: https://bamboo.ets.berkeley.edu/bamboo/browse/MYB-CPT157-3

Follow up to #4188, which caused side-effect failures in specs running Oec::SisImportTask. Our default fake_now is still set to 2013, and Berkeley::Terms won't return information on terms two years in the future.

If a nil term is encountered, output no date information and move on.


paulkerschen:
Green Bamboo.

PR #4190, opened 2015-09-10
CLC-5743 OEC: Better handling for non-participating departments

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5743

Oec::SisImportTask is getting a lot of options, but I think we want the ability to toggle this one.

When `import_all` is switched off (the default), a query on a non-participating department will return an empty array instead of blowing up.

When `import_all` is switched on, the `include_in_oec` flag in our database is ignored, and a full course listing is returned.


johncrossman:
Thanks much, Sir Paul

PR #4192, opened 2015-09-10
SISRP-7380 Workaround for bogus data in responses: force_encoding to UTF-8

ctweney:
temporary workaround to https://jira.berkeley.edu/browse/SISRP-7380 

I think this will work, but the problem doesn't manifest for me locally or on Bamboo builds. It only appears in the calcentral-sis-dev environment. I think that has to do with Ruby falling back on the OS default charset, but I'm not sure. Anyway, this PR should be relatively harmless to merge and see if it unblocks @christianv for dev.


christianv:
@ctweney thanks!

PR #4193, opened 2015-09-10
CLC-5748 Surface user-friendly department names where possible

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5748

Also tweaks the concise-name regex to catch "Charles and Louise Travers Department of Political Science", and rolls back some of the apostrophe escaping on Google API calls. (It turns out that you do need the escaping when building a query string, but that escaping metadata parameters in the Ruby API will add an extra backslash. Obviously....)

### DIFF
--- a/app/models/berkeley/departments.rb
+++ b/app/models/berkeley/departments.rb
@@ -8,7 +8,7 @@ module Berkeley
     end
 
     def self.shortened(name)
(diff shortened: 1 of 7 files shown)


johncrossman:
Very friendly.

PR #4195, opened 2015-09-11
CLC-5191 Mark bCourses user accounts which no longer match a CalNet UID

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-5191

### DIFF
--- a/app/models/canvas_csv/maintain_users.rb
+++ b/app/models/canvas_csv/maintain_users.rb
@@ -3,12 +3,14 @@ module CanvasCsv
   # Used by CanvasCsv::RefreshAllCampusData to maintain officially enrolled students/faculty
   # See CanvasCsv::AddNewUsers for maintenance of new active CalNet users within Canvas
   class MaintainUsers < Base
-    attr_accessor :sis_user_id_changes
+    include ClassLogger
+    attr_accessor :sis_user_id_changes, :user_email_deletions
 
     # Returns true if user hashes are identical
     def self.provisioned_account_eq_sis_account?(provisioned_account, sis_account)
+      # Canvas interprets an empty 'email' column as 'Do not change.'
       matched = provisioned_account['login_id'] == sis_account['login_id'] &&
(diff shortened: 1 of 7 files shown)


johncrossman:
There is one happy Travis. I'll review.

app/models/canvas_csv/maintain_users.rb
```diff
@@ -92,21 +100,56 @@ def handle_changed_sis_user_ids
...
+        end
+      end
+    end
+
     def categorize_user_account(existing_account, campus_user_rows)
       # Convert from CSV::Row for easier manipulation.
       old_account_data = existing_account.to_hash
       login_id = old_account_data['login_id']
+      if (inactive_account = /^inactive-([0-9]+)$/.match login_id)
+        login_id = inactive_account[1]
```
johncrossman: Is inactive_account[1] correct or do we want [0]?

raydavis: Yeah, I always have to look that up... [0] means "the whole original string"; [1] means "the first matched group".

app/models/canvas_csv/maintain_users.rb
```diff
@@ -60,6 +63,11 @@ def initialize(known_uids, sis_user_import_csv)
     def refresh_existing_user_accounts
       check_all_user_accounts
       handle_changed_sis_user_ids
+      if Settings.canvas_proxy.delete_bad_emails.present?
```
johncrossman: I should know this... If Settings.canvas_proxy.delete_bad_emails is 'false' in yml then Settings.canvas_proxy.delete_bad_emails.present? will equal false?

raydavis: Yep, ".blank?" means false, [], '', {}, or nil, and ".present?" means "not blank?".

johncrossman:
This looks good to me, @raydavis  I'll merge once I get your answer on 'delete_bad_emails' config

PR #4196, opened 2015-09-11
CLC-5546 Lower the log level of "nothing to do" messages

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-5546


johncrossman:
Travis is having a rough morning. Again, we're good on one out of two. Thanks @raydavis

PR #4198, opened 2015-09-11
CLC-5720, round one of ReportDiffTask has diff-marker per dept

johncrossman:
DO NOT MERGE until Bamboo passes https://bamboo.ets.berkeley.edu/bamboo/browse/MYB-CJT72-1 

Before pushing diff spreadsheets to Google Drive I want to verify basic logic of "is there a diff?". Notice that rows are compared when they align according to id generated by extract_id(). This code still looks out for "annotated" course ids; for example, "2015-D-12345_GSI" in which "GSI" is the annotation.

Follow up tasks:
* https://jira.ets.berkeley.edu/jira/browse/CLC-5718
* Push diff reports to Google Drive

### DIFF
--- a/app/models/oec/report_diff_task.rb
+++ b/app/models/oec/report_diff_task.rb
@@ -1,24 +1,77 @@
 module Oec
   class ReportDiffTask < Task
 
+    attr_accessor :dept_with_diff_set
+    attr_accessor :errors_per_dept
+
     def run_internal
-      Oec::CourseCode.by_dept_code(@course_code_filter).each do |dept_code, course_codes|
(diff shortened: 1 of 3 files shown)


paulkerschen:
@johncrossman This makes sense to me. Are tests for the actual error-reporting function part of the follow-up?

Also, how long does it take to run this spec? I see the value of exercising the code against a real Drive connection, but I've been cautious about running a whole SisImportTask against a live connection, out of concern for lengthening our Bamboo runs.

johncrossman:
@paulkerschen I agree. SisImportTask is too expensive. I'll create a new Jira to mock that aspect of the spec.

paulkerschen:
Bamboo is happy, follow-ups to follow, alles gut.

PR #4205, opened 2015-09-15
CLC-5638 CLC-5747 OEC: Format course, instructor and enrollment data for export

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5638
https://jira.ets.berkeley.edu/jira/browse/CLC-5747

This is part one of the job that takes department-confirmed course data, merges it with campus enrollment data, and outputs normalized data and relationship tables for export to the vendor.

Based on the latest discussion with the service lead, we assume that after the close of the department confirmation period, department-confirmed course data will be merged into a single `merged_course_confirmations` sheet in the root of the `departments` folder, where it may undergo further editing by the service lead. This sheet has the same denormalized structure as our initial SIS import sheets, and is handled by the same subclass of Oec::Worksheet.

Merging of supervisor data will come in a follow-up, as will further validation.


johncrossman:
...Merging code that does merging.

PR #4206, opened 2015-09-16
CLC-5756, ReportDiffTask spec covers different scenarios

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5756

Includes two bug fixes:
* For any given row in a spreadsheet, instructor_func should NOT be part of composite key
* Include row in diff report even if it has validation errors


paulkerschen:
@johncrossman `"instructor_func": "5"` = "Not teaching the class, but endowed the chair."

PR #4207, opened 2015-09-16
CLC-5191 Improve logging of SIS User ID changes

raydavis:
Researching https://jira.ets.berkeley.edu/jira/browse/CLC-5192 convinced me that it would be helpful to log the existing SIS User ID of the bCourses account alongside the projected new SIS User ID.


johncrossman:
All good.

PR #4208, opened 2015-09-16
CLC-5779, skip ReportDiff if SisImportTask fails

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5779

If Oec::Task does not return 'true' then abort loop in rake oec:sis_import.


paulkerschen:
Looks good @johncrossman ... in fact I forgot that Oec::Task was already returning a success code.

PR #4211, opened 2015-09-17
NOJIRA, v65 bump plus db/migrate/20150813225141_create_oec_course_codes.rb

johncrossman:
Per summary.


paulkerschen:
Not a lot to argue with here.

PR #4212, opened 2015-09-18
CLC-5640 OEC: validate export data before output

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5640

Validation for exported data, as demoed this morning.

The pretty formatting in Oec::ReportDiffTask is extracted to Oec::Validator for reusability. Error message collections are changed from arrays to hashes with an integer count as value, making it easier to log repeated errors compactly.

A bit of metaprogramming in Oec::Worksheet allows its subclasses to set their own validation rules with ActiveRecord-style macros.

### DIFF
--- a/app/models/oec/course_instructors.rb
+++ b/app/models/oec/course_instructors.rb
@@ -9,5 +9,7 @@ module Oec
       )
     end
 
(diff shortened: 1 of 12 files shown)


johncrossman:
Very nice. A big step towards readability.

PR #4213, opened 2015-09-18
CLC-5782 OEC: Filter get_courses query on section_cancel flag

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5782


johncrossman:
Poof! They're gone. Waiting on T-dog.

PR #4214, opened 2015-09-21
CLC-5700, publish_task invokes sftp to push to Explorance

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5788
https://jira.ets.berkeley.edu/jira/browse/CLC-5787

Notes:
* Publish task pulls exports from remote drive, SFTPs them to Explorance and writes stdout/stderr to log file in remote drive 'reports' dir.
* I'm using *LFTP* (see https://en.wikipedia.org/wiki/Lftp) which gives more options in using *SFTP*. For example, scriptable way of providing password.
* If password is blank in yml the code will assume ssh keys are at work and exclude password from command line script.

### DIFF
--- a/app/models/oec/publish_task.rb
+++ b/app/models/oec/publish_task.rb
@@ -1,0 +1,66 @@
+module Oec
+  class PublishTask < Task
+
+    def run_internal
+      download_dir = download_exports_from_drive
+      # SFTP stdout and stderr will go to a log file.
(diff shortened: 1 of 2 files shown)


paulkerschen:
Looks great, @johncrossman. I didn't know about lftp - Snappy Wrappers for Old Protocols.

PR #4215, opened 2015-09-21
SISRP-7708 Add URL param that will expire CS cache groups

ctweney:
https://jira.berkeley.edu/browse/SISRP-7708

Example Calls:

```
/dashboard?ucUpdateCache=finaid 
/dashboard?ucUpdateCache=profile
```


christianv:
lgtm and will cover the current use cases, in the future we might have to have multiple (e.g. both profile / finaid) but this should suffice for now.

PR #4191, opened 2015-09-10
CLC-5720, fix find_dept_courses_spreadsheet; optional date_time in Task

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5720
Bamboo is churning: https://bamboo.ets.berkeley.edu/bamboo/browse/MYB-CJT71-2

Includes:
* spreadsheets_by_title accepts optional parent_id
* configurable date_time in Task, useful in testing
* more readable (I hope) logging

@paulkerschen  The find_dept_courses_spreadsheet change was reverted

PR #4222, opened 2015-09-21
CLC-5799 OEC: Include timestamp in imports and exports subfolders

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5799

Happily, #4191 already gave Oec::Task a configurable DateTime, which lets us match timestamps between folder names and report/log files.

With all these timestamps now about, spaces and colons are preferred in remote item names where feasible.

### DIFF
--- a/app/models/oec/export_task.rb
+++ b/app/models/oec/export_task.rb
@@ -81,9 +81,9 @@ module Oec
       end
 
       if valid?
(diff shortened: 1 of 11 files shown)


johncrossman:
A definite improvement, @paulkerschen

PR #4224, opened 2015-09-22
CLC-5796 Restore caching of Course Capture feed

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-5796

Also enable logging of the max-exceeded caching error.


paulkerschen:
Here's hoping the evaporated milk fits in the can.

PR #4225, opened 2015-09-22
CLC-5773, script delegates to OEC rake tasks

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5773

This script can run any of the OEC rake tasks, given the proper command line args. See usage info.

### DIFF
--- a/script/run-oec-task.sh
+++ b/script/run-oec-task.sh
@@ -27,9 +27,10 @@ then
   export RAILS_ENV=production
   export LOGGER_STDOUT=only
   export LOGGER_LEVEL=INFO
+  export JRUBY_OPTS="-Xcext.enabled=true -J-client -X-C"
   echo "$(date): Run oec:${REQUESTED_TASK} task on $(hostname -s)" | ${LOGIT}
 
-  echo 'cd deploy'
+  cd deploy
   bundle exec rake oec:${REQUESTED_TASK} "${2}" | ${LOGIT}
 
   echo "$(date): The oec:${REQUESTED_TASK} task is done" | ${LOGIT}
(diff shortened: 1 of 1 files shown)


script/run-oec-task.sh
```diff
@@ -0,0 +1,49 @@
...
+  # Enable rvm and use the correct Ruby version and gem set.
+  [[ -s "${HOME}/.rvm/scripts/rvm" ]] && . "${HOME}/.rvm/scripts/rvm"
+  source .rvmrc
+
+  export RAILS_ENV=production
+  export LOGGER_STDOUT=only
+  export LOGGER_LEVEL=INFO
+  echo "$(date): Run oec:${REQUESTED_TASK} task on $(hostname -s)" | ${LOGIT}
+
+  echo 'cd deploy'
```
paulkerschen: @johncrossman Is this supposed to be inside an `echo`?

paulkerschen:
@johncrossman Do we need any of the JRUBY_OPTS that the other scripts set? I would not say I'm well informed about what all those options do.

script/run-oec-task.sh
johncrossman: Good catch, @paulkerschen  A remnant of my debugging

johncrossman:
@paulkerschen I would expect that env var to be picked up by `source .bash_profile` but I'll throw it in. Seems silly to have that copy 'n pasted across all scripts. :)

paulkerschen:
@johncrossman Good point, it should come in with the profile. I guess we should check in our ample spare time whether all this explicit option-setting is needed.

I like the flexibility of this wrapper. Superstitiously waiting on T though I wouldn't expect T to touch this one.

PR #4226, opened 2015-09-22
CLC-5802 Feature-flag bCourses user inactivation to deal with unreliable campus data

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-5802

Our database continues to miss many active accounts. And even after it's fixed, we know that what Oracle has done before, Oracle will someday do again.


johncrossman:
Syd Barrett in test data?! Shine on you crazy diamond.

PR #4227, opened 2015-09-22
CLC-5790, local_write for publish_task and nil check on get export csv

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5790

Support local_write standard on publish_task and remove that arg from spec.


paulkerschen:
@johncrossman A good fallback to have if the automated push ever runs into problems.

PR #4236, opened 2015-09-23
CLC-5811, some errors are NOT fatal in report_diff_task

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5811

If a row has bad data then log error and move on. If a failure occurs when loading csv then abort!


johncrossman:
We passed one Travis out of two, @paulkerschen

paulkerschen:
Boom, crash, all the way down the stairs. 

This is all good, but note that the Travis failure came from the diff spec expecting a log timestamp one second off from what it got. We should keep an eye on this in case it's an effect of reclocking the logs.

PR #4241, opened 2015-09-25
CLC-5812, oec:report_diff uses latest (SIS) imports folder by default

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5812

Allows for report_diff without generating more sis_imports. We always use most recent SIS data.


paulkerschen:
A nice solution to the date-matching problem.

PR #4242, opened 2015-09-26
CLC-5769 Use template sheets for departmental confirmations

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5769

- Departmental confirmation sheets are now a single Google Sheet with multiple worksheets. Oec::CreateConfirmationSheetsTask will look for a 'TEMPLATE' sheet in the departmental folder and copy over its styling and validation for departmental sheets. If the 'TEMPLATE' sheet is not found, it will create a new Google Sheet without styling or validation.
- Updating cells in place is accomplished with the new SheetsManager#update_worksheet, re-implementing the functionality of GoogleDrive::Spreadsheet#save with acceptable performance.
- Other tasks are changed to expect a multi-worksheet confirmation sheet. Calling the CSV export URL for a multi-worksheet Google Sheet will generate a CSV for the first worksheet only. As it happens, this is what ReportDiffTask requires anyway ("Courses," not "Report Viewers"), so that code had to change very little. MergeConfirmationSheetsTask requires both worksheets and has to be explicitly told to pull them both out.
- Oec::Worksheet#from_csv now discards decorative header rows, such as we get from the fancy-formatted department sheets, until it reaches a row with the header data that it's expecting.
- A misplaced validation is taken out of Oec::CourseConfirmation. (As written, that validation prevents any rows without instructor data from being written to confirmation sheets, even though the whole point of the confirmation sheets is to fill in blank data).
- https://jira.ets.berkeley.edu/jira/browse/CLC-5808 is also addressed.

### DIFF
--- a/app/models/google_apps/sheets_manager.rb
+++ b/app/models/google_apps/sheets_manager.rb
@@ -9,9 +9,12 @@ module GoogleApps
     end
 
     def export_csv(file)
-      unless file.exportLinks && (csv_export_uri = file.exportLinks['text/csv'])
-        raise Errors::ProxyError, "No CSV export path found for file ID: #{file.id}"
-      end
+      csv_export_uri = if file.respond_to? :csv_export_url
+                     file.csv_export_url
+                   elsif file.exportLinks && file.exportLinks['text/csv']
+                     file.exportLinks['text/csv']
+                   end
+      raise Errors::ProxyError, "No CSV export path found for file ID: #{file.id}" unless csv_export_uri
       result = @session.execute!(uri: csv_export_uri)
       log_response result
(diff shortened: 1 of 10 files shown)


app/models/google_apps/sheets_manager.rb
```diff
@@ -48,6 +51,52 @@ def spreadsheets_by_title(title, opts={})
       nil
     end
 
+    # An alternative implementation of GoogleDrive::Spreadsheet#save without the expensive XML parsing.
+    def update_worksheet(worksheet, updates)
```
johncrossman: Nicely done.

johncrossman:
I see no wrongs. A nice improvement.

PR #4243, opened 2015-09-28
CLC-5813, diff now reports on EVALUATION_TYPE, START_DATE, etc.

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5813
Includes:
* more columns checked during diff
* no more check on INSTRUCTOR_FUNC
* more readable block for data validation


paulkerschen:
いいですよ

PR #4244, opened 2015-09-28
CLC-5819 OEC: Include all courses in merged confirmation sheets; do not cancel merge on validation errors

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5819

Checking the EVALUATE column is moved from MergeConfirmationSheetsTask to ExportTask. A new `strict` option, which defaults to true, is turned off for MergeConfirmationSheetsTask only, allowing it to report errors while still creating the merged sheet.


johncrossman:
Très bien.

PR #4245, opened 2015-09-28
CLC-5820 Generate CROSS_LISTED_NAME even when cross-listing is filtered out

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5820


johncrossman:
Thanks @paulkerschen  The change in spec was helpful in code reading.

PR #4246, opened 2015-09-28
CLC-5824, one diff_report to rule them all

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5824

Our course-eval spa account has sample diff, in this new format, in folder: 2014-D/departments


paulkerschen:
I, for one, am for it.

PR #4248, opened 2015-09-28
CLC-5818, remove BLUE_ROLE from all spreadsheets

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5818

Don't be blue. Role with it.


paulkerschen:
Sacrebleu!

PR #4249, opened 2015-09-28
CLC-5815, all columns in diff report; remove INSTRUCTOR_FUNC

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5815

* Removal of INSTRUCTOR_FUNC in worksheet headers, test data, etc. 
* Diff report: Add missing headers and change prefix "DB_" -->  "sis:" for sake of readability  

Get an impression of new column naming in  '2014-D diff report' (spa-course-eval).


spec/models/oec/sis_import_task_spec.rb
```diff
@@ -165,10 +165,6 @@ def load_fixture_courses
 
       include_examples 'expected CSV structure'
 
-      it 'should not include course supervisor assignments' do
-        expect(subject.map{ |row| row['INSTRUCTOR_FUNC'] }).not_to include '3'
-      end
-
```
paulkerschen: @johncrossman This test no longer works as written, but since we're still filtering out INSTRUCTOR_FUNC 3 prior to sheet generation, it would be good to keep testing that behavior. Could the test be changed to expect that the course supervisor in our fixture data ("Super Visor," UID 12345678) doesn't show up in the generated sheet?

https://github.com/ets-berkeley-edu/calcentral/blob/master/fixtures/oec/courses_for_dept_STAT.json#L122

paulkerschen:
Every time you remove a column from your dataset, an angel gets its wings. One comment on a test; looks good to me otherwise.

spec/models/oec/sis_import_task_spec.rb
johncrossman: Yes @paulkerschen, will do.

PR #4251, opened 2015-09-29
CLC-5700, enhance run-oec-task.sh

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5700

Bash script enhancements:
* 'report_diff' is a task option
* more granular timestamp on log file
* override RAILS_ENV if you want
* ${2} is not needed in script. The **oec.rake** code will pick up cmd line args via ENV['foo']


paulkerschen:
Thanks much @johncrossman.

PR #4252, opened 2015-09-30
CLC-5774 Create Junction API endpoints wrapping OEC tasks

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5774

Part one:
- creates an `index` endpoint serving task, term and department information useful to the front end;
- creates a `run` endpoint that accepts a slimmed-down set of options and shunts OEC tasks over to a Torquebox background job;
- authorizes these endpoints to superusers or the new `administrator_uid` in YAML, which will be the same as the Google account ID in production but may not be in dev/QA.

Specs to follow.

### DIFF
--- a/app/controllers/oec_tasks_controller.rb
+++ b/app/controllers/oec_tasks_controller.rb
@@ -1,0 +1,36 @@
+class OecTasksController < ApplicationController
+  include ClassLogger
+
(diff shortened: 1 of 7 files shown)


johncrossman:
image

PR #4253, opened 2015-09-30
CLC-5793, toss out lftp; publish_task is always SSH

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5793

Changes:
* dev and qa environments use SSH keys so the code can stop relying on *lftp*
* we use sftp's batch-mode
* name of ssh_private_key_file must be explicit (see Settings.oec) because other apps might have a need for other keys in ~/.ssh

### DIFF
--- a/app/models/oec/publish_task.rb
+++ b/app/models/oec/publish_task.rb
@@ -2,7 +2,12 @@ module Oec
   class PublishTask < Task
 
     def run_internal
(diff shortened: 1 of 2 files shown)


paulkerschen:
Best regards, lftp; it was a brief and productive acquaintance.

PR #4254, opened 2015-09-30
CLC-5774 OEC: specs for API endpoints and task wrappers

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5774

Specs to match #4252.


johncrossman:
image

PR #4258, opened 2015-09-30
CLC-5830 OEC: ExportTask merges in previous course supervisors if present

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5830


johncrossman:
+1

PR #4262, opened 2015-10-01
CLC-5833, user can validate-only or export-and-publish

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5833

* New ExportAndPubish class so ApiWrapper can do both in one task
* 'validation_without_export' opt for ExportTask
* reordering in oec.rake according to real-world workflow
* PublishTask will not remove tmp_dir when local_write


johncrossman:
The Travis glass is half full.

paulkerschen:
@johncrossman Looks good! Maybe as a follow-up, ApiTaskWrapper could be told to fill in the `:validate_without_export` option when it's told to launch ExportTask alone?

PR #4266, opened 2015-10-01
CLC-5341 Shell script to report LTI app visibility

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-5341
Missed this little file in the previous PR...

### DIFF
--- a/script/report-lti-usage.sh
+++ b/script/report-lti-usage.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+# Script to generate two CSV reports on LTI applications configured during a term in bCourses.
+
+# Make sure the normal shell environment is in place, since it may not be
+# when running as a cron job.
+source "$HOME/.bash_profile"
+
+cd $( dirname "${BASH_SOURCE[0]}" )/..
+
+LOG=`date +"$PWD/log/report_lti_usage_%Y-%m-%d.log"`
+LOGIT="tee -a $LOG"
+
+# Enable rvm and use the correct Ruby version and gem set.
+[[ -s "$HOME/.rvm/scripts/rvm" ]] && . "$HOME/.rvm/scripts/rvm"
+source .rvmrc
+
+export RAILS_ENV=production
+export LOGGER_STDOUT=only
+export LOGGER_LEVEL=INFO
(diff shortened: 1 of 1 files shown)


script/report-lti-usage.sh
```diff
@@ -0,0 +1,28 @@
...
+cd $( dirname "${BASH_SOURCE[0]}" )/..
+
+LOG=`date +"$PWD/log/report_lti_usage_%Y-%m-%d.log"`
+LOGIT="tee -a $LOG"
+
+# Enable rvm and use the correct Ruby version and gem set.
+[[ -s "$HOME/.rvm/scripts/rvm" ]] && . "$HOME/.rvm/scripts/rvm"
+source .rvmrc
+
+export RAILS_ENV=production
```
johncrossman: with the OEC-related shell script I used the following. Should we allow devs to override RAILS_ENV (e.g., 'testext')?
``export RAILS_ENV=${RAILS_ENV:-production}``

script/report-lti-usage.sh
```diff
@@ -0,0 +1,28 @@
...
+export LOGGER_LEVEL=INFO
+export JRUBY_OPTS="-Xcext.enabled=true -J-client -X-C"
+
+echo | $LOGIT
+echo "------------------------------------------" | $LOGIT
+echo "`date`: About to run the LTI usage reporting script..." | $LOGIT
+
+cd deploy
+
+bundle exec rake canvas:report_lti_usage | $LOGIT
```
johncrossman: Another nit-pick (and prolly doesn't matter): IMO, a line break at end-of-file is a good standard.

script/report-lti-usage.sh
raydavis: Hmm, good question. Our existing pattern sets "production" directly in the shell scripts, but it's true that our shared servers don't have a RAILS_ENV by default. How about if we leave this as is and do a follow-up to cover all the scripts?

johncrossman: Fine by me, @raydavis

raydavis: Added CLC-5842 for the RAILS_ENV enabler.

johncrossman:
Looks good, @raydavis

PR #4267, opened 2015-10-01
CLC-5841, if standalone ExportTask then implicitly add :validate_without_export

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5841

Clients of ApiTaskWrapper don't need to know about the :validate_without_export config.


paulkerschen:
Thanks @johncrossman !

PR #4270, opened 2015-10-01
CLC-5832 OEC: Add new validations on export

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5832

Code to pull default term dates from supplemental_sources/courses is extracted into the base Oec::Task class so that validations can check against it. Also constrains values on MODULAR_COURSE and EVALUATION_TYPE.


johncrossman:
+1

PR #4271, opened 2015-10-02
CLC-5831, call default_term_dates only once

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5831

Call it once.


paulkerschen:
I should say!

PR #4272, opened 2015-10-02
CLC-5528 No more special TurnItIn bCourses subaccount

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-5528


johncrossman:
Welcome to the club, TurnItIn

PR #4273, opened 2015-10-02
CLC-5341 Keep "129182,129410" from displaying as "129,182,129,410"

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-5341

Why do you think they call it Excel?


paulkerschen:
Xquisite.

PR #4274, opened 2015-10-02
SISRP-7959 Use the new all-purpose address post endpoint

ctweney:
https://jira.berkeley.edu/browse/SISRP-7959


christianv:
@ctweney thanks!

PR #4276, opened 2015-10-02
CLC-5783 Defang bCourses account inactivation & give it new teeth

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-5783

There are two commits here:

1. Prevent a replay of the earlier inactivation fiasco by double-checking missing accounts against CalNet LDAP, rather than assuming an infallible DB copy.

2. But also take advantage of a new flag in the DB copy that tells us when a CalNet account is no longer active, meaning that this user can no longer log into CalCentral or use bCourses.


paulkerschen:
Makes sense to me. I picture expired person-type 'Z' as someone who's been asleep in the back row so long that their academic involvement just can't continue.

PR #4279, opened 2015-10-02
CLC-5846, if OEC uid=12004714 is obsolete then delete oauth2_data

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5846

* Environments still using uid=12004714: local, testext, dev, qa 
* Prod will use a new uid based on yaml settings (prepared prior to v66 release)


paulkerschen:
Tricky business! Thanks @johncrossman.

PR #4280, opened 2015-10-02
CLC-5839 OEC control panel: hide departments dropdown when not relevant

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5839

For tasks where department selection has no effect:
- Show text "All participating departments" in place of dropdown;
- If the departmentCode option was previously set, strip it on submit.


johncrossman:
This looks good to me. @christianv - do you care to comment?

johncrossman:
A follow up PR can address concerns, if any.

PR #4283, opened 2015-10-02
CLC-5849 OEC: Do not create course_instructors override sheet on term setup

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5849


johncrossman:
We like removals. I'll merge it's done cooking.

PR #4287, opened 2015-10-02
CLC-5831, give overrides/instructors precedent over SIS

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5831

I wrangled a bit with rewriting spec. I generalized and renamed the merge_overrides_data() method.

### DIFF
--- a/app/models/oec/sis_import_task.rb
+++ b/app/models/oec/sis_import_task.rb
@@ -6,6 +6,7 @@ module Oec
       log :info, "Will import SIS data for term #{@term_code}"
       imports_now = find_or_create_now_subfolder('imports')
       Oec::CourseCode.by_dept_code(@course_code_filter).each do |dept_code, course_codes|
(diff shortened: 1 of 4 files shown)


app/models/oec/sis_import_task.rb
```diff
@@ -6,6 +6,7 @@ def run_internal
       log :info, "Will import SIS data for term #{@term_code}"
       imports_now = find_or_create_now_subfolder('imports')
       Oec::CourseCode.by_dept_code(@course_code_filter).each do |dept_code, course_codes|
+        @term_dates ||= default_term_dates
```
paulkerschen: +1 That's how I shoulda solved that problem.

paulkerschen:
Travis is half happy, the generalized method looks like the Right Thing to me. Thanks @johncrossman !

image

PR #4288, opened 2015-10-02
SISRP-8139 Make phone and other deletes work

ctweney:
https://jira.berkeley.edu/browse/SISRP-8139

There are new paths to all delete endpoints -- look in routes.rb for their structure.


christianv:
@ctweney lgtm

PR #4301, opened 2015-10-06
CLC-5864, configurable local_write; replace Rails.root

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5864

I swap in ```Pathname.new``` because it plays well with existing code like ```LOG_DIRECTORY.join ...```

### DIFF
--- a/app/models/oec/publish_task.rb
+++ b/app/models/oec/publish_task.rb
@@ -3,6 +3,8 @@ module Oec
 
     include MergedSheetValidation
 
+    attr_accessor :staging_dir
+
     def run_internal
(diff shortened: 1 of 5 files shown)


johncrossman:
@paulkerschen I'll commit spec which verifies expected behavior. Stand by.

johncrossman:
Good catch, @paulkerschen  With a relative path, I'm not sure the *batch_file.sftp* would have landed in the nebulous knob directory. Nonetheless, this latest commit should leave no doubt.

paulkerschen:
Beware the knebulous knob! Bueno.

PR #4303, opened 2015-10-06
SISRP-8275 Adapt to student API payload changes

ctweney:
https://jira.berkeley.edu/browse/SISRP-8275

Student API payload changed slightly, so I made those adjustments. 

It no longer sends us student ID or LDAP UID, so we now look in the crosswalk for the student ID, and make use of the UID that we already have anyway.


christianv:
@ctweney lgtm

PR #4307, opened 2015-10-07
CLC-5852 OEC: Show task status and in-progress log in control panel

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5852

This approach is based on our progress-bar wiring for Canvas tasks, but requires less ceremony. When an OEC task is started, the front end receives an ID linked to a cache key. It then polls the back end every two seconds for status and logs until the task is complete, displaying the latest to the user as it goes.


paulkerschen:
Half of Travis is happy; the other half is still grumpy about badges...

johncrossman:
Nice work, @paulkerschen  I've scrutinized it with the best of my ability. LGTM

PR #4315, opened 2015-10-07
CLC-5869 OEC: Do not log stacktrace on non-fatal diff errors

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5869


johncrossman:
Assuming Travis does NOT fail, I will NOT NOT merge.

PR #4320, opened 2015-10-08
CLC-5759 CLC-5858 OEC: Allow tasks to define their own follow-ups

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5858

Oec::Task subclasses can define a follow-up task to run on success. This allows consistent behavior between the command-line and browser interfaces.

A task passes its own option set to the follow-up, as well as its own log under a `:previous_task_log` option. This allows the follow-up to seamlessly append its own logs in the control panel UI, but keep logfiles separate for upload to Drive. 

This PR also addresses https://jira.ets.berkeley.edu/jira/browse/CLC-5759 by having SisImportTask call ReportDiffTask only when `local_write` is turned off.

### DIFF
--- a/app/models/oec/sis_import_task.rb
+++ b/app/models/oec/sis_import_task.rb
@@ -1,6 +1,8 @@
 module Oec
   class SisImportTask < Task
 
+    on_success_run Oec::ReportDiffTask, if: proc { !@opts[:local_write] }
+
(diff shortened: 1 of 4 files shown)


raydavis:
A good beach read.

PR #4321, opened 2015-10-08
SISRP-8426 - FA Net Cost Card - Should display message if student not configured for access

christianv:
https://jira.berkeley.edu/browse/SISRP-8426


ctweney:
looks good!

PR #4326, opened 2015-10-08
CLC-5865 OEC: Show list of participating departments in control panel

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5865

The department list from the API gets a `participating` boolean, which is used to display a list of participating departments when the "All participating departments" option is selected.

For tasks that don't accept a department option, the irrelevant "All participating departments" text is removed.

Oec::CourseCode::included_dept_names is a middleman of a method and doesn't need to be in there.


johncrossman:
An ever-enhancing UI.

PR #4331, opened 2015-10-09
CLC-5873 Do not stub Time.now in specs

paulkerschen:
Granted, it's only one spec, but the title can serve as a PSA. This was causing the failures in https://jira.ets.berkeley.edu/jira/browse/CLC-5873 whenever it happened to be spec that got first dibs at Timeshifter.


johncrossman:
+1 Mergable once T is satisfied.

PR #4333, opened 2015-10-11
CLC-5881 Add bCourses user refresh script

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-5881
The fixed known_uids bug would create a second user account for an expired enrollment.
Also carries CLC-5873 fix from https://github.com/ets-berkeley-edu/calcentral/pull/4331.

### DIFF
--- a/app/models/canvas_csv/refresh_all_campus_data.rb
+++ b/app/models/canvas_csv/refresh_all_campus_data.rb
@@ -8,12 +8,15 @@ module CanvasCsv
     def initialize(batch_or_incremental)
       super()
       @users_csv_filename = "#{@export_dir}/canvas-#{DateTime.now.strftime('%F')}-users-#{batch_or_incremental}.csv"
-      @term_to_memberships_csv_filename = {}
-      @batch_mode = (batch_or_incremental == 'batch')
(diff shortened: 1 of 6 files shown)


app/models/canvas_csv/refresh_all_campus_data.rb
```diff
@@ -70,15 +75,17 @@ def import_csv_files
...
+        unless @accounts_only
+          @term_to_memberships_csv_filename.each do |term_id, csv_filename|
+            if csv_filename.present?
+              if @batch_mode
+                import_proxy.import_batch_term_enrollments(term_id, csv_filename)
+              else
+                import_proxy.import_all_term_enrollments(term_id, csv_filename)
+              end
             end
+            logger.warn "Enrollment import for #{term_id} succeeded"
```
johncrossman: This all looks good @raydavis  Minor question on this logging statement: Should we (1) report on mode ("batch" or "all") and (2) only log "succeeded" if csv_filename.present?

raydavis: Good suggestion, thanks!

johncrossman:
Thanks, @raydavis

PR #4334, opened 2015-10-12
CLC-5890 Use Rails timezone across CanvasCsv::TermEnrollments

paulkerschen:
See https://jira.ets.berkeley.edu/jira/browse/CLC-5890 for the motivation, and for the reason that a passing Travis at this hour is particularly important.


johncrossman:
+1

PR #4336, opened 2015-10-12
SISRP-8836 - Only show profile to students & hide the bConnected card for now in profile

christianv:
https://jira.berkeley.edu/browse/SISRP-8836


ctweney:
looks good.

PR #4338, opened 2015-10-12
CLC-5882, google spec testext must not collide with other testing

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5882
Bamboo passed: https://bamboo.ets.berkeley.edu/bamboo/browse/MYB-CJT75-2

Before I retire the 'ets.course.eval.qa' account and point textext specs at 'course-eval' I want to guarantee that test-data (e.g., term_date = 1066-D) does not collide with manual testing of QA dept.

### DIFF
--- a/spec/models/oec/remote_drive_spec.rb
+++ b/spec/models/oec/remote_drive_spec.rb
@@ -30,6 +30,7 @@ describe Oec::RemoteDrive do
       before {
         worksheet = Oec::SisImportSheet.new(dept_code: dept_code)
         course_codes = [Oec::CourseCode.new(dept_name: 'SPANISH', catalog_id: '', dept_code: dept_code, include_in_oec: true)]
+        # Import of '2015-D' is intentional because we want valid data as input.
         Oec::SisImportTask.new(:term_code => '2015-D').import_courses(worksheet, course_codes)
         @folder = subject.create_folder transient_folder_name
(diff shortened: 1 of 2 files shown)


spec/models/oec/remote_drive_spec.rb
```diff
@@ -1,48 +1,46 @@
...
+
       before {
         worksheet = Oec::SisImportSheet.new(dept_code: dept_code)
         course_codes = [Oec::CourseCode.new(dept_name: 'SPANISH', catalog_id: '', dept_code: dept_code, include_in_oec: true)]
-        Oec::SisImportTask.new(:term_code => term_code).import_courses(worksheet, course_codes)
-        @imports = subject.find_nested([term_code, 'imports'], on_failure: :error)
-        tomorrow_folder = subject.find_first_matching_folder(tomorrow, @imports) ||
-          subject.create_folder(tomorrow, @imports.id)
-        @dept_import = subject.check_conflicts_and_upload(worksheet, dept_code, Oec::Worksheet, tomorrow_folder)
+        Oec::SisImportTask.new(:term_code => '2015-D').import_courses(worksheet, course_codes)
```
johncrossman: We intentionally use '2015-D' here because we want valid data as input.  I'll put a comment in the code for future generations.

paulkerschen:
And we never had another collision, except with Harold II.

PR #4341, opened 2015-10-12
Second Part

christianv:
Don't merge


app/models/finaid/merged.rb
```diff
@@ -0,0 +1,36 @@
...
+      feed = {}
+      begin
+        self.proxies.each { |proxy_class|
+          proxy = proxy_class.new @uid
+          proxy.append_feed!(feed)
+        }
+      rescue => e
+        self.class.handle_exception(e, self.class.cache_key(@uid), {
+          id: @uid,
+          user_message_on_exception: "Remote server unreachable",
```
johncrossman: If we're serving user the message, "Remote server unreachable", then I'd suggest something a bit more friendly / Finaid related. User won't know which "Remote server" we're talking about.

app/models/hub_edos/my_person.rb
```diff
@@ -0,0 +1,20 @@
...
+    include Cache::JsonAddedCacher
+    include User::Student
+
+    def get_feed_internal
+      cs_id = lookup_campus_solutions_id
+      if cs_id.present?
+        HubEdos::Student.new({user_id: @uid}).get
+      else
+        HubEdos::Person.new({user_id: @uid}).get
+      end
```
johncrossman: IMO, use of ternary operator - in cases like this - is more readable.

app/models/finaid/merged.rb
ctweney: The question of what text should replace it would be up to the functional owners, who are busy with the SIS replacement of this feature. This old MyFinAid code is slated to be removed in GoLive4. I don't think we should mess with it.

app/models/hub_edos/my_person.rb
ctweney: I'll revise, but the ternary after a ? method seems odd. 

```
cs_id.present? ? HubEdos::Student.new({user_id: @uid}).get : HubEdos::Person.new({user_id: @uid}).get
```

Is that really better?

app/models/hub_edos/student.rb
```diff
@@ -0,0 +1,39 @@
+module HubEdos
+  class Student < Proxy
+
+    def initialize(options = {})
+      super(Settings.hub_edos_proxy, options)
+      initialize_mocks if @fake
+    end
+
+    def url
+      "#{@settings.base_url}/#{@campus_solutions_id}"
```
johncrossman: The *url* method relies on *@campus_solutions_id* and yet that var is set in HubEdos::Proxy#get_internal. Should *@campus_solutions_id = lookup_campus_solutions_id* move to *initialize* to avoid unwanted nil value?

app/models/hub_edos/my_person.rb
johncrossman: I agree, it does cause a double-take. Your call. My thought was:
```
person_klass = cs_id.present? ? HubEdos::Student : HubEdos::Person
persion_klass.new({user_id: @uid}).get
```

app/models/hub_edos/user_attributes.rb
```diff
@@ -0,0 +1,174 @@
...
+    end
+
+    def extract_names(edo, result)
+      # preferred name trumps primary name if present
+      find_name('PRI', edo, result) unless find_name('PRF', edo, result)
+    end
+
+    def find_name(type, edo, result)
+      edo[:names].each do |name|
+        if name[:type][:code] == type
```
johncrossman: Two questions: Is *name[:type]* ever nil?  Do we need case-insensitive compare on type?

app/models/hub_edos/user_attributes.rb
```diff
@@ -0,0 +1,174 @@
...
+      end
+    end
+
+    def extract_emails(edo, result)
+      edo[:emails].each do |email|
+        if email[:primary] == true
+          result[:email_address] = email[:emailAddress]
+        end
+        if email[:type].present? && email[:type][:code] == 'CAMP'
+          result[:official_bmail_address] = email[:emailAddress]
```
johncrossman: I'll assume 'official_bmail_address' is correct. Odd to not use '_email_'.

app/models/hub_edos/user_attributes.rb
```diff
@@ -0,0 +1,174 @@
...
+      if edo[:residency].present?
+        # TODO this data only supported in GoLive5
+        if edo[:residency][:official][:code] == 'RES'
+          result[:cal_residency_flag] = 'Y'
+        else
+          result[:cal_residency_flag] = 'N'
+        end
+        if term_transition?
+          result[:california_residency] = nil
+          result[:reg_status][:transitionTerm] = true
```
paulkerschen: The term-transition check in `CampusOracle::UserAttributes` had to do with residency information from Oracle being unavailable during term transitions. I don't know whether any corresponding scenario is likely to come up in the Campus Solutions world.

app/models/hub_edos/student.rb
ctweney: No, lookup_campus_solutions_id is an expensive external call so we don't want to do it from the constructor. Instead, the initialize_mocks call needs to be made again from get_internal. That change is reflected in #4346 .

app/models/hub_edos/my_person.rb
ctweney: Your way is nicer. I'll make it so.

app/models/hub_edos/user_attributes.rb
ctweney: Good point -- I'll add some defensive checks here.

johncrossman:
I'm done reviewing PR "Second Part". Comments / questions posted.

lib/cache/related_cache_key_tracker.rb
```diff
@@ -0,0 +1,39 @@
+module Cache
+  module RelatedCacheKeyTracker
```
raydavis: Nice to finally have a helper for this!

PR #4347, opened 2015-10-13
SISRP-8866 More fields from CS on the task feed

ctweney:
https://jira.berkeley.edu/browse/SISRP-8866


christianv:
@ctweney lgtm, thanks for adding these!

PR #4350, opened 2015-10-13
SISRP-8901 - In Firefox, when returning from Campus Solutions, the Notifications area never completes a refresh and does not display.

christianv:
https://jira.berkeley.edu/browse/SISRP-8901


ctweney:
This one's weird. Is it really Firefox-specific?

christianv:
@ctweney not at all FF specific, just depends on when things are loading.
So it's an async thing that happens and is probably a rare bug that already happens in the current CalCentral master.

PR #4356, opened 2015-10-13
SISRP-8911 Fall back on Oracle field-by-field if they come back null …

ctweney:
…from the Hub EDO

https://jira.berkeley.edu/browse/SISRP-8911


paulkerschen:
This looks good. As a follow-up, @raydavis's JIRA comments point out that we may need to get even more granular in the case of roles and affiliations, since the new population of grad students is likely to include some non-student roles only accessible from Oracle.

PR #4363, opened 2015-10-14
QA: fix bad fix

pfarestveit:
https://jira.ets.berkeley.edu/jira/browse/CLC-5874


johncrossman:
More true, less truthiness.

PR #4369, opened 2015-10-15
SISRP-9328 Put completedDate on tasks feed for completed SIS checklis…

ctweney:
…t items

https://jira.berkeley.edu/browse/SISRP-9328


johncrossman:
Assuming that ```result[:statusDt])``` is never nil... LGTM.

PR #4376, opened 2015-10-16
SISRP-9407 Backend logic for profile-fallback 

ctweney:

New feature flag, cs_profile_visible_for_legacy_users, does what it says. When the flag is true (the default), everyone gets the profile feature. When false, only new students in CS get it. 

@christianv will need to followup with the front-end changes for this: Hide Profile UI when showSisProfileUI is false in the /api/my/status feed.

### DIFF
--- a/app/models/campus_solutions/profile_feature_flagged.rb
+++ b/app/models/campus_solutions/profile_feature_flagged.rb
@@ -5,8 +5,8 @@ module CampusSolutions
     end
     alias_method(:is_cs_profile_feature_enabled, :is_feature_enabled)
 
-    def is_profile_hidden_for_legacy_users
-      Settings.features.cs_profile_hidden_for_legacy_users
+    def is_profile_visible_for_legacy_users
+      Settings.features.cs_profile_visible_for_legacy_users
     end
   end
 end
--- a/app/models/campus_solutions/profile_feature_flagged.rb
+++ b/app/models/campus_solutions/profile_feature_flagged.rb
@@ -4,5 +4,9 @@ module CampusSolutions
       Settings.features.cs_profile
     end
     alias_method(:is_cs_profile_feature_enabled, :is_feature_enabled)
+
+    def is_profile_hidden_for_legacy_users
+      Settings.features.cs_profile_hidden_for_legacy_users
+    end
   end
 end
--- a/app/models/user/api.rb
+++ b/app/models/user/api.rb
@@ -138,7 +138,7 @@ module User
         :uid => @uid,
         :sid => @student_id,
         :isCampusSolutionsStudent => is_campus_solutions_student,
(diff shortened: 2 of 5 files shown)


app/models/user/api.rb
```diff
@@ -137,7 +137,8 @@ def get_feed_internal
         :roles => get_campus_attribute(:roles),
         :uid => @uid,
         :sid => @student_id,
-        :isCampusSolutionsStudent => is_campus_solutions_student
+        :isCampusSolutionsStudent => is_campus_solutions_student,
+        :showSisProfileUI => (is_profile_hidden_for_legacy_users ? (is_campus_solutions_student && is_cs_profile_feature_enabled) : is_cs_profile_feature_enabled)
```
johncrossman: Despite my support of ternary operator, is the following accurate and more readable?
```
:showSisProfileUI => 
   is_cs_profile_feature_enabled && (is_campus_solutions_student || !is_profile_hidden_for_legacy_users)
```

app/models/campus_solutions/profile_feature_flagged.rb
```diff
@@ -4,5 +4,9 @@ def is_feature_enabled
       Settings.features.cs_profile
     end
     alias_method(:is_cs_profile_feature_enabled, :is_feature_enabled)
+
+    def is_profile_hidden_for_legacy_users
+      Settings.features.cs_profile_hidden_for_legacy_users
```
johncrossman: In this case, it'd be nice to have feature-mode rather than feature-flag. For example,
```
Settings.features.cs_profile=[all-users | none | only-new-users]
```
Nonetheless, LGTM

app/models/user/api.rb
ctweney: @johncrossman  Don't merge, I'm working on a different flag semantic that will make it easier to read.

johncrossman: roger, @ctweney

ctweney:
New feature flag is now cs_profile_visible_for_legacy_users.

johncrossman:
LGTM. What you say, @raydavis?

raydavis:
Thanks, @ctweney. This gives the front-end a hook to decide whether to show the new app page to existing users. The other side of our request is to keep the existing user profile data as is through the rest of the codebase. In other words, the fallback isn't just to drop the "My Profile" page, but also to continue to draw user attributes from our legacy DB sources. In the very-near-term the legacy DB sources is what reflects day-to-day reality for existing users, and so I think that's the most stable fallback while we develop and test a more complex merge of old and new.

raydavis:
Adding the new toggle to the split-brain in "get_campus_attribute(field)" might do the trick. What think you?

christianv:
@ctweney can you create a follow up on this?

christianv:
Front-end PR at #4377

PR #4377, opened 2015-10-16
SISRP-9407 - Front-end - Add minimal viable fallback for GL3 profile data in CalCentral

christianv:


raydavis:
Your local ETS nervous-nellie thanks you, @christianv .

PR #4380, opened 2015-10-17
CLC-5893, after OEC publish write files to /exports folder as plain test/csv

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5893
Bamboo: https://bamboo.ets.berkeley.edu/bamboo/browse/MYB-CJT78

The *course_students* sheet has 100-200K rows and Google is unhappy, it exceeds max rows. With this PR, we archive published files to /exports folder as 'text/csv'. It's lightweight and Google is playing nice.

### DIFF
--- a/app/models/google_apps/sheets_manager.rb
+++ b/app/models/google_apps/sheets_manager.rb
@@ -35,8 +35,8 @@ module GoogleApps
           raise Errors::ProxyError, "spreadsheet_by_id failed with id=#{id}. Error: #{result.data['error']}"
       end
       file
-    rescue Google::APIClient::ClientError => e
-      logger.error "spreadsheet_by_id failed with id=#{id}. Exception: #{e}"
+    rescue Google::APIClient::TransmissionError => e
+      log_transmission_error(e, "spreadsheet_by_id failed with id=#{id}")
       nil
     end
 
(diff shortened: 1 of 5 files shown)


app/models/google_apps/sheets_manager.rb
```diff
@@ -35,8 +35,8 @@ def spreadsheet_by_id(id)
           raise Errors::ProxyError, "spreadsheet_by_id failed with id=#{id}. Error: #{result.data['error']}"
       end
       file
-    rescue Google::APIClient::ClientError => e
-      logger.error "spreadsheet_by_id failed with id=#{id}. Exception: #{e}"
+    rescue Google::APIClient::TransmissionError => e
+      log_transmission_error(e, "spreadsheet_by_id failed with id=#{id}")
       nil
```
johncrossman: **Note:** ClientError is subclass of TransmissionError and has access to *Google::APIClient::Result* body.

paulkerschen:
@johncrossman This looks like the right solution to me. My one question is whether anything will need to change in TermSetupTask, now that the `supervisors` and `course_supervisors` in the previous export row are CSVs rather than sheets.

johncrossman:
Thanks @paulkerschen  That work is now in progress: https://jira.ets.berkeley.edu/jira/browse/CLC-5909

johncrossman:
@paulkerschen 'supervisors' is pulled from _previous_term/overrides_; 'course_supervisors' is pulled from _previous_term/exports_ and that is the code I modified.

paulkerschen:
Right you are, @johncrossman; I forgot we were only pulling one file out of `exports`.

johncrossman:
Commit for 'course_supervisors' needs to be 'course_supervisors.csv'. Doh, file extensions!

johncrossman:
If source is 'text/csv' then don't *copy_file*; upload it w/ convert=true to target dir. Cc: @paulkerschen

paulkerschen:
A good workaround for an unpleasant problem. Thanks @johncrossman!

PR #4385, opened 2015-10-19
CLC-5892 OEC: include unmatched and conflicting SIS data in confirmation-sheet merge

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5892

Failure to match confirmation rows with SIS import rows still logs validation errors, but no longer screens them out of the merge. Downstream validation requires problems to be cleared up before export.

Also fixes option handling in `Validator#validate_and_add`. For some reason I thought that Rails considered boolean `false` to be "present," but it is very definitely "blank."


johncrossman:
All rows are welcome to the party.

PR #4386, opened 2015-10-19
SISRP-9429 Add API for work experience

ctweney:
https://jira.berkeley.edu/browse/SISRP-9429

### DIFF
--- a/fixtures/json/work_experience.json
+++ b/fixtures/json/work_experience.json
@@ -1,0 +1,119 @@
+{
+  "studentResponse": {
+    "correlationId": "5713af86-b04e-4e03-bef5-6a6196040e0d",
+    "source": "UCB-SIS-STUDENT",
+    "students": {
+      "students": [
+        {
+          "identifiers": [
+            {
+              "type": "student-id",
+              "id": "CC00000001",
+              "primary": true,
+              "disclose": false
+            }
+          ],
+          "confidential": false,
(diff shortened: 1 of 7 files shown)


fixtures/json/work_experience.json
```diff
@@ -0,0 +1,119 @@
+{
```
christianv: @ctweney just as a small fyi, they are still adding a few extra properties to this list.

christianv:
@ctweney lgtm, but would be good for @raydavis, @paulkerschen or @johncrossman to have a look at it as well.

spec/models/hub_edos/work_experience_spec.rb
```diff
@@ -0,0 +1,30 @@
+require 'spec_helper'
```
paulkerschen: @ctweney No biggie but we're still trying to garden these requires out of the specs.

ctweney: ok @paulkerschen  I'll remove. Don't merge yet, I'll push a followup.

ctweney: Actually, scratch that -- I'll open a separate PR. There are a LOT of 'requires' in my code!

paulkerschen:
Looks good to me too. Thanks @ctweney for the comment pre-empting any questions about whether the API response really is structured like that.

PR #4387, opened 2015-10-19
NOJIRA Remove requires, because require's not required

ctweney:
Requires, begone! CMD-SHIFT-R FTW.


johncrossman:
I'm required to wait for Travis. LGTM

PR #4389, opened 2015-10-20
CLC-5901 Show REMOVED transcript lines in My Academics history

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5901


johncrossman:
Whether memo_or_title, it looks good. Waiting on T.

PR #4390, opened 2015-10-20
SISRP-9472 Base is_campus_solutions_student on the correct field

ctweney:
https://jira.berkeley.edu/browse/SISRP-9472

### DIFF
--- a/app/models/user/api.rb
+++ b/app/models/user/api.rb
@@ -101,7 +101,7 @@ module User
 
     def is_campus_solutions_student?
       # no, really, BCS users are identified by having 10-digit IDs.
-      @student_id.to_s.length >= 10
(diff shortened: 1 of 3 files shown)


app/models/user/api.rb
```diff
@@ -144,6 +144,7 @@ def get_feed_internal
         :roles => get_campus_attribute(:roles),
         :uid => @uid,
         :sid => @student_id,
+        :campusSolutionsID => get_campus_attribute('campus_solutions_id'),
```
johncrossman: @ctweney : should *campus_solutions_id* should be sym?

ctweney: No, because see how it gets used in the get_campus_attribute method.

johncrossman:
Makes sense. Thanks.

PR #4391, opened 2015-10-20
NOJIRA Remove more requires, because "Require's" not required, no matter how quoted!

ctweney:
This should get rid of the last spec_helper reference.


ctweney:
Wait for green T. This passed tests on my localhost, but I'm not sure if Travis sees the .rspec file or not.

johncrossman:
Epic and simple. Glad to see 'em go.

PR #4395, opened 2015-10-20
SISRP-9461 Followup - use show_notifications_archive_link feature fla…

ctweney:
…g for DashboardUrl feed

https://jira.berkeley.edu/browse/SISRP-9461


christianv:
@ctweney lgtm

Also, fyi, in this case we don't really need a front-end feature flag. If you don't provide it in the API, the front-end won't show it.

johncrossman:
LGTM2

PR #4400, opened 2015-10-21
SISRP-9480 - Label for "gender of record" displays as "gender identity", this is not the correct label

christianv:
https://jira.berkeley.edu/browse/SISRP-9480


johncrossman:
"Identity" theft!

PR #4403, opened 2015-10-21
Make CalCentral use the SAML Assertion from CAS to retrieve IDs where it can

ctweney:
https://jira.berkeley.edu/browse/SISRP-9421

We use a fork of omniauth-cas (https://github.com/ets-berkeley-edu/omniauth-cas) that was graciously provided by Steve Hansen to wrangle the SAML assertion into a form we can use. 

If the SAML assertion gave us a student ID and/or CS EMPLID, we store them in cache, which will reduce the load on the crosswalk service. 

Before merging, the team should duly consider the risk involved in maintaining our own fork of omniauth-cas, and weigh it against the benefits.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -26,9 +26,9 @@ gem 'activerecord-jdbch2-adapter', '~> 1.3.16'
 # http://flori.github.com/json/
 gem 'json', '~> 1.8.0'
 
-# CAS Strategy for OmniAuth
-# https://rubygems.org/gems/omniauth-cas
-gem 'omniauth-cas', '~> 1.1.0'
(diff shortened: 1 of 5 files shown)


johncrossman:
Looks fine. I'll merge but I recommend that @paulkerschen and @raydavis have a look.

lexlimo:
I'm glad this is working.  I didn't realize the Ruby client for CAS was so out of date.  But it doesn't look like a terribly active project, so you may want to reach out to Steven and any other UCB developers and discuss the merits of keeping it alive locally anyway.

PR #4405, opened 2015-10-21
SISRP-9453 - Profile - non-required address field appears to require a selection

christianv:
https://jira.berkeley.edu/browse/SISRP-9453


johncrossman:
<good />

PR #4404, opened 2015-10-21
CLC-5881 Check for reactivated user accounts; explain what to do with duplicates

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-5881

Another two issues found via the daily incremental-refresh...

PR #4413, opened 2015-10-22
QA CLC-5881 Check for reactivated user accounts; explain duplicates

raydavis:
Local testing of master went OK, so here's the QA version of https://github.com/ets-berkeley-edu/calcentral/pull/4404


johncrossman:
Looks good, as before.

PR #4415, opened 2015-10-22
CLC-5928 Do not attempt to merge transcript terms with no courses

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5928

What I would give for a static analysis tool that says, "Hey genius, this one might be nil too."


raydavis:
Thank you for the fast diagnosis & fix.

johncrossman:
Static analysis tool? Bah! That's so 2005.

PR #4418, opened 2015-10-22
SISRP-9882 - CC00000006 has a SIR item that's in the checklist but not in the SIR Config

christianv:
https://jira.berkeley.edu/browse/SISRP-9882


johncrossman:
Good. The inline docs help! Thanks @christianv

PR #4419, opened 2015-10-22
CLC-5916 OEC: group cross-listings together; otherwise sort by course code

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5916

This task was bigger than expected, both because Oec::Worksheet has had no real concept of row order to date, and because the requirement for an intuitive sort on course code was tricky to square with the requirement that cross-listings remain grouped together. `Oec::SisImportSheet#sorted_rows` contains the solution I arrived at, I hope comprehensibly expressed.

Note that the `@rows` instance variable in Oec::Worksheet has changed from a vanilla Hash to an ActiveSupport::OrderedHash, guaranteeing that rows will come out in the same order they went in. I don't think this will actually change behavior under Ruby 1.9, but it's a safety harness.


app/models/oec/course_code.rb
```diff
@@ -14,11 +14,19 @@ def self.catalog_id_specific_mapping(dept_name, catalog_id)
...
+      self.where(dept_code: dept_code).pluck(:dept_name)
+    end
+
     def self.included?(dept_name, catalog_id)
       course_code = find_code(dept_name, catalog_id)
       course_code.present? && course_code.include_in_oec
     end
 
+    def self.participating_dept_names
+      @participating_dept_names ||= self.where(include_in_oec: true).pluck(:dept_name).uniq
```
johncrossman: Singleton pattern? Will @participating_dept_names always pick up the latest changes to the *include_in_oec* set?  I.e., recent changes in db.

app/models/oec/sis_import_sheet.rb
```diff
@@ -36,5 +36,42 @@ def headers
...
+                         end
+          [row_priority, row['DEPT_NAME'].to_s, row['CATALOG_ID'].try(:match, /\d+/).to_s.to_i]
+        end
+        # Move one representative from each cross-listed group into the set of non-cross-listed rows for sorting.
+        non_cross_listed_rows << rows.shift
+      end
+
+      non_cross_listed_rows.sort_by! do |row|
+        # Sort by department name first and numeric portion of catalog id second.
+        [row['DEPT_NAME'].to_s, row['CATALOG_ID'].try(:match, /\d+/).to_s.to_i]
```
johncrossman: Would it make sense to sort by section number third?

johncrossman:
As you say, tricky code. A little tricky to read and yet elegantly done. I'll merge once I get your comments on above.

app/models/oec/sis_import_sheet.rb
paulkerschen: @johncrossman I thought about that. Sorting of sections within course codes is already done in the Oracle query, and in my local tests, all the new course-code ordering seems to preserve the section ordering as it came out of Oracle. We could re-express that sorting here for a better guarantee, but the requirements are actually pretty verbose (primary LEC comes before secondary DIS, etc.)

app/models/oec/course_code.rb
paulkerschen: @johncrossman I used the singleton pattern just because there seemed to be so many redundant database hits in my tests. But you're right, it won't pick up recent database changes, and I think the performance impact is much less than I originally thought. I'll amend to take that out.

app/models/oec/sis_import_sheet.rb
raydavis: I second @paulkerschen on this. Sorting correctly outside the Oracle query is really tough, and I'll gladly jump through hoops to let its ordering pass through the layers.

paulkerschen:
Thanks for review @johncrossman; singleton has been de-singled and Travis doesn't seem to mind.

johncrossman:
Travis happy. I happy.

PR #4421, opened 2015-10-22
CLC-5791 CLC-4299 Add Google API scope on demand to existing OAuth2 tokens

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-5791
https://jira.ets.berkeley.edu/jira/browse/CLC-4299

* A new API request parameter grants non-standard API authorizations for devs, admins, & SPA account use. The include_granted_scopes option ensures that previously added scopes are not lost.
* A new API endpoint shows the authorization scope for the currently logged-in user. (This is mostly for QA right now, but I can picture later use in the UX.)

See CLC-5791 for a walkthrough and screenshot.

### DIFF
--- a/app/controllers/google_auth_controller.rb
+++ b/app/controllers/google_auth_controller.rb
@@ -10,7 +10,10 @@ class GoogleAuthController < ApplicationController
   def request_authorization
     expire
     @scope = Settings.google_proxy.scope
-    @scope += " #{params['scope']}" if params['scope'].present?
+    if params['scope'].present?
+      additional_scope = params['scope'].is_a?(Array) ? params['scope'].join(' ') : params['scope']
+      @scope += " #{additional_scope}"
+    end
     final_redirect = params['final_redirect'] || '/'
     if params['force_domain'].present? && params['force_domain'] == 'false'
       client = get_client(final_redirect, force_domain = false)
--- a/app/controllers/google_auth_controller.rb
+++ b/app/controllers/google_auth_controller.rb
@@ -9,6 +9,8 @@ class GoogleAuthController < ApplicationController
 
   def request_authorization
     expire
+    @scope = Settings.google_proxy.scope
+    @scope += " #{params['scope']}" if params['scope'].present?
     final_redirect = params['final_redirect'] || '/'
     if params['force_domain'].present? && params['force_domain'] == 'false'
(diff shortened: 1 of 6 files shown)


app/controllers/google_auth_controller.rb
```diff
@@ -9,6 +9,8 @@ class GoogleAuthController < ApplicationController
 
   def request_authorization
     expire
+    @scope = Settings.google_proxy.scope
+    @scope += " #{params['scope']}" if params['scope'].present?
```
johncrossman: Is validation necessary on *params['scope']* value?   Valid scopes: https://developers.google.com/identity/protocols/googlescopes

johncrossman: Is adding more than one scope a use case?

raydavis: Handling multiple scopes seemed like overkill to me, given how rarely this will get done. Do you think it should be supported? Wouldn't be tough.

raydavis: I looked for an easy way to validate the input against a canonical list of scopes, but so far I've only found web documents which have to be copied/pasted by human beings. It would be nice, though!

johncrossman: As you know, scopes are quite granular. With OEC work we needed a minimum of "/drive", "/drive.file" and "/drive. metadata". I assume the more-than-one use case will come along. Of course, modifying code as needs arise is what we do.

raydavis: Good point, @johncrossman . I'll amend the PR.

BTW, the new testext test managed to survive the latest Bamboo bloodbath...

johncrossman:
Once your amendment is in, I'll merge. I think your instinct is right on validation: not needed until proven otherwise.

raydavis:
Duh, I just remembered my original-and-then-forgotten thought on multiple added scopes: they can be specified exactly the same way they're sent to Google, as a string with blank space between each ID:

http://localhost:3000/api/google/request_authorization?scope=this%20that

"scope[]=this&scope[]=that" is more readable, though, even if it is extra parsing on the server side.

raydavis:
PR amended, @johncrossman

johncrossman:
Looks good, @raydavis  All eyes on Travis...

PR #4422, opened 2015-10-22
CLC-5920 OEC: Validate supervisors sheet before export

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5920

The supervisors sheet was evading validation because it happened already to be in the format we send to Explorance, so was simply getting copied over.


johncrossman:
Green on green.

PR #4425, opened 2015-10-23
CLC-5918 OEC: validate DEPT_FORM against participating department names

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5918


johncrossman:
Mergaroni

PR #4426, opened 2015-10-23
CLC-5916 OEC: Fix Bamboo failures on newly sorted sheets

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5916

Follow-up to #4419 just for Bamboo, chugging at https://bamboo.ets.berkeley.edu/bamboo/browse/MYB-CPT203-1.


paulkerschen:
Bamboo is still sad about Campus Solutions but now happy about OEC.

PR #4416, opened 2015-10-22
SISRP-9872 Don't send nulls to CS APIs

ctweney:
https://jira.berkeley.edu/browse/SISRP-9872

PR #4428, opened 2015-10-23
SISRP-9937 - CalCentral API: Sending over a null to the back-end for address POST converts it into bad XML for CS

christianv:
https://jira.berkeley.edu/browse/SISRP-9937


christianv:
Also reverts #4416

ctweney:
@christianv Actually I think I'd like to keep the change from #4416 -- it's a defense against accidental nulls sent from the front-end -- but also have the front-end do its value-or-blank logic in this particular case.

christianv:
@ctweney K, I'll follow up

christianv:
@ctweney see follow up, just changed the front-end code.

ctweney:
Looks great, thanks for the change!

PR #4429, opened 2015-10-23
CLC-5919 OEC: Improve date validation

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5919

See spec for the complete messages surfaced.


johncrossman:
If year not after year before start of last term.... Looks good.

PR #4430, opened 2015-10-23
SISRP-9932 - Profile Contact Info - Save button should be disabled while address inputs are loading

christianv:
https://jira.berkeley.edu/browse/SISRP-9932


ctweney:
LGTM.

PR #4431, opened 2015-10-23
CLC-5924 OEC: Carry forward course_instructors from previous term if present

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5924


johncrossman:
+1

PR #4433, opened 2015-10-23
SISRP-9893 - Add "This will be editable in April 2016" to profile cards that aren't editable yet.

christianv:
https://jira.berkeley.edu/browse/SISRP-9893


ctweney:
Looks good.

PR #4435, opened 2015-10-23
CLC-5906, dynamically flip key configurations without redeploy/restart

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5906

Note:
* Code does before and after diff on feature flags (litmus test)
* Message to the user guesses at success / failure. Ultimately, user must verify success with his/her own smoke-testing.
* Overhauled the spec

### DIFF
--- a/lib/calcentral_config.rb
+++ b/lib/calcentral_config.rb
@@ -14,23 +14,24 @@ module CalcentralConfig
   end
 
   def reload_settings
-    settings_hash = CalcentralConfig.load_settings
-    old_level = Rails.logger.level
-    new_level = (settings_hash && settings_hash.logger && settings_hash.logger.level)
-    valid = !new_level.nil? &&
-      new_level != old_level &&
-      new_level.is_a?(Integer) &&
-      (0...Log4r::LNAMES.length).include?(new_level)
-    if valid
-      old_level_name = Log4r::LNAMES[old_level]
-      new_level_name = Log4r::LNAMES[new_level]
-      Rails.logger.level = new_level
-      Rails.logger.warn "Rails.logger.level changed (old -> new): #{old_level_name} -> #{new_level_name}"
-    else
-      Rails.logger.warn "Do nothing. Log levels have not changed (old -> new): #{old_level} -> #{new_level}"
+    new_settings = load_settings
+    if Kernel.const_defined? :Settings
+      Rails.logger.warn 'Preparing to reload application settings via Kernel'
+      Kernel.const_set(:Settings, new_settings)
+      new_level = Log4r::LNAMES.index Settings.logger.level
+      old_level = Rails.logger.level
+      if new_level.nil?
+        Rails.logger.warn "The new logger level is invalid: #{Settings.logger.level}"
+      elsif new_level == old_level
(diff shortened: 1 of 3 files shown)


lib/calcentral_config.rb
```diff
@@ -14,24 +14,25 @@ def load_settings
...
+      else
+        Rails.logger.level = new_level
+        Rails.logger.warn "Logger level changed (old -> new): #{Log4r::LNAMES[old_level]} -> #{Log4r::LNAMES[Rails.logger.level]}"
+      end
+      Rails.logger.warn 'YAML settings reloaded. Ask yourself, should I clear the cache?'
     end
-    valid
   end
 
+
```
johncrossman: I'm going to remove this extra line. Also, add another assertion to spec and improve wording on 'warn' msg. Commit 2 coming in moments.

lib/calcentral_config.rb
```diff
@@ -14,22 +14,22 @@ def load_settings
...
+      new_level = Log4r::LNAMES.index Settings.logger.level
+      old_level = Rails.logger.level
+      if new_level.nil?
+        Rails.logger.warn "The new logger level is invalid: #{Settings.logger.level}"
+      elsif new_level == old_level
+        Rails.logger.warn "No change to logger level: #{Log4r::LNAMES[Rails.logger.level]}"
+      else
+        Rails.logger.level = new_level
+        Rails.logger.warn "Logger level changed (old -> new): #{Log4r::LNAMES[old_level]} -> #{Log4r::LNAMES[Rails.logger.level]}"
+      end
```
paulkerschen: @johncrossman It seems a little odd to have the bulk of this method contain code specific to log levels, now that it does so much more. Do you think there would be value in extracting lines 21-30 to a new `CalcentralLogging.reload` method, or is it not worth the back-and-forth?

johncrossman: @paulkerschen I had same observation. I avoided a dedicated method because log-level should only change via YAML change – a separate utility method is prone to misuse. Nonetheless, I don't like "odd" code. I'll change it up; a private util method perhaps.

paulkerschen: @johncrossman Makes sense; risk of misuse argues for private method in this module, not over in the logging module.

paulkerschen: (Not that Ruby modules respect privacy, a quirk I always forget...)

johncrossman:
New ```update_rails_logger_level``` method is in. Travis is green. Ready for re-review.

paulkerschen:
Flip it, into shape. Shape it up, get straight.

ctweney:
This looks right to me on a very cursory review, but it only changes settings for the web app subprocess. The message processing service (another part of Torquebox) will keep running with the old settings. That service is where cache warming via HotPlate happens, so it's possible that data derived from HotPlate processing would diverge from data derived from web UI processing. 

We need to figure out a way to restart the messaging subsystem without disrupting the web system.

johncrossman:
Thanks, @ctweney. I'll create a separate Jira. I'd like to avoid the dreaded hard-to-reproduce bug due to out-of-sync settings.

PR #4436, opened 2015-10-23
CLC-5926 OEC: Avoid repetitively hardcoding folder names

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5926

This part-one PR does no renaming of folders but centralizes folder names in a constant Hash. Once this is cleared, part two will switch the names over.

### DIFF
--- a/app/models/oec/remote_drive.rb
+++ b/app/models/oec/remote_drive.rb
@@ -48,11 +48,6 @@ module Oec
       item
     end
 
-    def find_dept_courses_spreadsheet(term_code, dept_code)
-      dept_name = Berkeley::Departments.get(dept_code, concise: true)
-      find_nested [term_code, 'departments', dept_name, 'Courses']
-    end
-
     def find_first_matching_folder(title, parent=nil)
       find_folders_by_title(title, folder_id(parent)).first
     end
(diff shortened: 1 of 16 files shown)


app/models/oec/remote_drive.rb
```diff
@@ -3,6 +3,15 @@ class RemoteDrive < GoogleApps::SheetsManager
 
     HUMAN_URL = 'https://drive.google.com/drive/my-drive'
 
+    FOLDER = {
+      confirmations: 'departments',
+      merged_confirmations: 'departments',
```
paulkerschen: These two will be different folders post-rename.

johncrossman:
Much better, @paulkerschen  My only critique is the verbosity of ```Oec::RemoteDrive::FOLDER[:overrides]``` and such. I have not used proper enum pattern in Ruby. What is feasibility of...
```
class Folder
  include Enum

  Folder.define :confirmations, 'departments',
  ...
end
```
Then we have ```Oec::Folder::confirmations```. *Bottom line:* what you have looks great.

paulkerschen:
@johncrossman - That's a good call. I'll amend for concision.

paulkerschen:
@johncrossman A solution is pushed. Ruby 1.9 has no enums as such, but it's not hard to fake them. In addition to the enum-type methods, it's still possible to access the FOLDER_TITLES Hash for use in loops and such.

johncrossman:
Yum, home-brewed enums. Looks good.

PR #4438, opened 2015-10-23
SISRP-9984 Don't log 404 as an error from work experiences API

ctweney:
https://jira.berkeley.edu/browse/SISRP-9984


ctweney:
Don't merge! I goofed on this one.

PR #4440, opened 2015-10-23
SISRP-9985 - 400 Response on the SIR POST API

christianv:
https://jira.berkeley.edu/browse/SISRP-9985


christianv:
Basically reverts #4416 - we thought that bug originated on by CalCentral sending over `nil=true` but in fact, it was a CS issue. I've also re-tested https://jira.berkeley.edu/browse/SISRP-9937 and everything still works correctly after this change.

PR #4443, opened 2015-10-26
CLC-5938 MyAcademics::CanvasSites: match terms by slug, not index

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5938

The particular cause of this bug was a Canvas group site associated with entries in both the `:semesters` and `:teachingSemesters` arrays of a student's feed. The code identified the relevant term by its index in the `:semesters` array; attempting to access this index in the `:teachingSemesters` array caused a crash because the arrays were of different lengths.

The fix is to track terms not by index but by slug, which stays consistent between arrays.


johncrossman:
Looks good, @paulkerschen  Waiting on Mr. T

PR #4445, opened 2015-10-27
CLC-5936 - QA: move "requires" to helper file

pfarestveit:
https://jira.ets.berkeley.edu/jira/browse/CLC-5936

At the suggestion of devs, the selenium_helper.rb file is in spec/support, and it orders its requires as needed given dependencies within the classes.  I am assuming when reviewing this PR, the reviewer will make sure none of this does any damage to specs outside of the ui_selenium suite.


johncrossman:
That's a lovely bloodbath of code removal. Waiting on T

PR #4448, opened 2015-10-28
SISRP-9984 Don't treat 404 from work experience API as an error

ctweney:
https://jira.berkeley.edu/browse/SISRP-9984

Missing data will appear to the client UI as a 404 but won't get logged as an error in the app log.


johncrossman:
Weirdness of array structure looks right.

PR #4450, opened 2015-10-28
SISRP-10457 Use smart_fetch_from_cache to cache errors in lookup for …

ctweney:
…a short time

https://jira.berkeley.edu/browse/SISRP-10457


paulkerschen:
Sensible!

PR #4451, opened 2015-10-29
SISRP-9988 - Support html in teaser messages in CalCentral?

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-9988


johncrossman:
+1

PR #4452, opened 2015-10-29
SISRP-9884 - Profile Contact Info - phone number max chars not being validated

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-9884


johncrossman:
24 it is. Waiting on T

PR #4455, opened 2015-10-30
CLC-5952: fix FinAid element locator in Selenium tests

pfarestveit:
https://jira.ets.berkeley.edu/jira/browse/CLC-5952


johncrossman:
I say good. What does Travis say?

PR #4458, opened 2015-11-02
SISRP-10567 - JS errors that show up when going to /dashboard without being logged in

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-10567


johncrossman:
+1

PR #4459, opened 2015-11-02
SISRP-10583 Work around hub API's lack of charset header

ctweney:
https://jira.berkeley.edu/browse/SISRP-10583


johncrossman:
Special characters rejoice!

PR #4464, opened 2015-11-03
SISRP-10583 Followup - force encoding also needed in log statement

ctweney:
https://jira.berkeley.edu/browse/SISRP-10583


johncrossman:
Decipherable log statements are good.

PR #4466, opened 2015-11-03
SISRP-10610 Protect against NoMethodError with nil check

ctweney:
https://jira.berkeley.edu/browse/SISRP-10610


johncrossman:
T happy, me happy.

PR #4468, opened 2015-11-04
SISRP-10609 - Provide proxy pass through for finaid funding sources

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-10609


raydavis:
Let it be so provided.

PR #4469, opened 2015-11-04
SISRP-10524 - Profile - required address fields not marked "required"

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-10524


christianvuerings:
Definitely would recommend this one going into GL3. /cc @ctweney

johncrossman:
After 7 hours we have no objection.  Merged...

PR #4471, opened 2015-11-04
SISRP-10589 Set charset in post to CS to preserve Unicode chars

ctweney:
https://jira.berkeley.edu/browse/SISRP-10589


paulkerschen:
хорошо.

PR #4472, opened 2015-11-04
CLC-5904 Do not upcase task subtitles by default

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5904


raydavis:
Thanks, @paulkerschen !

PR #4476, opened 2015-11-05
SISRP-9437 Merge rather than overwriting Oracle-derived roles

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-9437
https://jira.ets.berkeley.edu/jira/browse/CLC-5948
https://jira.ets.berkeley.edu/jira/browse/CLC-5957

Start with Oracle roles if we have them, then merge in any available data from Hub affiliations. So far all we know about is undergraduate and graduate student status.

Inline comments explain some of the fake-proxy fiddling.

### DIFF
--- a/app/models/hub_edos/proxy.rb
+++ b/app/models/hub_edos/proxy.rb
@@ -11,6 +11,14 @@ module HubEdos
     APP_ID = 'integrationhub'
     APP_NAME = 'Integration Hub'
 
+    def initialize(settings, options)
+      super
+      if @fake
+        @campus_solutions_id = lookup_campus_solutions_id
+        initialize_mocks
+      end
+    end
+
     def instance_key
       @uid
     end
@@ -45,7 +53,7 @@ module HubEdos
     end
 
     def get_internal
-      @campus_solutions_id = lookup_campus_solutions_id
+      @campus_solutions_id ||= lookup_campus_solutions_id
       if @campus_solutions_id.nil?
         logger.info "Lookup of campus_solutions_id for uid #{@uid} failed, cannot call Campus Solutions API"
         {
@@ -53,7 +61,6 @@ module HubEdos
         }
       else
         logger.info "Fake = #{@fake}; Making request to #{url} on behalf of user #{@uid}; cache expiration #{self.class.expires_in}"
-        initialize_mocks if @fake
         response = get_response(url, request_options)
         logger.debug "Remote server status #{response.code}, Body = #{response.body.force_encoding('UTF-8')}"
         feed = build_feed response
--- a/app/models/hub_edos/student.rb
+++ b/app/models/hub_edos/student.rb
@@ -7,7 +7,6 @@ module HubEdos
 
     def initialize(options = {})
       super(Settings.hub_edos_proxy, options)
-      initialize_mocks if @fake
     end
 
     def url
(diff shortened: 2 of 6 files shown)


app/models/hub_edos/proxy.rb
```diff
@@ -11,6 +11,14 @@ class Proxy < BaseProxy
...
     APP_NAME = 'Integration Hub'
 
+    def initialize(settings, options)
+      super
+      if @fake
+        @campus_solutions_id = lookup_campus_solutions_id
+        initialize_mocks
+      end
+    end
+
```
paulkerschen: This replaces the `initialize_mocks` call that appeared in `get_internal` right before the request. Initialization was delayed because `@campus_solutions_id` wasn't immediately available for URL matching; the problem is that re-initializing mocks right before the call made it impossible to use `override_json` down in the HubEdos::UserAttributes spec.

Getting `@campus_solutions_id` early in the case of a fake proxy lets mock initialization happen along with all the other initialization, and allows overrides to work as expected.

app/models/hub_edos/student.rb
```diff
@@ -7,7 +7,6 @@ class Student < Proxy
 
     def initialize(options = {})
       super(Settings.hub_edos_proxy, options)
-      initialize_mocks if @fake
```
paulkerschen: Here and below, now handled in superclass.

app/models/hub_edos/user_attributes.rb
```diff
@@ -6,19 +6,14 @@ class UserAttributes < Proxy
 
     def initialize(options = {})
       super(Settings.hub_edos_proxy, options)
-      initialize_mocks if @fake
     end
 
     def self.test_data?
-      @fake
+      Settings.hub_edos_proxy.fake.present?
```
paulkerschen: This method was always returning nil, since Ruby is perfectly happy to let class methods use their own instance variables, which nonetheless have nothing to do with the instance variables used by instances.

app/models/hub_edos/user_attributes.rb
```diff
@@ -6,19 +6,14 @@ class UserAttributes < Proxy
...
     end
 
-    def url
-      ''
-    end
-
-    def json_filename
-      'user_attributes.json'
```
paulkerschen: There is no 'user_attributes.json', which makes sense since this class calls no endpoints; letting it share in `initialize_mocks` behavior was causing nil responses to get in the way of other proxies.

spec/models/user/api_with_campus_solutions_enabled_spec.rb
```diff
@@ -117,22 +117,27 @@
...
   end
 
-  it "should say random student gets the academics tab", if: HubEdos::UserAttributes.test_data? do
+  it 'should say random student gets the academics tab', if: HubEdos::UserAttributes.test_data? do
     user_data = User::Api.new(@random_id).get_feed
-    user_data[:hasAcademicsTab].should be_truthy
+    expect(user_data[:hasAcademicsTab]).to eq true
   end
 
-  it "should say a staff member with no academic history does not get the academics tab", if: HubEdos::UserAttributes.test_data? do
```
paulkerschen: The always-nil `HubEdos::UserAttributes.test_data?` concealed this failing test from us.

raydavis:
Looks good to me! I'll wait a bit for @ctweney to peer at it, though.

ctweney:
Looks right to me! Thanks for the fake proxy fidlding.

PR #4480, opened 2015-11-06
SISRP-5143 - Dev/Implement Front-End: Funding sources Card

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-5143


raydavis:
I've never used "flex" -- do you think it might help with our outstanding vertical alignment issue at https://jira.ets.berkeley.edu/jira/browse/CLC-3552 ?

christianvuerings:
@raydavis that might indeed be the case. I would recommend people to start using it almost every time when you're thinking "Should this be a float or ..."

https://css-tricks.com/snippets/css/a-guide-to-flexbox/ is a great reference
http://caniuse.com/#search=flex

PR #4484, opened 2015-11-07
CLC-5932 - Hyperactive spinners

christianvuerings:
https://jira.ets.berkeley.edu/jira/browse/CLC-5932

### DIFF
--- a/src/assets/stylesheets/_useful.scss
+++ b/src/assets/stylesheets/_useful.scss
@@ -83,11 +83,9 @@
     transform: rotate(1turn);
   }
 }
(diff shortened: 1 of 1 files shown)


raydavis:
That is some head-spinning code... Is there an easy way of explaining the need for the "even"/"uneven" split?

christianvuerings:
@raydavis also added a link in the jira: http://stackoverflow.com/questions/15868431/keyframe-animations-not-in-sync

In Chrome, we can't rely on the CSS animations being in sync.

PR #4486, opened 2015-11-09
SISRP-10788 Fix conditional decision to show profile link

ctweney:
https://jira.berkeley.edu/browse/SISRP-10788

This JIRA blocks all performance testing so I plan to self-merge once T is happy.

### DIFF
--- a/src/assets/templates/widgets/gear_popover.html
+++ b/src/assets/templates/widgets/gear_popover.html
@@ -13,7 +13,7 @@
       <h4 data-ng-bind="api.user.profile.fullName"></h4>
     </div>
     <ul class="cc-popover-items">
-      <li class="cc-popover-item" data-ng-if="api.user.profile.showSisProfileUI && api.user.profile.roles.student">
+      <li class="cc-popover-item" data-ng-if="api.user.profile.showSisProfileUI">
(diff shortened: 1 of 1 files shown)


ctweney:
To clarify, this makes gear_popover.html use the same conditional used by profile.html. That means the backend makes the decision of show/not-show profile, as it should do.

src/assets/templates/widgets/gear_popover.html
```diff
@@ -13,7 +13,7 @@
       <h4 data-ng-bind="api.user.profile.fullName"></h4>
     </div>
     <ul class="cc-popover-items">
-      <li class="cc-popover-item" data-ng-if="api.user.profile.showSisProfileUI && api.user.profile.roles.student">
+      <li class="cc-popover-item" data-ng-if="api.user.profile.showSisProfileUI">
```
christianvuerings: @ctweney this change lgtm but does this mean we aren't setting the `roles.student` correctly?

ctweney: Yes. That will be a followup JIRA.

ctweney: The followup is https://jira.berkeley.edu/browse/SISRP-10789

PR #4487, opened 2015-11-09
CLC-5969, UITEST needs the Xvfb check before the 'if' block

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5969

```bamboo-tests.sh``` is used by Bamboo to run testext or run testext + uitests. See comments in CLC-5969 that highlight the oddity Xvfb placement. This change guarantees that Xvfb check happens when uitests are run.


raydavis:
Ping @pfarestveit !

PR #4488, opened 2015-11-09
SISRP-10799 - Awards / Funding sources follow up

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-10799


johncrossman:
Makes sense.

PR #4489, opened 2015-11-10
CLC-5965 Make current logic handle new CALCENTRAL_STUDENT_TERM_VW rows

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-5965

Our current code assumes only one term being stored in the view. We now have access to all of them, so on the First Do No Harm principle, we need to adapt.


paulkerschen:
Dizzying vistas open.

PR #4492, opened 2015-11-10
SISRP-6337 - CC - Going to /finances/finaid directly results in a 404

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-6337


raydavis:
Makes sense from here!

PR #4490, opened 2015-11-10
CLC-5996, front-end wiring for 'Welcome Delegate' page and widget to link accounts

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5996
https://jira.ets.berkeley.edu/jira/browse/CLC-5973

The magic URL would be /welcome_delegate   The next step is to put some meat on the bones of `widgets/delegated_access/link_accounts.html`

You'll notice I also tossed in feature flag that will be used in upcoming PRs

PR #4493, opened 2015-11-10
CLC-5996, front-end wiring for 'Welcome Delegate' page and widget to link accounts

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5996
https://jira.ets.berkeley.edu/jira/browse/CLC-5973

The magic URL would be /welcome_delegate The next step is to put some meat on the bones of widgets/delegated_access/link_accounts.html  Plus, 'cs_delegated_access ' feature flag.

(See relevant comments on closed PR #4490)

### DIFF
--- a/src/assets/templates/welcome_delegate.html
+++ b/src/assets/templates/welcome_delegate.html
@@ -1,0 +1,7 @@
+<div class="cc-page-welcome-delegate">
+
+  <div class="cc-widget cc-widget-delegated-access" data-ng-if="api.user.profile.features.cs_delegated_access">
+    <div data-ng-include src="'widgets/delegated_access/link_accounts.html'"></div>
+  </div>
+
+</div>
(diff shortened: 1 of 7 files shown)


johncrossman:
Travis is 50% green

paulkerschen:
Our Story Begins.

src/assets/templates/welcome_delegate.html
```diff
@@ -0,0 +1,7 @@
+<div class="cc-page-welcome-delegate">
```
christianvuerings: @johncrossman would it be possible to put the `ng-if` on here instead? That way none of the html below should be loaded.

`data-ng-if="api.user.profile.features.cs_delegated_access"`

johncrossman: @christianvuerings - I will make that change in upcoming PR w/ widget work. Thanks

PR #4494, opened 2015-11-10
CLC-5982 Backend wiring for delegate access creation

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5982

This only covers our best guess as to what the happy path will look like. Error and failure handling will require a working API to develop against.


paulkerschen:
Wrong branch, closed.

PR #4497, opened 2015-11-11
SISRP-10875 - Profile - odd horizontal bar during updates

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-10875


johncrossman:
Spin on.

PR #4498, opened 2015-11-11
SISRP-10880 - Profile - very long strings should wrap

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-10880


johncrossman:
Wordy wrappinghood

PR #4499, opened 2015-11-11
CLC-6005 - Weird behavior when opening the side menu on an iOS9 device & latest safari

christianvuerings:
https://jira.ets.berkeley.edu/jira/browse/CLC-6005


christianvuerings:
Follow up at #4502

PR #4501, opened 2015-11-11
SISRP-8922 - FA Profile Card - need a way to change T4 Authorization

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-8922


johncrossman:
Looks good

PR #4507, opened 2015-11-12
SISRP-10872 Fetch and invalidate Student API in smaller chunks

ctweney:
https://jira.berkeley.edu/browse/SISRP-10872

### DIFF
--- a/app/models/hub_edos/demographics.rb
+++ b/app/models/hub_edos/demographics.rb
@@ -1,0 +1,23 @@
+module HubEdos
+  class Demographics < Student
+
+    include Cache::UserCacheExpiry
+
+    def initialize(options = {})
+      super(options)
+    end
+
+    def url
+      "#{@settings.base_url}/#{@campus_solutions_id}/demographic"
+    end
+
+    def json_filename
+      'demographics.json'
+    end
+
+    def include_fields
+      %w(ethnicities languages usaCountry foreignCountries)
+    end
+
+  end
+end
--- a/app/models/hub_edos/student.rb
+++ b/app/models/hub_edos/student.rb
@@ -20,19 +20,21 @@ module HubEdos
     end
 
     def build_feed(response)
-      transformed_response = redact_sensitive_keys(transform_address_keys(parse_response(response)))
+      transformed_response = filter_fields(redact_sensitive_keys(transform_address_keys(parse_response(response))))
       {
-        'student' => transformed_response['studentResponse']['students']['students'][0]
+        'student' => transformed_response
       }
     end
 
     def transform_address_keys(response)
       # this should really be done in the Integration Hub, but they won that argument due to time constraints.
       response['studentResponse']['students']['students'].each do |student|
-        student['addresses'].each do |address|
-          address['state'] = address.delete('stateCode')
-          address['postal'] = address.delete('postalCode')
-          address['country'] = address.delete('countryCode')
+        if student['addresses'].present?
+          student['addresses'].each do |address|
+            address['state'] = address.delete('stateCode')
+            address['postal'] = address.delete('postalCode')
+            address['country'] = address.delete('countryCode')
+          end
         end
       end
       response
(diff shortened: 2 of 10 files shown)


paulkerschen:
Looks good to me. Will leave open in case @raydavis or @johncrossman want a look-see.

johncrossman:
I'll have a quick look

app/models/hub_edos/student.rb
```diff
@@ -50,5 +52,21 @@ def redact_sensitive_keys(response)
       response
     end
 
+    def filter_fields(response)
+      # only include the fields that this proxy is responsible for
+      if include_fields.nil?
+        return response['studentResponse']['students']['students'][0]
```
johncrossman: The ```['students']['students']``` is intentional?

app/models/hub_edos/demographics.rb
```diff
@@ -0,0 +1,23 @@
...
+    def url
+      "#{@settings.base_url}/#{@campus_solutions_id}/demographic"
+    end
+
+    def json_filename
+      'demographics.json'
+    end
+
+    def include_fields
+      %w(ethnicities languages usaCountry foreignCountries)
```
christianvuerings: Just as a reference, these are the fields that are part of neither `contacts` or `demographic`:

* birth
* affiliations
* gender

@ctweney does that mean we'll no longer be able to show the gender to the student?

app/models/hub_edos/student.rb
johncrossman: Now I see... Test data would indicate yes

johncrossman:
LGTM. @christianvuerings can merge once he has an answer.

app/models/hub_edos/demographics.rb
christianvuerings: @ctweney for instance, this is where we're currently using it: http://git.io/v8hnQ

christianvuerings:
It'll be interesting to see whether this will actually be way more performant. If I understand this correctly, we're calling 2 API's now instead of 1 and not including birth, affiliations & gender (and we do need the last one).

So does this then result in 2 less calls to Campus Solutions (for birth and affiliations)? Or did I miss something?

app/models/hub_edos/demographics.rb
ctweney: Correct: With this change we no longer get birth, affiliations, or gender. I'll file a JIRA to have the integrations team add those to one of the endpoints (or make a new endpoint). Right now they have their hands 150% full making efficiency improvements to pass the load tests, so it may take some time.

ctweney:
@christianvuerings This results in fewer calls to CS when we re-fetch /api/edos/student after the user has modified their profile data. The re-fetch will only hit Hub's Contacts API, not the All API. That's the efficiency gain. For the user-just-logged-in case, we decrease efficiency a tiny bit by making 2 calls instead of 1.

app/models/hub_edos/student.rb
ctweney: I said, I said, ayup, ayup, that is correct sir. </foghornLeghorn>

christianvuerings:
@ctweney so do we still feel comfortable with going ahead with this merge and showing 'Not Available' for gender even though we do have that data in `/all`?

Also, for GL4, I will be adding edit capabilities to demographic information, so in that case we do need to re-fetch the demographics API.
https://jira.berkeley.edu/browse/SISRP-10779

ctweney:
Requested granular access to the missing fields here: https://jira.berkeley.edu/browse/SISRP-10974

And yes, I think we should go ahead with this merge on the assumption that integration will be able to solve SISRP-10974 in time for GL3. Otherwise, we will have a lot of trouble passing the load tests.

christianvuerings:
@ctweney How about accessing the `/all` on first request and then only fetching `/contacts` after that? That way we wouldn't have to introduce the bug where gender information is missing.

ctweney:
@christianvuerings  That would be incredibly hard to engineer with the way our cache expiration works. Brittle and error prone.

christianvuerings:
@ctweney Do we have another solution that would be better, works with our cache expiration and avoids introducing the `gender` bug?

ctweney:
@christianvuerings  No, the integration team will have to fix https://jira.berkeley.edu/browse/SISRP-10779

christianvuerings:
As discussed with @ctweney, we'll be holding off on this change until next week.

PR #4509, opened 2015-11-12
CLC-6010 OEC: Display Google authorization errors in UI

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6010


johncrossman:
@paulkerschen do we care if local_write=true?

johncrossman:
Ignore that concern. We were not protected from Google dependency before.

paulkerschen:
Bamboozling at https://bamboo.ets.berkeley.edu/bamboo/browse/CPB-CPT215-1

PR #4510, opened 2015-11-12
SISRP-10915 Consume Student API as XML instead of JSON

ctweney:
https://jira.berkeley.edu/browse/SISRP-10915

Introduces Nori, the only XML parsing lib I could find that dealt intelligently with typecasting XML strings of "false" and "true" into boolean true and false. Notice also the horrible convert_plurals tomfoolery which is required when you go from XML to JSON. Previously, that jiggery-pokery was on the Hub; now we own the argle-bargle. 

Front-end will experience a tiny change in collections that can be null. It didn't break anything, when I tried it, but it bears testing with (for example) users with no language, 1 language, and multiple languages. 

Slight change in strategy: We want to load test this XML-instead-of-JSON change first (on Monday) because the other change (from SISRP-10872 / PR #4507 ) breaks the birth, gender, and affiliations data. Resolving that breakage requires some additional code from the integrations team. We'll test that change on Tuesday in a separate build.


christianvuerings:
@ctweney lgtm!

If this does fix the Hub issue, imo it would be worth keeping it.
However, if it doesn't have big performance gains, we should probably revert this change and let the Hub do the JSON parsing.

paulkerschen:
@ctweney Why is `HubEdos::Person` going away as part of this? Is is that the class isn't hooked up to the UI for GL3, so it's not necessary to convert it over to XML?

ctweney:
@paulkerschen Yeah, Person is junk at this point (there's no API corresponding to it in the Hub) so I got rid of it for now.

PR #4511, opened 2015-11-13
Revert "SISRP-10915 Consume Student API as XML instead of JSON"

ctweney:
This reverts commit 07e1aea93a6fb83743310a4883b0ed22bbf7307d.

As soon as this is merged I'll issue a new pull which has a fix for SISRP-10872 (so that we can test that change separately from this change).


raydavis:
Nori! Come back!

PR #4513, opened 2015-11-13
SISRP-10579 - Profile - address input is being truncated after the user clicks Save

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-10579


johncrossman:
<max-goodness />

PR #4518, opened 2015-11-13
CLC-5971, real delegate_terms_and_conditions wiring; rename student_link to delegate_link

johncrossman:
**PLEASE MERGE AFTER QA BRANCH IS CUT AT 3PM**

https://jira.ets.berkeley.edu/jira/browse/CLC-5972

SIS has not given us API but all other wiring is now ready. I took this opportunity to do some renaming: ```delegate/linking.html``` and ```link-delegate``` is congruent with backend naming.


johncrossman:
Branch has been cut. Green light to review 'n merge.

src/assets/templates/widgets/delegate/linking.html
```diff
@@ -0,0 +1,44 @@
...
+        <input type="text" id="cc-page-widget-link-delegate-email" data-ng-model="currentObject.data.email" aria-required="true" />
+      </div>
+      <div>
+        <input type="checkbox" id="cc-page-widget-link-delegate-terms" data-ng-model="currentObject.data.agreeToTerms">
+        <label for="cc-page-widget-link-delegate-terms">I agree to the Terms and Conditions.</label>
+        <button class="cc-button-link" data-ng-if="!showTermsAndConditions" data-ng-click="getTermsAndConditions()">Show</button>
+        <p class="cc-page-widget-link-delegate-agreement" data-ng-if="showTermsAndConditions" data-ng-bind="feed.daAgreeToTerms.termsAndConditions"></p>
+      </div>
+      <div>
+        <button class="cc-button cc-button-blue" type="submit" data-ng-submit="linkAccounts()" data-ng-disabled="isSaving">
```
paulkerschen: @johncrossman: I wasn't familiar with the `currentObject` pattern but I see it's being used in the profile front end.

Following the models in the profile code, it looks like `$scope.currentObject` and $`scope.currentObject.data` both need to be initialized in the Angular controller (and we may not need a separate `currentObject.data` nested inside `currentObject`). That object should then be passed to the `linkAccounts()` call above on submit. Does that match how you're thinking about it?

johncrossman: Yes, that pattern is my intention.

src/assets/javascripts/angular/factories/delegateLinkingFactory.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+var angular = require('angular');
+
+/**
+ * POST to the Campus Solutions API which links delegate to a student account
+ */
+angular.module('calcentral.factories').factory('delegateLinkingFactory', function($http) {
+  var urlDelegateAccessAPI = '/api/campus_solutions/delegate_access';
```
paulkerschen: @johncrossman Here's a question: do you think the terms and conditions should come from a GET call to `/api/campus_solutions/delegate_access`, or would it be more logical to use that URL for the list of students and roles returned to the delegate?

johncrossman: @paulkerschen my PR comes first so it's the winner!  Seriously tho', is there a precedent for overloading #get such that we don't need another controller?  I'll take a moment to code review

paulkerschen: We could add three one-line methods to DelegateAccessController - say `get_students`, `get_terms_and_conditions` and `get_campus_solutions_url` - and hook them up to distinct paths in routes.rb.

johncrossman: thanks @paulkerschen  makes more sense... commit coming shortly

src/assets/templates/widgets/delegate/linking.html
```diff
@@ -0,0 +1,44 @@
...
+          Email Address [1]: <span class="cc-required-field-indicator">*</span>
+        </label>
+      </div>
+      <div>
+        <input type="text" id="cc-page-widget-link-delegate-email" data-ng-model="currentObject.data.email" aria-required="true" />
+      </div>
+      <div>
+        <input type="checkbox" id="cc-page-widget-link-delegate-terms" data-ng-model="currentObject.data.agreeToTerms">
+        <label for="cc-page-widget-link-delegate-terms">I agree to the Terms and Conditions.</label>
+        <button class="cc-button-link" data-ng-if="!showTermsAndConditions" data-ng-click="getTermsAndConditions()">Show</button>
```
christianvuerings: @johncrossman Since we're loading this asynchronously, would it be possible to show a spinner for this as well? It's probably pretty fast locally but not on a slow 3G connection.
Should be testable by wrapping a `$timeout` around the request.

src/assets/templates/widgets/delegate/linking.html
```diff
@@ -0,0 +1,44 @@
...
+    <form>
+      <h3 class="cc-widget-link-delegate-title">Link Your Account with Your Student</h3>
+      <div>Enter information from the email message you received:</div>
+      <div>
+        <label for="cc-page-widget-link-delegate-key">
+          Security Key: <span class="cc-required-field-indicator">*</span>
+        </label>
+      </div>
+      <div>
+        <input type="text" id="cc-page-widget-link-delegate-key" data-ng-model="currentObject.data.securityKey" aria-required="true" />
```
christianvuerings: @johncrossman if this is required, we should probably also add `required="true"`

src/assets/javascripts/angular/factories/delegateLinkingFactory.js
```diff
@@ -0,0 +1,22 @@
...
+
+  var getTermsAndConditions = function() {
+    return $http.get(urlDelegateAccessAPI);
```
christianvuerings: @johncrossman for `.get` please use our http service instead `return apiService.http.request(options, url);`.
The main reason is that it allows for better cache handling.

src/assets/templates/widgets/delegate/linking.html
johncrossman: Jira to do that in follow up PR: https://jira.ets.berkeley.edu/jira/browse/CLC-6014

raydavis:
Travis-CI reports:

Failures:
  1) CampusSolutions::DelegateAccessController authenticated user should get terms and conditions
     Failure/Error: get :get
     ActionController::UrlGenerationError:
       No route matches {:action=>"get", :controller=>"campus_solutions/delegate_access"}
     # ./spec/controllers/campus_solutions/delegate_access_controller_spec.rb:28:in `(root)'

johncrossman:
Thanks @raydavis  The move away from generic :get deserved another run of spec. I was in a rush on friday... to social hour.

paulkerschen:
In you go, then.

PR #4525, opened 2015-11-16
CLC-5971, delegateService.js provides end-to-end POST op

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5967

Proper wiring to server-side #post with conventional delegateService.js wired in. Rename delegateLinkingFactory.js -> delegateFactory.js for possible use with /toolbox content.


src/assets/javascripts/angular/services/delegateService.js
```diff
@@ -0,0 +1,33 @@
+'use strict';
+
+var angular = require('angular');
+
+angular.module('calcentral.services').service('delegateService', function() {
+  /**
+   * Fired after an action (e.g., save) has finished
+   */
+  var actionCompleted = function($scope, data, callback) {
```
christianvuerings: @johncrossman instead of using callbacks, could we use a `promise` here?
In Angular, we do that with the [$q service](https://docs.angularjs.org/api/ng/service/$q)

src/assets/stylesheets/_welcome_delegate.scss
```diff
@@ -1,4 +1,11 @@
 .cc-page-welcome-delegate {
+  .cc-page-widget-link-delegate-content-error {
```
christianvuerings: @johncrossman it looks like in this file we use a list of prefixes:

* `cc-page-welcome-delegate`
* `cc-page-widget-link-delegate-content`
* `cc-delegate`
* `cc-widget-link-delegate`

Would it be possible to pick one?

johncrossman:
@christianvuerings renaming was good idea - thanks. Let me know if promise looks right.

paulkerschen:
If @christianvuerings is available to confirm that the $q service looks right (not my area of expertise either), the PR looks good to me.

src/assets/javascripts/angular/services/delegateService.js
```diff
@@ -0,0 +1,33 @@
+'use strict';
+
+var angular = require('angular');
+
+angular.module('calcentral.services').service('delegateService', function($q) {
+  /**
+   * Fired after an action (e.g., save) has finished
+   */
+  var actionCompleted = function(data) {
+    if (data.errorMessage) {
```
christianvuerings: @johncrossman could you try to do it the following way:

```
// for the purpose of this example let's assume that variables `$q` and `okToGreet`
// are available in the current lexical scope (they could have been injected or passed in).

function asyncGreet(name) {
  var deferred = $q.defer();

  setTimeout(function() {
    deferred.notify('About to greet ' + name + '.');

    if (okToGreet(name)) {
      deferred.resolve('Hello, ' + name + '!');
    } else {
      deferred.reject('Greeting ' + name + ' is not allowed.');
    }
  }, 1000);

  return deferred.promise;
}
```

src/assets/stylesheets/_welcome_delegate.scss
```diff
@@ -18,19 +16,23 @@
...
+    .cc-delegate-linking-error {
+      background-color: $cc-color-red-light;
+      border-bottom: 1px solid $cc-color-red;
+      margin: -10px -10px 10px;
+      padding: 10px;
+      white-space: pre-line;
     }
-    .cc-widget-link-delegate {
-      margin: 7px 5px 0;
+    .cc-delegate-linking-title {
```
christianvuerings: @johncrossman thanks for making these consistent!

christianvuerings:
Lgtm as soon as we did the `$q` changes.

paulkerschen:
Merci @johncrossman.

PR #4527, opened 2015-11-17
[QA] CLC-6016 Do not allow cache warmer to trigger disabled features

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6016


johncrossman:
Half happy is good enough for me. Thanks @paulkerschen

ctweney:
Whoa, this is a problematic change. This creates a difference in the data between what cache warming would do, and what regular invocation would do. That is going to create some very difficult inconsistencies. I would recommend rethinking this change and putting the feature-flag check on the model level instead of the warmer level.

PR #4529, opened 2015-11-17
**DO NOT MERGE** Implemented Loan Payment Estimate widget onto Dashboard

davidlimcheng:
**DO NOT MERGE**

@christianvuerings - PR for SISRP-4440, implement Loan Payment Estimate widget for practice purposes only.  I probably could've done the CSS a lot more efficiently - nevertheless, looking forward to your feedback.  Thanks!


christianvuerings:
@davidlimcheng looks like travis is failing - probably because of one of our linting tools.
https://travis-ci.org/ets-berkeley-edu/calcentral/builds/91538298

Feel free to push up a fix after.

Gemfile.lock
```diff
@@ -515,3 +515,6 @@ DEPENDENCIES
   torquebox-messaging (~> 3.1.2)
   torquebox-server (~> 3.1.2)
   webmock (~> 1.20.4)
+
+BUNDLED WITH
+   1.10.4
```
christianvuerings: @davidlimcheng Do you know why this has been added, don't think we ever needed that before.

public/dummy/json/loan_payment_estimate.json
```diff
@@ -0,0 +1,12 @@
+{
+  "estimatedBalance": 38772,
+  "monthlyPayment": 114,
+  "paymentWithInterest": 500,
+  "years": 15,
+  "paymentPlan": {
+    "year": [2015, 2020, 2025, 2030],
+    "balance": [
+    [40000, 32000, 20000, 0]
```
christianvuerings: @davidlimcheng just as a fyi, we usually format it in a different way.
I usually use [jsonlint](http://jsonlint.com/) and then replace it with 2 spaces.

public/dummy/json/loan_payment_estimate.json
```diff
@@ -0,0 +1,12 @@
+{
+  "estimatedBalance": 38772,
+  "monthlyPayment": 114,
+  "paymentWithInterest": 500,
+  "years": 15,
+  "paymentPlan": {
+    "year": [2015, 2020, 2025, 2030],
+    "balance": [
```
christianvuerings: @davidlimcheng what is the reason we needed an array with 1 element that has 4 elements?

src/assets/javascripts/angular/controllers/widgets/loanPaymentEstimateController.js
```diff
@@ -0,0 +1,30 @@
...
+
+var angular = require('angular');
+
+/**
+ * Loan Payment Estimate Controller
+ */
+angular.module('calcentral.controllers').controller('loanPaymentEstimateController', function(loanPaymentEstimateFactory, $scope) {
+  // Get JSON dummy data
+  var getLoanSummary = function() {
+    loanPaymentEstimateFactory.getLoanData().success(function(data) {
```
christianvuerings: @davidlimcheng would recommend adding a `return` here

src/assets/javascripts/angular/controllers/widgets/loanPaymentEstimateController.js
```diff
@@ -0,0 +1,30 @@
...
+      // Format monthlyPayment value
+      data.monthlyPayment = '$' + data.monthlyPayment.toString() + '/MO';
```
christianvuerings: @davidlimcheng do you need the `toString()` here and anywhere else? 
Basically does the following work as well:

```
data.monthlyPayment = '$' + data.monthlyPayment + '/MO';
```

src/assets/javascripts/angular/controllers/widgets/loanPaymentEstimateController.js
```diff
@@ -0,0 +1,30 @@
...
+      angular.extend($scope, data);
+    });
+  };
+  // Angular-Chart configuration
+  $scope.options = {
+    pointDot: false,
+    scaleLabel: function(label) {
+      var valueString = label.value.toString();
+      if (valueString.substring(valueString.length - 3, valueString.length) === '000') {
```
christianvuerings: @davidlimcheng can you add a comment on what you're trying to do here?

src/assets/templates/widgets/loan_payment_estimate.html
```diff
@@ -0,0 +1,26 @@
+<div data-ng-controller="loanPaymentEstimateController">
+  <div class="cc-widget">
+    <div class="cc-widget-title">
+      <div class="row collapse">
```
christianvuerings: @davidlimcheng do we need all the `row collapse` and `columns` here?

src/assets/templates/widgets/loan_payment_estimate.html
```diff
@@ -0,0 +1,26 @@
...
+        <div class="columns"><h2>Loan Payment Estimate</h2></div>
+      </div>
+    </div>
+    <div class="cc-balance-container cc-clearfix-container">
+      <div class="cc-left cc-estimated-balance">
+        <div class="cc-balance-left"> {{estimatedBalance | currency:$:0}} </div>
```
christianvuerings: @davidlimcheng try to use `data-ng-bind` and `data-ng-bind-template` instead of the curly brackets `{{`.
http://stackoverflow.com/questions/16125872/angularjs-why-ng-bind-is-better-than-in-angular

src/assets/templates/widgets/loan_payment_estimate.html
```diff
@@ -0,0 +1,26 @@
...
+      <div class="cc-left cc-estimated-monthly-payments">
+        <div class="cc-balance-right"> {{monthlyPayment}} </div>
+        <div class="cc-balance-title">ESTIMATED MONTHLY PAYMENTS</div>
+      </div>
+    </div>
+    <div class="cc-chart-container">
+      <canvas id="line" class="chart chart-line" chart-data="paymentPlan.balance" chart-labels="paymentPlan.year" chart-options="options"></canvas>
+    </div>
+    <div class="cc-loan-payment-summary-container">
+    <p> It's estimated that you'll pay <strong>{{paymentWithInterest | currency:$:0}}</strong> (including loan interest) over <strong>{{years}}</strong> years.
```
christianvuerings: @davidlimcheng can we intend this correctly?

Gemfile.lock
davidlimcheng: I believe it got added when I updated gems after fetching the repository:
http://stackoverflow.com/questions/30713922/how-to-stop-rails-from-adding-bundled-with-to-the-gemfile-lock

Seems like something that was added in Bundler v1.10 - Should I be doing a checkout of this file before committing?

public/dummy/json/loan_payment_estimate.json
davidlimcheng: @christianvuerings I believe it's because of the way chart.js parses the datasets.  Taken from the chart.js documentation:
"The data for line charts is broken up into an array of datasets. Each dataset has a colour for the fill, a colour for the line and colours for the points and strokes of the points."
So the "parent" array acts as a wrapper in the case of multiple datasets that we want to show on the same chart - allowing for multiple series of line graphs.

public/dummy/json/loan_payment_estimate.json
davidlimcheng: @christianvuerings gotcha!  Bookmarked.  Do you mean you add 2 spaces to the end of the outermost object?  Why?

src/assets/javascripts/angular/controllers/widgets/loanPaymentEstimateController.js
davidlimcheng: @christianvuerings would you recommend to generally add a return to the end of every function?  Something like this?
http://stackoverflow.com/questions/17337064/does-every-javascript-function-have-to-return-a-value

src/assets/javascripts/angular/controllers/widgets/loanPaymentEstimateController.js
davidlimcheng: @christianvuerings Definitely do not need the toString() method here - this being my first go at it, I thought it was a bit odd mixing & matching different primitive types.  But I totally see that it is an unnecessary process!

src/assets/javascripts/angular/controllers/widgets/loanPaymentEstimateController.js
davidlimcheng: @christianvuerings Here, I am trying to configure the y-axis of the chart to reflect the card design.  Chart.js's y-axis values default to a simple integer.  First, I check to see if the value ends with three zeroes - and if it does, remove it, since the appendage of "k" at the end will make them redundant (e.g., $40000 = $40k).  If the value does not end with three zeroes, it does nothing, and the formatting stays consistent (e.g., $0 = $0k).  

Talking through this, I realize that this would totally break if given a value like $500, which would return $500k instead of $.5k.

showTooltips is the option to turn on/off the data that is shown when hovering over the chart.  

It does seem like an overcomplicated way of formatting the y-axis.  Was there a better way to achieve this?

src/assets/templates/widgets/loan_payment_estimate.html
davidlimcheng: @christianvuerings Nope!  Could have gotten rid of those divs as well.  Remnants of the template that I used as a base - sorry, that was sloppy.

src/assets/templates/widgets/loan_payment_estimate.html
davidlimcheng: @christianvuerings Makes sense!

src/assets/templates/widgets/loan_payment_estimate.html
davidlimcheng: @christianvuerings what do you mean?

davidlimcheng:
@christianvuerings Thanks a bunch for the feedback.  Very much appreciated - as you know this is my first foray into development so any advice is welcome.  The Travis CI failing is an interesting one - I'll try to figure out what's going on and push another commit, along with all of the changes we discussed.

davidlimcheng:
@christianvuerings huge thanks again for the feedback + direction.  Should I go ahead and close this PR?

christianvuerings:
@davidlimcheng done, and you're welcome

PR #4528, opened 2015-11-17
[QA] CLC-6016 Move feature-flag blocking from cache to models

paulkerschen:
Follow-up to #4527 addressing @ctweney's comment.

https://jira.ets.berkeley.edu/jira/browse/CLC-6016

PR #4531, opened 2015-11-17
[QA] CLC-6016 Finaid-related Campus Solutions user models should observe finaid flag

paulkerschen:
I hope this is my last QA PR under https://jira.ets.berkeley.edu/jira/browse/CLC-6016.

A couple of Finaid-related models are switched from the CS profile flag (which I hastily put in yesterday) to the CS finaid flag. Using the wrong flag would presumably have created a separate leak between GL3 and GL4 ("...the _Titanic_ and her sister ship _Olympic_ were built with a cellular double bottom and divided into 16 major watertight compartments with 15 transverse watertight bulkheads that ran clear across the ship...").

Bamboo has built at https://bamboo.ets.berkeley.edu/bamboo/browse/CPB-CPT216-1. I believe the failures, including the CampusSolutions::AidYears failure, are red herrings; they're triggered by Campus Solutions errors farther down the stack and are showing up on all our branches at present.

(Parts one and two are at #4527 and #4528; if this last PR checks out, a squashed PR of all three into master will follow).

### DIFF
--- a/app/models/campus_solutions/my_aid_years.rb
+++ b/app/models/campus_solutions/my_aid_years.rb
@@ -4,10 +4,10 @@ module CampusSolutions
     include Cache::LiveUpdatesEnabled
     include Cache::FreshenOnWarm
     include Cache::JsonAddedCacher
-    include CampusSolutions::ProfileFeatureFlagged
+    include CampusSolutions::FinaidFeatureFlagged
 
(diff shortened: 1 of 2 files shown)


raydavis:
Sounds reasonable. Thanks for the detective work!

PR #4533, opened 2015-11-17
CLC-6020 Expect logger levels as integers, not strings

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6020

Rails initialization is always twisty, but the crux seems to be that our logging constants (from Log4r) are evaluated as integers 0 through 6, not the corresponding string values; so the integers are what to check for.

### DIFF
--- a/lib/calcentral_config.rb
+++ b/lib/calcentral_config.rb
@@ -73,17 +73,20 @@ module CalcentralConfig
     files
   end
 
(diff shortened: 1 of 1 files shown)


johncrossman:
I suspected numeric v. token...  Thank you, @paulkerschen  Waiting on Mr. T

johncrossman:
level = TOP_NOTCH

PR #4535, opened 2015-11-18
CLC-6021 - Roster photos: printing in Firefox only prints one page

christianvuerings:
https://jira.ets.berkeley.edu/jira/browse/CLC-6021


johncrossman:
+1

PR #4537, opened 2015-11-18
CLC-6018 Robustly handle Hub ID lookup failures

paulkerschen:
A Campus Solutions ID lookup failure from the Hub will return a feed of `{noStudentId: true}`, with no status code; so we oughtn't to blow up if a status code is missing.

https://jira.ets.berkeley.edu/jira/browse/CLC-6018

### DIFF
--- a/app/models/hub_edos/my_student.rb
+++ b/app/models/hub_edos/my_student.rb
@@ -17,13 +17,13 @@ module HubEdos
       return merged unless is_cs_profile_feature_enabled
 
       [HubEdos::Contacts, HubEdos::Demographics, HubEdos::Affiliations].each do |proxy|
-        feed = proxy.new({user_id: @uid}).get
-        if feed[:statusCode] > 400
+        hub_response = proxy.new({user_id: @uid}).get
+        if hub_response[:noStudentId] || hub_response[:statusCode] > 400
           merged[:statusCode] = 500
           merged[:errored] = true
           logger.error("Got errors in merged student feed on #{proxy} for uid #{@uid}")
         else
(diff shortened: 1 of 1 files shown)


app/models/hub_edos/my_student.rb
```diff
@@ -18,7 +18,7 @@ def get_feed_internal
 
       [HubEdos::Contacts, HubEdos::Demographics, HubEdos::Affiliations].each do |proxy|
         feed = proxy.new({user_id: @uid}).get
-        if feed[:statusCode] > 400
+        if !feed[:statusCode] || feed[:statusCode] > 400
```
johncrossman: @paulkerschen is the following a better communicate the biz logic:
```if internal_response[:noStudentId] || internal_response[:statusCode] > 400```

Also, the underlying campus_solutions/proxy class does not treat :noStudentId as an error. Is this code different in that respect?

paulkerschen: @johncrossman There are a lot of different ways to slice it. Our Bearfacts and financial aid proxies also have had to deal with the no-student-id case and aren't especially standardized about it; some classes return nil, some insert their own `statusCode` with a 4xx or 5xx value; some return `noStudentId: true` and expect code farther upstream to wrap it in an error response. 

For sure standardization across the board is desirable, but for the moment it seemed best just to protect this volatile Hub code against blowups in a world where feeds might not have the expected elements. What do you think?

paulkerschen: @johncrossman I am convinced! Will amend.

johncrossman:
The renaming is helpful. Thanks @paulkerschen  I'll merge when green T.

PR #4539, opened 2015-11-18
CLC-5927, for now, hide My Campus links to Berkeley Campus Solutions

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5927

This particular My Campus link has been set to 'false' for all users. What we do on master branch depends on answer to question in Jira above.


paulkerschen:
"Campus Solutions: too secret for staff."

PR #4540, opened 2015-11-18
CLC-6019 Decouple SIR and profile feature flags on back end

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6019


johncrossman:
You are an enabler, sir.

PR #4541, opened 2015-11-18
SISRP-10912 - Remove semester options in CalCentral

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-10912

**Note**: Calcentral-sis-dev is actually broken until this PR gets merged (since the API changed)


paulkerschen:
The semester option looks cleanly extracted as far as I can see. Let testing continue.

PR #4543, opened 2015-11-19
CLC-5968, Campus Solutions URL with which students can manage delegates

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5966

Proxy for getting URL from CS. Plus, front-end profile widget (disabled) so students can users can click over to CS to add / remove delegates.

### DIFF
--- a/app/models/campus_solutions/delegate_management_url.rb
+++ b/app/models/campus_solutions/delegate_management_url.rb
@@ -1,0 +1,22 @@
+module CampusSolutions
+  class DelegateManagementURL < DirectProxy
+
+    def initialize(options = {})
+      super options
+      initialize_mocks if @fake
+    end
+
+    def response_root_xml_node
+      'DELEGATED_ACCESS_URL'
+    end
+
+    def xml_filename
+      'delegate_management_url.xml'
+    end
+
+    def url
+      "#{@settings.base_url}/UC_CC_DELEGATED_ACCESS_URL.v1/get"
+    end
+
+  end
+end
--- a/app/models/campus_solutions/delegate_management_url.rb
+++ b/app/models/campus_solutions/delegate_management_url.rb
@@ -1,13 +1,11 @@
 module CampusSolutions
   class DelegateManagementURL < DirectProxy
 
+    include DelegatedAccessFeatureFlagged
+
     def initialize(options = {})
       super options
       initialize_mocks if @fake
-    end
-
-    def response_root_xml_node
-      'DELEGATED_ACCESS_URL'
     end
 
     def xml_filename
--- a/fixtures/xml/campus_solutions/delegate_management_url.xml
+++ b/fixtures/xml/campus_solutions/delegate_management_url.xml
@@ -1,0 +1,4 @@
+<?xml version="1.0"?>
+<DELEGATED_ACCESS_URL>
+  <URL><![CDATA[https://bcs-web-dev-03.is.berkeley.edu:8443/psc/bcsdev/EMPLOYEE/HRMS/c/SA_LEARNER_SERVICES.SS_CC_DA_SHAREINFO.GBL?Page=SS_CC_DA_HDR&Action=A&ForceSearch=Y&TargetFrameName=None]]></URL>
+</DELEGATED_ACCESS_URL>
--- a/fixtures/xml/campus_solutions/delegate_management_url.xml
+++ b/fixtures/xml/campus_solutions/delegate_management_url.xml
@@ -1,4 +1,4 @@
 <?xml version="1.0"?>
 <DELEGATED_ACCESS_URL>
-  <URL><![CDATA[https://bcs-web-dev-03.is.berkeley.edu:8443/psc/bcsdev/EMPLOYEE/HRMS/c/SA_LEARNER_SERVICES.SS_CC_DA_SHAREINFO.GBL?Page=SS_CC_DA_HDR&Action=A&ForceSearch=Y&TargetFrameName=None]]></URL>
+  <URL>https://bcs-web-dev-03.is.berkeley.edu:8443/psc/bcsdev/EMPLOYEE/HRMS/c/SA_LEARNER_SERVICES.SS_CC_DA_SHAREINFO.GBL?Page=SS_CC_DA_HDR&amp;Action=A&amp;ForceSearch=Y&amp;TargetFrameName=None</URL>
 </DELEGATED_ACCESS_URL>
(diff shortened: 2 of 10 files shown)


johncrossman:
**Note:** ```delegate.html``` is not styled yet.

fixtures/xml/campus_solutions/delegate_management_url.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0"?>
+<DELEGATED_ACCESS_URL>
+  <URL><![CDATA[https://bcs-web-dev-03.is.berkeley.edu:8443/psc/bcsdev/EMPLOYEE/HRMS/c/SA_LEARNER_SERVICES.SS_CC_DA_SHAREINFO.GBL?Page=SS_CC_DA_HDR&Action=A&ForceSearch=Y&TargetFrameName=None]]></URL>
```
christianvuerings: @johncrossman interesting, first time I'm seeing the `CDATA` in a fixture. Do we actually need it? Perhaps something worthwhile to check in with the CS API devs about.

src/assets/javascripts/angular/controllers/widgets/profile/delegateController.js
```diff
@@ -0,0 +1,18 @@
...
+var angular = require('angular');
+
+/**
+ * Controller for users wanting to manage delegates
+ */
+angular.module('calcentral.controllers').controller('DelegateController', function(delegateFactory, $scope) {
+  var loadInformation = function() {
+    $scope.isLoading = true;
+    delegateFactory.getManageDelegatesURL().then(function(data) {
+      angular.extend($scope, data.data.feed.root);
```
christianvuerings: @johncrossman would recommend using `_.get(data, 'data.feed.root')` instead, will avoid errors when e.g. the `feed` property wouldn't exist.

src/assets/javascripts/angular/controllers/widgets/profile/delegateController.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+var angular = require('angular');
+
+/**
+ * Controller for users wanting to manage delegates
+ */
+angular.module('calcentral.controllers').controller('DelegateController', function(delegateFactory, $scope) {
+  var loadInformation = function() {
+    $scope.isLoading = true;
```
christianvuerings: @johncrossman although we do have some legacy code that still has this, I would recommend using something like `$scope.delegate.isLoading` instead.

src/assets/templates/widgets/profile/delegate.html
```diff
@@ -0,0 +1,22 @@
+<div class="cc-widget-text cc-page-widget-profile-delegate"
+  data-ng-controller="DelegateController"
+  data-cc-spinner-directive="isLoading"
+>
+  <h2><a data-ng-href={{delegatedAccessUrl.url}}>Manage Delegates</a></h2>
```
christianvuerings: @johncrossman for each campus solutions URL, we should add something like the following properties:

```
          data-cc-campus-solutions-link-directive="delegatedAccessUrl.url"
          data-cc-outbound-enabled="false"
```

feel free to search through the code for more examples.

christianvuerings: @johncrossman please write something like:

```
<h2>
  <a>...</a>
</h2>
```

src/assets/templates/widgets/profile/delegate.html
```diff
@@ -0,0 +1,22 @@
...
+
+  <p>Delegates are individuals who can act on a student's behalf by paying bills, viewing academic information, etc.
```
christianvuerings: @johncrossman 

```
<p>
  Delegates...
</p>
```

src/assets/templates/widgets/profile/delegate.html
```diff
@@ -0,0 +1,22 @@
...
+    They are often parents or guardians of students. Students choose which privileges to delegate, and can change
+    those privileges at any time.</p>
+
+  <p>To created a new delegate or change the privileges of an existing one, click the link above.</p>
+
+  <h3>Other information:</h3>
+  <ul>
+    <li><b>Free/busy:</b> To give a delegate a free/busy view of your bCal calendar, follow
```
christianvuerings: @johncrossman please use `strong` instead of `b`.

Also
```
<ul>
  <li>
    <strong>Free/busy:</strong> To give a delegate a free/busy view of your bCal calendar, follow
    the <a href="https://support.google.com/calendar/answer/37082?hl=en">bCal instructions</a>.
  </li>
...
```

src/assets/templates/widgets/profile/delegate.html
```diff
@@ -0,0 +1,22 @@
...
+      the <a data-ng-href="https://support.google.com/calendar/answer/37082?hl=en">bCal instructions</a>.</li>
```
christianvuerings: @johncrossman we probably don't need `ng-href` here, `href` should suffice.

app/models/campus_solutions/delegate_management_url.rb
```diff
@@ -0,0 +1,22 @@
...
+  class DelegateManagementURL < DirectProxy
+
+    def initialize(options = {})
+      super options
+      initialize_mocks if @fake
+    end
+
+    def response_root_xml_node
+      'DELEGATED_ACCESS_URL'
+    end
```
paulkerschen: @johncrossman Opaquely enough, I don't think `response_root_xml_node` is used in CS proxies that fire GET requests; it's only subclasses of `CampusSolutions::PostingProxy` that make use of it.

fixtures/xml/campus_solutions/delegate_management_url.xml
johncrossman: @christianvuerings Ampersand is not allowed. We must use CDATA or replace it with \&amp;.

christianvuerings: @johncrossman seems like we're using `&amp;` in https://github.com/ets-berkeley-edu/calcentral/blob/1dbd774c36fc79e7ffffdece5d8b7db42f7e77d0/fixtures/xml/campus_solutions/financial_aid_data.xml#L615

Would be good to keep it consistent

christianvuerings:
lgtm as soon as we fixed the `&amp;` comment

fixtures/xml/campus_solutions/delegate_management_url.xml
johncrossman: Thanks @christianvuerings  I'll amend

paulkerschen:
Stamp of approval!

PR #4545, opened 2015-11-19
SISRP-10872 Followup - use the merged MyStudent feed from UserAttributes

ctweney:
Followup for https://jira.berkeley.edu/browse/SISRP-10872 eliminates the last usage of /students/all.


johncrossman:
LVGTM

PR #4546, opened 2015-11-19
SISRP-11243 - Fluid return text to CalCentral

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-11243


johncrossman:
Looks about right...

PR #4547, opened 2015-11-19
CLC-5974, rename files to put 'delegate_' first; js best practices

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-5974
* With this renaming we get delegated-access files in expected order
* I've tried to apply some Angular best practices...
* Added data-cc-spinner-directive


paulkerschen:
My but it cleans up right nice, @johncrossman.

PR #4553, opened 2015-11-20
CLC-6031 Protect merged models against provider-specific errors

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6031 briefly explains why this turned into such a big PR. As it turned out, all six of our big `My\w+::Merged` models required protection similar to that intended for `MyAcademics::Merged`. Once that was set up, there seemed to be no reason not to standardize a bit further.

The new `MergedModel` module allows classes to wrap their merging logic inside an error handler, just as our proxies wrap their network calls. Errors are logged and the feed does its best to keep building, with the names of any misbehaving providers added to an `errors` array. At present nothing is done with the values in that array, but they might come in handy down the road for front-end messaging, targeted cache expiration, etc.

Bamboo at work: https://bamboo.ets.berkeley.edu/bamboo/browse/CPB-CPT223-1/


spec/models/my_badges/merged_spec.rb
```diff
@@ -0,0 +1,133 @@
...
+      allow_any_instance_of(MyBadges::GoogleCalendar).to receive(:verify_and_format_date).and_raise(ArgumentError, 'foo')
+      suppress_rails_logging do
+        filtered_feed = badges.get_feed
+        expect(filtered_feed[:badges]['bcal'][:count]).to eq 0
+        expect(filtered_feed[:badges]['bdrive'][:count]).to eq 0
+        expect(filtered_feed[:badges]['bmail'][:count]).to be > 0
+      end
+    end
+
+    it 'should handle utter failure from a particular source' do
```
paulkerschen: Note that adding the 'utter failure' example is the only substantive change in this particular spec. Rename plus cleaning out deprecated Rspec syntax caused Github to lose track.

app/models/my_tasks/merged.rb
```diff
@@ -17,8 +14,12 @@ def initialize(uid, options={})
       @now_time = Time.zone.now
     end
 
-    def init
```
raydavis: I believe the enabled_sources check was put in "init" so that we wouldn't take a performance hit on "Canvas::Proxy.access_granted?" for cached feeds.

raydavis: Ah, but now the providers check is _always_ deferred... well played, @paulkerschen

paulkerschen: @raydavis: That makes sense. I was trying to preserve that gain below by keeping the `@providers` instance variable around, but is there a gap that I failed to plug?

raydavis:
Needs more refactor.

PR #4557, opened 2015-11-23
CLC-6007 Mention term in blinkered My Academics Profile

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6007

Aside from single-quote and Rspec cleanup, the changes are:
- Ensure that MyAcademics::CollegeAndLevel always returns a `termName` value, falling back to the CalCentral current term if it can't extract a term from Bearfacts;
- Surface a more detailed message in the profile template, including term name, when a student is considered active but not registered.


raydavis:
Thank you for this clarification, @paulkerschen, I think it will help students out a lot!

PR #4558, opened 2015-11-23
CLC-6039, /api/my/status -> hasToolboxTab when superuser, can-view-as

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-6039

Nothing surfaces on the UI yet. This is purely putting hasToolboxTab in the /api/my/status feed when user is "super" or can-view-as. Eventually, this tab will have view-as features and other admin-ish type cards.

### DIFF
--- a/app/models/user/api.rb
+++ b/app/models/user/api.rb
@@ -146,6 +146,7 @@ module User
           has_instructor_history || has_student_history
         ),
         :hasFinancialsTab => (roles[:student] || roles[:exStudent]),
+        :hasToolboxTab => current_user_policy.can_administrate? || current_user_policy.can_view_as?,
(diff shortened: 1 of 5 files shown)


johncrossman:
**Note:** I did an Rspec 3 overhaul on ```api_spec``` and then I found its copy-and-paste sibling, ```api_with_campus_solutions_enabled_spec.rb```. I duplicated my _hasToolboxTab_ specs in both. I suppose _api_spec.rb_ will eventually go away.

paulkerschen:
Which spec is the evil twin, is the question.

ctweney:
api_with_campus_solutions_enabled_spec will become api_spec at the point of no return when we abandon the cs_profile feature flag -- I guess at GoLive5? Some point after?

PR #4559, opened 2015-11-23
SISRP-7644 Add loan payment estimator link to My Campus

ctweney:
https://jira.berkeley.edu/browse/SISRP-7644


christianvuerings:
Lgtm but wanted @paulkerschen @johncrossman or @raydavis to have a look as well. We would want this as part of GL4.

johncrossman:
This looks right to me. If it were an immediate need we could just modify in prod per https://confluence.ets.berkeley.edu/confluence/x/RIoUAg

PR #4561, opened 2015-11-24
SISRP-11283 - GL3 Remove hardcoded SIR response messages

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-11283


paulkerschen:
@christianvuerings With this change, it looks like the contents of the two div.cc-widget-text elements are identical. Would it be better to combine the data-ng-if conditions for these two; or is that condition perhaps no longer needed?

christianvuerings:
@paulkerschen thank you! Really good point, I'll make it update.

christianvuerings:
@paulkerschen pushed the update

paulkerschen:
LGTM @christianvuerings, waiting for T.

PR #4568, opened 2015-11-24
CLC-5840 Distribute LTI Usage / TurnItIn reports through Google Drive

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-5840

The first commit untangles OEC-specific classes from the more generic GoogleApps module, allows SPA accounts (for example) to find shared folders, and adds a convenient upload-to-a-worksheet method.

The second commit uploads LTI and TurnItIn usage reports to a folder shared between PDG and TLS.

Bamboo testext under way, which may call for amendments.


app/models/google_apps/sheets_manager.rb
```diff
@@ -97,30 +115,11 @@ def update_worksheet(worksheet, updates)
...
-            csv << csv_row
-          else
-            csv << row.values_at(*headers)
-          end
-        end
-      end
-      upload_to_spreadsheet(title, description, StringIO.new(content), parent_id)
-    end
-
-    def upload_to_spreadsheet(title, description, path_or_io, parent_id)
```
raydavis: It looks as if Canvas & OEC clients both rely a lot more on worksheet titles than on Google file "descriptions". I have a hard time even finding them in the current Drive UX.

app/models/google_apps/sheets_manager.rb
```diff
@@ -97,30 +115,11 @@ def update_worksheet(worksheet, updates)
       result.body
     end
 
-    def upload_worksheet(title, description, worksheet, parent_id = 'root', opts={})
-      content = CSV.generate do |csv|
-        headers = worksheet.headers
-        csv << headers
-        worksheet.each_sorted do |row|
```
raydavis: This method is specific to Oec::Worksheet, which is a completely different animal from the GoogleDrive::Worksheet used elsewhere in this same class.

app/models/google_apps/sheets_manager.rb
paulkerschen: Quite so, @raydavis; to my recollection, all we were ever did with the file description was dutifully pass an empty string.

raydavis:
Oh yeah, _definitely_ some test amendments coming up. Unfortunately I also have a meeting coming up & an emergency at home, so this PR will have to sit on ice until tomorrow.

PR #4570, opened 2015-11-25
CLC-5840 Distribute LTI Usage / TurnItIn reports through Google Drive

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-5840

The first commit untangles OEC-specific classes from the more generic GoogleApps module, allows SPA accounts (for example) to find shared folders, and adds a convenient upload-to-a-worksheet method.

The second commit uploads LTI and TurnItIn usage reports to a folder shared between PDG and TLS.

The third fixes my new textext test and removes some confusingly ignored arguments from existing ones.

### DIFF
--- a/app/models/google_apps/sheets_manager.rb
+++ b/app/models/google_apps/sheets_manager.rb
@@ -1,6 +1,24 @@
 module GoogleApps
   class SheetsManager < DriveManager
 
+    # Wrapper to avoid extra copies of large CSV tables.
+    class CsvAsUpdates
+      def initialize(csv)
+        @csv = csv
+      end
+      def each(&block)
+        row = 1
+        @csv.each do |row_values|
+          col = 1
+          row_values.each do |value|
+            yield [row, col], value
+            col += 1
+          end
+          row += 1
+        end
+      end
+    end
+
     def initialize(uid, options = {})
       super uid, options
       auth = get_google_api.authorization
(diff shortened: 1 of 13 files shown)


app/models/oec/worksheet.rb
```diff
@@ -113,6 +113,24 @@ def sorted_rows
       @rows.values
     end
 
+    def to_csv_string
+      CSV.generate do |csv|
+        csv << headers
+        each_sorted do |row|
+          csv_row = headers.map do |header|
+            # A trick to force plaintext formatting in Google Sheets.
+            # TODO Why not use the CSV :force_quotes option?
```
paulkerschen: Good question, @raydavis; perhaps, lack of initiative!

app/models/google_apps/sheets_manager.rb
```diff
@@ -1,6 +1,24 @@
...
+      end
+      def each(&block)
+        row = 1
+        @csv.each do |row_values|
+          col = 1
+          row_values.each do |value|
+            yield [row, col], value
+            col += 1
+          end
+          row += 1
```
paulkerschen: FWIW, I find `each_with_index` a helpful shorthand for this stuff, and the differently punctuated `each.with_index` perhaps even more helpful, since it will accept a starting value other than zero.

paulkerschen:
This looks really good to me, @raydavis; thanks for the untangling work and removal of unused arguments.

PR #4571, opened 2015-11-25
CLC-6024 Present filtered My Academics to delegate viewers

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6024

Road map:
- In general, stuff to be hidden from delegates is so deeply intertwined in the My Academics feed that it seemed best to construct an entirely new feed rather than diving deep into the existing structure and unmerging all the merged components. Hence the new MyAcademics::FilteredForDelegate.
- Fortunately, the two flavors of MyAcademics::FilteredForDelegate (with and without grades) are similar enough that additional grade filtering can use the same filter-for-view-as logic that already exists in My Badges and elsewhere.
- MyAcademics::Merged and MyAcademics::FilteredForDelegate are cached separately. Feed components that supply both without alteration (CollegeAndLevel, GpaUnits, and TransitionTerm) are now individually cached as well, though this is still a far cry from the fine-grained separation we'd like to set up in the My Academics feed.
- All decisions depend on a couple of to-be-implemented methods in AuthenticationState. My assumption is that this class will be used to return delegate status and permissions from session state, as we currently do with View-As. This is the major bit of wiring that remains outstanding for the delegated access feature.

### DIFF
--- a/app/controllers/my_academics_controller.rb
+++ b/app/controllers/my_academics_controller.rb
@@ -3,7 +3,11 @@ class MyAcademicsController < ApplicationController
   before_filter :api_authenticate
 
   def get_feed
-    render :json => MyAcademics::Merged.from_session(session).get_feed_as_json
+    if current_user.authenticated_as_delegate?
+      render json: MyAcademics::FilteredForDelegate.from_session(session).get_feed_as_json
+    else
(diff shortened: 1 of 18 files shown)


christianvuerings:
Front-end lgtm

raydavis:
Nicely refactored! Too bad Semesters is such a mess, but GIGO...

PR #4573, opened 2015-11-26
CLC-6052 - Fix some things on the toolbox page

christianvuerings:
https://jira.ets.berkeley.edu/jira/browse/CLC-6052

* Fix spacing on the settings page
* Fix spacing for the last column on the toolbox page
* Fix background color on the toolbox page (we shouldn't make it white as soon as we have a `cc-widget` on it.
* Bring back the error messages (like `That does not appear to be a valid UID or SID.`) for the UID / SID Lookup
* Add `label` to `UID/SID Lookup` for accessibility reasons
* Fix weird spinner behavior on Chrome when running the CalCentral API Test
* Fix alignment of check & spinner icons with their text next to it (flex-base)
* Allow for re-running the API tests
* Fix the main spinner for `CalCentral API Test` (only hide the contents on the widget, not the widget container)
* Refactor `CalCentral API Test` - and fix the dependency on the counter / index.
* Use data-ng-attr for dynamic items

### DIFF
--- a/src/assets/javascripts/angular/controllers/widgets/apiTestController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/apiTestController.js
@@ -1,11 +1,12 @@
 'use strict';
 
 var angular = require('angular');
+var _ = require('lodash');
 
 /**
  * API Test controller
  */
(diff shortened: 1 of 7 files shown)


johncrossman:
Thanks @christianvuerings  I appreciate the cleanup ;)

PR #4577, opened 2015-12-01
SISRP-10417 - UX Design: Add message to SIR experience noting that payments will take time to process and may not be reflected on your account immediately

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-10417


christianvuerings:
FYI people are expecting this to still be part of GL3.

PR #4579, opened 2015-12-01
CLC-6065, avoid HubEdos::Proxy error for non-student by handling 404 from sis-integration/.../affiliations call

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-6065

404 status is not an error condition. It's what sis-integration returns to us when user is non-student. Response body:
```
{"StudentResponse":{"source":"UCB-SIS-STUDENT","message":"Student Not Found","students":[""]}} 
```


raydavis:
Thanks for the fix, @johncrossman !

johncrossman:
@paulkerschen The follow-up: https://jira.ets.berkeley.edu/jira/browse/CLC-6066

PR #4580, opened 2015-12-02
CLC-6064 Apply dynamic logging changes to actual logging outputters

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6064

I believe we were outsmarted by Log4r's fancy multi-outputter setup. This change sets things right for me locally.


raydavis:
Yes, that was my first guess too, having had to trudge through those before. Thank you for the fast fix!

PR #4581, opened 2015-12-02
SISRP-10824 Use Crosswalk for view-as admin lookup

ctweney:
https://jira.berkeley.edu/browse/SISRP-10824

This is a partial fix. It lets you view-as a new user who's not in Oracle. But there's a minor bug where such users won't record in the view-as admin history (SISRP-11917).

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -273,9 +273,9 @@ cache:
     CampusSolutions::Translate: <%= 24.hours %>
     CampusSolutions::DashboardUrl: <%= 29.days %>
 
-    CalnetCrosswalk::Proxy: <%= 29.days %>
-    CalnetCrosswalk::ByUid: <%= 29.days %>
-    CalnetCrosswalk::BySid: <%= 29.days %>
+    CalnetCrosswalk::Proxy: <%= 24.hours %>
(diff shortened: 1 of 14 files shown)


config/settings.yml
```diff
@@ -274,6 +274,8 @@ cache:
     CampusSolutions::DashboardUrl: <%= 29.days %>
 
     CalnetCrosswalk::Proxy: <%= 29.days %>
+    CalnetCrosswalk::ByUid: <%= 29.days %>
```
christianvuerings: Although this is fine with me, I do wonder about the case when an employee becomes a student. They'll already have a UID but not SID yet.

ctweney: Hm, maybe 1-day is more appropriate. What do you think @raydavis ?

raydavis: 1-day sounds good to me regardless, given the schedule uncertainties about which upstream systems will do what when.

raydavis:
Thanks, @ctweney ! The code looks good to me. I expect we'll be generalizing our fallbacks soon, but seems wise not to overanticipate.

ctweney:
I pushed a new commit shortening the crosswalk's cache time to 24 hours.

PR #4582, opened 2015-12-02
SISRP-11903 Tighten filters on Hub merge into User::Api

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-11903

While this doesn't take the place of error handling farther down the stack, it should provide some more immediate rigor as far as what's allowed into the /my/status feed.

### DIFF
--- a/app/models/user/api.rb
+++ b/app/models/user/api.rb
@@ -15,32 +15,44 @@ module User
       if is_cs_profile_feature_enabled
         @edo_attributes ||= HubEdos::UserAttributes.new(user_id: @uid).get
       end
-      @default_name ||= get_campus_attribute('person_name')
+      @default_name ||= get_campus_attribute('person_name', :string)
       @first_login_at ||= @calcentral_user_data ? @calcentral_user_data.first_login_at : nil
-      @first_name ||= get_campus_attribute('first_name') || ''
-      @last_name ||= get_campus_attribute('last_name') || ''
+      @first_name ||= get_campus_attribute('first_name', :string) || ''
+      @last_name ||= get_campus_attribute('last_name', :string) || ''
       @override_name ||= @calcentral_user_data ? @calcentral_user_data.preferred_name : nil
-      @student_id = get_campus_attribute('student_id')
+      @student_id = get_campus_attribute('student_id', :numeric_string)
     end
 
     # split brain until SIS GoLive5 makes registration data available
-    def get_campus_attribute(field)
-      value = nil
-      if is_sis_profile_visible? && @edo_attributes[:noStudentId].blank?
-        value = @edo_attributes[field.to_sym]
+    def get_campus_attribute(field, format)
+      if is_sis_profile_visible? && @edo_attributes[:noStudentId].blank? && (edo_attribute = @edo_attributes[field.to_sym])
+        begin
+          validated_edo_attribute = validate_attribute(edo_attribute, format)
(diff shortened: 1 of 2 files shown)


app/models/user/api.rb
```diff
@@ -15,32 +15,39 @@ def init
...
-        value = @oracle_attributes[field]
+      validate_attribute(edo_attribute, format) || @oracle_attributes[field]
+    end
+
+    def validate_attribute(value, format)
+      case format
+        when :string
+          value if value.is_a?(String) && value.present?
+        when :numeric
+          value if value.respond_to?(:to_i) && value.to_i > 0
```
raydavis: Wouldn't the usual validation be something like:

```
begin
  value = Integer(value, 10)
rescue ArgumentError
  nil
end
```

app/models/user/api.rb
```diff
@@ -15,32 +15,39 @@ def init
...
       end
-      if value.nil?
-        value = @oracle_attributes[field]
+      validate_attribute(edo_attribute, format) || @oracle_attributes[field]
+    end
+
+    def validate_attribute(value, format)
+      case format
+        when :string
+          value if value.is_a?(String) && value.present?
```
johncrossman: Is there value in logging when value type does not match expectation?

app/models/user/api.rb
```diff
@@ -15,32 +15,39 @@ def init
...
+        when :numeric
+          value if value.respond_to?(:to_i) && value.to_i > 0
       end
-      value
     end
 
-    # special-case brain split for roles
+    # Conservative merge of roles from EDO
+    WHITELISTED_EDO_ROLES = [:student]
```
raydavis: Good idea!

paulkerschen: +1 @johncrossman @raydavis; I'll amend.

app/models/user/api.rb
raydavis: ... if not is_a? Numeric, that is...

spec/models/user/api_with_campus_solutions_enabled_spec.rb
```diff
@@ -140,45 +140,122 @@
...
     context 'a non-student' do
-      let(:test_roles) {student_roles[:non]}
+      let(:test_roles) {:non}
       it {should be_falsey}
     end
     context 'an ex-student' do
-      let(:test_roles) {student_roles[:expired]}
-      it {should be_truthy}
+      let(:test_roles) {:expired}
+      it {pp User::Api.new(@random_id).get_feed; should be_truthy}
```
raydavis: Looks like some stray development-only code creeped in...

raydavis:
Looks really good, @paulkerschen , OK by me to merge as soon as you're done fiddling. I especially appreciate the WHITELISTED_EDO_ROLES addition.

app/models/user/api.rb
paulkerschen: Trick is that we're checking for a numeric string, not a number; and assuming validation passes, we want to keep the original string rather than a converted value. Ruby has 15 ways to do it, of course:

http://stackoverflow.com/questions/5661466/test-if-string-is-a-number-in-ruby-on-rails

I'll accept that the integer cast with exception handling is a conventional way to do it, though.

paulkerschen:
Fiddled for clarity and logging, @raydavis @johncrossman.

PR #4586, opened 2015-12-03
CLC-6066, less brittle HubEdos module via safe response handling

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-6066
https://jira.berkeley.edu/browse/SISRP-11933

This brings a new home for reusable code and, hopefully, a bit more resiliency to the HubEdos module.

### DIFF
--- a/app/models/hub_edos/response_handler.rb
+++ b/app/models/hub_edos/response_handler.rb
@@ -1,0 +1,13 @@
+module HubEdos
+  module ResponseHandler
+
+    def get_students(response)
+      return [] unless response.is_a? Hash
+      container = response['studentResponse']
+      # Verify that target element is capable of being iterated. And yes, the structure really is this weird.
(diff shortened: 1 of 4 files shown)


app/models/hub_edos/response_handler.rb
```diff
@@ -0,0 +1,12 @@
+module HubEdos
+  module ResponseHandler
+
+    def get_students(response)
+      container = response['studentResponse']
```
paulkerschen: @johncrossman Do we also need a nil check on `response` itself? It's a little hard to tell from the logs, but I think a nil parsed response caused the blowup in https://jira.berkeley.edu/browse/SISRP-11933.

johncrossman: One can never be too safe, @paulkerschen   Comin' right up!

paulkerschen:
One comment above; a welcome improvement overall.

paulkerschen:
Happy T. Thanks @johncrossman!

PR #4587, opened 2015-12-03
SISRP-11943 - Profile backend breakages are not indicated in UX

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-11943


paulkerschen:
True to pattern.

PR #4588, opened 2015-12-03
CLC-6061, redirect to dashboard after view-as + reauth

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-6061

We must redirect because you are now viewing-as a user that might not have access to /toolbox page.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -31,9 +31,11 @@ class ApplicationController < ActionController::Base
     end
   end
 
-  def reauthenticate
+  def reauthenticate(opts = {})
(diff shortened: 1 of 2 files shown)


raydavis:
This is a model of succinct diagnosis & cure, @johncrossman -- thank you! But it looks like a test needs to be updated, & for easy reference it would be great to get the diagnosis into the JIRA thread.

johncrossman:
Thank you, @raydavis  I'll make sure Jira is always the first to know w.r.t diagnosis. The specs were tweaked; green is the T.

raydavis:
"Merge pull request", that greeny button...

PR #4590, opened 2015-12-03
SISRP-11950 Improve logging of HubEdos / Crosswalk errors

raydavis:
https://jira.berkeley.edu/browse/SISRP-11950

PR #4592, opened 2015-12-03
SISRP-11950 Fix exception when lookup of CS ID fails

raydavis:
Eagle-eyed @paulkerschen spotted this nasty error path in my previous https://github.com/ets-berkeley-edu/calcentral/pull/4590


paulkerschen:
The many meanings of 'empty'. Looks good!

PR #4593, opened 2015-12-03
CLC-6075 Point to new Support page for GL3

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6075

("http" is used to avoid browser warnings in certain combinations of test environments between Java and Drupal).


christianvuerings:
lgtm

paulkerschen:
I choose to agree with the happy half of Travis.

PR #4596, opened 2015-12-04
CLC-6072, 'gulp build' expects port 3001 but underlying code would deduce 3000

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-6072

See latest comment in Jira. The change in `sessions_controller.rb` fixes [same problem when logging out on localhost:3001](https://jira.ets.berkeley.edu/jira/browse/CLC-6072?focusedCommentId=232320&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-232320)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -34,7 +34,7 @@ class ApplicationController < ActionController::Base
   def reauthenticate(opts = {})
     delete_reauth_cookie
     url = '/auth/cas?renew=true'
-    url << "&url=#{opts[:redirect_path]}" if opts[:redirect_path]
+    url << "&url=#{url_for_path opts[:redirect_path]}" if opts[:redirect_path]
     redirect_to url_for_path url
   end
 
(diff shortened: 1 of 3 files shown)


app/controllers/application_controller.rb
```diff
@@ -151,7 +160,8 @@ def access_log
   # first to "http:" and then to "https:". This method makes relative paths safer to use.
   def url_for_path(path)
     if (protocol = default_url_options[:protocol])
-      protocol + request.host_with_port + path
+      url = protocol + request.host_with_port + path
+      ApplicationController.correct_port(url, env)
```
paulkerschen: @johncrossman: While the partial-string-sub does seem very conservative, it feels a little off a way to change a port number in a URL. Would it make sense to call `request.host` and `request.port_string` instead of `request.host_with_port`, change the port string from ':3000' to ':3001' if conditions obtain, and then concatenate them ourselves?

paulkerschen: My bad, @johncrossman: I was looking at the wrong library. Assuming this request is a `Rack::Request`, then `host_with_port` is the right method. But it still might be safer to do the string sub on the returned host string rather than the whole URL?

johncrossman: Thanks, @paulkerschen. As you suggest, I can't split up the `host_with_port` logic as found in  `Rack::Request`. Nonetheless, I'll make the `correct_port` usage more targeted.

johncrossman:
@paulkerschen Submitted for your approval...
image

paulkerschen:
It's the vindictive crochet that really makes it. Thanks @johncrossman!

christianvuerings:
@johncrossman thanks!

PR #4598, opened 2015-12-04
SISRP-12043 - Undergrad applicants have a SIR card with only a 'Yes' option - this generates a JS error

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-12043


src/assets/javascripts/angular/controllers/widgets/sir/sirItemController.js
```diff
@@ -111,7 +111,7 @@ angular.module('calcentral.controllers').controller('SirItemController', functio
    */
   var selectFirstResponseReason = function() {
     $scope.$watch('item.responseReasons', function(value) {
-      if (!value) {
```
christianvuerings: FYI `value` was an empty array when we didn't have any reasons. So we also need to check for the `length`

PR #4606, opened 2015-12-07
CLC-6078, qa branch only: WorkExperience is not GL3 so is_feature_enabled=false

johncrossman:
QA BRANCH ONLY. There is no corresponding PR on 'master' branch. 
See analysis in https://jira.ets.berkeley.edu/jira/browse/CLC-6078

Tests passed on localhost... Let's see what Travis thinks.


raydavis:
The squeaky wheel gets the merge, I guess...

PR #4607, opened 2015-12-08
CLC-6062: a few more broken Selenium tests fixed

pfarestveit:
https://jira.ets.berkeley.edu/jira/browse/CLC-6062


johncrossman:
Looks good. Waiting on T

PR #4608, opened 2015-12-08
CLC-6053, spec for has_toolbox_tab in AuthenticationStatePolicy

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-6053


paulkerschen:
@johncrossman Full of sense.

PR #4612, opened 2015-12-08
SISRP-10909 - In Calcentral no Notifications are generated

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-10909


christianvuerings:
*Note* This is a GL3 fix so definitely has priority - QA PR is in the making

PR #4613, opened 2015-12-08
SISRP-10909 QA - In Calcentral no Notifications are generated

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-10909

*Note* - we need to confirm that #4612 works on calcentral-dev first.


christianvuerings:
@johncrossman & @paulkerschen please review - Janhavi confirmed that this is now working in calcentral-dev.

PR #4619, opened 2015-12-09
SISRP-12145 Do not log Crosswalk ID lookup failures as errors

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-12145

This PR doesn't aim to change how CalnetCrosswalk::Proxy talks to other classes; all the public lookup methods still return nil on a 404 response. However, since 404s are expected in the normal course of business, they shouldn't be logged as errors. Two changes are made:
- Since subclasses of CalnetCrosswalk::Proxy may search by either UID or SID, the misleading hard-coded "UID" is removed from log messages and a class method substituted in.
- In the event of a 404 response, a `:notFound` key is added to the response before caching. Like the `:errored` key for other failures, this will cause public methods to return nil, but it won't trigger error logging.

### DIFF
--- a/app/models/calnet_crosswalk/proxy.rb
+++ b/app/models/calnet_crosswalk/proxy.rb
@@ -7,6 +7,10 @@ module CalnetCrosswalk
     APP_ID = 'calnetcrosswalk'
     APP_NAME = 'Calnet Crosswalk'
 
+    def self.id_type
+      self.name.demodulize.sub(/\ABy/,'').upcase
+    end
+
     def initialize(options = {})
       super(Settings.calnet_crosswalk_proxy, options)
       initialize_mocks if @fake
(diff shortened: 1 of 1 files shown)


app/models/calnet_crosswalk/proxy.rb
```diff
@@ -39,10 +43,10 @@ def get
       internal_response = self.class.smart_fetch_from_cache(id: instance_key) do
         get_internal
       end
-      if internal_response[:noStudentId] || internal_response[:statusCode] < 400
```
johncrossman: Was `noStudentId` a copy-and-paste carryover from another proxy class?

johncrossman:
Looks good. Thanks, @paulkerschen

app/models/calnet_crosswalk/proxy.rb
paulkerschen: @johncrossman Yes, I think it must have been; since ain't no code upstream sending that key.

PR #4620, opened 2015-12-09
[QA] SISRP-12145 Do not log Crosswalk ID lookup failures as errors

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-12145

QA version of #4619.


johncrossman:
As before, so qa.

PR #4623, opened 2015-12-10
SISRP-12147 - SIR card is beneath Tasks and Notifications in mobile view - expected?

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-12147

*Note* GL3 change

PR #4624, opened 2015-12-10
QA - SISRP-12147 - SIR card is beneath Tasks and Notifications in mobile view - expected?

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-12147


christianvuerings:
*Note* Only merge when #4623 is verified by @pfarestveit

PR #4634, opened 2015-12-10
SISRP-12219 - Show 'your information is not available' when we don't get any iHub data back at all.

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-12219

### DIFF
--- a/app/models/hub_edos/my_student.rb
+++ b/app/models/hub_edos/my_student.rb
@@ -26,6 +26,14 @@ module HubEdos
           merged[:feed][:student].merge!(hub_response[:feed]['student'])
         end
       end
+
+      # When we don't have any identifiers for this student, we should send a 404 to the front-end
+      if !merged[:errored] && !merged[:feed][:student]['identifiers']
+        merged[:statusCode] = 404
+        merged[:errored] = true
+        logger.warn("No identifiers found for student feed uid #{@uid} with feed #{merged}")
+      end
+
       merged
     end
 


christianvuerings:
FYI - This is a GL3 item

christianvuerings:
QA PR at #4635

app/models/hub_edos/my_student.rb
```diff
@@ -26,6 +26,14 @@ def get_feed_internal
           merged[:feed][:student].merge!(hub_response[:feed]['student'])
         end
       end
+
+      # When we don't have any identifiers for this student, we should send a 404 to the front-end
+      if !merged[:errored] && !merged[:feed][:student][:identifiers]
+        merged[:statusCode] = 404
+        merged[:errored] = true
```
johncrossman: In the `HubEdos::Proxy` class, 404 is synonymous with `studentNotFound: true`. Also, this `MyStudent` class associates `merged[:errored] = true` with code 500. Should we stay consistent with existing patterns? I.e., one of the following, in this new code:
1. statusCode = 500, or
2. replace `merged[:errored] = true` with `studentNotFound: true`?

johncrossman:
Looks like `hub_edo_controller_spec` failed.

app/models/hub_edos/my_student.rb
```diff
@@ -26,6 +26,14 @@ def get_feed_internal
...
+        logger.warn("No identifiers found for student feed uid #{@uid} with feed #{merged}")
```
johncrossman: Is there a risk that we'll log sensitive data here?  I.e., contents of `#{merged}`

johncrossman:
@christianvuerings, I'll have a look at the spec failure.

app/models/hub_edos/my_student.rb
```diff
@@ -26,6 +26,14 @@ def get_feed_internal
           merged[:feed][:student].merge!(hub_response[:feed]['student'])
         end
       end
+
+      # When we don't have any identifiers for this student, we should send a 404 to the front-end
+      if !merged[:errored] && !merged[:feed][:student][:identifiers]
```
johncrossman: Ah ha, the `identifiers` key is not a sym. So we must reference it as:
`if !merged[:errored] && !merged[:feed][:student]['identifiers']`

johncrossman: @christianvuerings, with the change above the spec will pass.

johncrossman:
@christianvuerings My thought on how to code this new logic:
```
# When we don't have any identifiers for this student, we should send a 404 to the front-end
if !merged[:errored] && !merged[:feed][:student]['identifiers']
  merged[:statusCode] = 404
  merged[:studentNotFound] = true
  logger.warn("No identifiers found for student feed uid #{@uid} with affiliations: #{merged[:feed][:student]['affiliations']}")
end
```
cc: @paulkerschen

app/models/hub_edos/my_student.rb
christianvuerings: @johncrossman follow up for that at https://jira.berkeley.edu/browse/SISRP-12240
In the front-end we basically would want 1 parameter to know whether to show the student information. Right now that's the `errored` piece, but we could definitely change that in the future.

christianvuerings:
@johncrossman implemented your recommended changes, however, didn't change the `logger.warn` message since that "student" probably won't have any CS affiliations either.

johncrossman:
Thanks, @christianvuerings  My concern w/ dumping `#{merged}` to the log is that we'll write *genderOfRecord* or some other sensitive data. I recommend a follow up PR to fix it. Nonetheless, I'll merge this in order to keep the ball rolling.  cc @raydavis @paulkerschen

PR #4635, opened 2015-12-10
QA - SISRP-12219 - Show 'your information is not available' when we don't get any iHub data back at all.

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-12219

### DIFF
--- a/app/models/hub_edos/my_student.rb
+++ b/app/models/hub_edos/my_student.rb
@@ -26,6 +26,14 @@ module HubEdos
           merged[:feed][:student].merge!(hub_response[:feed]['student'])
         end
       end
+
+      # When we don't have any identifiers for this student, we should send a 404 to the front-end
+      if !merged[:errored] && !merged[:feed][:student]['identifiers']
+        merged[:statusCode] = 404
+        merged[:errored] = true
+        logger.warn("No identifiers found for student feed uid #{@uid} with feed #{merged}")
+      end
+
       merged
     end
 


christianvuerings:
*Note* Only merge this when #4634 is verified.

app/models/hub_edos/my_student.rb
```diff
@@ -26,6 +26,14 @@ def get_feed_internal
           merged[:feed][:student].merge!(hub_response[:feed]['student'])
         end
       end
+
+      # When we don't have any identifiers for this student, we should send a 404 to the front-end
+      if !merged[:errored] && !merged[:feed][:student]['identifiers']
+        merged[:statusCode] = 404
+        merged[:errored] = true
+        logger.warn("No identifiers found for student feed uid #{@uid} with feed #{merged}")
```
johncrossman: @christianvuerings Since this is our release candidate, can you remove `with feed #{merged}`. If we want more informative logging we can make adjustments on master branch.

johncrossman: After discussing in person, @christianvuerings convinced me that `#{merged}` will not include sensitive user data when `merged[:feed][:student]['identifiers']` is nil. And because this scenario is not well understood (i.e., what does upstream data look like when `!error` and empty `identifiers`?) the logging of `merged` might be a benefit.

johncrossman:
In cc-dev, @pfarestveit verified the important use case of SISRP-12219 so we proceed to qa.

PR #4636, opened 2015-12-10
CLC-6084 - more Selenium test fixes

pfarestveit:
https://jira.ets.berkeley.edu/jira/browse/CLC-6084

A few of the Selenium tests fail occasionally, mainly because DOM updates happen in a way the tests don't expect.  This is the first of a set of changes to make the tests more robust.


johncrossman:
Looks good 2 me.

PR #4638, opened 2015-12-11
CLC-6095 Allow fake Google Drive proxy to respond to v2 or v3 API calls

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6095

For how the fix works, see https://github.com/ets-berkeley-edu/calcentral/blob/master/lib/proxies/mock_http_interaction.rb#L34:
- When defining a mock request, we usually pass in a `uri_matching` key with a partial URI, which Proxies::Mockable then turns into a nice URI-matching regexp.
- However, if a class needs the flexibility of building its own regexp, it can be passed in with just `uri` as the key.

### DIFF
--- a/app/models/google_apps/drive_list.rb
+++ b/app/models/google_apps/drive_list.rb
@@ -12,7 +12,8 @@ module GoogleApps
     end
 
     def mock_request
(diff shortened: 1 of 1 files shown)


johncrossman:
Gold star, @paulkerschen!

PR #4643, opened 2015-12-13
SISRP-7698 - Front-End Dev: Service Indicators "Holds": Placement and design(s)

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-7698


app/models/campus_solutions/holds.rb
```diff
@@ -0,0 +1,33 @@
...
+    def xml_filename
+      'holds.xml'
+    end
+
+    def build_feed(response)
+      return {} if response.parsed_response.blank?
+      feed = response.parsed_response
+      feed['SERVICE_INDICATORS'].each do |indicator|
+        # Convert to a CalCentral date
+        indicator['START_DATE'] = format_date(strptime_in_time_zone(indicator['START_DATE'], '%Y-%m-%d'))
```
johncrossman: Can we rely on START_DATE to always be present?

app/models/campus_solutions/holds.rb
```diff
@@ -0,0 +1,33 @@
...
+      end
+      feed
+    end
+
+    def url
+      "#{@settings.base_url}/UC_CC_SERVC_IND.v1/Servc_ind/Get?/EMPLID=#{@campus_solutions_id}"
```
johncrossman: Do we want `Get?EMPLID=` instead of `Get?/EMPLID=`

app/models/campus_solutions/proxy.rb
```diff
@@ -82,7 +82,7 @@ def url
     end
 
     def is_errored?(feed)
-      feed[:errmsgtext].present?
+      feed.kind_of?(Hash) && feed[:errmsgtext].present?
```
johncrossman: I'm curious why this is necessary. If we're not getting a hash then we might be masking a bug(?)

src/assets/javascripts/angular/controllers/widgets/academicsStatusAndBlocksController.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+var angular = require('angular');
+
+/**
+ * Academics status, holds & blocks controller
+ */
+angular.module('calcentral.controllers').controller('AcademicsStatusHoldsBlocksController', function($scope) {
```
johncrossman: I would say we want `AcademicsStatusAndBlocksController` instead of `AcademicsStatusHoldsBlocksController`?  But we don't want the word "Blocks" at all, right?

app/models/campus_solutions/proxy.rb
christianvuerings: @johncrossman because in this case it was an array, and when we have `errmsgtext` it should always be a hash. Do you have a different / better way to fix this?

app/models/campus_solutions/holds.rb
christianvuerings: @johncrossman that's the URL they provided us, so that's what we should be using for now.

app/models/campus_solutions/holds.rb
christianvuerings: @johncrossman pushed up a fix

src/assets/javascripts/angular/controllers/widgets/academicsStatusAndBlocksController.js
christianvuerings: @johncrossman we actually do want status, holds and blocks in there, since it does have all of those items. Later on (when Bearfacts goes away), we will only have status & holds. I'll also update the name.

app/models/campus_solutions/proxy.rb
johncrossman: That makes sense although it's not clear why we didn't hit this problem with check_list_items XML which is also type='array'. I'll have a closer look for my own education

johncrossman: I ran `checklist_spec` with breakpoint on this line. The feed is a Hash with array under `feed[:checkListItems]`. So why don't we get `feed[:serviceIndicators]` from `holds.xml`?  My main concern: If Campus Solutions were to give us `<ERRMSGTEXT>` as child element of `< SERVICE_INDICATORS>` would we miss it?

cc: @paulkerschen @raydavis for second opinion

raydavis: I'm still trying to understand the specifics on this one, but one general warning vis-a-vis Ruby classes in parsed XML: be sure to test _both_ the case where there are multiple items in an array _and_ the case where there is only one item. One item "lists" will generally show up as a Hash rather than as a single-item Array.

raydavis: @christianvuerings , you say "in this case it was an array" -- can you provide a sample "fixtures" XML file or dummy data in a spec to illustrate? Working examples are usually the best explanation.

raydavis: Cancel my first worry. I just ran a quick test against dummy data & was glad to see that the XML's ``type="array"`` attribute did in fact enforce single-item Array results.

raydavis: It would also be helpful to have at least one example somewhere of the explicit error response. Does the XML looks something like

```
<?xml version="1.0"?>
<ERRMSGTEXT>Terrible news!</ERRMSGTEXT>
```

raydavis: NM, I see that @johncrossman just added a JIRA for such a test: CLC-6099

christianvuerings: @johncrossman & @raydavis just reverted this change - and everything does still seem to work, you were both right.

christianvuerings:
@raydavis @paulkerschen @johncrossman made the changes - would be great if someone could have a look

johncrossman:
Thanks @christianvuerings Looks good.

PR #4644, opened 2015-12-15
Merge master into sis-master

christianvuerings:


christianvuerings:
We need this in order to have a green bamboo - @paulkerschen fixed the google tests on master.
https://bamboo.ets.berkeley.edu/bamboo/browse/MYB-CSTW-313

PR #4645, opened 2015-12-15
CLC-6101 Explicitly version all Google API calls

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6101

Bamboo https://bamboo.ets.berkeley.edu/bamboo/browse/CPB-CPT230-2 is happy on the Google front.


johncrossman:
:+1:

PR #4648, opened 2015-12-15
CLC-6087 Log Google worksheet update errors more thoroughly

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6087

### DIFF
--- a/app/models/google_apps/sheets_manager.rb
+++ b/app/models/google_apps/sheets_manager.rb
@@ -100,19 +100,26 @@ module GoogleApps
             </feed>
       EOS
 
-      result = @session.execute!(
-        http_method: :post,
-        uri: "#{cells_feed_url}/batch",
-        body: xml,
-        headers: {
-          'Content-Type' => 'application/atom+xml;charset=utf-8',
-          'If-Match' => '*'
-        }
-      )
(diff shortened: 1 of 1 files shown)


app/models/google_apps/sheets_manager.rb
```diff
@@ -74,45 +74,51 @@ def update_worksheet(worksheet, updates)
...
 
       cells_feed_url = CGI.escapeHTML worksheet.cells_feed_url.to_s
-      cells_feed_url_base = cells_feed_url.gsub(/\/private\/full\Z/, '')
-      xml = <<-EOS
-            <feed xmlns="http://www.w3.org/2005/Atom"
-                  xmlns:batch="http://schemas.google.com/gdata/batch"
-                  xmlns:gs="http://schemas.google.com/spreadsheets/2006">
-              <id>#{cells_feed_url}</id>
-      EOS
+      begin
```
johncrossman: *Nit-pick:* Why not have the `begin` just before the `result = @session.execute!` statement?

paulkerschen: @johncrossman Huh. Why not indeed?

PR #4650, opened 2015-12-15
CLC-6106 Feature-flag GL4-specific merged models against cache warming

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6106

### DIFF
--- a/app/models/my_academics/filtered_for_delegate.rb
+++ b/app/models/my_academics/filtered_for_delegate.rb
@@ -4,6 +4,7 @@ module MyAcademics
     include Cache::LiveUpdatesEnabled
     include Cache::FreshenOnWarm
     include Cache::JsonAddedCacher
+    include CampusSolutions::DelegatedAccessFeatureFlagged
     include MergedModel
 
     def self.providers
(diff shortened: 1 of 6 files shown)


app/models/my_academics/filtered_for_delegate.rb
```diff
@@ -20,11 +21,13 @@ def get_feed_as_json(force_cache_write=false)
         super(force_cache_write)
       else
         feed = get_feed(force_cache_write)
-        filter_grades(feed).to_json
+        filter_grades(feed) if feed.present?
```
johncrossman: Why have `if feed.present?` It seems unnecessary.

johncrossman:
LGTM.  One question posted.

app/models/my_academics/filtered_for_delegate.rb
paulkerschen: In tests, the to-and-from serialized JSON conversion was causing the feed to come back nil when the feature flag was off, which made it blow up when looking for a `:semesters` key. But you're right, the nil check looks weird there and would be better in the `filter_grades` method. Amending!

PR #4651, opened 2015-12-16
CLC-6084: additional Selenium test fixes

pfarestveit:
https://jira.ets.berkeley.edu/jira/browse/CLC-6084

A second PR, mainly aimed at refactoring the WebDriver tests for Tasks so that they can interact with any task in any time bucket from any source and so that the tests themselves are more readable.  The refactoring is important for testing checklist items in future tests.  Also a couple minor fixes.


johncrossman:
It's as if the QA dept has its own Ruby dialect. I will *wait_for_travis_then_merge*

PR #4653, opened 2015-12-16
CLC-6097 Optionally download Canvas CSVs to the file system

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6097

Full parsing of large bCourses reports consumes large amounts of memory, CPU, and time.
When one-row-at-a-time parsing is viable, CSV.foreach('filename.csv') is much more efficient.
This PR should help us move in that direction.

### DIFF
--- a/app/models/canvas/report/base.rb
+++ b/app/models/canvas/report/base.rb
@@ -6,6 +6,7 @@ module Canvas
       def initialize(options = {})
         super options
         @account_id = options.delete(:account_id) || settings.account_id
+        @download_to_file_name = options.delete(:download_to_file)
         @options = options
       end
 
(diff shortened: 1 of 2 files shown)


app/models/canvas/report/base.rb
```diff
@@ -65,7 +72,12 @@ def get_account_csv(report_type, object_type, term_id)
             logger.error "Unable to download report #{report_id} : #{csv_response}"
             return nil
           end
-          csv = CSV.parse(csv_response[:body], {headers: true})
+          if @download_to_file_name.blank?
+            csv = CSV.parse(csv_response[:body], {headers: true})
+          else
+            File.open(@download_to_file_name, 'w') {|fp| fp.write(csv_response[:body])}
```
johncrossman: @raydavis my understanding is that `fp.puts` would append a newline and `fp.write` does not. Which do you desire?

raydavis: Ah, there's the weak link in the unit test! Thanks, @johncrossman , I'll give it a live test (and possible PR amendment) before asking for a merge.

raydavis:
{{fp.write}} is what we want in this case, @johncrossman . Thanks for checking.

johncrossman:
Thank you much, @raydavis

PR #4658, opened 2015-12-17
SISRP-12357 - Update Title4 text for people that have not previously authorized

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-12357

### DIFF
--- a/src/assets/templates/widgets/profile/title4.html
+++ b/src/assets/templates/widgets/profile/title4.html
@@ -29,9 +29,17 @@
           I have authorized the use of my Title IV awards to be applied to pay all outstanding charges. If you would like to revoke this authorization, please uncheck the box below.
         </p>
       </div>
-      <div data-ng-if="!title4.isApproved">
+      <div data-ng-if="title4.isApproved === false">
         <p>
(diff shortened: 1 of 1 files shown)


src/assets/templates/widgets/profile/title4.html
```diff
@@ -29,11 +29,19 @@
           I have authorized the use of my Title IV awards to be applied to pay all outstanding charges. If you would like to revoke this authorization, please uncheck the box below.
         </p>
       </div>
-      <div data-ng-if="!title4.isApproved">
+      <div data-ng-if="title4.isApproved === false">
```
johncrossman: @christianvuerings When I look at raw source I see the following ng-if in which `!title4.isApproved` would be true for both nil and false values. Do we want to give it the same consideration as line above?
```
<span data-ng-if="!title4.isApproved">
  I have not authorized the University to...
</span>
```

christianvuerings: @johncrossman in that case we actually want both the `null` and `false`.

johncrossman:
...and thus we merge.

PR #4662, opened 2015-12-18
CLC-6112 Only show status header when there is status messaging

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6112

This refactor of the status popover front end replaces the enormous compound Boolean with a couple of scope variables to clarify intent. In brief, status header and messaging should only show if:
- there are active alerts or warnings;
- or the user has non-null registration data;
- or the user has non-null financial data.

### DIFF
--- a/src/assets/javascripts/angular/controllers/widgets/statusController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/statusController.js
@@ -17,6 +17,9 @@ angular.module('calcentral.controllers').controller('StatusController', function
 
     $scope.studentInfo = data.studentInfo;
 
+    if (_.get(data, 'studentInfo.regStatus.code') !== null) {
+      $scope.hasRegistrationData = true;
+    }
     if (_.get(data, 'studentInfo.regStatus.needsAction') && apiService.user.profile.features.regstatus) {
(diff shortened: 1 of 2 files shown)


src/assets/javascripts/angular/controllers/widgets/statusController.js
```diff
@@ -44,6 +47,9 @@ angular.module('calcentral.controllers').controller('StatusController', function
     }
     $scope.totalPastDueAmount = data.summary.totalPastDueAmount;
     $scope.minimumAmountDue = data.summary.minimumAmountDue;
+    if ($scope.minimumAmountDue !== null) {
```
christianvuerings: @paulkerschen could we write this as:

```
$scope.hasBillingData = ($scope.minimumAmountDue !== null)
```

christianvuerings:
lgtm - only 1 small suggestion but a non blocker

src/assets/javascripts/angular/controllers/widgets/statusController.js
paulkerschen: @christianvuerings Sure, I'll amend.

PR #4663, opened 2015-12-18
CLC-6111 - Combined Profile/Status/Settings button is accessibility concern

christianvuerings:
https://jira.ets.berkeley.edu/jira/browse/CLC-6111


src/assets/templates/widgets/gear_popover.html
```diff
@@ -10,15 +10,16 @@
         <i class="fa fa-user"></i>
       </span>
       <span data-ng-bind="api.user.profile.first_name" class="cc-popover-gear-first-name" data-ng-class="{'cc-popover-gear-has-alerts': hasAlerts}"></span>
+      <span class="cc-visuallyhidden">.</span>
```
christianvuerings: ... still feels a bit odd to me but so far the best solution that we've found.

paulkerschen:
Looks. Good. To. Me.

PR #4664, opened 2015-12-18
CLC-5895 - Update activity_list_link with ng-bind

davidlimcheng:
https://jira.ets.berkeley.edu/jira/browse/CLC-5895


christianvuerings:
@davidlimcheng lgtm - make sure to change the status to resolved and set the correct sprint version (71)

PR #4667, opened 2015-12-18
CLC-6097 CLC-6119 bCourses email address deletion

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6097

- Work around change to Canvas API permissions.
- Add separate Rake task to delete emails for inactive accounts, since it can take 12+ hours.
- Reduce memory/CPU cost of user account maintenance.
- Fix CSV download-to-file-system UTF-8 bug.
- Fix unit test bug CLC-6119.

### DIFF
--- a/app/models/canvas/communication_channels.rb
+++ b/app/models/canvas/communication_channels.rb
@@ -8,17 +8,17 @@ module Canvas
     end
 
     def list
-      wrapped_get request_path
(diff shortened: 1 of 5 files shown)


paulkerschen:
Looks right to me!

raydavis:
D'oh! Just saw a bone-headed mistake on a piece of code that would have been dauntingly time-consuming to test. Amendment coming!

PR #4668, opened 2015-12-18
CLC-6109 - QA: fix Selenium tests post mega-merge for GL4

pfarestveit:
https://jira.ets.berkeley.edu/jira/browse/CLC-6109

What should be the final set of changes...


paulkerschen:
So long, Settings!

PR #4671, opened 2015-12-18
SISRP-12544 - Communications Not Surfacing CC-DEV yet Surfacing in CC-SIS-DEV

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-12544

### DIFF
--- a/app/models/my_activities/campus_solutions_messages.rb
+++ b/app/models/my_activities/campus_solutions_messages.rb
@@ -28,7 +28,7 @@ module MyActivities
               cs: {}
             }
 
-            if message[:adminFunction] && message[:adminFunction][:varSeqData] && (Finaid::Shared::ADMIN_FUNCTION.include? message[:adminFunction][:adminFunction])
+            if message[:adminFunction] && message[:adminFunction].is_a?(Hash) && message[:adminFunction][:adminFunction] && message[:adminFunction][:varSeqData] && (Finaid::Shared::ADMIN_FUNCTION.include? message[:adminFunction][:adminFunction])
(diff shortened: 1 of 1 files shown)


christianvuerings:
FYI - the issue is because they haven't moved the API changes over to QAT yet.
For us this will be an issue for the interim release (since they won't have moved the fix to prod by then).

app/models/my_activities/campus_solutions_messages.rb
```diff
@@ -28,7 +28,7 @@ def self.get_feed(uid)
               cs: {}
             }
 
-            if message[:adminFunction] && message[:adminFunction][:varSeqData] && (Finaid::Shared::ADMIN_FUNCTION.include? message[:adminFunction][:adminFunction])
+            if message[:adminFunction] && message[:adminFunction].is_a?(Hash) && message[:adminFunction][:adminFunction] && message[:adminFunction][:varSeqData] && (Finaid::Shared::ADMIN_FUNCTION.include? message[:adminFunction][:adminFunction])
```
paulkerschen: @christianvuerings I don't think we need the explicit nil-or-false checks on `message[:adminFunction]` or `message[:adminFunction][:adminFunction]`. A nil or false `message[:adminFunction]` will return false on `is_a?(Hash)`, and a nil or false `message[:adminFunction][:adminFunction]` presumably won't be included in 'Finaid::Shared::ADMIN_FUNCTION`, right?

christianvuerings: @paulkerschen thanks! Learned something new.

PR #4677, opened 2015-12-21
CLC-6131 - fix bad element locator in Selenium test

pfarestveit:
https://jira.ets.berkeley.edu/jira/browse/CLC-6131


johncrossman:
Yay, a bit of xpath gone.

PR #4678, opened 2015-12-21
CLC-6110 - Header profile photo is too itsy bitsy teensy weensy

christianvuerings:
https://jira.ets.berkeley.edu/jira/browse/CLC-6110


johncrossman:
An itsy bitsy PR.

PR #4680, opened 2015-12-21
CLC-6129 Log and override stray 0 cache expiration times

raydavis:
Attempt at gathering diagnostics on the trigger of https://jira.ets.berkeley.edu/jira/browse/CLC-6129

Also fixes a bad config reference in RelatedCacheKeyTracker.

### DIFF
--- a/lib/cache/cacheable.rb
+++ b/lib/cache/cacheable.rb
@@ -8,10 +8,11 @@ module Cache
     # to the cache.
     def fetch_from_cache(id=nil, force_write=false)
       key = cache_key id
-      Rails.logger.debug "#{self.name} cache_key will be #{key}, expiration #{self.expires_in}, forced: #{force_write}"
+      expiration = self.expires_in
+      Rails.logger.debug "#{self.name} cache_key will be #{key}, expiration #{expiration}, forced: #{force_write}"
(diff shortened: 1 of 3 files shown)


lib/cache/cacheable.rb
```diff
@@ -72,7 +73,12 @@ def expires_in
       expirations = Settings.cache.expiration.marshal_dump
       exp = expirations[self.name.to_sym] || expirations[:default]
       exp = parse_expiration_setting(exp)
-      [exp, Settings.cache.maximum_expires_in].min
+      if exp.blank? || exp == 0
+        logger.error "Got 0 expiration for #{self.name}; resetting to default #{expirations[:default]}"
```
paulkerschen: @raydavis: Since diagnostics are the focus here, do you think there'd be value in raising and rescuing an error so that we can log a stack trace as well; or do you think this log in context will give us enough to go on?

raydavis: I was wondering about that too, @paulkerschen . Wonderment * 2 = PR amendment.

raydavis:
Requesting your eagle eyes again, @paulkerschen !

paulkerschen:
Looks good, @raydavis; let's catch that rabbit.

PR #4683, opened 2015-12-23
CLC-5589 New ActiveRecord model for service alerts

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5589

Note that the `preview` method plus rails_admin configuration should allow alerts to be created, previewed and published right from within ccadmin.

### DIFF
--- a/app/models/service_alerts/alert.rb
+++ b/app/models/service_alerts/alert.rb
@@ -1,0 +1,39 @@
+module ServiceAlerts
+  class Alert < ActiveRecord::Base
+    include DatedFeed
+    include HtmlSanitizer
+
+    self.table_name = 'service_alerts'
+
+    attr_accessible :title, :snippet, :body, :publication_date, :display
+
+    validates :title, presence: true
+    validates :body, presence: true
+    validates :publication_date, presence: true
+
+    after_initialize :set_default_publication_date
+
+    def self.get_latest
+      self.where(display: true).order(created_at: :desc).first
+    end
+
+    def set_default_publication_date
+      self.publication_date ||= Time.zone.today.in_time_zone.to_datetime
+    end
+
+    def preview
+      body.try :html_safe
+    end
+
+    def to_feed
+      {
+        title: title,
+        link: '/alert',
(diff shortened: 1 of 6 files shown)


paulkerschen:
Bamboo https://bamboo.ets.berkeley.edu/bamboo/browse/CPB-CPT237-5/

raydavis:
Thank you for taking on this important but much deferred task, @paulkerschen . This looks like a great start. Two starting questions:

* Our current UX shows the "title" as a banner alert and at the top of the front-page display and the "body" will show up on the alert display page and (HTML-sanitized) in the front-page display. But where will the required "summary" field be used?

* What would you think about adding a defaulted-but-editable "publication_date" field (to be prefixed to the title in display)? The Drupal version has occasionally been handy when preparing a service alert in advance.

paulkerschen:
@raydavis: "publication_date" is a good idea; I'll amend to add it.

"Summary" is intended to be used as "snippet" is used in the front-page display on the current UX:

https://github.com/ets-berkeley-edu/calcentral/blob/master/src/assets/templates/splash.html#L22

My understanding was that the snippet text displayed on the front page doesn't necessarily include the entire contents of the alert body.

raydavis:
I believe that was the original idea, @paulkerschen , but in practice it almost always turns out just to be the body of the alert. If we do decide to keep that flexibility (for when the body text only makes sense with full HTML formatting, for example), we'll want the "summary" to be either not-required or filled in automatically. How about making it not-required?

paulkerschen:
Makes sense to me, @raydavis!

paulkerschen:
Thanks for the feedback, @raydavis. PR amended with changes:
- Snippet field made optional;
- Defaulted-but-editable publication date added;
- Feed properties made consistent with EtsBlog::RSSProxy for easier merge into existing structures. Includes a `link` property pointing to the to-be-implemented `/alert` path.

raydavis:
Lookin' sharp, @paulkerschen -- thank you again for taking this on!

PR #4685, opened 2015-12-24
CLC-6130 - Amount Due Number After Name in Header Looks Odd

davidlimcheng:
https://jira.ets.berkeley.edu/jira/browse/CLC-6130

I *think* this is correct.  Also, would it make sense to add a "if .errored" to loadFinances() in statusController.js?

### DIFF
--- a/src/assets/stylesheets/_header.scss
+++ b/src/assets/stylesheets/_header.scss
@@ -260,8 +260,8 @@ body {
   }
   .cc-icon-status-person-layover-white {
     background: $cc-color-white;
-    color: $cc-color-dark-grey;
     border-radius: 50%;
+    color: $cc-color-dark-grey;
     cursor: pointer;
     font-size: 8px;
     height: 15px;
--- a/src/assets/stylesheets/_header.scss
+++ b/src/assets/stylesheets/_header.scss
@@ -256,13 +256,13 @@ body {
     position: absolute;
     top: -2px;
     z-index: 1;
-    &.red {
-      background: $cc-color-red-dark;
-    }
-    &.white {
-      background: $cc-color-white;
-      color: $cc-color-dark-grey;
-    }
+  }
+  .cc-icon-status-person-layover-red {
+    background: $cc-color-red-dark;
+  }
+  .cc-icon-status-person-layover-white {
+    background: $cc-color-white;
+    color: $cc-color-dark-grey;
   }
   .cc-popover-gear {
     .cc-header-icon {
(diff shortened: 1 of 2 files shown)


davidlimcheng:
Ahhh - pretty certain the Travis is failing because I had to disable Atom's linter-scss-lint.  Spent a very long time troubleshooting it because it wouldn't work, and didn't want it to be a blocker to actually getting some stuff done today.  My mistake - back to the drawing board!

Should I close this PR in the meantime?

christianvuerings:
@davidlimcheng feel free to leave it open but just push a couple of fixes.

I would also recommend having a `.cc-icon-status-person-layover` class which has the base CSS & a `.cc-icon-status-person-layover-white` & `.cc-icon-status-person-layover-red` with some color specific CSS. That way it would be a bit more DRY (Don't Repeat Yourself)

As for travis - it looks like only 1 of the 2 builds failed. I just restarted the failed one.

davidlimcheng:
Will do, @christianvuerings .  That makes a lot of sense.  Thanks!

Any idea why the Travis would stall like that and have to be restarted?  Also - is it possible for me to restart it if I see it stall like that?

christianvuerings:
@davidlimcheng we sometimes have false negatives when e.g. one of the many dependencies we're downloading is not available. That's the reason we're doing the same build twice.

And as for restarting it yourself I think you need to have push/commit powers to the repo. Another way might be to close / open the PR - see http://stackoverflow.com/questions/17606874/trigger-a-travis-ci-rebuild-without-pushing-a-commit

davidlimcheng:
@christianvuerings I see.  Got the scss-lint to finally start working, so we should be good!

src/assets/stylesheets/_header.scss
```diff
@@ -257,6 +256,13 @@ body {
     position: absolute;
     top: -2px;
     z-index: 1;
+    &.red {
```
christianvuerings: @davidlimcheng we actually want to avoid using qualifying selectors `&.classname`, they're less flexible and efficient:
* https://css-tricks.com/whats-the-difference/
* http://csswizardry.com/2012/07/quasi-qualified-css-selectors/

Also, we always namespace any of our classes (no `red` or `white`).

So Instead, would it be possible to use `.cc-icon-status-person-layover-red` & `.cc-icon-status-person-layover-white`?

davidlimcheng:
@christianvuerings for sure - I was having some trouble with that approach but I'll give the literature a good read.  Sorry this is taking so long, but this is very good to know best practices vs. what works!

davidlimcheng:
~~3rd~~ 4th times a charm.

PR #4687, opened 2015-12-24
CLC-6137 Bad property name in MyClasses front-end updates

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6137


paulkerschen:
The distinction between "data" and "feed" may escape the subtlest minds. Thanks for the fix @raydavis.

PR #4692, opened 2015-12-31
CLC-6141 Disambiguate mock proxy URIs which are subpaths of other proxies

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6141

### DIFF
--- a/app/models/bearfacts/profile.rb
+++ b/app/models/bearfacts/profile.rb
@@ -9,5 +9,11 @@ module Bearfacts
       "/student/#{@student_id}"
     end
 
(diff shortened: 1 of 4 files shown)


paulkerschen:
Sneaky bug, @raydavis, and the fix looks good. Down the road, I could see a case for clarifying mock proxy behavior by:
- replacing the `uri_matching` key with some assortment of more verbose and explicit keys (`uri_starting_with`, `uri_entire_path`...)
- Having some sort of callback to Webmock that would unregister URIs once a mock proxy goes out of scope.

PR #4694, opened 2016-01-04
SISRP-12545 Allow ex-student status from Oracle views to override student status from Hub

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-12545


johncrossman:
The brain is more minced than split. Looks good, @paulkerschen. I'll let @raydavis (second pair of eyes) pull the trigger on merge.

raydavis:
We may be expanding on this, but it's better than what we have now. Thanks, @paulkerschen !

PR #4697, opened 2016-01-04
CLC-6146, Profile - 'Add' button for email is showing accessibility text

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-6146

After this fix I grep'd for similar typos. No more violations found.
 ```grep -R visually src/ | grep -v cc-visuallyhidden```

Note: The grep above _excludes_ valid usages of cc-visuallyhidden.


raydavis:
Fast work!

PR #4700, opened 2016-01-04
SISRP-12376 - Remove the 'Do Not Display' code from CalCentral

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-12376

### DIFF
--- a/app/models/hub_edos/student.rb
+++ b/app/models/hub_edos/student.rb
@@ -4,8 +4,6 @@ module HubEdos
     include Cache::UserCacheExpiry
     include ResponseHandler
 
-    SENSITIVE_KEYS = %w(addresses names phones emails emergencyContacts)
-
     def initialize(options = {})
       super(Settings.hub_edos_proxy, options)
     end
@@ -21,7 +19,7 @@ module HubEdos
     end
 
     def build_feed(response)
-      transformed_response = filter_fields(redact_sensitive_keys(transform_address_keys(parse_response(response))))
+      transformed_response = filter_fields(transform_address_keys(parse_response(response)))
       {
         'student' => transformed_response
       }
--- a/app/models/hub_edos/student.rb
+++ b/app/models/hub_edos/student.rb
@@ -45,18 +45,6 @@ module HubEdos
       response
     end
 
-    def redact_sensitive_keys(response)
-      # TODO: more stuff the Integration Hub should be doing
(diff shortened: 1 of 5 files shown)


davidlimcheng:
@christianvuerings - For the 'def redact_sensitive_keys' definition in student.rb, is this something I should leave in case it's used in the future for any reason?  If not, I can go ahead and push a new commit up after deleting it.

christianvuerings:
@davidlimcheng since we don't use `redact_sensitive_keys` anywhere anymore, I would definitely suggest removing it as part of this PR.

Also have a look at the Travis CI failure.

davidlimcheng:
@christianvuerings Sounds good, I'll delete it and push a new commit.  I think the Travis was another false negative since there are no logs, I'll try to rerun it and see what comes up.

davidlimcheng:
Drats failed again.  I'll take a deeper look at it.

paulkerschen:
@davidlimcheng It looks like all the "missing template" failures on Travis are a red herring - they only appear in one of the two build. The test failure that appears in both builds is the following:

```
HubEdos::Contacts mock proxy returns data with the expected structure
     Failure/Error: expect(subject[:feed]['student']['addresses'][0]['postal']).to eq '94720'
       
       expected: "94720"
            got: "94708"
       
       (compared using ==)
     # ./spec/models/hub_edos/contacts_spec.rb:14:in `(root)'
```

paulkerschen:
@davidlimcheng The solution might be to edit `fixtures/json/student_contacts.json` to remove the "Do Not Display" code, same as you did with `fixtures/json/student_api_via_hub.json`.

davidlimcheng:
@paulkerschen Thank you!  I'll give it a shot.  I presume I'll also have to update the consistency check in `contacts_spec.rb`, i.e. 

```ruby
expect(subject[:feed]['student']['addresses'][0]['postal']).to eq '94720'
```

Should be eq to 94708.

davidlimcheng:
Along with `formattedAddress`

davidlimcheng:
@christianvuerings @paulkerschen Hurrah!

christianvuerings:
lgtm - @pfarestveit definitely something we should check during QA - setting an address / name / phone / ... to `Do Not Display` in CS and seeing whether it shows up in CalCentral.

PR #4701, opened 2016-01-05
CLC-6145 - QA: update profile info element locators

pfarestveit:
https://jira.ets.berkeley.edu/jira/browse/CLC-6145

Gnarly XPath = brittle element locator


johncrossman:
Gnice fix to a gnarly xpath. One green Travis.

PR #4705, opened 2016-01-05
NOJIRA, bump to v72 and latest db/migrate

johncrossman:
We bump version +2 because it was not incremented three weeks ago.


paulkerschen:
@johncrossman, for my own edification, why are we setting `previous_release_db_schema.txt` to the schema for v72 and not v71? I may not understand how our emergency db rollbacks work.

paulkerschen:
@johncrossman: Never mind, this is master, not QA, and that _is_ the schema for v71. All good!

PR #4706, opened 2016-01-05
SISRP-12175 - UX Develop: Preferred Name – Restrict character set for input in Profile

dfkaye:
https://jira.berkeley.edu/browse/SISRP-12175

### DIFF
--- a/src/assets/stylesheets/_profile_page.scss
+++ b/src/assets/stylesheets/_profile_page.scss
@@ -47,7 +47,6 @@
     .cc-page-widget-profile-section-input-error {
       background-color: $cc-color-red;
       border: 1px solid $cc-color-red;
-      border-radius: 3px;
       color: $cc-color-white;
       margin-bottom: 10px;
     }
--- a/src/assets/stylesheets/_profile_page.scss
+++ b/src/assets/stylesheets/_profile_page.scss
@@ -44,21 +44,22 @@
       padding: 10px;
       white-space: pre-line;
     }
-    .cc-page-widget-profile-section-input-error {
+    .cc-input-error {
       background-color: $cc-color-red;
       border: 1px solid $cc-color-red;
-      border-radius: 3px;
       color: $cc-color-white;
       margin-bottom: 10px;
     }
-    .cc-page-widget-profile-section-input-error-message {
+    .cc-input-error-message {
       background-color: $cc-color-red-light;
       border-bottom: 1px solid $cc-color-red;
       border-top: 1px solid $cc-color-red;
+      color: $cc-color-red;
       margin: 0 -10px;
       padding: 10px;
       white-space: pre-line;
     }
+
     .cc-page-widget-profile-section-content-item:not(:last-child) {
       margin-bottom: 15px;
     }
--- a/src/assets/stylesheets/_profile_page.scss
+++ b/src/assets/stylesheets/_profile_page.scss
@@ -44,6 +44,18 @@
       padding: 10px;
       white-space: pre-line;
     }
+    .cc-page-widget-profile-section-input-error {
+      background-color: $cc-color-red;
+      border: 1px solid $cc-color-red;
+      color: $cc-color-white;
+    }
+    .cc-page-widget-profile-section-input-error-message {
+      background-color: $cc-color-red-light;
+      border-bottom: 1px solid $cc-color-red;
+      margin: -10px 0 0;
+      padding: 10px;
+      white-space: pre-line;
+    }
     .cc-page-widget-profile-section-content-item:not(:last-child) {
       margin-bottom: 15px;
     }
--- a/src/assets/stylesheets/_profile_page.scss
+++ b/src/assets/stylesheets/_profile_page.scss
@@ -44,22 +44,21 @@
       padding: 10px;
       white-space: pre-line;
     }
-    .cc-input-error {
+    .cc-page-widget-profile-section-input-error {
       background-color: $cc-color-red;
       border: 1px solid $cc-color-red;
+      border-radius: 3px;
       color: $cc-color-white;
       margin-bottom: 10px;
     }
-    .cc-input-error-message {
+    .cc-page-widget-profile-section-input-error-message {
       background-color: $cc-color-red-light;
       border-bottom: 1px solid $cc-color-red;
       border-top: 1px solid $cc-color-red;
-      color: $cc-color-red;
       margin: 0 -10px;
       padding: 10px;
       white-space: pre-line;
     }
-
     .cc-page-widget-profile-section-content-item:not(:last-child) {
       margin-bottom: 15px;
     }
--- a/src/assets/stylesheets/_profile_page.scss
+++ b/src/assets/stylesheets/_profile_page.scss
@@ -44,14 +44,14 @@
       padding: 10px;
       white-space: pre-line;
     }
-    .cc-page-widget-profile-section-input-error {
+    .cc-input-error {
       background-color: $cc-color-red;
       border: 1px solid $cc-color-red;
       border-radius: 3px;
       color: $cc-color-white;
       margin-bottom: 10px;
     }
-    .cc-page-widget-profile-section-input-error-message {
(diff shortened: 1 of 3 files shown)


src/assets/templates/widgets/profile/basic_preferred_name_editor.html
```diff
@@ -11,7 +11,14 @@
...
       </div>
       <div>
-        <input type="text" id="cc-page-widget-basic-preferred-first-name" data-ng-model="currentObject.data.givenName" required aria-required="true" maxlength="30" />
+        <p>Preferred names are limited to alphabetical characters (A-Z), spaces, and hyphens. Consecutive spaces or hyphens are not allowed.</p>
+        <input type="text" id="cc-page-widget-basic-preferred-first-name" name="preferred_first_name" required aria-required="true" maxlength="30"
+          data-ng-class="{'cc-page-widget-profile-section-input-error': cc_page_widget_basic_preferred_name.preferred_first_name.$invalid}"
+          data-ng-model="currentObject.data.givenName" data-ng-pattern="preferredNamePattern" data-ng-trim="" />
+        <div class="cc-page-widget-profile-section-input-error-message cc-text-red" data-ng-if="cc_page_widget_basic_preferred_name.preferred_first_name.$invalid">
+          <i class="cc-icon-red fa fa-exclamation-circle"></i>
+          <strong>Invalid entry: </strong><span data-ng-bind="preferredNameErrorMessage"></span>
```
christianvuerings: @dfkaye since we only use `preferredNameErrorMessage` once, could we just include it here?

```
<span>Preferred name can only consist of letters, spaces, and hyphens. Please re-enter.</span>
```

src/assets/stylesheets/_profile_page.scss
```diff
@@ -44,6 +44,21 @@
...
       white-space: pre-line;
     }
+    .cc-page-widget-profile-section-input-error {
+      background-color: $cc-color-red;
+      border: 1px solid $cc-color-red;
+      border-radius: 3px;
+      color: $cc-color-white;
+      margin-bottom: 10px;
+    }
+    .cc-page-widget-profile-section-input-error-message {
```
christianvuerings: @dfkaye we can leave this as is but we should definitely think about a global error message class in the future. Looks like this is a near copy paste of `cc-delegate-linking-error` https://github.com/ets-berkeley-edu/calcentral/blob/d7dfeea78c7a40f7a23a1a8d1622c7832e2fbe99/src/assets/stylesheets/_delegate_welcome.scss#L19

src/assets/stylesheets/_profile_page.scss
```diff
@@ -44,6 +44,21 @@
       padding: 10px;
       white-space: pre-line;
     }
+    .cc-page-widget-profile-section-input-error {
+      background-color: $cc-color-red;
+      border: 1px solid $cc-color-red;
+      border-radius: 3px;
```
christianvuerings: @dfkaye even though it's in the design, we never used a border radius on an input field, it makes it inconsistent with our other styling. Could you check in with the designers and see whether we really want this?

src/assets/templates/widgets/profile/basic_preferred_name_editor.html
```diff
@@ -11,7 +11,14 @@
         </label>
       </div>
       <div>
-        <input type="text" id="cc-page-widget-basic-preferred-first-name" data-ng-model="currentObject.data.givenName" required aria-required="true" maxlength="30" />
+        <p>Preferred names are limited to alphabetical characters (A-Z), spaces, and hyphens. Consecutive spaces or hyphens are not allowed.</p>
+        <input type="text" id="cc-page-widget-basic-preferred-first-name" name="preferred_first_name" required aria-required="true" maxlength="30"
+          data-ng-class="{'cc-page-widget-profile-section-input-error': cc_page_widget_basic_preferred_name.preferred_first_name.$invalid}"
+          data-ng-model="currentObject.data.givenName" data-ng-pattern="preferredNamePattern" data-ng-trim="" />
```
christianvuerings: @dfkaye could you explain the empty `data-ng-trim`?

In the [angular input docs](https://docs.angularjs.org/api/ng/input/input%5Btext%5D), it mentions that `ngTrim` is enabled by default. Could you verify whether this still works when we remove the attribute on this `input` field?

src/assets/templates/widgets/profile/basic_preferred_name_editor.html
dfkaye: ok I'll do that.

src/assets/stylesheets/_profile_page.scss
dfkaye: Rachel asked for it only in the error state to distinguish it from the error message that appears below it.

src/assets/templates/widgets/profile/basic_preferred_name_editor.html
dfkaye: I left that in there as I'd used it for testing the validation against trimmed vs. non-trimmed input values.  I'll remove it.

src/assets/stylesheets/_profile_page.scss
dfkaye: I'll look at that to see what makes sense as a base or global field error state class.  Something like `cc-input-invalid` perhaps?

What about the error message class I've put in there (just below it)?

    .cc-page-widget-profile-section-input-error-message {
      background-color: $cc-color-red-light;
      border-bottom: 1px solid $cc-color-red;
      border-top: 1px solid $cc-color-red;
      margin: 0 -10px;
      padding: 10px;
      white-space: pre-line;
    }

That's is a one-off of 

    .cc-page-widget-profile-section-content-error {
      background-color: $cc-color-red-light;
      border-bottom: 1px solid $cc-color-red;
      margin: -10px -10px 10px;
      padding: 10px;
      white-space: pre-line;
    }

with a different `margin` and adds a `border-top`

Not sure we can generalize it yet but if we can, how about a name like `cc-input-invalid-message`?

dfkaye: @christianvuerings ~ Just a thought here.  Are all forms and input elements descendants of the `.cc-widget` class?  If so, might make sense to define `.cc-input-error` and `.cc-input-error-message` as descendants of that in the `_widgets.scss` file...

src/assets/stylesheets/_profile_page.scss
dfkaye: @christianvuerings ~ border-radius will be removed ~ after conversation w/Rachel, here's our understanding:

+ Buttons have border radius.
+ Text input fields do not have border radius.
+ Validation messages can be distinguished from text input fields by their wider width (zero margin).

src/assets/stylesheets/_profile_page.scss
christianvuerings: @dfkaye at this very moment they are but I would suggest just making it global instead. We just need to make sure we also have the `margin` & `border-top` consistent, however, I would suggest doing that in the future and not as part of this PR.

src/assets/stylesheets/_profile_page.scss
```diff
@@ -44,6 +44,21 @@
       padding: 10px;
       white-space: pre-line;
     }
+    .cc-input-error {
```
christianvuerings: @dfkaye I should have been more concrete.

If we leave it within this file, we should leave the name in this file as it was before (`cc-page-widget-profile-section-input-error` and `cc-page-widget-profile-section-input-error-message`.

However, if we decide to move it up in the chain (e.g. as part of `useful.scss`), then we should use `cc-input-error` and `cc-input-error-message`. And if we do that, we also need to make sure it's reused in all the other places (e.g. `cc-delegate-linking-error`).

My suggestion would be to just go for the old naming for now and fix the refactoring item in a separate PR / ticket.

src/assets/stylesheets/_profile_page.scss
dfkaye: @christianvuerings - I've reverted those css error-state names - ready for review again :)

PR #4707, opened 2016-01-06
CLC-6151 Show registration alerts only to active students

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6151


johncrossman:
Makes sense to me. Thanks @paulkerschen

PR #4709, opened 2016-01-06
CLC-6153 Handle nil term values in Oracle queries

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6153

Belt and suspenders for a world in which terms can be nil. The calendar model should not include nil values in its terms array; nor should the Oracle connection freak out if someone does send it a terms array with nil values.


raydavis:
Sometimes... in a world without belts... we look to one suspender...

PR #4711, opened 2016-01-06
SISRP-12558, revert change on data-cc-compile-directive = task.notes (SISRP-12156)

johncrossman:
https://jira.berkeley.edu/browse/SISRP-12558

[Git commit 7cf549c](https://github.com/christianvuerings/calcentral/commit/7cf549c6c5b425355597bbdc541adf680da9f517) fixed https://jira.berkeley.edu/browse/SISRP-12156 via data-cc-compile-directive. In this PR, I revert commit 7cf549c.

I'll reopen SISRP-12156 and, meanwhile, revert this change on master and qa.

PR #4712, opened 2016-01-06
SISRP-12558, QA PR: revert change on data-cc-compile-directive = task.notes (SISRP-12156)

johncrossman:
https://jira.berkeley.edu/browse/SISRP-12558

This is QA follow up to PR #4711 which I verified in cc-sis-qa.


paulkerschen:
If it's worth mergin' once, it's worth mergin' twice.

PR #4714, opened 2016-01-07
CLC-5589 Replace RSS service alert proxy with Postgres-derived feed

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5589 

The JIRA issue has descriptions and screenshots of the new behavior. Aside from losing a bunch of proxy-related code, the architectural changes in this PR have to do with decoupling service alerts from the My Badges feed, since the merge with user-specific data has been causing sticky-cache issues.

Service alerts used to appear in two different feeds: both in My Badges, and merged in with release notes at the controller level. This PR removes the alerts from My Badges, and replaces the merge in BlogFeedController with one of our more usual cached-as-JSON merged models (ServiceAlerts::Merged). Cache time remains set to 2 minutes. While we're no longer making an expensive network call for alerts, and the release notes have their own caching at the default 35-minute mark, this merged feed is going to get served up with every session, so we may as well take a little load off.

Decoupling service alerts from My Badges means that stale alerts will no longer get served up at login. The flipside is that service alerts no longer get the benefit of live updates. Our live update code expects to deal with user-specific data and does a lot of caching by UID, and hooking up all that machinery to a non-user-specific feed felt messy and wrong. Still, I'm open to looking at workarounds if others feel that this would be too great a regression.

Snippets are no longer HTML-sanitized, since they're coming from the same back end as the HTML body, and the splash page looks much nicer with paragraphs and links present.


src/assets/templates/splash.html
```diff
@@ -17,9 +17,16 @@
       </div>
       <div class="medium-10 columns small-12">
         <p class="cc-page-splash-blog-title">
-          <a data-ng-href="{{splashNote.link}}">{{splashNote.title}}</a>
+          <a data-ng-if="splashNote.link" data-ng-href="{{splashNote.link}}" data-ng-bind="splashNote.title"></a>
+          <span data-ng-if="!splashNote.link" data-ng-bind="splashNote.title"></span>
         </p>
-        <p class="cc-page-splash-blog-summary" data-ng-bind-html="splashNote.snippet"></p>
+        <div data-ng-show="splashNote.snippet && !showFullBody">
```
raydavis: "showFullBody" sounds interesting, but where's it coming from?

raydavis: Ah, I see, it's coming from two lines down!

raydavis:
Thanks, @paulkerschen , I & @pfarestveit look forward to kicking the new tires.

src/assets/templates/splash.html
paulkerschen: @raydavis It's a curious pattern we seem to have in use for "Show more" messages. Example - the "showReminder" variable here: 
https://github.com/ets-berkeley-edu/calcentral/blob/master/src/assets/templates/no_bconnected.html
Since the controller never sets it up, it starts out undefined and then toggles to true.

PR #4716, opened 2016-01-07
CLC-6155, script to verify API endpoints now includes Delegated Access

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-6155

In addition to adding new DA endpoints to the script (e.g., "/UC_CC_DELEGATED_ACCESS_URL.v1/get") I also:
* validate cmd line input
* write response body to log file
* grep response for known errors

Change to `.gitignore` allows us to write log files to a subdir of `/log/` 

cc: @dkawase

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,7 +11,7 @@
 /db/*.sqlite3
 
 # Ignore all logfiles and tempfiles.
-/log/*.log
+/log/**/*.log
 /log/*.gz
 /tmp/*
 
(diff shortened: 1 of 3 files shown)


johncrossman:
**Note:** https://jira.berkeley.edu/browse/SISRP-13068 reports unexpected results from DA API.

paulkerschen:
Getting fancy!

PR #4718, opened 2016-01-07
CLC-6158 Show transition-term profile header to active students only

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6158


johncrossman:
It's good to be active.

PR #4719, opened 2016-01-08
CLC-6155, add 'not authorized' to grep'ing response for errors

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-6155

When testing the Delegated Access endpoints I identified another error condition to grep for: 'not authorized'.  I removed "/UC_DA_T_C.v1/get" because it's a placeholder and invalid.


paulkerschen:
Travis is half happy. All good to me, though I bet we're not done finding errors to grep for.

PR #4723, opened 2016-01-08
CLC-6085 Do not attempt to parse bizarro Hub 404s as JSON

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6085

Hub outages sometimes result in unexpected 404 responses with unexpected payloads that defeat our JSON parser. As of #4590 we're logging the payload, so we might as well return our predetermined `empty_feed` rather than letting the JSON parser fail and return the same result.

Everything else is formatting fixes in the same class.

### DIFF
--- a/app/models/hub_edos/proxy.rb
+++ b/app/models/hub_edos/proxy.rb
@@ -44,8 +44,8 @@ module HubEdos
           internal_response
         else
           internal_response.merge({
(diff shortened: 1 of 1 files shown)


johncrossman:
Thanks @paulkerschen  No mo' bizarro

PR #4726, opened 2016-01-10
SISRP-13172 - Negative service indicators should only show for applicants & students

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-13172


raydavis:
Thanks, @christianvuerings , those alerts were getting a bit monotonous.

christianvuerings:
@raydavis as a fyi - this doesn't fix it yet in anything that is not linked to BCSDEV - see https://jira.berkeley.edu/browse/SISRP-12461 (we need to migrate the CS project to the other servers)

PR #4728, opened 2016-01-10
NOJIRA Stop logging CS request options even at DEBUG level

raydavis:
They're quite dull.

@christianvuerings , wouldn't you agree?


johncrossman:
Dull as ditchwater. Thanks @raydavis

christianvuerings:
@raydavis thanks ;)

PR #4731, opened 2016-01-11
CLC-6142 Add -Xcompile.invokedynamic=false for Java 8

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6025

This is safe to land before the clusters are moved to 1.8.


raydavis:
One green Travis.

PR #4734, opened 2016-01-12
SISRP-13249 Allow JmsWorker to handle multiple connections and queues

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-13249

JmsConnection now handles multiple queues, and JmsWorker handles multiple JmsConnection instances (as we might need to listen to the new Hub connection and our old Bearfacts connection simultaneously).

This implementation sticks with a single handler instance, since our Notifications::Handler seems already well set up to dispatch received messages among multiple processors.

NB: Changing connections and queues to arrays breaks YAML compatibility. Please ping me before merge so that YAML edits can be synchronized with deploys. Developer machines should also update their YAML, since the incompatibility will probably secretly persist until the next time one tries to run Torquebox.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -290,13 +290,14 @@ cache_warmer:
 
 # Bearfacts JMS event queue
 ist_jms:
-  url: tcp://localhost:61616
-  queue: testqueue
-#  username: USERNAME
-#  password: PASSWORD
-  freshen_recording: false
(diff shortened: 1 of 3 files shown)


raydavis:
Looks great, @paulkerschen , and it even looks safe for me to edit existing config files before the new code lands.

PR #4735, opened 2016-01-13
CLC-5752 - Recent Activity: item title can overlap the toggle

davidlimcheng:
https://jira.ets.berkeley.edu/jira/browse/CLC-5752

### DIFF
--- a/src/assets/stylesheets/_activities.scss
+++ b/src/assets/stylesheets/_activities.scss
@@ -107,7 +107,7 @@
     margin-left: 30px;
     .cc-widget-activities-title {
       display: inline-block;
-      width: 94%;
+      margin-right: 20px;
     }
     .cc-widget-activities-status-line {
       margin-top: 4px;
--- a/src/assets/stylesheets/_activities.scss
+++ b/src/assets/stylesheets/_activities.scss
@@ -106,7 +106,8 @@
   .cc-widget-activities-text-margin {
     margin-left: 30px;
     .cc-widget-activities-title {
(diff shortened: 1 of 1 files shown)


src/assets/stylesheets/_activities.scss
```diff
@@ -106,7 +106,8 @@
   .cc-widget-activities-text-margin {
     margin-left: 30px;
     .cc-widget-activities-title {
-      margin-right: 20px;
+      display: inline-block;
+      width: 94%;
```
christianvuerings: @davidlimcheng is there any other way to solve this?
Does it always work, even when we scale it down?
Do we know why the `margin-right` didn't fix it?

davidlimcheng:
@christianvuerings I believe `margin-right` didn't fix it because `.cc-widget-activities-title` was wrapped in a `<strong>` element, and things like `margin-right` and `width` don't apply to inline elements unless `display: inline-block` is called to turn that element into a block.  That being said, `margin-right` also fixes it as well, and I'm realizing now that it's probably better to use than `width` to keep consistency with our other css elements.

`width` also does the trick, and my original train of thought was that using a % would be more responsive, but after doing some research, I see that it's really limiting as opposed to px where you can move things however much you like.

Will push a new commit now!

christianvuerings:
Makes total sense, thanks for digging deep and figuring out a good fix.

This lgtm as soon as travis is showing some :green_heart:.
(other devs, feel free to merge this)

johncrossman:
...A'merging we go!

PR #4737, opened 2016-01-13
SISRP-13362 - CalCentral-QA ToolBox View As UID/SID Field

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-13362


raydavis:
The commit log message is pretty opaque, @christianvuerings , but the change itself is easy to understand. :)

PR #4738, opened 2016-01-13
CLC-5648 Remove obsolete service alerts proxy configs

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5648


johncrossman:
grep tells me this is the proper amount of surgery. We'll see if Travis agrees.

PR #4739, opened 2016-01-13
CLC-6165 Fix bug when cache refreshes on the exact minute it expired

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6165

This has been plaguing us for many months, but the logging in Monday's release finally pinpointed the cause.


paulkerschen:
What a sneaky one. Thanks @raydavis.

PR #4742, opened 2016-01-14
SISRP-13418 - CalCentral - Remove minimal-ui from our viewport meta tag

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-13418


johncrossman:
Very meta.

PR #4743, opened 2016-01-14
SISRP-12156 - Can links within the text of the task (Checklist Item verbiage) open in a new window?

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-12156


christianvuerings:
@johncrossman might have an interest in having a look at this one.

johncrossman:
Thanks for the inline comment, @christianvuerings

PR #4747, opened 2016-01-14
SISRP-13434 CS finaid proxies: add default year, improve error handling

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-13434

Correct a false assumption that only PostingProxy will return UC_CM_FAULT_DOC; financial aid proxies are returning it also.

When CampusSolutions::MyFinancialAidData is called by LiveUpdatesWarmer, set it to the user's default aid year. It would be nice to have it warm up other years as well, but that would require architectural changes to the class.

### DIFF
--- a/app/models/campus_solutions/direct_proxy.rb
+++ b/app/models/campus_solutions/direct_proxy.rb
@@ -15,5 +15,9 @@ module CampusSolutions
       })
     end
 
+    def error_response_root_xml_node
+      'UC_CM_FAULT_DOC'
+    end
+
   end
 end
--- a/app/models/campus_solutions/my_aid_years.rb
+++ b/app/models/campus_solutions/my_aid_years.rb
@@ -11,5 +11,13 @@ module CampusSolutions
       CampusSolutions::AidYears.new({user_id: @uid}).get
     end
 
+    def default_aid_year
+      feed = self.get_feed
+      years = feed && feed[:feed] && feed[:feed][:finaidSummary] && feed[:feed][:finaidSummary][:finaidYears]
+      if years && (default_year = years.find { |y| y[:default] })
+        default_year[:id]
+      end
+    end
+
   end
 end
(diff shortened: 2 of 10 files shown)


app/models/campus_solutions/my_aid_years.rb
```diff
@@ -11,5 +11,13 @@ def get_feed_internal
       CampusSolutions::AidYears.new({user_id: @uid}).get
     end
 
+    def default_aid_year
+      feed = self.get_feed
+      years = feed && feed[:feed] && feed[:feed][:finaidSummary] && feed[:feed][:finaidSummary][:finaidYears]
+      if years && (default_year = years.find { |y| y[:default] })
+        default_year[:id]
```
johncrossman: We can rely on years to always have a `:default`?

paulkerschen: If any element in `years` doesn't have a `:default`, then `y[:default]` will evaluate to nil and it will just move on to the next one, returning nil for the block if nothing turns up.

app/models/campus_solutions/direct_proxy.rb
```diff
@@ -15,5 +15,9 @@ def request_options
       })
     end
 
+    def error_response_root_xml_node
+      'UC_CM_FAULT_DOC'
+    end
+
```
johncrossman: Does `AidYears` class need change, too?

paulkerschen: `AidYears` doesn't take any parameters other than ID, so it's harder to trigger a UC_CM_FAULT_DOC on it. It seems like what PostingProxy has in common with the finaid classes is that it accepts extra parameters and throws a fault if they don't work out.

johncrossman:
Looks good, @paulkerschen  There are a few finAid related proxies (AidYears, TermsAndConditions, Title4) that did not get treatment in this PR. If those are not of concern then I'll proceed with merge.

PR #4750, opened 2016-01-15
SISRP-13257, controller to manage session during delegate-view-as mode

johncrossman:
https://jira.berkeley.edu/browse/SISRP-13257

You'll see that `DelegateActAsController` extends `ActAsController`. If shared logic becomes a problem we'll split off the `delegate-act-as` code. After this merge I'll drop in `footer.html` and related.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -129,6 +129,8 @@ Calcentral::Application.routes.draw do
   end
 
   # Act-as endpoints
+  post '/delegate_act_as' => 'delegate_act_as#start', :via => :post
+  post '/stop_delegate_act_as' => 'delegate_act_as#stop', :via => :post
   post '/act_as' => 'act_as#start', :via => :post
   post '/stop_act_as' => 'act_as#stop', :via => :post
   get '/stored_users' => 'stored_users#get', :via => :get, :defaults => { :format => 'json' }
--- a/src/assets/javascripts/angular/controllers/widgets/delegateActingAsController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/delegateActingAsController.js
@@ -1,0 +1,17 @@
+'use strict';
+
+var angular = require('angular');
+
+/**
+ * Controller for delegate users in act-as mode.
+ */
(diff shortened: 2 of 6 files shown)


src/assets/javascripts/angular/factories/adminFactory.js
```diff
@@ -53,12 +63,14 @@ angular.module('calcentral.factories').factory('adminFactory', function(apiServi
   };
 
   return {
+    delegateActAs: delegateActAs,
```
christianvuerings: @johncrossman fyi - usually we try to alphabetize these

src/assets/javascripts/angular/controllers/widgets/delegateActingAsController.js
```diff
@@ -0,0 +1,17 @@
...
+ * Controller for delegate users in act-as mode.
+ */
+angular.module('calcentral.controllers').controller('DelegateActingAsController', function(adminFactory, apiService, $scope) {
+  $scope.delegate = {};
+
+  /**
+   * Stop acting as someone else
+   */
+  $scope.delegate.stopActAs = function() {
+    adminFactory.stopDelegateActAs().success(apiService.util.redirectToToolbox).error(apiService.util.redirectToToolbox);
```
christianvuerings: @johncrossman if `success` and `error` is the same, we should probably use `.then` instead.

johncrossman:
Thanks @christianvuerings  I've amended the commit.

config/routes.rb
```diff
@@ -129,6 +129,8 @@
   end
 
   # Act-as endpoints
+  post '/delegate_act_as' => 'delegate_act_as#start', :via => :post
+  post '/stop_delegate_act_as' => 'delegate_act_as#stop', :via => :post
```
paulkerschen: @johncrossman FYI, all these `:via` options in our routes are redundant holdovers from an older version of Rails. I'll create a backlog ticket to remove them in our leisure time.

paulkerschen:
images

johncrossman:
Yay! Merged! Thanks, @paulkerschen   
gettin-ziggy-with-it

PR #4756, opened 2016-01-15
CLC-6098, informative logging in campus_solution proxy, plus hash check

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-6098

In `campus_solutions/proxy` the call to `convert_feed_keys(feed)` can return either an array or a hash. However, our `is_errored?` method has assumed hash so this PR includes a sanity check.

In the case of error, we now log the cs_id and raw response.


paulkerschen:
{plus: 1}

PR #4757, opened 2016-01-15
SISRP-13426 - Update Title IV authorization shows not authorized even after selecting to Authorize

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-13426


paulkerschen:
@christianvuerings: From what I can tell in tracking the `refreshCache` option, it seems to be clearing client-side cache only. Is that correct, and if so, do we also need server-side work to expire the aid years feed?

christianvuerings:
@paulkerschen Great question, but if I'm not mistaken, that should already happen, see https://jira.berkeley.edu/browse/SISRP-6245

They way I tested this is by reloading the browser, which did fix the issue mentioned in SISRP-13426.

paulkerschen:
Alles gut, @christianvuerings.

PR #4761, opened 2016-01-15
CLC-6168 Flatten caching in Campus Solutions proxies

paulkerschen:
Part two of https://jira.ets.berkeley.edu/jira/browse/CLC-6168. This was a little more refactoring than I would have preferred, but caching was baked pretty deeply into the proxy classes.

- CampusSolutions::DirectProxy, an unneeded intermediary between CampusSolutions::Proxy and all subclasses, is removed. Proxies now inherit directly from CampusSolutions::Proxy or from the subclass CampusSolutions::CachedProxy.
- Caching is retained for proxies that talk to a controller, and for a couple of special cases such as the long-lived CampusSolutions::DashboardUrl.
- CampusSolutions::PostingProxy and subclasses had expiration set to one second in order to work around the baked-in caching behavior. Now that caching has been relegated to a different subclass, the workaround is no longer needed.
- The unused CampusSolutions::IntegrationHubProxy module is removed.


raydavis:
Thank you for cutting through this tangle, @paulkerschen ! I'll merge & start resolving conflicts with my working branch.

PR #4767, opened 2016-01-19
SISRP-13484, delegateActingAsUid according to actingAsUid usage

johncrossman:
https://jira.berkeley.edu/browse/SISRP-13484

Delegate-act-as mode must play by the rules of standard act-as (no bConnected, etc.). In this PR we get coverage of concerns. Future PRs will do more to "filter" the delegate experience.


christianvuerings:
@johncrossman thanks!

PR #4768, opened 2016-01-19
CLC-6175 Reduce log level on OEC unexpected-data errors

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6175

For a large class of OEC errors involving user-modifiable data on Google Drive, surface the problem to users, but log at WARN instead of ERROR and don't include a stack trace.

For more serious problems, Google connection errors, etc., retain current behavior.


johncrossman:
Makes complete sense. Thanks @paulkerschen

PR #4772, opened 2016-01-20
SISRP-13261, Welcome to CalCentral for delegate users

johncrossman:
https://jira.berkeley.edu/browse/SISRP-13261
https://jira.berkeley.edu/browse/SISRP-13381

* New card for Toolbox page. See PDF attached to https://jira.berkeley.edu/browse/SISRP-6601
* Hide delegate cards when feature flag


src/assets/javascripts/angular/controllers/widgets/delegateIntroController.js
```diff
@@ -0,0 +1,19 @@
...
+  $scope.delegateIntro = {
+    isLoading: true,
+    showComplete: false
+  };
+
+  var loadInformation = function() {
+    $scope.delegateIntro.isLoading = false;
+  };
+
+  loadInformation();
```
paulkerschen: @johncrossman If I'm reading right, when this controller is loaded, `$scope.delegateIntro.isLoading` will be initialized as true, then set to false an eyeblink later. Is there any need for that pattern here? Generally we use it to set the spinner directive while APIs are loading from the back end, but it seems we have no such APIs for this card.

src/assets/templates/widgets/toolbox/delegate_intro_to_calcentral.html
```diff
@@ -0,0 +1,49 @@
+<div
+  class="cc-widget cc-toolbox-widget"
+  data-ng-controller="DelegateIntroController"
+  data-cc-spinner-directive="delegateIntro.isLoading"
```
paulkerschen: @johncrossman See above; might be we don't need this.

src/assets/javascripts/angular/controllers/widgets/delegateIntroController.js
johncrossman: Yes, you're right @paulkerschen I'm going to close this PR in favor of simpler setup. Thank you.

johncrossman: New PR is #4774

PR #4774, opened 2016-01-20
SISRP-13261, Welcome to CalCentral for delegate users

johncrossman:
https://jira.berkeley.edu/browse/SISRP-13261
* New card for Toolbox page
* Hide delegate cards when feature flag

### DIFF
--- a/src/assets/templates/widgets/toolbox/delegate_intro_to_calcentral.html
+++ b/src/assets/templates/widgets/toolbox/delegate_intro_to_calcentral.html
@@ -1,0 +1,45 @@
+<div class="cc-widget cc-toolbox-widget">
+  <div class="cc-widget-title">
+    <h2>Welcome to CalCentral</h2>
+  </div>
+  <div class="cc-widget-padding">
+    <h3>You are a Delegate</h3>
+    <p>
+      CalCentral is the portal for the University of California, Berkeley. If you are seeing this view of CalCentral,
+      you have been designated as a Delegate by a Berkeley student. Delegates are often parents, guardians or other
+      trusted adults.
+    </p>
+    <p>
+      You may be a Delegate for more than one student.
+    </p>
+    <h3>Using CalCentral</h3>
+    <h4>Your Privileges</h4>
+    <p>
+      A student (your Delegator) has delegated privileges to you. Your privileges may include features such as paying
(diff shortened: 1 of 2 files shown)


paulkerschen:
Simplicity! Looks good @johncrossman

src/assets/templates/widgets/toolbox/delegate_intro_to_calcentral.html
```diff
@@ -0,0 +1,45 @@
...
+    </p>
+    <p>
+      You may be a Delegate for more than one student.
+    </p>
+    <h3>Using CalCentral</h3>
+    <h4>Your Privileges</h4>
+    <p>
+      A student (your Delegator) has delegated privileges to you. Your privileges may include features such as paying
+      bills and viewing the student's schedule.<br />
+      <button class="cc-button-link" data-ng-show="!showDelegateWelcome" data-ng-click="showDelegateWelcome=true">Show more</button>
```
christianvuerings: @johncrossman definitely a minor optimization but would suggest trying out using `data-ng-if` instead of `data-ng-show` and `data-ng-show` in this file.

johncrossman: I'll revise in upcoming PR. Thanks @christianvuerings

paulkerschen: @christianvuerings @johncrossman In fact I recently tried switching to `data-ng-if` with this same "Show more" pattern in #4714, and it didn't work because of scope issues. I suspect that's why we use `data-ng-show` for the pattern elsewhere.

johncrossman: Thanks @paulkerschen  Truth be told, I tried `data-ng-if` while working on this widget and it did not work. I grep'd and found pattern @paulkerschen mentions.

christianvuerings: @johncrossman @paulkerschen indeed, that's why I mentioned to try it out, if it doesn't, no worries.

The way we should be able to get around this is by having a `.` in our model http://stackoverflow.com/questions/17606936/angularjs-dot-in-ng-model and specifying it in a controller.

PR #4775, opened 2016-01-21
SISRP-13375, delegate should never see student's preferred name

johncrossman:
https://jira.berkeley.edu/browse/SISRP-13375

*Note:* From what I can tell, `preferred_name` and `last_name` are not used in templates but I kept them in the feed, with camelCase.  See Jira comments for more info.

### DIFF
--- a/app/controllers/user_api_controller.rb
+++ b/app/controllers/user_api_controller.rb
@@ -8,9 +8,9 @@ class UserApiController < ApplicationController
   end
 
   def am_i_logged_in
-    response.headers["Cache-Control"] = "no-cache, no-store, private, must-revalidate"
-    response.headers["Pragma"] = "no-cache"
-    response.headers["Expires"] = "-1"
+    response.headers['Cache-Control'] = 'no-cache, no-store, private, must-revalidate'
+    response.headers['Pragma'] = 'no-cache'
+    response.headers['Expires'] = '-1'
     render :json => {
       :amILoggedIn => !!session['user_id']
     }.to_json
@@ -37,7 +37,7 @@ class UserApiController < ApplicationController
         :delegateActingAsUid => directly_authenticated ? false : delegate_acting_as_uid,
         :youtubeSplashId => Settings.youtube_splash_id
       })
-      status.merge!(User::Api.from_session(session).get_feed)
+      status.merge! User::Api.from_session(session).get_feed
     else
       status.merge!({
         :isBasicAuthEnabled => Settings.developer_auth.enabled,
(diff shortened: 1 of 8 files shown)


app/models/hub_edos/user_attributes.rb
```diff
@@ -55,17 +55,24 @@ def extract_names(edo, result)
...
+            if name[:type][:code].upcase == 'PRI'
+              result[:given_name] = name[:givenName]
+              result[:family_name] = name[:familyName]
+            end
+            if name[:type].present? && name[:type][:code].present? && name[:type][:code].upcase == type.upcase
+              result[:first_name] = name[:givenName]
+              result[:last_name] = name[:familyName]
+              result[:person_name] = name[:formattedName]
+              found_match = true
+            end
```
johncrossman: This code is convoluted so I tried to avoid a significant refactor. Bottom line: even when user has a 'PRF' name we must inspect 'PRI' for givenName and familyName.

johncrossman:
One green Travis.

paulkerschen:
@johncrossman: I'm not up on the specific requirements for this feature, but it seems that there might be a problem in that we're still sending the preferred name to the front end, even if it doesn't show up in the templates; so the hypothetical nosy delegate can still open up their browser console and see the hidden data. Our other filtered-for-delegates or filtered-for-view-as feeds have had to do their censoring in the back end, replacing feed values as appropriate, rather than switching out keys in the template. Am I missing anything in this particular case?

johncrossman:
Thank you @paulkerschen, filtering is a must. An amendment is coming soon.

johncrossman:
Okay, finally back to coding... Travis is in effect.

app/controllers/user_api_controller.rb
```diff
@@ -37,7 +37,8 @@ def mystatus
         :delegateActingAsUid => directly_authenticated ? false : delegate_acting_as_uid,
         :youtubeSplashId => Settings.youtube_splash_id
       })
-      status.merge!(User::Api.from_session(session).get_feed)
+      feed = User::Api.from_session(session).get_feed
+      status.merge! feed
```
johncrossman: Actually, I'll amend this bit. Residue from original location of filter call.

app/controllers/user_api_controller.rb
```diff
@@ -37,7 +37,7 @@ def mystatus
         :delegateActingAsUid => directly_authenticated ? false : delegate_acting_as_uid,
         :youtubeSplashId => Settings.youtube_splash_id
       })
-      status.merge!(User::Api.from_session(session).get_feed)
+      status.merge! User::Api.from_session(session).get_feed
```
johncrossman: Okay, one-liner restored.

app/controllers/user_api_controller.rb
```diff
@@ -46,9 +46,18 @@ def mystatus
         :youtubeSplashId => Settings.youtube_splash_id
       })
     end
+    filter_user_api_for_delegator status if delegate_acting_as_uid
     render :json => status.to_json
   end
 
+  def filter_user_api_for_delegator(feed)
+    # Remove preferred name and sensitive data. Extraordinary privileges are set to false.
+    feed.reject! { |k, v| [:firstName, :lastName, :fullName, :preferredName, :firstLoginAt].include? k }
```
paulkerschen: @johncrossman As part of this filter, what would you think about not deleting `:firstName and :fullName`, but instead replacing them with the values of `:givenFirstName` and `:givenFullName`? That way it wouldn't be necessary to add all those ternary operators into the Angular attributes.

johncrossman: A good idea. Thx @paulkerschen. Please stand by.

paulkerschen:
Looks solid to me, @johncrossman, and as far as I can see, the controller-level filtering should sidestep cache conflicts. Thanks!

PR #4780, opened 2016-01-21
CLC-5910 Ensure visibility of concurrent-enrollment academics profile message

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5910


johncrossman:
Like The Jeffersons, concurrent-enrollment message is movin' on up. Looks good, @paulkerschen

PR #4792, opened 2016-01-26
SISRP-13232 Enrollment card MVP: back-end work

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-13232

Proxies and support structure for the 'enrollment term' and 'enrollment terms' APIs. The URLs are fake and the code is likely to change once we have real endpoints to consume.

Since caching requirements are not yet nailed down, for now I've given the two user-specific models default lifetimes and hooked up MyEnrollmentTerm (but not MyEnrollmentTerms) to LiveUpdates. This means that MyEnrollmentTerm needs the ability to set a default term id, similar to the fix for MyFinancialAidData in #4747.

@christianvuerings, note that the dummy JSON has changed a bit to match the actual back-end response and the Angular controller is looking one element deeper for the feed. Whether or not this structure is strictly necessary, I wanted to keep it consistent with the other CS proxies.


christianvuerings:
@paulkerschen thanks - also created https://jira.berkeley.edu/browse/SISRP-13947 for the actual hookup.

src/assets/templates/widgets/enrollment/adjust.html
```diff
@@ -1,7 +1,7 @@
 <div
   data-cc-enrollment-card-header-directive
   data-count="$index + 1"
-  data-date="After to {{enrollmentTerm.enrollmentPeriodsById.phase1.date.epoch * 1000 | date:'MMM d'}}"
```
christianvuerings: @paulkerschen thanks ;)

christianvuerings:
@paulkerschen and looking at the PR, looks like you already took care of it. However, I'll still leave that jira open to verify whether everything works, once we actually have the API.

paulkerschen:
Amended to fix conflict with #4793.

christianvuerings:
Lgtm but might be good for @johncrossman or @raydavis to have a look at this as well.

johncrossman:
I've reviewed back-end code... LGTM, @paulkerschen

PR #4795, opened 2016-01-26
SISRP-13564 QA - Add sentence to the SIR received message to notify students when they can log in and see updates

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-13564


johncrossman:
Hurry! Put to QA before @pfarestveit is back from his meeting.

PR #4798, opened 2016-01-27
CLC-5842 Allow override of RAILS_ENV in shell scripts

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5842

SPOILER: there are no surprises.


johncrossman:
SPOILER: SIS project is swallowed by Sharknado and GoLive 5 never happens.

PR #4801, opened 2016-01-27
SISRP-9941 - Profile Contact Info - don't show the preferred checkbox when the item is preferred

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-9941

PR #4802, opened 2016-01-27
SISRP-9941 QA - Profile Contact Info - don't show the preferred checkbox when the item is preferred

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-9941

### DIFF
--- a/src/assets/javascripts/angular/services/profileService.js
+++ b/src/assets/javascripts/angular/services/profileService.js
@@ -9,7 +9,7 @@ angular.module('calcentral.services').service('profileService', function() {
    */
   var actionCompleted = function($scope, data, callback) {
     if (data.data.errored) {
-      $scope.errorMessage = data.data.feed.errmsgtext;
+      $scope.errorMessage = _.get(data, 'data.feed.errmsgtext') || 'An error occurred while saving your data.';
     } else {
       $scope.closeEditor();
       callback({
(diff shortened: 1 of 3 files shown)


src/assets/javascripts/angular/services/profileService.js
```diff
@@ -9,7 +9,7 @@ angular.module('calcentral.services').service('profileService', function() {
    */
   var actionCompleted = function($scope, data, callback) {
     if (data.data.errored) {
-      $scope.errorMessage = data.data.feed.errmsgtext;
+      $scope.errorMessage = _.get(data, 'data.feed.errmsgtext') || 'An error occured while saving your data.';
```
christianvuerings: @paulkerschen @johncrossman FYI this was happening when QAT was down (503). So we'll show the `errormsgtext` if we get it, otherwise, we show this alternate error message.

johncrossman:
@christianvuerings Has someone proclaimed this a must-fix for v72?

christianvuerings:
@johncrossman the jira mentions GL4, so that's what I'm following. It looks like @benhubbard set it. I would also suggest trying the bug out yourself, the current behavior (without this fix) is very annoying.

johncrossman:
Annoying? I agree. Must fix? I vote no. Let's not merge until those on high say we must.

johncrossman:
@christianvuerings After merging #4801 to master, I tested add/edit "Other" email on `cc-sis-qa`. I discovered that https://jira.berkeley.edu/browse/SISRP-13870 (a very related problem) is not fixed; PaulF is reopening that Jira with log snippet. BenH agreed on postponing this fix to v73 release, at which time SISRP-13870 and proper error handling will be done. Bottom line: we should close this PR.

PR #4807, opened 2016-01-27
SISRP-7689, strengthen act_as_authorization and give front-end uid, too

johncrossman:
https://jira.berkeley.edu/browse/SISRP-7689

The user cannot delegate-view-as unless s/he has one of: :financial, :viewEnrollments, :viewGrades. Delegates with only :phone privileges will be denied. Also, front-end needs `ldap_uid` in delegate_students card.


paulkerschen:
A place for every ID. Every ID in its place.

PR #4812, opened 2016-01-28
SISRP-13362, ToolBox > View As > 10 digit CS ID

johncrossman:
https://jira.berkeley.edu/browse/SISRP-13362

Travis might fail on first pass but I want to get code diff out there for review.


johncrossman:
@paulkerschen  I did a hurried test on localhost using 10 digit CS ID and it worked. Next question: what's the risk?

PR #4813, opened 2016-01-28
CLC-6174 -  QA test fixes for v72 (profile tests)

pfarestveit:
https://jira.ets.berkeley.edu/jira/browse/CLC-6174

Changes:
- Alters validation tests for preferred name
- Ditches tests that will hit "preferred" flag validation errors
- Uses basic auth instead of CAS authentication so that any type of UID can be tested


johncrossman:
Preferred!

PR #4815, opened 2016-01-29
SISRP-13442 - UX DESIGN: View-As Card with Names for Advising Pt 1

winstonkim:
https://jira.berkeley.edu/browse/SISRP-13442


johncrossman:
Thank you, @winstonkim  Such improvements for view-as card are overdue. @christianvuerings / @winstonkim I'm concerned about doing work this on `master` and risking conflicts in the mega-merge. How about redirecting this PR to `sis-master`, side-by-side with the ongoing Delegated Access work.

christianvuerings:
@johncrossman Do you recall when when is the mega merge planned? Also apologize to @winstonkim since I asked him to make it to master.

src/assets/templates/widgets/toolbox/view_as.html
```diff
@@ -28,17 +33,38 @@
...
+            <table>
+              <thead>
+                <tr>
+                  <th width="20%" scope="col">
+                    UID
+                  </th>
+                  <th width="20%" scope="col">
+                    SID
+                  </th>
+                  <th width="60%" scope="col">
```
christianvuerings: @winstonkim We should actually add a visuallyhidden span in here with something like `Delete` or `Save`

christianvuerings:
lgtm but we should make sure we're addressing @johncrossman's concern - probably a PR against `sis-master` would be better for now. /cc @winstonkim

johncrossman:
@christianvuerings mega-merge is scheduled for Wed Feb 3. The only concern is the GoLive 4.1 deploy on Feb 8. I'm hoping to stick w/ planned mega-merge date.

christianvuerings:
@johncrossman thanks
@winstonkim would you be able to address the 1 comment and make the PR against `sis-master`?

src/assets/templates/widgets/toolbox/view_as.html
winstonkim: @christianvuerings sorry, could you clarify which line needs the visuallyhidden span? can't seem to find the line that needs it

christianvuerings: @winstonkim inside the `th`, we should not have any empty headers in a table (even though you won't see it, only hear it).

So something like:
```
<th width="60%" scope="col">
  <span class="cc-visuallyhidden">Save</span>
</th>
```
(However this will probably need to be dynamic, I'll leave that up to you ;) )

winstonkim:
@christianvuerings I've made the changes you requested and I think this PR can be closed. I've opened a new PR to sis-master here: https://github.com/ets-berkeley-edu/calcentral/pull/4817

PR #4817, opened 2016-01-30
SISRP-13442 - UX DESIGN: View-As Card with Names for Advising Pt 1

winstonkim:
https://jira.berkeley.edu/browse/SISRP-13442


christianvuerings:
Lgtm - thanks for making the quick updates. I would also suggest creating a new jira (and link it to the old one) to add in the names.

winstonkim:
@christianvuerings sure, i've created a new ticket here: https://jira.berkeley.edu/browse/SISRP-14135

What do you mean by add in the names? Should I make a new branch corresponding to this ticket number instead of the UX jira ticket?

christianvuerings:
@winstonkim right now we've mainly updated the design, added a link and did some accessibility fixes.
In that new PR, we would add the student's/person's name to the table. In order to do that we'll need to tweak the back-end a bit.

PR #4820, opened 2016-02-01
SISRP-13247 Add enrollment card dependencies to API verify script

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-13247


johncrossman:
Looks good, @paulkerschen  I think we need a conditional in the script which runs those tests when feature flag is true. Future Jira...

PR #4822, opened 2016-02-02
NO JIRA, merge master to sis-master on 2016-02-01 

johncrossman:
Just because.


paulkerschen:
Keepin' it tidy, @johncrossman

PR #4823, opened 2016-02-02
SISRP-14236, act-as pattern for new advisor use case

johncrossman:
https://jira.berkeley.edu/browse/SISRP-14236

Following the pattern established by delegate-view-as. I created a distinct advisor-act-as controller and session variable because overloading the delegate-act-as logic would be prone to regressions.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -129,10 +129,12 @@ Calcentral::Application.routes.draw do
   end
 
   # Act-as endpoints
-  post '/delegate_act_as' => 'delegate_act_as#start', :via => :post
-  post '/stop_delegate_act_as' => 'delegate_act_as#stop', :via => :post
-  post '/act_as' => 'act_as#start', :via => :post
-  post '/stop_act_as' => 'act_as#stop', :via => :post
+  post '/advisor_act_as' => 'advisor_act_as#start'
+  post '/stop_advisor_act_as' => 'advisor_act_as#stop'
+  post '/delegate_act_as' => 'delegate_act_as#start'
+  post '/stop_delegate_act_as' => 'delegate_act_as#stop'
+  post '/act_as' => 'act_as#start'
+  post '/stop_act_as' => 'act_as#stop'
   get '/stored_users' => 'stored_users#get', :via => :get, :defaults => { :format => 'json' }
   post '/store_user/saved' => 'stored_users#store_saved_uid', via: :post, defaults: { format: 'json' }
   post '/delete_user/saved' => 'stored_users#delete_saved_uid', via: :post, defaults: { format: 'json' }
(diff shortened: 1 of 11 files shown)


johncrossman:
One happy Travis.

config/routes.rb
```diff
@@ -129,10 +129,12 @@
   end
 
   # Act-as endpoints
-  post '/delegate_act_as' => 'delegate_act_as#start', :via => :post
-  post '/stop_delegate_act_as' => 'delegate_act_as#stop', :via => :post
-  post '/act_as' => 'act_as#start', :via => :post
-  post '/stop_act_as' => 'act_as#stop', :via => :post
+  post '/advisor_act_as' => 'advisor_act_as#start'
```
christianvuerings: @johncrossman Although this makes sense for delegates to have a different URL here, will that also be the case with advisors? Is there perhaps a way we can tell on the `/act_as` endpoint whether they are a CalCentral super user or advisor?

src/assets/javascripts/angular/factories/adminFactory.js
```diff
@@ -64,12 +66,14 @@ angular.module('calcentral.factories').factory('adminFactory', function(apiServi
 
   return {
     actAs: actAs,
+    advisorActAs: advisorActAs,
```
christianvuerings: @johncrossman this seems to expose just a url in a factory instead of the promise. Also see comment above.

config/routes.rb
johncrossman: @christianvuerings the benefit of a dedicated url path is (1) readability and (2) a dedicated `act_as_authorization` method in the `AdvisorActAsController` subclass which defines access rule. I don't see a lot of maintenance overhead – can you elaborate?

src/assets/javascripts/angular/factories/adminFactory.js
johncrossman: Yeah, whoops! Fixed. (See my reply above to reusability concern.)

config/routes.rb
christianvuerings: @johncrossman makes sense, I was hoping to re-use the current act as widget that we currently use for admins. We could also do the check in the front-end, if you're an admin, use /act_as. If you're an advisor use /advisor_act_as. So we should be good.

christianvuerings:
Lgtm

config/routes.rb
johncrossman: Yes, that is the plan, @christianvuerings  I.e., re-use the widget.

spec/policies/authentication_state_spec.rb
```diff
@@ -137,6 +144,18 @@
...
       end
     end
+    context 'when in advisor-view-as mode' do
+      let(:user_id) { random_id }
+      let(:fake_session) {{
+        'user_id' => user_id,
+        'original_advisor_user_id' => random_id
+      }}
+      before do
+      end
```
paulkerschen: @johncrossman Empty block

paulkerschen:
Thanks @johncrossman - this all looks very good to me; just two stray lines above.

johncrossman:
Thanks @paulkerschen Gone is the empty block.

PR #4826, opened 2016-02-02
SISRP-13260 - Build "Students" card in delegate user's default view ("Toolbox")

dfkaye:
https://jira.berkeley.edu/browse/SISRP-13260

WIP: 
+ `src/assets/javascripts/angular/controllers/widgets/delegateStudentsController.js` contains temporary logic to set a `student.fakeProfilePhotoAPI`
+ `src/assets/templates/widgets/toolbox/delegate_students.html` uses the temporary `student.fakeProfilePhotoAPI`

### DIFF
--- a/src/assets/javascripts/angular/controllers/widgets/delegateStudentsController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/delegateStudentsController.js
@@ -11,37 +11,66 @@ angular.module('calcentral.controllers').controller('DelegateStudentsController'
     isLoading: true
   };
 
-  var linkablePrivileges = [
+  /**
+   * `delegateAccessPrivileges` here mean __only__ those that allow the delegate
+   * "linkable" access to the student's profile. The `phone` privilege by itself
+   * does not grant that access.
+   */
+  var delegateAccessPrivileges = [
     'financial',
+    'phone',
     'viewEnrollments',
     'viewGrades'
   ];
 
   /**
-   * TEMPORARY LOGIC ~ TO BE REPLACED BY BACK-END PROPERTY SETTING.
-   * setFakeStudentPhotoAPI() adds `profilePhoto` property as a fictious API for
-   * retrieving a delegate student's profile photo.
+   * setDelegateAccess() adds `delegateAccess` property on student, if and only
+   * if the student has granted at least one `viewable` (i.e., other than phone)
+   * privilege.
    */
-  var setFakeStudentPhotoAPI = function(student) {
-    student.fakeProfilePhotoAPI = '/api/student/photo/' + student.uid;
+  var setDelegateAccess = function(student) {
+    var phone = 'phone';
+    var viewable = _.some(delegateAccessPrivileges, function(key) {
+      return student.privileges[key] && key !== phone;
+    });
+
+    student.delegateAccess = viewable;
+
+    /**
+     * If at least one student grants no privileges, this flag let's us show the
+     * global 'No privileges' explanatory paragraph in the template.
+     */
+    if (!viewable && !$scope.showNoPrivilegesMessage && !student.privileges[phone]) {
+      $scope.showNoPrivilegesMessage = true;
+    }
   };
 
   /**
-   * setStudentViewability() adds `viewable` property on student, if and only if student
-   * has granted at least one delegate viewing privilege.
+   * TODO:
+   * REPLACE THIS PROPERTY-SETTING METHOD WITH BACK-END PROPERTY SETTING ON THE
+   * STUDENT OBJECT.
+   *
+   * setStudentPhotoApi() adds the `profilePhotoUrl` property for a fictitious
+   * API to retrieve a delegate student's profile photo.
+   *
+   * For testing purposes, you can commment out the single assignment to display
+   * the img-not-available case in the template.
    */
-  var setStudentViewability = function(student) {
-    var viewable = _.some(linkablePrivileges, function(key) {
-      return student.privileges[key];
-    });
-    student.viewable = viewable;
+  var setStudentPhotoApi = function(student) {
+    student.profilePhotoApi = '/api/student/photo/' + student.uid;
   };
 
   var getStudents = function() {
     return delegateFactory.getStudents().then(function(data) {
       angular.extend($scope, _.get(data, 'data.feed'));
-      _.each($scope.students, setFakeStudentPhotoAPI);
-      _.each($scope.students, setStudentViewability);
+
+      /**
+       * TODO: replace setStudentPhotoApi() function with back-end photo URL
+       * property (see above).
+       */
+      _.each($scope.students, setStudentPhotoApi);
+
+      _.each($scope.students, setDelegateAccess);
       $scope.delegateStudents.isLoading = false;
     });
   };
--- a/src/assets/javascripts/angular/controllers/widgets/delegateStudentsController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/delegateStudentsController.js
@@ -11,9 +11,37 @@ angular.module('calcentral.controllers').controller('DelegateStudentsController'
     isLoading: true
   };
 
+  var linkablePrivileges = [
+    'financial',
+    'viewEnrollments',
+    'viewGrades'
+  ];
+
+  /**
+   * TEMPORARY LOGIC ~ TO BE REPLACED BY BACK-END PROPERTY SETTING.
+   * setFakeStudentPhotoAPI() adds `profilePhoto` property as a fictious API for
+   * retrieving a delegate student's profile photo.
+   */
+  var setFakeStudentPhotoAPI = function(student) {
+    student.fakeProfilePhotoAPI = '/api/student/photo/' + student.uid;
+  };
+
+  /**
+   * setStudentViewability() adds `viewable` property on student, if and only if student
+   * has granted at least one delegate viewing privilege.
+   */
+  var setStudentViewability = function(student) {
+    var viewable = _.some(linkablePrivileges, function(key) {
+      return student.privileges[key];
+    });
+    student.viewable = viewable;
+  };
+
   var getStudents = function() {
     return delegateFactory.getStudents().then(function(data) {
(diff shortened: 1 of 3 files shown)


src/assets/javascripts/angular/controllers/widgets/delegateStudentsController.js
```diff
@@ -11,9 +11,37 @@ angular.module('calcentral.controllers').controller('DelegateStudentsController'
...
+  var linkablePrivileges = [
+    'financial',
+    'viewEnrollments',
+    'viewGrades'
+  ];
+
+  /**
+   * TEMPORARY LOGIC ~ TO BE REPLACED BY BACK-END PROPERTY SETTING.
+   * setFakeStudentPhotoAPI() adds `profilePhoto` property as a fictious API for
+   * retrieving a delegate student's profile photo.
```
johncrossman: @dfkaye Minor requests:
1. I'd use "TODO:" with instructions for future dev work
2. I'd remove "fake" from variable and method names. There's a risk that the naming will remain even after the permanent logic lands.
3. "fictitious" instead of "fictious" :)

src/assets/javascripts/angular/controllers/widgets/delegateStudentsController.js
```diff
@@ -11,9 +11,37 @@ angular.module('calcentral.controllers').controller('DelegateStudentsController'
     isLoading: true
   };
 
+  var linkablePrivileges = [
```
johncrossman: I'd rename this `delegatePrivileges` in the interest of self-documenting code.

src/assets/javascripts/angular/controllers/widgets/delegateStudentsController.js
```diff
@@ -11,9 +11,37 @@ angular.module('calcentral.controllers').controller('DelegateStudentsController'
...
+    'financial',
+    'viewEnrollments',
+    'viewGrades'
```
johncrossman: phone?

src/assets/templates/widgets/toolbox/delegate_students.html
```diff
@@ -3,12 +3,38 @@
...
   <div class="cc-widget-text" data-cc-spinner-directive="delegateStudents.isLoading">
-    <ul>
-      <li data-ng-repeat="student in students">
-        <div>
-          <button class="cc-button-link" data-ng-bind="student.name" data-ng-click="delegateStudents.actAs(student.emplid)"></button>
-        </div>
-      </li>
-    </ul>
+    <div id="cc_toolbox_widget_delegate_students" data-ng-cloak data-ng-if="students.length">
+      <ul data-cc-widget-toolbox-delegate-students-list>
```
johncrossman: we seem to be missing `class="..."`

src/assets/templates/widgets/toolbox/delegate_students.html
```diff
@@ -3,12 +3,38 @@
...
+        <li class="cc-flex cc-flex-align-center" data-ng-repeat="student in students"
+        data-ng-if="student.viewable || student.privileges.phone">
```
johncrossman: This conditional is confusing because all students should be rendered. I.e., there is no use case in which we skip a student in the list.

src/assets/templates/widgets/toolbox/delegate_students.html
```diff
@@ -3,12 +3,38 @@
...
+          <div class="cc-toolbox-widget-delegate-student-picture">
+
+            <!-- WIP: photo display/no-display logic TO BE REPLACED -->
```
johncrossman: I'd use TODO instead of WIP and give instructions on future implementation.

src/assets/templates/widgets/toolbox/delegate_students.html
```diff
@@ -3,12 +3,38 @@
...
+              <img data-cc-image-loaded-directive="profilePictureLoading" alt="{{student.fullName}}'s Profile Picture" data-ng-src="{{student.fakeProfilePhotoAPI}}" width="72" height="96" border="1">
+            </div>
+            <div data-ng-if="!student.fakeProfilePhotoAPI" class="cc-toolbox-widget-delegate-student-picture-not-available">
+              <i class="fa fa-user"><!-- WIP: pending new default/fallback image --></i>
+            </div>
+            <!-- END photo display/no-display logic ... -->
+
+          </div>
+          <div class="cc-widget-padding">
+            <button class="cc-button-link" data-ng-click="delegateStudents.actAs(student.emplid)" data-ng-if="student.viewable"><span class="cc-text-big" data-ng-bind="student.fullName"></span></button>
```
johncrossman: We still want to render student name when `!student.viewable`. Won't your `data-ng-if` hide button _and_ fullName?

src/assets/templates/widgets/toolbox/delegate_students.html
```diff
@@ -3,12 +3,38 @@
...
+            </div>
+          </div>
+        </li>
+      </ul>
+      <hr>
+      <div data-cc-toolbox-widget-delegate-students-adding-others>
+        <strong>Adding other students:</strong> If you are a delegate for other students, to add them here please start with the email message you received for each student.
+      </div>
+    </div>
+    <div data-cc-toolbox-widget-delegate-students-list-empty data-ng-if="!students.length">
```
johncrossman: Need `class="..."`.  And should we be using shorter class names?

johncrossman:
Thanks @dfkaye  I've dropped in comments but I'd like @christianvuerings to have a pass, too.

src/assets/templates/widgets/toolbox/delegate_students.html
```diff
@@ -3,12 +3,38 @@
...
   </div>
   <div class="cc-widget-text" data-cc-spinner-directive="delegateStudents.isLoading">
-    <ul>
-      <li data-ng-repeat="student in students">
-        <div>
-          <button class="cc-button-link" data-ng-bind="student.name" data-ng-click="delegateStudents.actAs(student.emplid)"></button>
-        </div>
-      </li>
-    </ul>
+    <div id="cc_toolbox_widget_delegate_students" data-ng-cloak data-ng-if="students.length">
```
christianvuerings: @dfkaye do we need the `data-ng-cloak` here? Especially since it looks like this is wrapped in a spinner.

src/assets/templates/widgets/toolbox/delegate_students.html
```diff
@@ -3,12 +3,38 @@
...
+      <ul data-cc-widget-toolbox-delegate-students-list>
+        <li class="cc-flex cc-flex-align-center" data-ng-repeat="student in students"
+        data-ng-if="student.viewable || student.privileges.phone">
+          <div class="cc-toolbox-widget-delegate-student-picture">
+
+            <!-- WIP: photo display/no-display logic TO BE REPLACED -->
+            <div data-ng-if="student.fakeProfilePhotoAPI">
+              <img data-cc-image-loaded-directive="profilePictureLoading" alt="{{student.fullName}}'s Profile Picture" data-ng-src="{{student.fakeProfilePhotoAPI}}" width="72" height="96" border="1">
```
christianvuerings: @dfkaye could we use `data-ng-attr-alt` instead of `alt`?
https://docs.angularjs.org/guide/interpolation#-ngattr-for-binding-to-arbitrary-attributes

src/assets/templates/widgets/toolbox/delegate_students.html
christianvuerings: @dfkaye would be easier to read if we have `<span class="cc-text-big" data-ng-bind="student.fullName"></span>` on a separate line.

src/assets/templates/widgets/toolbox/delegate_students.html
christianvuerings: @johncrossman fyi long classnames should be fine, but yes, we do need add a `class=` if we want to style this.

src/assets/templates/widgets/toolbox/delegate_students.html
```diff
@@ -3,12 +3,38 @@
...
+            <button class="cc-button-link" data-ng-click="delegateStudents.actAs(student.emplid)" data-ng-if="student.viewable"><span class="cc-text-big" data-ng-bind="student.fullName"></span></button>
+            <div data-ng-if="!student.viewable && student.privileges.phone">
+              <strong class="cc-text-big" data-ng-bind="student.fullName"></strong>
+              <div><i class="cc-text-small">(TBD person-or-phone-only access message)</i></div>
+            </div>
+          </div>
+        </li>
+      </ul>
+      <hr>
+      <div data-cc-toolbox-widget-delegate-students-adding-others>
```
christianvuerings: @dfkaye same here, do we need a `class=` instead?

johncrossman:
Lgtm. @christianvuerings : please review.

src/assets/javascripts/angular/controllers/widgets/delegateStudentsController.js
```diff
@@ -11,9 +11,66 @@ angular.module('calcentral.controllers').controller('DelegateStudentsController'
...
+
   var getStudents = function() {
     return delegateFactory.getStudents().then(function(data) {
       angular.extend($scope, _.get(data, 'data.feed'));
+
+      /**
+       * TODO: replace setStudentPhotoApi() function with back-end photo URL
+       * property (see above).
+       */
+      _.each($scope.students, setStudentPhotoApi);
```
christianvuerings: @dfkaye not a merge blocker but it might be good to do the following optimization later.

We should change the `$scope` only when we really need to.
So in this case it might be better to have something like:
```
var students = _.get(data, 'data.feed.students');
_.each(students, setStudentPhotoApi);
_.each(students, setDelegateAccess);
$scope.students = students;
```

This way we don't update each student as part of the `$scope` every time. There might be DOM changes on each or a `$watch` on each change.

dfkaye: @christianvuerings - very good suggestion - I'll do that

PR #4828, opened 2016-02-02
SISRP-14262, advisors get Toolbox; clean up data-ng-if statements in toolbox.html

johncrossman:
https://jira.berkeley.edu/browse/SISRP-14262
* move `has_toolbox_tab?` to `User::Api` because logic needs user roles 
* `has_toolbox_tab?` is false if user is in view-as mode 
* delegate users lose status when feature flag is false
* clean up conditionals in `toolbox.html`


paulkerschen:
Thanks @johncrossman; that much cleaner.

PR #4829, opened 2016-02-02
CLC-5935 Reload YAML settings in Torquebox message processors

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-5935

This doesn't solve the problem of accessing the web subsystem across multiple nodes, but I'm curious to see if it works with message processors across the cluster.


johncrossman:
Whew. The feature has been realized! Thank you @paulkerschen

PR #4830, opened 2016-02-03
SISRP-14262, Advisor gets standard view-as widgets with restricted access to user data

johncrossman:
https://jira.berkeley.edu/browse/SISRP-14262
* Created generic `can_view_as_for_all_uids?` for advisor and users w/ the more powerful `can_view_as?`.
* remove `delegateActingAsController.js` in favor of general purpose `actingAsController.js`
* for now, advisorActingAsUid gets same restricted access of delegateActingAsUid. I realize there is a need for refactoring in the templates (e.g., expression below) but this PR is not the place.
```
data-ng-show="!api.user.profile.delegateActingAsUid && !api.user.profile.advisorActingAsUid && (controllerName === 'DashboardController')"
``` 
* custom footer for advisor-view-as mode


christianvuerings:
@johncrossman thank you - as a FYI - looks like this will be coming up soon:
https://jira.berkeley.edu/browse/SISRP-14214

/cc @paulkerschen @raydavis

johncrossman:
Jshint prob. I'll fix

src/assets/templates/profile.html
```diff
@@ -1,4 +1,4 @@
-<div class="cc-page-profile" data-ng-if="api.user.profile.showSisProfileUI && !api.user.profile.delegateActingAsUid">
+<div class="cc-page-profile" data-ng-if="api.user.profile.showSisProfileUI && !api.user.profile.delegateActingAsUid && !api.user.profile.advisorActingAsUid">
```
christianvuerings: @johncrossman advisors actually do need to see the profile but we need to hide certain elements - see https://jira.berkeley.edu/browse/SISRP-14283

src/assets/templates/footer.html
```diff
@@ -1,5 +1,5 @@
 <footer class="row" data-ng-controller="FooterController" data-ng-cloak role="complementary">
-  <div data-ng-class="{'cc-footer-acting-as':(api.user.profile.actingAsUid || api.user.profile.delegateActingAsUid)}">
+  <div data-ng-class="{'cc-footer-acting-as':(api.user.profile.actingAsUid || api.user.profile.delegateActingAsUid || api.user.profile.advisorActingAsUid)}">
```
christianvuerings: @johncrossman not a blocker, but it would be so nice if we could have something like `isActingAs` being returned in the feed.

src/assets/templates/profile.html
johncrossman: I've moved SISRP-14283 (assigned to me) into the sprint. I'll revise per requirements in future PR. Okay?

src/assets/templates/footer.html
johncrossman: I totally agree. (See my comment in description of this PR.) I've created a Jira to follow your advice: https://jira.berkeley.edu/browse/SISRP-14285

src/assets/templates/profile.html
christianvuerings: @johncrossman sounds good

src/assets/templates/footer.html
christianvuerings: @johncrossman makes sense, thanks

PR #4834, opened 2016-02-03
SISRP-14107 - fix whitespace issue in QA FinAid test

pfarestveit:
https://jira.berkeley.edu/browse/SISRP-14107


johncrossman:
One outta two ain't bad.

PR #4836, opened 2016-02-03
SISRP-13947 Add academic careers and student ID to enrollment terms proxy

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-13947

Adjusting proxy behavior to reflect data coming from a real endpoint.

Note that this proxy is no longer a straight passthrough; the `ACAD_CAREER` element is being denormalized to individual terms, as I think this will make things easier for the front end.

### DIFF
--- a/app/models/campus_solutions/enrollment_terms.rb
+++ b/app/models/campus_solutions/enrollment_terms.rb
@@ -10,9 +10,17 @@ module CampusSolutions
     end
 
     def build_feed(response)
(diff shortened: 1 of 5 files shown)


christianvuerings:
@paulkerschen lgtm and indeed, the payload looks ways more consumable in the front-end which we can use for the College Scheduler link. Thank you!

johncrossman:
yes, thank you @paulkerschen  Waiting on a happy T

PR #4837, opened 2016-02-04
SISRP-14283, filter out EFC when advisor-view-as

johncrossman:
https://jira.berkeley.edu/browse/SISRP-14283

FinancialAidDataController invokes filtered feed, which excludes 'Expected Family Contribution' and similar, when advisor-view-as mode.

### DIFF
--- a/app/models/campus_solutions/my_financial_aid_filtered_for_advisor.rb
+++ b/app/models/campus_solutions/my_financial_aid_filtered_for_advisor.rb
@@ -1,0 +1,55 @@
+module CampusSolutions
+  class MyFinancialAidFilteredForAdvisor < UserSpecificModel
+
+    include ClassLogger
+    include Cache::CachedFeed
+    include Cache::JsonAddedCacher
+    include Cache::RelatedCacheKeyTracker
+    include CampusSolutions::FinaidFeatureFlagged
+
+    attr_accessor :aid_year
+
+    def get_feed_as_json(force_cache_write=false)
+      feed = get_feed force_cache_write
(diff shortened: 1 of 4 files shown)


app/models/campus_solutions/my_financial_aid_filtered_for_advisor.rb
```diff
@@ -0,0 +1,55 @@
...
+      end
+      feed
+    end
+
+    def instance_key
+      "#{@uid}-#{aid_year}"
+    end
+
+    def has_family_information(item)
+      [item[:title].to_s, item[:value].to_s].any? { |s| s =~ /family/i || s =~ /efc/i || s =~ /parent/i }
```
paulkerschen: @johncrossman Very minor, but it would probably be quicker to instantiate a single regexp for this: `s =~ /family|efc|parent/i` if it's not too hard to read that way?

app/models/campus_solutions/my_financial_aid_filtered_for_advisor.rb
```diff
@@ -0,0 +1,55 @@
...
+        end
+      end
+      feed
+    end
+
+    def instance_key
+      "#{@uid}-#{aid_year}"
+    end
+
+    def has_family_information(item)
```
paulkerschen: @johncrossman Add a question mark to this method name as per convention?

johncrossman:
Mo' better and conciser. Amended @paulkerschen

paulkerschen:
Looks good, @johncrossman - will merge on green T

PR #4838, opened 2016-02-04
SISRP-14422, When advisor-view-as mode, show Profile but hide Demographics tab

johncrossman:
https://jira.berkeley.edu/browse/SISRP-14422

Advisor-view-as mode can get to /profile but will not see student's:
* demographics
* bConnected
* "Assign a delegate" call-to-action

I'm going to add *delegateActingAsUid* references to the `profileMenuService.js` logic just in case we decide to show limited profile to Delegates.

### DIFF
--- a/src/assets/javascripts/angular/services/profileMenuService.js
+++ b/src/assets/javascripts/angular/services/profileMenuService.js
@@ -39,7 +39,8 @@ angular.module('calcentral.services').factory('profileMenuService', function(api
           featureFlag: 'csDelegatedAccess',
           roles: {
             student: true
-          }
+          },
+          hideWhenActAsModes: ['advisorActingAsUid']
         },
         {
           id: 'information-disclosure',
@@ -77,7 +78,8 @@ angular.module('calcentral.services').factory('profileMenuService', function(api
       categories: [
         {
           id: 'bconnected',
-          name: 'bConnected'
+          name: 'bConnected',
+          hideWhenActAsModes: ['advisorActingAsUid']
         }
       ]
     }
@@ -149,6 +151,30 @@ angular.module('calcentral.services').factory('profileMenuService', function(api
     return defer(navigation, filterFeatureFlagsInNavigation);
   };
 
+  var filterActAsInCategory = function(categories) {
+    var userProfile = apiService.user.profile;
+    return _.filter(categories, function(category) {
+      if (!_.get(category, 'hideWhenActAsModes.length')) {
+        return true;
+      } else {
+        return !_.some(category.hideWhenActAsModes, function(hideWhenActAsMode) {
+          return userProfile[hideWhenActAsMode];
+        });
+      }
+    });
+  };
+
+  var filterActAsInNavigation = function(navigation) {
+    return filterCategories(navigation, filterActAsInCategory);
+  };
+
+  /**
+   * Filter based on active view-as mode (admin, advisor, delegate, etc.)
+   */
+  var filterActAs = function(navigation) {
+    return defer(navigation, filterActAsInNavigation);
+  };
+
   var filterEmptyInNavigation = function(navigation) {
     return _.filter(navigation, function(item) {
       return !!_.get(item, 'categories.length');
@@ -174,6 +200,7 @@ angular.module('calcentral.services').factory('profileMenuService', function(api
       .then(initialNavigation)
       .then(filterRoles)
       .then(filterFeatureFlags)
+      .then(filterActAs)
       .then(filterEmpty);
   };
 
--- a/src/assets/templates/widgets/profile/demographic.html
+++ b/src/assets/templates/widgets/profile/demographic.html
@@ -5,9 +5,9 @@
 
   <div data-ng-if="!demographicInformation.isErrored">
     <!-- Editable soon -->
-    <div data-ng-include="'widgets/profile/editable_soon.html'"></div>
+    <div data-ng-include="'widgets/profile/editable_soon.html'" data-ng-if="!api.user.profile.advisorActingAsUid"></div>
     <!-- Gender -->
-    <div data-ng-include="'widgets/profile/demographic_gender.html'"></div>
+    <div data-ng-include="'widgets/profile/demographic_gender.html'" data-ng-if="!api.user.profile.advisorActingAsUid"></div>
     <!-- Ethnicity -->
     <div data-ng-include="'widgets/profile/demographic_ethnicity.html'"></div>
     <!-- Veteran Status -->
--- a/src/assets/templates/profile.html
+++ b/src/assets/templates/profile.html
@@ -1,4 +1,4 @@
-<div class="cc-page-profile" data-ng-if="api.user.profile.showSisProfileUI && !api.user.profile.delegateActingAsUid && !api.user.profile.advisorActingAsUid">
+<div class="cc-page-profile" data-ng-if="api.user.profile.showSisProfileUI && !api.user.profile.delegateActingAsUid">
 
   <div class="medium-9 medium-push-3 columns cc-sidebar-content">
     <h1 class="cc-heading-page-title" data-ng-bind="header"></h1>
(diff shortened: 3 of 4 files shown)


christianvuerings:
@johncrossman An advisor does actually need to see the demographics tab (Gender of Record / Ethnicity & Veteran status). I know the matrix might be confusing but it specifically lists all the elements we need to filter out on the back-end side (we need to make sure this data doesn't come across the wire).

src/assets/javascripts/angular/services/profileMenuService.js
```diff
@@ -26,7 +26,8 @@ angular.module('calcentral.services').factory('profileMenuService', function(api
         },
         {
           id: 'demographic',
-          name: 'Demographic Information'
+          name: 'Demographic Information',
+          hideWhenActAsModes: [ 'advisorActingAsUid', 'delegateActingAsUid' ]
```
christianvuerings: @johncrossman I don't understand the `delegateActingAsUid` in here, shouldn't the profile be completely hidden in that case? (see matrix)

src/assets/javascripts/angular/services/profileMenuService.js
```diff
@@ -39,7 +40,8 @@ angular.module('calcentral.services').factory('profileMenuService', function(api
           featureFlag: 'csDelegatedAccess',
           roles: {
             student: true
-          }
+          },
+          hideWhenActAsModes: [ 'advisorActingAsUid', 'delegateActingAsUid' ]
```
christianvuerings: @johncrossman we also (should) never have space after `[`, not sure whether we currently have a linting rule for that.

src/assets/javascripts/angular/services/profileMenuService.js
johncrossman: Yes, that "completely hide profile from delegates" logic is in place, @christianvuerings. I added `delegateActingAsUid` here as safety net but I'll remove given slight performance concern.

christianvuerings:
The most important piece of this change should be to filter out the data from the back-end on `api/edos/student` and perhaps `api/my/status`.

src/assets/javascripts/angular/services/profileMenuService.js
johncrossman: @christianvuerings Fixing `jscs` error now.

johncrossman:
Thanks @christianvuerings  The server-side api filtering will be verified in https://jira.berkeley.edu/browse/SISRP-14423

src/assets/javascripts/angular/services/profileMenuService.js
```diff
@@ -149,6 +151,30 @@ angular.module('calcentral.services').factory('profileMenuService', function(api
     return defer(navigation, filterFeatureFlagsInNavigation);
   };
 
+  var filterActAsInCategory = function(categories) {
+    return _.filter(categories, function(category) {
+      if (!category.hideWhenActAsModes) {
```
christianvuerings: @johncrossman very small suggestion here since we're comparing against an array, could we check for the `length` property. However that'll result in errors since `category.hideWhenActAsModes` might not exist.

How about `if (!_.get(category, 'hideWhenActAsModes.length')) {`?

johncrossman:
Okay, I've created separate Jira to make sure no profile data is editable by advisor: https://jira.berkeley.edu/browse/SISRP-14424. This PR is ready for another review. cc: @christianvuerings

src/assets/javascripts/angular/services/profileMenuService.js
johncrossman: +1

src/assets/javascripts/angular/services/profileMenuService.js
```diff
@@ -149,6 +151,30 @@ angular.module('calcentral.services').factory('profileMenuService', function(api
...
+        return true;
+      } else {
+        var userProfile = apiService.user.profile;
```
christianvuerings: @johncrossman even though we only use it in this else section, could we move this to the beginning of the `filterActAsInCategory` function?

```
var filterActAsInCategory = function(categories) {
var userProfile = apiService.user.profile;
```

That way the JS runtime doesn't need to create the `var` every time it runs over this `_filter` loop.

src/assets/templates/widgets/profile/demographic.html
```diff
@@ -5,9 +5,9 @@
 
   <div data-ng-if="!demographicInformation.isErrored">
     <!-- Editable soon -->
-    <div data-ng-include="'widgets/profile/editable_soon.html'"></div>
+    <div data-ng-include="'widgets/profile/editable_soon.html'" data-ng-if="!api.user.profile.advisorActingAsUid"></div>
     <!-- Gender -->
-    <div data-ng-include="'widgets/profile/demographic_gender.html'"></div>
+    <div data-ng-include="'widgets/profile/demographic_gender.html'" data-ng-if="!api.user.profile.advisorActingAsUid"></div>
```
christianvuerings: @johncrossman which gender are we currently getting through from the Hub? We still need to the show the official current gender (M/F) but not the `Sex at Birth` / `Gender Identity` and `Sexual Orientation`.

christianvuerings:
Lgtm after the gender question has been addressed.
Thanks @johncrossman for the super quick follow up on those questions!

src/assets/templates/widgets/profile/demographic.html
johncrossman: Yes, I have the matrix at my side. :) In SISRP-14423, I will filter out on server-side:
* sexAtBirth
* genderIdentity

As far as I know - and grep'ing - we're getting sexual orientation from Hub/CS.

johncrossman: *Typo correction*: we are NOT getting sexual orientation from Hub/CS

christianvuerings: @johncrossman interesting, so we're not getting the current gender in the API, might be good to create a new jira and assign it to Adam C if that's the case.

johncrossman: @christianvuerings  We are getting both `genderOfRecord` and `genderIdentity`. The former does not need to be filtered out.

paulkerschen: @christianvuerings Perhaps because we aren't calling the Hub /gender endpoint (https://jira.berkeley.edu/browse/SISRP-14049)

christianvuerings: @johncrossman indeed, so we probably shouldn't completely hide `widgets/profile/demographic_gender.html` in this case, just on the pieces we need to filter out.

I'm ok if we do this in a follow up PR.

johncrossman: We do not call `/gender`. We get the following from `/all`:
```
          "gender": {
            "genderOfRecord": {
              "code": "M",
              "description": "Male"
            },
            "discloseGenderOfRecord": false,
            "sexAtBirth": {
              "code": "",
              "description": ""
            },
            "discloseSexAtBirth": false,
            "genderIdentity": {
              "code": "",
              "description": ""
            },
            "discloseGenderIdentity": false,
            "fromDate": "1902-02-01"
          },
```

christianvuerings: @johncrossman but at this very second in production, we do neither, correct? We don't call `/all` and we don't call `gender`?

Since that's the case, we'll actually never show the gender to the user, so that's why we have https://jira.berkeley.edu/browse/SISRP-14049 and should try to fix it before GL5.

johncrossman: @christianvuerings It is an oddity. The `url` method of class below is in our code but never invoked. It's overridden by subclasses every time. Fwiw, SISRP-14049 is in the sprint but the "wire up demographics tab" work is implicit, at best. (I've linked to this PR in the Jiras we've referenced.) 
```
  class Student < Proxy

    def url
      "#{@settings.base_url}/#{@campus_solutions_id}/all"
    end
```

paulkerschen:
Looks good to me, @johncrossman. If follow-up is needed on gender, I think it makes sense to do it once we actually have gender information in the feed.

src/assets/templates/widgets/profile/demographic.html
christianvuerings: @johncrossman & @paulkerschen thanks for investigating and working this out.

PR #4839, opened 2016-02-04
SISRP-14142 - In View As mode - Edit / add buttons should be disabled

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-14142

Since there is so much overlap, this also includes

SISRP-14421 - Hide links to CS (classic and fluid) when acting as a delegate or advisor

https://jira.berkeley.edu/browse/SISRP-14421


johncrossman:
Looks good, @christianvuerings   Nice to see this coming together.

PR #4840, opened 2016-02-04
SISRP-13947 Update enrollment card to match API structure

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-13947

Adjustments to reflect the final shape of the enrollment term APIs, which diverge from originally expected behavior.

In particular, the "get all enrollment terms" endpoint currently returns many terms for each student, some of which will return no enrollment information via the "get specific enrollment term" endpoint. This requires a change in front-end behavior: instead of showing individual spinners for each term, we have to keep the main spinner active until all terms have loaded. Otherwise individual term cards would briefly appear with spinners, then disappear after coming back empty.

### DIFF
--- a/app/models/campus_solutions/enrollment_term.rb
+++ b/app/models/campus_solutions/enrollment_term.rb
@@ -7,26 +7,31 @@ module CampusSolutions
 
     def initialize(options = {})
       super options
+      @term_id = options[:term_id]
       initialize_mocks if @fake
     end
 
     def build_feed(response)
-      return {} if response.parsed_response.blank?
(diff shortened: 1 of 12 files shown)


christianvuerings:
@paulkerschen lgtm

Ideally we would ask to only get the terms which actually contain any data for a student. Would you be able to create a ticket for that (assigned to Adam & cc Babu) and add me as a watcher?

For now though, this is fine and should allow us to start testing.

paulkerschen:
@christianvuerings https://jira.berkeley.edu/browse/SISRP-14458 created.

christianvuerings:
@paulkerschen beautiful, thanks!

PR #4849, opened 2016-02-05
SISRP-14284 Back end for College Scheduler redirect

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-14284

College Scheduler URLs are dynamically generated, one use only, and expire after two minutes. For these reasons, instead of serving up a prefetched URL to the front end, we define an endpoint that grabs a new URL and redirects on request.

The CS API is giving us old skool all-caps URLs. Reproduced without comment.


app/controllers/campus_solutions/college_scheduler_controller.rb
```diff
@@ -0,0 +1,15 @@
+module CampusSolutions
+  class CollegeSchedulerController < CampusSolutionsController
+
+    # GET /college_scheduler/:acad_career/:term_id
+    def get
```
johncrossman: @paulkerschen reminder to add new CS URLs to the verify-sis script, when you have chance.

app/models/campus_solutions/college_scheduler_url.rb
```diff
@@ -0,0 +1,29 @@
...
+      response = self.get
+      [:feed, :scheduleplannerssolink, :url].inject(response) { |hash, key| hash[key] if hash }
+    end
+
+    def xml_filename
+      'college_scheduler_url.xml'
+    end
+
+    def url
+      "#{@settings.base_url}/UC_SR_COLLEGE_SCHDLR_URL.v1/get?EMPLID=#{@campus_solutions_id}&STRM=#{@term_id}&ACAD_CAREER=#{@acad_career}&INSTITUTION=UCB01"
```
johncrossman: @paulkerschen (correction on code ref) reminder to add new CS URLs to the verify-sis script, when you have chance.
 Add a line note

johncrossman:
Lgtm @paulkerschen - You can update sis-verify script in future PR.

app/models/campus_solutions/college_scheduler_url.rb
paulkerschen: Thanks @johncrossman; follow-up in #4849.

app/models/campus_solutions/college_scheduler_url.rb
```diff
@@ -0,0 +1,29 @@
+module CampusSolutions
+  class CollegeSchedulerUrl < Proxy
```
christianvuerings: @paulkerschen I'm such a ruby noob but want to be sure, we're not doing any caching on this, correct?

paulkerschen: @christianvuerings Correct, no caching here. In #4761, we removed caching behavior from most of the Campus Solutions proxies and created the CachedProxy subclass for those that needed it.

PR #4850, opened 2016-02-05
SISRP-14284 Enrollment card front-end updates

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-14284

Follows up on #4849 with front-end code, and also an updated SIS verify script.

The changes in the Angular controller are necessary because the plural "enrollment terms" feed includes the required `acadCareer` parameter, but the singular "enrollment term" feed does not. Data from the latter must extend the former rather than replacing it.

### DIFF
--- a/src/assets/templates/widgets/enrollment/schedule.html
+++ b/src/assets/templates/widgets/enrollment/schedule.html
@@ -6,5 +6,8 @@
 ></div>
 
 <div class="cc-enrollment-card-section-content" data-ng-if="section.show">
-  Customized semester scheduling is unavailable at this time.
+  <a data-ng-href="/college_scheduler/{{enrollmentTerm.acadCareer}}/{{enrollmentTerm.term}}" target="_self">
+    Organize and preview your upcoming semester
+  </a>
+  based on your scheduling needs.
 </div>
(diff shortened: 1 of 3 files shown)


johncrossman:
Thanks for sis-verify script change, @paulkerschen

src/assets/templates/widgets/enrollment/schedule.html
```diff
@@ -6,5 +6,8 @@
 ></div>
 
 <div class="cc-enrollment-card-section-content" data-ng-if="section.show">
-  Customized semester scheduling is unavailable at this time.
+  <a data-ng-href="/college_scheduler/{{enrollmentTerm.acadCareer}}/{{enrollmentTerm.term}}" target="_self">
```
christianvuerings: @paulkerschen could we remove the `target="_self"`?

paulkerschen: @christianvuerings Without `target="_self"`, Angular tries to match the route itself rather than sending it back to the Rails controller, and the result is a 404 page.

christianvuerings: @paulkerschen makes sense, let's leave it as is. I'm wondering if it actually opens in a new window now or not. And how it works with our `outboundLinkDirective`

paulkerschen: @christianvuerings Of course, good catch. It won't open in a new window, and I assume it's expected to. If you get a chance, do you want to create a new JIRA for one of us to take this?

christianvuerings: @paulkerschen Saw you created https://jira.berkeley.edu/browse/SISRP-14634 so assigned it to myself.

PR #4853, opened 2016-02-05
SISRP-13173 - Remove Slate checklist code from CalCentral

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-13173

I *think* this should be good...it seems like any Slate checklist code that existed on "collective" files (e.g. `MyTasks::Merged`) were removed in a past PR.


johncrossman:
@davidlimcheng Nice. I feel lighter already! Can you remove `features.slate_checklist` flag from `settings.yml` and elsewhere.

davidlimcheng:
@johncrossman Yep! Also removed the `spec.rb`, and confirmed that flag is nowhere to be found in `other environments.yml`.

christianvuerings:
@davidlimcheng Travis.ci failure seems real.

```
1) MyTasks::Merged should return Google tasks when Canvas service is unavailable
     Failure/Error: expect([GoogleApps::Proxy::APP_ID, Slate::Checklist::APP_ID, CampusSolutions::Proxy::APP_NAME]).to include task[:emitter]
     NameError:
       uninitialized constant Slate
     # ./spec/models/my_tasks/merged_spec.rb:187:in `(root)'
     # ./spec/models/my_tasks/merged_spec.rb:186:in `each'
     # ./spec/models/my_tasks/merged_spec.rb:186:in `(root)'
```

Could you make sure to search for `Slate` and make sure we don't miss any occurrences?

davidlimcheng:
@christianvuerings Yes!  I found an occurence in `merged_spec.rb` and have removed it, hopefully that will do it.

But, I also found this occurence in `dashboard_task_loop.html`:

```
      <div
        class="cc-icon cc-icon-slate"
        data-ng-if="task.emitter=='slate'">
      </div>
```

It looks like since the checklist item will be coming from CS, it looks like we won't be able to visually differentiate between Slate and CS tasks - is this true?  If so, I'll remove this code as well.

christianvuerings:
@davidlimcheng indeed, we should remove that line as well. You actually already removed the code in `slate_tasks.rb` which sets that `slate` emitter.

davidlimcheng:
Sounds good.  @christianvuerings @johncrossman, should be good to go once Travis gives the OK.

src/assets/stylesheets/_icons.scss
```diff
@@ -48,9 +48,6 @@
 .cc-icon-oskicat {
   background-image: url('../images/icon_oskicat_16x16.png');
 }
-.cc-icon-slate {
-  background-image: url('../images/svg/icon_slate_256x256.svg');
```
christianvuerings: @davidlimcheng thanks for spotting these ones as well!

PR #4854, opened 2016-02-08
SISRP-14433, any user, regardless of affiliations, can have delegating students

johncrossman:
https://jira.berkeley.edu/browse/SISRP-14433
https://jira.berkeley.edu/browse/SISRP-14531
https://jira.berkeley.edu/browse/SISRP-14530

Few changes wrapped in one PR:
* move delegate_permissions check out of authentication_state
* all users get delegate_students check in `User::Api`
* `api.user.profile.hasDashboardTab`
* spec changes accordingly

### DIFF
--- a/app/models/my_academics/filtered_for_delegate.rb
+++ b/app/models/my_academics/filtered_for_delegate.rb
@@ -52,7 +52,14 @@ module MyAcademics
     end
 
     def show_grades?
-      delegate_permissions.include? 'View grades'
+      response = CampusSolutions::DelegateStudents.new(user_id: authentication_state.original_delegate_user_id).get
+      if response[:feed] && (students = response[:feed][:students])
+        campus_solutions_id = CalnetCrosswalk::ByUid.new(user_id: @uid).lookup_campus_solutions_id
+        student = students.detect { |s| campus_solutions_id == s[:campusSolutionsId] }
+        student && student[:privileges] && student[:privileges][:viewGrades]
+      else
+        false
+      end
     end
   end
 end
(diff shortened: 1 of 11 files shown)


app/models/my_academics/filtered_for_delegate.rb
```diff
@@ -17,7 +17,8 @@ def self.providers
     end
 
     def get_feed_as_json(force_cache_write=false)
-      if show_grades?
+      @show_grades ||= show_grades?
+      if @show_grades
```
paulkerschen: @johncrossman Why the instance variable here? I don't see that we're using the stored value anywhere (and of course if the stored value is `false`, then `||=` will kick in and call the method anyway).

johncrossman: Thanks, @paulkerschen Amended. A holdover from previous work-in-progress.

app/models/user/api.rb
```diff
@@ -24,8 +24,27 @@ def init
       @given_first_name = (@edo_attributes && @edo_attributes[:given_name]) || @first_name || ''
       @family_name = (@edo_attributes && @edo_attributes[:family_name]) || @last_name || ''
       @student_id = get_campus_attribute('student_id', :numeric_string)
-      delegate_permissions = authentication_state.delegate_permissions
-      @delegate_privileges = delegate_permissions && delegate_permissions[:privileges]
+      @delegate_students = get_delegate_students
+      @delegate_privileges = get_delegate_privileges
```
paulkerschen: @johncrossman I wonder if it's a bit brittle to have two instance variables relating to delegates, one dependent on the other. (If the order of these two assignments is reversed, `@delegate_privileges` will end up a surprise nil.) Is there a structure that would make the dependency clearer at a glance?

johncrossman: @paulkerschen I'd say yes. `@delegate_privileges` is only relevant when delegate is in view-as mode whereas `@delegate_students` is relevant *prior* to view-as. I think I can make that more obvious. Stand by.

johncrossman:
`User::Api` amended. Purpose and context of `delegateViewAsPrivileges` is more obvious, I hope.  cc: @paulkerschen

paulkerschen:
Thanks @johncrossman! Looks good to me.

PR #4855, opened 2016-02-08
SISRP-14515 Enrollment card: back end for academic plan

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-14515

Note that the XML fixture includes an unparseable array attribute for the `ACADEMICPLAN` element, as currently returned by the CS API (bug reported in https://jira.berkeley.edu/browse/SISRP-14585). Still, `feed.updateAcademicPlan.url` is coming through, so we should be able to surface the deeplink in the front end if that is, in fact, a GL5 requirement.


johncrossman:
You can have as many ACADEMICPLANs as you'd like... as long as it's <= 1. Makes sense. Thanks, @paulkerschen

PR #4862, opened 2016-02-09
SISRP-14578 Ensure default aid year for user-specific finaid models

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-14578

A couple of classes were missed in #4747.


johncrossman:
+1

PR #4863, opened 2016-02-09
SISRP-14578 Do not explode on missing SERVICE_INDICATORS element

paulkerschen:
Further log diving on https://jira.berkeley.edu/browse/SISRP-14578 is turning up several unrelated bugs. Here's another one.


johncrossman:
Makes sense. Thanks @paulkerschen

PR #4867, opened 2016-02-09
SISRP-14530, delegate-view-as will not have dashboard so default to next best

johncrossman:
https://jira.berkeley.edu/browse/SISRP-14530

Default home is `/dashboard` but delegates in view-as mode must have default based on privileges.

### DIFF
--- a/src/assets/javascripts/angular/services/userService.js
+++ b/src/assets/javascripts/angular/services/userService.js
@@ -38,10 +38,18 @@ angular.module('calcentral.services').service('userService', function($http, $lo
   /**
    * Redirect user to the dashboard when you're on the splash page
    */
-  var redirectToDashboard = function() {
+  var redirectToHome = function() {
     if ($location.path() === '/') {
       analyticsService.sendEvent('Authentication', 'Redirect to dashboard');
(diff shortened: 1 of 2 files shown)


src/assets/javascripts/angular/services/userService.js
```diff
@@ -38,10 +38,18 @@ angular.module('calcentral.services').service('userService', function($http, $lo
...
-      utilService.redirect('dashboard');
+      if (profile.hasDashboardTab === true) {
+        utilService.redirect('dashboard');
+      } else if (profile.hasAcademicsTab === true) {
+        utilService.redirect('academics');
+      } else if (profile.hasFinancialsTab === true) {
+        utilService.redirect('finances');
+      } else {
+        utilService.redirect('campus');
+      }
```
johncrossman: Is there a better way to do this, @christianvuerings?

christianvuerings: @johncrossman I think this is fine, however I would suggest to remove ` === true` everywhere in this PR.

johncrossman:
Thanks, @christianvuerings  Done.

PR #4868, opened 2016-02-09
SISRP-14653 - UX Develop: Prototype - Send a redirect parameter to the CS logout when a user logs out of CalCentral

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-14653

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -84,9 +84,9 @@ class SessionsController < ApplicationController
   def destroy
     logout
     url = request.protocol + ApplicationController.correct_port(request.host_with_port, request.env['HTTP_REFERER'])
-    if Settings.features.cs_logout
-      url = "#{Settings.campus_solutions_logout.url}&redirect_url=#{CGI.escape url}"
-    end
+
+    url = "#{Settings.campus_solutions_proxy.logout_url}&redirect_url=#{CGI.escape url}" if Settings.features.cs_logout
+
     render :json => {
       :redirectUrl => "#{Settings.cas_logout_url}?service=#{CGI.escape url}"
     }.to_json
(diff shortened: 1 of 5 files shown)


app/controllers/sessions_controller.rb
```diff
@@ -84,6 +84,9 @@ def basic_lookup
   def destroy
     logout
     url = request.protocol + ApplicationController.correct_port(request.host_with_port, request.env['HTTP_REFERER'])
+    if Settings.features.cs_logout
+      url = "#{Settings.campus_solutions_logout.url}&redirect_url=#{CGI.escape url}"
+    end
     render :json => {
```
johncrossman: @christianvuerings  This can be a one-liner:
```
url = "#{Settings.campus_solutions_logout.url}&redirect_url=#{CGI.escape url}" if Settings.features.cs_logout
```

config/settings.yml
```diff
@@ -407,6 +408,9 @@ campus_solutions_proxy:
   username: secret
   password: secret
 
+campus_solutions_logout:
+  url: 'https://bcs-web-dev-03.is.berkeley.edu:8443/psp/bcsdev/EMPLOYEE/HRMS/?cmd=logout'
+
```
johncrossman: @christianvuerings  I suggest putting this under `campus_solutions_proxy` in our YAML because it has same domain/port as `campus_solutions_proxy.base_url`. We do a lot of toggling between bcs-dev and bcs-qat so having all bcs configs in one grouping makes config mgmt easier.

app/controllers/sessions_controller.rb
christianvuerings: @johncrossman thanks - changed it to that instead

config/settings.yml
christianvuerings: @johncrossman I agree and made the update. We do keep the complete URL in there though since it's quite different from the API endpoint.

johncrossman: Thanks @christianvuerings  And yes, having complete URL makes sense.

johncrossman:
Thx @christianvuerings  And away we go...

PR #4869, opened 2016-02-09
SISRP-13892 Make RelatedCacheKeyTracker aware of JSON keys

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-13892

Refactor to make Cache::RelatedCacheKeyTracker work with Cache::JsonAddedCacher. Also changes RelatedCacheKeyTracker to group related keys by uid+class instead of just uid, so that expiring one set of related keys doesn't expire all other related keys for the user.

### DIFF
--- a/lib/cache/json_added_cacher.rb
+++ b/lib/cache/json_added_cacher.rb
@@ -16,6 +16,10 @@ module Cache
       end
     end
 
+    def extended_instance_keys
+      super << self.class.json_key(instance_key)
+    end
+
     module ClassMethods
       def json_key(id)
         "json-#{id}"
--- a/lib/cache/related_cache_key_tracker.rb
+++ b/lib/cache/related_cache_key_tracker.rb
@@ -1,13 +1,20 @@
 module Cache
   module RelatedCacheKeyTracker
 
+    def get_feed(force_cache_write=false)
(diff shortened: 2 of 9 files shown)


paulkerschen:
Half of Travis is still upset about textbooks.

lib/cache/related_cache_key_tracker.rb
```diff
@@ -1,13 +1,20 @@
...
+    end
+
     def self.included base
       base.extend ClassMethods
     end
 
     module ClassMethods
       def user_key(uid)
-        "related-cache-keys-#{uid}"
+        cache_key "related-cache-keys-#{uid}"
```
paulkerschen: This is the bit that adds the class name to the related-key collection.

lib/cache/json_added_cacher.rb
```diff
@@ -16,6 +16,10 @@ def get_feed_as_json(force_cache_write=false)
       end
     end
 
+    def extended_instance_keys
+      super << self.class.json_key(instance_key)
+    end
```
paulkerschen: Written this way to make it easier in the future if we have more work with multiply cached models.

raydavis:
Nice labyrinth threading.

PR #4870, opened 2016-02-09
SISRP-14578 Zealously block empty aid_year from LiveUpdates

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-14578

The wall built in #4747 was leaky. Also, remove needless default zero for aid_year.


johncrossman:
Bonus points for "zealously" in PR summary. And thanks for spec tidying, @paulkerschen

PR #4872, opened 2016-02-10
SISRP-14648 - Enrollment Card - where should 'Discovery Courses' go?

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-14648


paulkerschen:
Where? There!

PR #4877, opened 2016-02-10
SISRP-14767: Show 404 for privilege url when viewing as delegate w/o corresponding privilege

dfkaye:
https://jira.berkeley.edu/browse/SISRP-14767

### DIFF
--- a/src/assets/templates/academics.html
+++ b/src/assets/templates/academics.html
@@ -31,5 +31,5 @@
       <div data-ng-include="'academics_university_requirements.html'" data-ng-if="requirements && isUndergraduate"></div>
     </div>
   </div>
+  <div data-ng-include="'404.html'" data-ng-if="!canViewAcademics"></div>
 </div>
-<div data-ng-include="'404.html'" data-ng-if="!canViewAcademics"></div>
--- a/src/assets/templates/academics.html
+++ b/src/assets/templates/academics.html
@@ -31,5 +31,5 @@
       <div data-ng-include="'academics_university_requirements.html'" data-ng-if="requirements && isUndergraduate"></div>
     </div>
   </div>
-  <div data-ng-include="'404.html'" data-ng-if="!canViewAcademics"></div>
 </div>
+<div data-ng-include="'404.html'" data-ng-if="!canViewAcademics"></div>
--- a/src/assets/templates/academics.html
+++ b/src/assets/templates/academics.html
@@ -1,6 +1,4 @@
-<div
-  class="cc-page-academics"
-  data-cc-spinner-directive>
+<div class="cc-page-academics" data-cc-spinner-directive>
   <div data-ng-if="canViewAcademics">
     <div class="column">
       <h1 class="cc-heading-page-title">My Academics</h1>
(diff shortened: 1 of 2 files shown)


src/assets/templates/academics.html
```diff
@@ -1,6 +1,4 @@
-<div
-  class="cc-page-academics"
-  data-cc-spinner-directive>
+<div class="cc-page-academics" data-cc-spinner-directive>
   <div data-ng-if="canViewAcademics">
     <div class="column">
```
johncrossman: @dfkaye I think we want the `data-ng-if="!canViewAcademics"` only top-level div. I release that you did not touch that line but now is a good time to fix.

johncrossman: Note: I just corrected my typo above. :)  I meant to say we want `data-ng-if="canViewAcademics"`

src/assets/templates/academics.html
```diff
@@ -34,14 +32,4 @@ <h1 class="cc-heading-page-title">My Academics</h1>
...
   </div>
 </div>
-<div class="cc-page-academics" data-ng-if="!canViewAcademics" data-ng-cloak>
-  <div class="small-12 columns">
-    <h2>Data not available</h2>
-    <p>
-      This page displays academic status information for UC Berkeley students and faculty, and is not available to staff.
-    </p>
-    <p>
-      If you believe you are seeing this message in error, please <a href="http://www.ets.berkeley.edu/calcentral-support">let us know</a>.
```
johncrossman: @raydavis @christianvuerings  Do you see a problem in swapping out this "contact us" text? We've dropped in a `404.html` include, as used in `profile.html` and others.

src/assets/templates/academics.html
christianvuerings: @johncrossman Do you mean moving it up to the level of `cc-page-academics`? Won't that also hide the spinner?

src/assets/templates/academics.html
christianvuerings: @raydavis so we're ok with removing this functionality?

src/assets/templates/academics.html
christianvuerings: Looks ok now

johncrossman: @christianvuerings Do we need a spinner if we're presenting a 404?

christianvuerings: @johncrossman The current pattern is to redirect to a 404, that would be my suggestion.

christianvuerings: @johncrossman I see, we already do this for profile as well, so I'm actually ok with how it is right now.

johncrossman: @dfkaye I'll turn this over to @christianvuerings   He certainly knows our best practice.

christianvuerings: @johncrossman @dfkaye I can go either way on this. Usually we avoid `ng-if` statements on the same line as the spinner `data-cc-spinner-directive`.

However, in this case, since `canViewAcademics` is a part of `api/my/status`, which is the first XHR request we make on every page, it's fine to put it on the top level element.

src/assets/templates/academics.html
raydavis: Thanks for checking. From my point of view, we might as well assume that it's the same as any other unexpected I-can't-get-to-this-place error, so the normal 404 makes sense.

johncrossman:
@dfkaye  Once you amend this commit by putting `data-ng-if="canViewAcademics"` in top-level div, I will merge the PR.

src/assets/templates/academics.html
```diff
@@ -1,6 +1,5 @@
-<div
-  class="cc-page-academics"
-  data-cc-spinner-directive>
+
+<div class="cc-page-academics" data-cc-spinner-directive>
```
johncrossman: @dfkaye The line break at start of file should be removed.

johncrossman:
Thanks @dfkaye  Onward!

PR #4878, opened 2016-02-10
SISRP-14765 - Can't Accept Terms and Conditions in cc-sis-test in View-As mode

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-14765


johncrossman:
And I `canMerge`. Looks good, @christianvuerings

PR #4882, opened 2016-02-11
CLC-6199 Fix reactivation of bCourses account when SIS ID is also changed

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6199

I expect this will migrate to a module method at some point to be used from other classes, but didn't want to delay the fix.

### DIFF
--- a/app/models/canvas_csv/maintain_users.rb
+++ b/app/models/canvas_csv/maintain_users.rb
@@ -28,25 +28,31 @@ module CanvasCsv
       if (user_logins = response[:body])
         # We look for the login with a numeric "unique_id", and assume it is an LDAP UID.
         user_logins.select! do |login|
-          begin
-            Integer(login['unique_id'], 10)
-            true
-          rescue ArgumentError
-            false
-          end
+          parse_login_id(login['unique_id'])[:ldap_uid]
         end
         if user_logins.length > 1
           logger.error "Multiple numeric logins found for Canvas user #{canvas_user_id}; will skip"
         elsif user_logins.empty?
           logger.warn "No LDAP UID login found for Canvas user #{canvas_user_id}; will skip"
         else
-          login_id = user_logins[0]['id']
+          login_object_id = user_logins[0]['id']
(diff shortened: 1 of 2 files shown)


app/models/canvas_csv/maintain_users.rb
```diff
@@ -28,27 +28,34 @@ def self.change_sis_user_id(canvas_user_id, new_sis_user_id)
...
       end
       false
     end
 
+    def self.parse_login_id(login_id)
+      if (matched = /^(inactive-)?([0-9]+)$/.match login_id)
+        inactive_account = matched[1]
+        ldap_uid = matched[2].to_i
+      else
+      end
```
johncrossman: @raydavis Empty else block?

raydavis: Yeah. So what's your point?

... amendment coming, thanks!

app/models/canvas_csv/maintain_users.rb
```diff
@@ -131,22 +137,21 @@ def handle_email_deletions(canvas_user_ids)
...
       old_account_data = existing_account.to_hash
-      login_id = old_account_data['login_id']
-      if (inactive_account = /^inactive-([0-9]+)$/.match login_id)
-        login_id = inactive_account[1]
-      end
-      if (ldap_uid = Integer(login_id, 10) rescue nil)
-        @known_uids << login_id
+      parsed_login_id = self.class.parse_login_id old_account_data['login_id']
+      ldap_uid = parsed_login_id[:ldap_uid]
+      inactive_account = parsed_login_id[:inactive_account]
```
johncrossman: @raydavis Not a big deal: It seems we only care about `inactive_account` value if ldap_uid is not nil. Putting the expression above within the `if ldap_uid` block would make that clear. Not a merge blocker.

johncrossman:
Looks good to me, @raydavis

app/models/canvas_csv/maintain_users.rb
raydavis: I vacillated on that myself, @johncrossman , and decided to keep the parsed_login_id hash out of the inner block logic instead.

PR #4890, opened 2016-02-12
CLC-6192 Update Housing links in My Campus

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6192


johncrossman:
Combining into one looks good. Thanks @raydavis

PR #4893, opened 2016-02-12
SISRP-14845, If user can_administrate? then surface SIS dependencies in /api/config

johncrossman:
https://jira.berkeley.edu/browse/SISRP-14845

See Jira for sample screenshot.


paulkerschen:
Thank you @johncrossman! I haven't been this excited about a labor-saving device since the time I finally bought a dishwasher.

PR #4901, opened 2016-02-16
NOJIRA, bump master to v74

johncrossman:
No change to `versions/previous_release_db_schema.txt`


paulkerschen:
At least half of Travis is OK with this shocking version bump.

PR #4902, opened 2016-02-16
CLC-6207 Handle LDAP errors

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6207


paulkerschen:
Thanks for patching the hole, @raydavis.

PR #4904, opened 2016-02-16
SISRP-14914 Fall back to Oracle when LDAP contains conflicting student affiliations

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-14914


raydavis:
This troubles me, but it's not the PR's fault. Let's keep investigating that LDAP combo...

PR #4905, opened 2016-02-16
SISRP-10695 - Profile - update preferred name, gear still shows previous name

dfkaye:
https://jira.berkeley.edu/browse/SISRP-10695

This was interesting ~ Let me know what you think!

### DIFF
--- a/src/assets/javascripts/angular/controllers/widgets/profile/basicPreferredNameController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/profile/basicPreferredNameController.js
@@ -6,7 +6,7 @@ var _ = require('lodash');
 /**
  * Preferred Name Controller
  */
-angular.module('calcentral.controllers').controller('BasicPreferredNameController', function(apiService, profileFactory, $scope) {
+angular.module('calcentral.controllers').controller('BasicPreferredNameController', function(apiService, profileFactory, $scope, $rootScope) {
   angular.extend($scope, {
     emptyObject: {},
     items: {
@@ -50,6 +50,9 @@ angular.module('calcentral.controllers').controller('BasicPreferredNameControlle
   };
 
   var saveCompleted = function(data) {
+    // Notify other controllers about the preferredName update.
+    $rootScope.$broadcast('calcentral.custom.api.preferredname.update');
+
     $scope.isSaving = false;
     actionCompleted(data);
   };
--- a/src/assets/javascripts/angular/controllers/widgets/statusController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/statusController.js
@@ -6,7 +6,7 @@ var angular = require('angular');
 /**
  * Status controller
  */
-angular.module('calcentral.controllers').controller('StatusController', function(activityFactory, apiService, badgesFactory, financesFactory, holdsFactory, $scope, $q) {
+angular.module('calcentral.controllers').controller('StatusController', function(activityFactory, apiService, badgesFactory, financesFactory, holdsFactory, profileFactory, $http, $scope, $q) {
   // Keep track on whether the status has been loaded or not
   var hasLoaded = false;
 
@@ -85,6 +85,16 @@ angular.module('calcentral.controllers').controller('StatusController', function
     // Hides the spinner
     $scope.statusLoading = '';
   };
+
+  /**
+   * Listen for this event in order to make a refresh request which updates the
+   * displayed `api.user.profile.firstName` in the gear_popover.
+   */
+  $scope.$on('calcentral.custom.api.preferredname.update', function() {
+    apiService.user.fetch({
+      refreshCache: true
+    });
+  });
 
   $scope.$on('calcentral.api.user.isAuthenticated', function(event, isAuthenticated) {
     if (isAuthenticated && !hasLoaded) {
--- a/app/models/campus_solutions/my_person_name.rb
+++ b/app/models/campus_solutions/my_person_name.rb
@@ -1,7 +1,7 @@
 module CampusSolutions
   class MyPersonName < UserSpecificModel
 
-    include PersonDataUpdatingModel
+    include UserApiUpdatingModel
 
     def update(params = {})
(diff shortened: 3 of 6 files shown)


src/assets/javascripts/angular/controllers/widgets/profile/basicPreferredNameController.js
```diff
@@ -50,6 +50,9 @@ angular.module('calcentral.controllers').controller('BasicPreferredNameControlle
   };
 
   var saveCompleted = function(data) {
+    // Notify other controllers about this update.
+    $rootScope.$broadcast('preferredNameUpdate');
```
christianvuerings: @dfkaye would it be possible to use the naming standard for `$rootScope.$broadcast(`?

e.g. `$rootScope.$broadcast('calcentral.custom.api.preferredname.update', ...)`
We definitely want to avoid any collisions on the rootscope.

src/assets/javascripts/angular/controllers/widgets/statusController.js
```diff
@@ -82,10 +82,30 @@ angular.module('calcentral.controllers').controller('StatusController', function
...
     $scope.statusLoading = '';
   };
 
+  /**
+   * Listen for this event in order to make a refresh request to get person data
+   * and update the displayed `preferredName` in the gear_popover.
+   */
+  $scope.$on('preferredNameUpdate', function() {
+    var uid = apiService.user.profile.uid;
+    var deleteCachedUserUrl = '/api/cache/delete?key=User::Api/' + uid;
```
christianvuerings: @dfkaye we should definitely not do this. Only admins should be able to delete the cache.
Instead, it should just be fine to call the endpoint with `refreshCache` as you're doing.

If that doesn't work, we'll also need to make a back-end change. Which makes sure the cache is cleared when we update the preferred name.

src/assets/javascripts/angular/controllers/widgets/statusController.js
```diff
@@ -82,10 +82,30 @@ angular.module('calcentral.controllers').controller('StatusController', function
...
+
+    $http.get(deleteCachedUserUrl).then(function() {
+      profileFactory.getPerson({
+        refreshCache: true
+      }).then(function(data) {
+        $scope.preferredName = _.get(data, 'data.feed.student.names[0].givenName');
```
christianvuerings: Looks like we're combining `apiService.user.profile.firstName` and `data.feed.student.names[0].givenName`. Would it be possible to always use the one from the `apiService`?

src/assets/javascripts/angular/controllers/widgets/profile/basicPreferredNameController.js
dfkaye: @christianvuerings Sounds good ~ I'll get that in there tomorrow morning :)

src/assets/javascripts/angular/controllers/widgets/statusController.js
dfkaye: @christianvuerings Aha ~ you are correct ~ user must be a superuser to delete the cache ~ The `refreshCache` option does not work so yes, we have to make a back end change...

dfkaye: @christianvuerings ~ just fyi, @paulkerschen has agreed to guide me on this

src/assets/javascripts/angular/controllers/widgets/statusController.js
dfkaye: Don't think so ~ `apiService.user.profile.firstName` is static ~ once it's loaded in the angular, that's it.

I'm working passing the new `preferredName` from the editor as a `broadcast` argument (once the save is completed), and letting the status controller assign that to the $scope after refresh.

dfkaye: Alternatively, could reassign `apiService.user.profile.firstName` itself with the new preferredName...

Which approach makes more sense?

christianvuerings: @dfkaye I would suggest refetching the `api/my/status` feed by using `apiService.user.fetch()` which should set the correct preferred name again.

Also, have a look at https://github.com/ets-berkeley-edu/calcentral/blob/245104561107dab2cbf905b94d274c0c6a5666c2/src/assets/javascripts/angular/controllers/widgets/hookupGoogleReminderController.js#L14

src/assets/javascripts/angular/controllers/widgets/statusController.js
```diff
@@ -86,6 +86,16 @@ angular.module('calcentral.controllers').controller('StatusController', function
     $scope.statusLoading = '';
   };
 
+  /**
+   * Listen for this event in order to make a refresh request which updates the
+   * displayed `api.user.profile.firstName` in the gear_popover.
+   */
+  $scope.$on('calcentral.custom.api.preferredname.update', function() {
+    apiService.user.fetch({
```
christianvuerings: @dfkaye do we actually still need the `broadcast` and `on`? Could we not call `apiService.user.fetch` directly in the `BasicPreferredNameController`.

Honestly could go either way on this so whatever you think is the cleanest/best approach.

dfkaye: @christianvuerings ~ good question ~ I thought of that, then @paulkerschen asked me the same thing.  

I think we should stick with the current impl for the following reasons:

+ The gear_popover template is controlled by the statusController, so that controller should make the call to refresh the data used by the template (even if it's the globally available `api.user.profile` data).  
+ The event-driven approach makes the separation of concerns between controllers/components a little easier to understand (to my mind, at least).  In this case, the preferred name controller just notifies listeners that a change was made, allowing them to take what action they need on their data and view...

Sorry for sounding like an MVC essay, but that's mainly it - separation of concerns.

christianvuerings: @dfkaye thanks for explaining and I agree with you.

PR #4906, opened 2016-02-16
SISRP-13071 Remove CS ListOfValues; make SirResponse test pending

raydavis:
https://jira.berkeley.edu/browse/SISRP-13071


paulkerschen:
In the end, it cannot be claimed that ListOfValues led an especially happy or useful life; still, it did less harm than many of us might have done in its place. Vale.

christianvuerings:
@raydavis thanks for doing this cleanup!

PR #4912, opened 2016-02-17
SISRP-15057, CalCentral, when querying CS API for students per delegate, should use ByCsId(empl_id) to get LDAP_UID

johncrossman:
https://jira.berkeley.edu/browse/SISRP-15057


paulkerschen:
Sensible, @johncrossman

PR #4913, opened 2016-02-17
SISRP-14454, flush User::Api cache after delegate claims student

johncrossman:
https://jira.berkeley.edu/browse/SISRP-14454

User::Api flush is needed to get `hasToolboxTab: true` and then 'My Students' card.


johncrossman:
Formatting fix on `app/controllers/config_controller.rb` snuck in. Why not?

PR #4915, opened 2016-02-18
SISRP-15082, Acting as a super user with the advisor affiliation gives the wrong experience

johncrossman:
https://jira.berkeley.edu/browse/SISRP-15082

You get limited Advisor-view-as session if you are not SuperUser and do not have standard view-as permission.

### DIFF
--- a/src/assets/javascripts/angular/factories/adminFactory.js
+++ b/src/assets/javascripts/angular/factories/adminFactory.js
@@ -29,7 +29,8 @@ angular.module('calcentral.factories').factory('adminFactory', function(apiServi
   };
 
   var actAs = function(user) {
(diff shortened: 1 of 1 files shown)


christianvuerings:
@johncrossman thank you so much!

This doesn't block the merge, but don't we need the same logic in `stopActAs` in this file?
If we do need it there as well, might be good to add it to `setExtraProperties` in the `userService`.

johncrossman:
@christianvuerings I don't think we need a change to {{stopActAs}}. If {{advisorActingAsUid}} is present then there's no doubt the view-as session was started as an Advisor-view-as session. Let me know if I'm missing something.

PR #4916, opened 2016-02-18
SISRP-15077 - Show enrollment card even though we don't have the period data yet

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-15077

Also fixes
SISRP-14820 - Show waitlisted classes as soon as we have test data in BCSDEV

https://jira.berkeley.edu/browse/SISRP-14820

### DIFF
--- a/app/models/campus_solutions/enrollment_term.rb
+++ b/app/models/campus_solutions/enrollment_term.rb
@@ -13,8 +13,7 @@ module CampusSolutions
 
     def build_feed(response)
       return {} unless response &&
-        (enrollment_term = response['UC_SR_CLASS_ENROLLMENT']) &&
-        enrollment_term['ENROLLMENT_PERIOD'].present?
+        (enrollment_term = response['UC_SR_CLASS_ENROLLMENT'])
 
       # Though its name is in the singular, ENROLLMENT_PERIOD is an array containing multiple periods.
       enrollment_term['ENROLLMENT_PERIOD'].each do |period|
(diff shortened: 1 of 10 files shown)


app/models/campus_solutions/enrollment_term.rb
```diff
@@ -13,8 +13,7 @@ def initialize(options = {})
 
     def build_feed(response)
       return {} unless response &&
-        (enrollment_term = response['UC_SR_CLASS_ENROLLMENT']) &&
-        enrollment_term['ENROLLMENT_PERIOD'].present?
+        (enrollment_term = response['UC_SR_CLASS_ENROLLMENT'])
 
       # Though its name is in the singular, ENROLLMENT_PERIOD is an array containing multiple periods.
       enrollment_term['ENROLLMENT_PERIOD'].each do |period|
```
paulkerschen: @christianvuerings Even if we're letting empty ENROLLMENT_PERIOD through, a missing ENROLLMENT_PERIOD will error on this line, so we probably ought to handle that case somehow. OK to address this in a follow-up, though, since it doesn't seem the API is behaving that way at present.

paulkerschen:
Looks good @christianvuerings. Left one comment for follow-up, but I'll go ahead and merge so testing can proceed.

app/models/campus_solutions/enrollment_term.rb
christianvuerings: @paulkerschen actually they always seem to return `ENROLLMENT_PERIOD` but sometimes it's empty. Would you still like me to do a follow up on this?

PR #4917, opened 2016-02-18
SISRP-15111 Do not surface /api/my/photo to delegates

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-15111


johncrossman:
Not fair! I want to see my daughter's pink hair and nose ring!

PR #4927, opened 2016-02-22
SISRP-15334, delegates should not see Academics if privilege is only 'financial'

johncrossman:
https://jira.berkeley.edu/browse/SISRP-15334

When delegate-view-as mode the `get_my_students query` should use uid = original_delegate_user_id


paulkerschen:
Makes sense, @johncrossman.

johncrossman:
For the record, this PR resolves https://jira.berkeley.edu/browse/SISRP-15335, too.

PR #4929, opened 2016-02-22
CLC-6214 Fix reverso-blunder in Canvas Login ID parsing

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6214

True, shmue, so long as it's Boolean...


paulkerschen:
A not inconsiderable change!

PR #4930, opened 2016-02-22
SISRP-15342 - Enrollment card - change to Add / Swap links instead of Update

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-15342


christianvuerings:
Needs 1 more change - let's hold off for a second.

PR #4933, opened 2016-02-23
SISRP-13265, Delegate My Students card > No link to view-as when privilege is only by phone

johncrossman:
https://jira.berkeley.edu/browse/SISRP-13265

CS API change on ROLENAME of phone-only: `UC_CS_DA_THIRD_PARTY_RELEASE`


paulkerschen:
Release!

PR #4935, opened 2016-02-23
SISRP-15340, When delegate ends view-as session, flush cache per student UID

johncrossman:
https://jira.berkeley.edu/browse/SISRP-15340

There are two scenarios to consider – see Jira comments. Note: `UserApiController.expire` is a no-op but we add it to `UserApiExpiry` just in case.

### DIFF
--- a/app/controllers/act_as_controller.rb
+++ b/app/controllers/act_as_controller.rb
@@ -23,11 +23,12 @@ class ActAsController < ApplicationController
   end
 
   def stop
-    return redirect_to root_path unless session['user_id'] && session[@act_as_session_key]
-
-    # TODO: We might have to be aggressive with cache invalidation.
-    Cache::UserCacheExpiry.notify session['user_id']
-    logger.warn "Stop: #{session[@act_as_session_key]} act as #{session['user_id']}"
+    exiting_uid = session['user_id']
+    return redirect_to root_path unless exiting_uid && session[@act_as_session_key]
+    # Cached user data might have been filtered during view-as session so we flush to reset.
+    Cache::UserCacheExpiry.notify exiting_uid
+    CampusSolutions::UserApiExpiry.expire exiting_uid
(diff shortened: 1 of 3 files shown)


paulkerschen:
@johncrossman, this makes sense as a fix for the testing scenarios described in JIRA, but I think you may have uncovered a bigger architectural issue. At root, the bug is that the cached User::Api model changes behavior depending on authentication_state, and now for the first time it's possible to call it with differing authentication_states, causing behavior to leak between sessions. What happens if a student is using CalCentral at the same time as a delegate viewing that student? I think that data might unpredictably slip from one user's experience to the other's, even with flush-on-expiration behavior added.

I'm wondering if User::Api actually needs to come in different flavors (as we do in CampusSolutions::MyFinancialAidFilteredForAdvisor, CampusSolutions::MyFinancialAidFilteredForDelegate, MyAcademics::FilteredForDelegate) in order to plug the leak. Or as a speedier alternative, perhaps we could change the format of the cache key depending on authentication state. 

What do you think?

johncrossman:
@paulkerschen I think that makes a lot of sense. My hope is that we merge this PR and I follow up on your smarter plan w/ a subsequent Jira. SIT is breathing down my neck and this interim fix is an important one. Thoughts?  cc: @raydavis

raydavis:
Thanks for the ping, @johncrossman & @paulkerschen -- sorry to be late to this party.

Collision between censored View-As feeds and real-person uncensored feeds is a serious problem which we've already had to face for the Google parts of MyTasks. See "Cache::FilteredViewAsFeed".

johncrossman:
Thanks @raydavis  I'll amend this PR to get the right fix in.  cc: @paulkerschen

paulkerschen:
Per consensus, the fix goes in; we'll follow up on the tricky cache conflicts.

johncrossman:
@raydavis @paulkerschen  Follow up Jira is https://jira.berkeley.edu/browse/SISRP-15421

PR #4940, opened 2016-02-23
SISRP-13250 Expire user data by functional area based on JMS event topic

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-13250

The wiring of event topics to expiry modules is incomplete and will require functional validation. However, it should be easy from here to wire up the right behavior once we're certain what that behavior is.

### DIFF
--- a/app/models/notifications/sis_expiry_processor.rb
+++ b/app/models/notifications/sis_expiry_processor.rb
@@ -1,0 +1,47 @@
+module Notifications
+  class SisExpiryProcessor
+    include ClassLogger
+
+    def process(event, timestamp)
+      return false unless accept? event
+      logger.debug "Processing event: #{event}; timestamp = #{timestamp}"
+      if (expiry_module = get_expiry event) && (uid = get_uid event)
+        expiry_module.expire uid
+      end
+    end
+
+    private
+
+    def accept?(event)
+      event && EXPIRY_BY_TOPIC.keys.include?(event['topic'])
+    end
+
+    def get_expiry(event)
+      EXPIRY_BY_TOPIC[event['topic']]
+    end
+
+    def get_uid(event)
+      if (campus_solutions_id = event['payload'] && event['payload']['student'] && event['payload']['student']['StudentId'])
+        uid = CalnetCrosswalk::ByCsId.new(user_id: campus_solutions_id).lookup_ldap_uid
+        logger.error "No UID found for Campus Solutions ID #{campus_solutions_id}" unless uid
+      else
+        logger.error "Could not parse Campus Solutions ID from event #{event}"
+      end
+      uid
+    end
+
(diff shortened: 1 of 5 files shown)


app/models/notifications/sis_expiry_processor.rb
```diff
@@ -0,0 +1,46 @@
+module Notifications
+  class SisExpiryProcessor
+    include ClassLogger
+
+    def process(event, timestamp)
+      return false unless accept? event
+      logger.debug "Processing event: #{event}; timestamp = #{timestamp}"
+      if (expiry_module = get_expiry event) && (uid = get_uid event)
```
johncrossman: @paulkerschen Not a big deal but it seems we could do without `accept?` method. Instead, have `get_expiry` return nil if event is nil. That gives us bonus of debug logging when `event['topic']` matches no key.

paulkerschen: Makes sense, @johncrossman: this class started out inheriting from AbstractEventProcessor and then didn't. Amendment coming.

app/models/notifications/sis_expiry_processor.rb
```diff
@@ -0,0 +1,46 @@
...
+        uid = CalnetCrosswalk::ByCsId.new(user_id: campus_solutions_id).lookup_ldap_uid
+        logger.error "No UID found for Campus Solutions ID #{campus_solutions_id}" unless uid
+      else
+        logger.error "Could not parse Campus Solutions ID from event #{event}"
+      end
+      uid
+    end
+
+    EXPIRY_BY_TOPIC = {
+      'sis:staff:advisor' => nil,
```
johncrossman: Are the `=> nil` entries here to show they were considered? Worthy of a comment perhaps?

app/models/notifications/sis_expiry_processor.rb
paulkerschen: Actually, @johncrossman, I was mistaken; the reason the `accept?` method is useful here is because Notifications::JmsMessageHandler is set up to send every event to every processor. So every processor needs to check topic right away to see if a given event is its concern. Otherwise we'd get debug logging from this class for every JMS event, whether it belongs to this framework or one of our other JMS queues, such as the BearFacts or Final Grades notifications.

app/models/notifications/sis_expiry_processor.rb
paulkerschen: @johncrossman The nil entries are because the functional requirements for cache expiry aren't yet clear and have to be worked out from spreadsheets. My hope is that remaining work on the event notification framework will be more or less limited to putting the right values in this hash.

johncrossman:
Thanks @paulkerschen  Alley-oop!

PR #4941, opened 2016-02-23
SISRP-13710, delegate users do not need to reauth when view-as

johncrossman:
https://jira.berkeley.edu/browse/SISRP-13710

`current_user.viewing_as?` is true when delegate enters view-as so we must get more granular. Only require re-auth if it's standard-view-as or advisor-view-as.


paulkerschen:
Looks good @johncrossman; hurdle removed.

PR #4947, opened 2016-02-24
SISRP-15401 Only show CS-based Profile data to active students or applicants

raydavis:
https://jira.berkeley.edu/browse/SISRP-15401

Still needs review by the UX designers.

### DIFF
--- a/src/assets/templates/widgets/profile/basic_nonstudent.html
+++ b/src/assets/templates/widgets/profile/basic_nonstudent.html
@@ -1,0 +1,46 @@
+<div
+  class="cc-widget-text cc-page-widget-profile-sections"
+  data-ng-controller="BasicController"
+  data-cc-spinner-directive="basicInformation.isLoading">
+
+  <div class="cc-page-widget-profile-note">
+    <i class="fa fa-exclamation-triangle cc-icon cc-icon-gold" aria-hidden="true"></i>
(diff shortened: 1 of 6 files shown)


src/assets/templates/widgets/profile/basic_nonstudent.html
```diff
@@ -0,0 +1,46 @@
...
+
+  <!-- Name -->
+  <div class="cc-page-widget-profile-section">
+    <div class="cc-page-widget-profile-section-label">Name</div>
+    <div class="cc-page-widget-profile-section-content">
+      <div data-ng-bind="api.user.profile.fullName"></div>
+    </div>
+  </div>
+  <!-- Student ID -->
+  <div data-ng-include="'widgets/profile/basic_student_id.html'"></div>
```
christianvuerings: @raydavis Just checking, do we want to keep this for `basic_nonstudent`?

christianvuerings:
lgtm - posted 1 question

src/assets/templates/widgets/profile/basic_nonstudent.html
raydavis: Thanks, @christianvuerings. When the Student ID is available for ex-students, we show it on My Academics / Profile, so there's precedent. I also thought it might be useful for diagnosis in cases where the user _does_ need access to CS and the new code is mistakenly not giving it.

PR #4955, opened 2016-02-24
SISRP-15467 Do not break api/my/status on unexpected Hub payloads

raydavis:
https://jira.berkeley.edu/browse/SISRP-15467

Junction policy is to march through upstream exceptions rather than showing a blank page.


app/models/hub_edos/user_attributes.rb
```diff
@@ -22,24 +24,26 @@ def get_edo
...
-        extract_education_level(edo, result)
-        extract_total_units(edo, result)
-        extract_special_program_code(edo, result)
-        extract_reg_status(edo, result)
-        extract_residency(edo, result)
-        result[:statusCode] = 200
-      else
-        logger.error "Could not get Student EDO data for uid #{@uid}"
-        result[:noStudentId] = true
+      handling_exceptions(@uid) do
```
johncrossman: Is there value in passing a `:user_message_on_exception` opt to `handling_exceptions()` to say "We were trying to parse a Hub response"

johncrossman: Such a message would be most useful to Ops, I suppose. A developer would quickly use line number in stacktrace to diagnose.

raydavis:
Some interesting Travis failures -- I'll have to withdraw this for now.

PR #4956, opened 2016-02-25
SISRP-15505, no 'Link Your Account with Your Student' page when delegate-view-as mode

johncrossman:
https://jira.berkeley.edu/browse/SISRP-15505

We don't want delegate to inadvertently go to this page while still viewing-as.


raydavis:
Makes sense!

PR #4958, opened 2016-02-25
SISRP-15467 Do not break api/my/status on unexpected Hub payloads

raydavis:
https://jira.berkeley.edu/browse/SISRP-15467

Exceptions in Hub API parsing currently trigger complete blankness. Service policy is that as much UX as possible should survive any API issues.


johncrossman:
Makes sense. Thanks @raydavis

PR #4961, opened 2016-02-25
SISRP-14832, verify_sis_api_endpoints loads config/settings.yml for default feature flag settings

johncrossman:
https://jira.berkeley.edu/browse/SISRP-14832

The intended precedence seems to work. I.e., local.yml wins when feature flag is set in both files.


paulkerschen:
Go yml go.

PR #4969, opened 2016-02-26
SISRP-15079 - Refactors for filtering of confidential group items

redconfetti:
https://jira.berkeley.edu/browse/SISRP-15079

Refactors for filtering of 'Expected Family Contribution' and 'Berkeley Family Contribution' only, allowing 'Dependency Status' and 'Family Members in College'

### DIFF
--- a/app/models/campus_solutions/my_financial_aid_filtered_for_advisor.rb
+++ b/app/models/campus_solutions/my_financial_aid_filtered_for_advisor.rb
@@ -35,11 +35,8 @@ module CampusSolutions
       return feed unless feed[:feed] && feed[:feed][:status] && (categories = feed[:feed][:status][:categories])
       categories.each do |category|
         if category[:itemGroups]
-          category[:itemGroups].each_with_index do |group, group_index|
-            category[:itemGroups][group_index].each_with_index do |item, item_index|
-              category[:itemGroups][group_index][item_index] = nil if has_confidential_information? item
-            end
-            category[:itemGroups][group_index].compact!
+          category[:itemGroups].each do |group|
+            group.reject! { |item| has_confidential_information? item }
           end
         end
       end
--- a/app/models/campus_solutions/my_financial_aid_filtered_for_advisor.rb
+++ b/app/models/campus_solutions/my_financial_aid_filtered_for_advisor.rb
@@ -28,19 +28,19 @@ module CampusSolutions
       logger.debug "Advisor #{advisor_uid} viewing user #{@uid} financial aid feed where aid_year = #{aid_year}"
       {
         filteredForAdvisor: true
-      }.merge(remove_family_information feed)
+      }.merge(remove_confidential_information feed)
     end
 
-    def remove_family_information(feed={})
+    def remove_confidential_information(feed={})
       return feed unless feed[:feed] && feed[:feed][:status] && (categories = feed[:feed][:status][:categories])
       categories.each do |category|
         if category[:itemGroups]
-          filtered_groups = []
-          category[:itemGroups].each do |group|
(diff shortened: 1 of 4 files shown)


app/models/campus_solutions/my_financial_aid_filtered_for_advisor.rb
```diff
@@ -28,19 +28,19 @@ def apply_filter(feed)
...
-    def remove_family_information(feed={})
+    def remove_confidential_information(feed={})
       return feed unless feed[:feed] && feed[:feed][:status] && (categories = feed[:feed][:status][:categories])
       categories.each do |category|
         if category[:itemGroups]
-          filtered_groups = []
-          category[:itemGroups].each do |group|
-            exclude_group = group.any? { |item| has_family_information? item }
-            filtered_groups << group unless exclude_group
+          category[:itemGroups].each_with_index do |group, group_index|
```
paulkerschen: @redconfetti Why `each_with_index` here? Seems more readable just to use `each` and refer to `group` instead of `category[:itemGroups][group_index]` within the loop.

app/models/campus_solutions/my_financial_aid_filtered_for_advisor.rb
```diff
@@ -28,19 +28,19 @@ def apply_filter(feed)
...
+            category[:itemGroups][group_index].each_with_index do |item, item_index|
+              category[:itemGroups][group_index][item_index] = nil if has_confidential_information? item
+            end
+            category[:itemGroups][group_index].compact!
```
paulkerschen: @redconfetti For this sort of selective clearing-out of arrays, you should actually be able to get away with a one-liner instead of four:

`category[:itemGroups][group_index].reject! { |item| has_confidential_information? item }`

app/models/campus_solutions/my_financial_aid_filtered_for_advisor.rb
redconfetti: @paulkerschen I needed the index to delete the item from the array. Your note below concerning the use of reject! is a great idea. I've used Array#reject often, but didn't remember Array#reject! existing

redconfetti: Also I just noticed that there is Array#each_index, so I don't have to unnecessarily involve the element itself in the loop.

app/models/campus_solutions/my_financial_aid_filtered_for_advisor.rb
redconfetti: :+1:

spec/models/campus_solutions/my_financial_aid_filtered_for_advisor_spec.rb
```diff
@@ -20,12 +20,23 @@
...
+          expect(json).to include 'SHIP Health Insurance'
+          expect(json).to include 'Student Standing'
+          expect(json).to include 'Estimated Cost of Attendance'
+          expect(json).to include 'Dependency Status'
+          expect(json).to include 'Family Members in College'
+        end
+        it 'should filter out \'Expected Family Contribution\'' do
+          expect(json).to_not include 'Expected Family Contribution'
+        end
+        it 'should filter out \'Berkeley Family Contribution\'' do
```
paulkerschen: @redconfetti Minor, but looks like a typo here and in the other spec below - it should say 'Berkeley Parent Contribution', right?

redconfetti: @paulkerschen Yes. The spec description is wrong. The assertion is correct. Thank you

paulkerschen:
Thanks @redconfetti!

PR #4970, opened 2016-02-26
SISRP-15421, formalize set of view-as related session keys

johncrossman:
https://jira.berkeley.edu/browse/SISRP-15421

This is a prelude to cache-key management scheme involving auth-state. There are many files touched but most are specs. No major refactoring, just formalizing session keys.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -122,6 +122,18 @@ class ApplicationController < ActionController::Base
 
   private
 
+  def get_active_view_as_session_type
+    SessionKey::VIEW_AS_TYPES.each do |key|
+      return key if session.has_key? key
+    end
+    nil
+  end
+
+  def get_original_viewer_uid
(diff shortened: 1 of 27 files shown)


raydavis:
Puttin' on my top hot, brushin' off my tails...

app/controllers/application_controller.rb
```diff
@@ -122,6 +122,18 @@ def self.correct_port(host_with_port, http_referer)
 
   private
 
+  def get_active_view_as_session_type
+    SessionKey::VIEW_AS_TYPES.each do |key|
+      return key if session.has_key? key
+    end
+    nil
+  end
```
paulkerschen: @johncrossman What about a one-liner here? `SessionKey::VIEW_AS_TYPES.find { |key| session.has_key? key }`

johncrossman: Mos' def, @paulkerschen. I'll include in subsequent PR.

PR #4973, opened 2016-02-26
SISRP-13245 Adapt legacy Bearfacts data to GL5 academic profile design

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-13245
JIRA with designs: https://jira.berkeley.edu/browse/SISRP-11939

While waiting for the Hub to finalize its academic status endpoint, we can go ahead and adapt our legacy academics feed to the GL5 academic profile design.
- All users, new and old, get the new design from the JIRA, _except_ that continuing students will not lose the "Including AP / Not Including AP" distinction if relevant.
- Our incomplete-profile messaging remains, but users with 10-digit Campus Solutions IDs will not be directed to Bearfacts.
- The back end replaces its concept of "standing" with possibly-plural "careers."
- "Colleges" becomes "majors," in order to make room for the "minors" data that our new source will provide.

### DIFF
--- a/app/models/my_badges/student_info.rb
+++ b/app/models/my_badges/student_info.rb
@@ -11,7 +11,7 @@ module MyBadges
       campus_attributes = CampusOracle::UserAttributes.new(user_id: @uid).get_feed
       result = {
         californiaResidency: campus_attributes[:california_residency],
-        isLawStudent: law_student,
+        isLawStudent: law_student?,
         regBlock: get_reg_blocks
       }
       if campus_attributes[:reg_status] && campus_attributes[:reg_status][:transitionTerm]
@@ -42,17 +42,13 @@ module MyBadges
       end
     end
 
-    # Set isLawStudent to true iff user's only college is School of Law
-    def law_student
-      college_feed = {}
-      MyAcademics::CollegeAndLevel.new(@uid).merge(college_feed)
-      if !(college_feed.empty? || college_feed[:empty])
-        colleges = college_feed[:collegeAndLevel][:colleges]
-        if !colleges.nil?
-          return colleges[0][:college].eql? "School of Law"
-        end
+    # True if user's first college is the School of Law.
+    def law_student?
+      if (college_feed = MyAcademics::CollegeAndLevel.new(@uid).merge({})) && college_feed[:majors].present?
+        college_feed[:majors].first[:college] == Berkeley::Departments.get('CLLAW')
+      else
+        false
       end
-      return false
     end
 
     def get_reg_blocks
--- a/src/assets/javascripts/angular/controllers/pages/academicsController.js
+++ b/src/assets/javascripts/angular/controllers/pages/academicsController.js
@@ -1,6 +1,7 @@
 /* jshint camelcase: false */
 'use strict';
 
+var _ = require('lodash');
 var angular = require('angular');
 
 /**
@@ -129,9 +130,9 @@ angular.module('calcentral.controllers').controller('AcademicsController', funct
     }
 
     $scope.isLSStudent = academicsService.isLSStudent($scope.collegeAndLevel);
-    $scope.isUndergraduate = ($scope.collegeAndLevel && $scope.collegeAndLevel.standing === 'Undergraduate');
+    $scope.isUndergraduate = _.contains(_.get($scope.collegeAndLevel, 'careers'), 'Undergraduate');
     $scope.isProfileCurrent = !$scope.transitionTerm || $scope.transitionTerm.isProfileCurrent;
-    $scope.showProfileMessage = (!$scope.isProfileCurrent || !$scope.collegeAndLevel || !$scope.collegeAndLevel.standing);
+    $scope.showProfileMessage = (!$scope.isProfileCurrent || !$scope.collegeAndLevel || _.isEmpty($scope.collegeAndLevel.careers));
 
     $scope.hasTeachingClasses = academicsService.hasTeachingClasses(data.teachingSemesters);
     if (data.teachingSemesters) {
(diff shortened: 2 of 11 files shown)


app/models/my_badges/student_info.rb
```diff
@@ -43,16 +43,12 @@ def get_transition_reg_status(code)
...
-      college_feed = {}
-      MyAcademics::CollegeAndLevel.new(@uid).merge(college_feed)
-      if !(college_feed.empty? || college_feed[:empty])
-        colleges = college_feed[:collegeAndLevel][:colleges]
-        if !colleges.nil?
-          return colleges[0][:college].eql? "School of Law"
-        end
+    def law_student?
+      if (college_feed = MyAcademics::CollegeAndLevel.new(@uid).merge({})) && college_feed[:majors].present?
+        college_feed[:majors].first[:college] == 'School of Law'
```
johncrossman: @paulkerschen Should we use `Berkeley::Departments.department_map['CLLAW']`?  Also, I suggest we not use actual variable/attribute name `isLawStudent` in comment above.

src/assets/javascripts/angular/controllers/pages/academicsController.js
```diff
@@ -129,9 +130,9 @@ angular.module('calcentral.controllers').controller('AcademicsController', funct
     }
 
     $scope.isLSStudent = academicsService.isLSStudent($scope.collegeAndLevel);
-    $scope.isUndergraduate = ($scope.collegeAndLevel && $scope.collegeAndLevel.standing === 'Undergraduate');
+    $scope.isUndergraduate = $scope.collegeAndLevel && _.contains($scope.collegeAndLevel.careers, 'Undergraduate');
```
johncrossman: @paulkerschen Doesn't _lodash_ allow us to remove null check on `$scope.collegeAndLevel`. I.e., the statement above can be shortened to:
`$scope.isUndergraduate = _.contains($scope.collegeAndLevel.careers, 'Undergraduate');`

src/assets/templates/academics_student_profile.html
```diff
@@ -134,29 +144,21 @@ <h3 data-ng-if="!isProfileCurrent && api.user.profile.roles.student && collegeAn
...
-              <span data-ng-if="isUndergraduate">Including AP</span>
-              <span data-ng-if="!isUndergraduate">Current</span>
-            </th>
-          </tr>
-        </thead>
         <tbody>
           <tr>
+            <th width="92px">Level</th>
+            <th class="cc-table-subheader" scope="col" data-ng-if="collegeAndLevel.nonApLevel">Including AP</th>
+            <td data-ng-if="!collegeAndLevel.nonApLevel"><strong data-ng-bind="collegeAndLevel.level"></strong></td>
```
johncrossman: @paulkerschen `th` instead of `td` here?

johncrossman:
Looks good. I'll wait on reply to comments above.

src/assets/templates/academics_student_profile.html
paulkerschen: @johncrossman Though the alignment is odd, in the case where we don't show AP level, this is supposed to be a data element (e.g. "Junior"), so I think `td` is the right semantic fit.

src/assets/javascripts/angular/controllers/pages/academicsController.js
paulkerschen: @johncrossman I don't think `_.contains` by itself will skirt the null check, but an added `_.get` would, which I will add.

app/models/my_badges/student_info.rb
paulkerschen: @johncrossman Oh, cool, I forgot about `Berkeley::Departments`.

paulkerschen:
Thanks @johncrossman, amendment pushed.

johncrossman:
Thanks @paulkerschen  Nice work.

src/assets/templates/academics_student_profile.html
```diff
@@ -87,46 +95,48 @@ <h3 data-ng-if="!isProfileCurrent && api.user.profile.roles.student && collegeAn
...
-            <th width="87px">General</th>
-            <th class="cc-table-subheader" scope="col">Standing</th>
-          </tr>
-        </thead>
-        <tbody data-ng-if="collegeAndLevel.standing">
+        <tbody>
           <tr>
-            <td></td>
-            <td><strong data-ng-bind="collegeAndLevel.standing"></strong></td>
+            <th width="92px" data-ng-pluralize count="collegeAndLevel.majors.length" when="{'1': 'Major', 'other': 'Majors'}"></th>
```
christianvuerings: @paulkerschen any possibility we could set this in CSS instead?

src/assets/templates/academics_student_profile.html
```diff
@@ -164,34 +166,33 @@ <h3 data-ng-if="!isProfileCurrent && api.user.profile.roles.student && collegeAn
...
-            <th width="87px">ID Number</th>
-            <th class="cc-table-subheader" scope="col">UID</th>
-          </tr>
-        </thead>
         <tbody>
           <tr>
-            <td></td>
-            <td><strong data-ng-bind="api.user.profile.uid"></strong></td>
+            <th width="92px">Earned Units</th>
+            <td><strong data-ng-bind="gpaUnits.totalUnits"></strong></td>
```
christianvuerings: @paulkerschen does this remove the show / hide functionality for the GPA as requested by the students? If so, might be worthwhile double checking with the designers.

src/assets/templates/academics_student_profile.html
paulkerschen: @christianvuerings To my surprise, I couldn't find a reliable CSS-based solution to fix table-cell width in local testing; I assume this is why the previous version of the template also relied a `width` attribute. If there's a trick I'm missing, though, please let me know.

christianvuerings: @paulkerschen thanks for clarifying. And even I learned something new today:

```
In CSS 2.1, the effect of 'min-width' and 'max-width' on tables, inline tables, table cells, table columns, and column groups is undefined.
```

The only way to get around this would be to add an extra inner `div` and set the `min-width` on that. However, that even feels worse, so let's leave it as is.

PR #4974, opened 2016-02-27
SISRP-14260 - Improves accessibility of link to manage delegates

redconfetti:
https://jira.berkeley.edu/browse/SISRP-14260

### DIFF
--- a/src/assets/stylesheets/_profile_page.scss
+++ b/src/assets/stylesheets/_profile_page.scss
@@ -4,8 +4,15 @@
     margin-right: 0;
     width: 540px;
   }
+
   .cc-page-widget-profile-note {
     padding: 0 10px 10px;
+  }
+
+  .cc-page-widget-profile-delegate {
+    .cc-page-widget-profile-delegate-slim-margin {
+      margin-bottom: 8px;
+    }
   }
   .cc-page-widget-profile-sections {
     .cc-page-widget-profile-section-info {
--- a/src/assets/stylesheets/_profile_page.scss
+++ b/src/assets/stylesheets/_profile_page.scss
@@ -10,10 +10,8 @@
   .cc-page-widget-profile-margin-top {
     margin-top: 5px;
   }
-  .cc-page-widget-profile-delegate {
-    .cc-page-widget-profile-delegate-slim-margin {
-      margin-bottom: 8px;
-    }
+  .cc-page-widget-profile-margin-bottom {
+    margin-bottom: 5px;
   }
   .cc-page-widget-profile-sections {
     .cc-page-widget-profile-section-info {
--- a/src/assets/stylesheets/_profile_page.scss
+++ b/src/assets/stylesheets/_profile_page.scss
@@ -4,11 +4,12 @@
     margin-right: 0;
     width: 540px;
   }
-
   .cc-page-widget-profile-note {
     padding: 0 10px 10px;
   }
-
+  .cc-page-widget-profile-margin-top {
+    margin-top: 5px;
+  }
   .cc-page-widget-profile-delegate {
     .cc-page-widget-profile-delegate-slim-margin {
       margin-bottom: 8px;
@@ -98,9 +99,6 @@
           margin-bottom: 5px;
         }
       }
-    }
-    .cc-page-widget-profile-section-content-item-margin-top {
-      margin-top: 5px;
     }
     .cc-page-widget-profile-section-not-available {
       font-style: italic;
--- a/src/assets/templates/widgets/profile/emergency_contact.html
+++ b/src/assets/templates/widgets/profile/emergency_contact.html
@@ -12,7 +12,7 @@
         <span data-ng-bind="phone.number"></span>
       </div>
       <div data-ng-bind="emergencyContact.email.emailAddress"></div>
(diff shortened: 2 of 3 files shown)


src/assets/stylesheets/_profile_page.scss
```diff
@@ -4,9 +4,16 @@
     margin-right: 0;
     width: 540px;
   }
+
   .cc-page-widget-profile-note {
     padding: 0 10px 10px;
   }
+
+  .cc-page-widget-profile-delegate {
```
christianvuerings: @redconfetti none of the CSS in this file should be specific to 1 page. Would it be possible to extract this in it's own CSS file?

src/assets/stylesheets/_profile_page.scss
```diff
@@ -4,9 +4,16 @@
...
+    .cc-page-widget-profile-delegate-slim-margin {
+      margin-bottom: 8px;
```
christianvuerings: @redconfetti this very much feels like a magic number, ok if we use `5px` or `10px` instead?

src/assets/stylesheets/_profile_page.scss
redconfetti: @christianvuerings I was trying to follow the convention that I see already in place, where styles are placed in this stylesheet specifically for the widgets placed on the profile page.

```
$ git grep "cc-page-widget-profile-note"
src/assets/stylesheets/_profile_page.scss:  .cc-page-widget-profile-note {
src/assets/templates/widgets/profile/basic_nonstudent.html:  <div class="cc-page-widget-profile-note">
src/assets/templates/widgets/profile/double_entry.html:<div class="cc-page-widget-profile-note">
src/assets/templates/widgets/profile/editable_soon.html:<div class="cc-page-widget-profile-note">
```

I'm not clear on what you are suggesting about extracting this to a separate CSS file.

If you would prefer that I remove this wrapper, so that the style can be shared in other areas, I would understand that.

src/assets/stylesheets/_profile_page.scss
redconfetti: @christianvuerings Can you clarify what you mean by "magic number"? The UX design that Flynn provided had this exact spacing for the link below the paragraph.

christianvuerings: @redconfetti have a look at https://css-tricks.com/magic-numbers-in-css/
And yes, but the designs we're getting are not pixel perfect, what did Flint say when you showed him the 5px and 10px versions? Would either of those work as well?

src/assets/stylesheets/_profile_page.scss
christianvuerings: @redconfetti indeed, `cc-page-widget-profile-note` is used in multiple profile sections. On which other section will `cc-page-widget-profile-delegate-slim-margin` be used?

If we have a style which is used only in 1 particular area, we should create a specific rule for it as we do here. However, it should also be specified in its separate CSS.

If we have a style which can be reused by other sections, then it's fine to specify it on the page level. And if we have a style which can be used even more globally, we have `_useful.scss`

src/assets/stylesheets/_profile_page.scss
redconfetti: @christianvuerings Flint did not specify that it must be pixel perfect, but I did align the design to fit the screenshot based mockup he provided. I'm reading further.

src/assets/stylesheets/_profile_page.scss
redconfetti: @christianvuerings If I look around the CalCentral app, I don't see a situation where we are placing a link below a paragraph in the same section, breaking the line, and having a bit of space so that it's clearly separate of the paragraph, yet still close enough to be considered part of the section... associated with the previous paragraph.

I've created [this jsfiddle](https://jsfiddle.net/redconfetti/48gz6tq3/) so we can discuss the proper approach to solving this problem. I've tried to put the link inside of the paragraph, to avoid it being affected by the bottom margin of the global paragraph style. To achieve what I want, I'd have to make the link a block element, with this `.5em` top margin. I'm doubting that you'd agree with this approach.

redconfetti:
@christianvuerings Since I'm moving the `cc-page-widget-profile-section-content-item-margin-top` style so it can be re-used later, I'm wanting to make sure that the emergency contact view that it's used for currently still renders properly.

Any tips on how I can load emergency contact info? I tried to use dummy JSON, but that didn't help, as there is only some student with languages data in place. I tried to enable and disable the proxies with false data, this didn't help. I even used real proxy data, but tried to force the HubEdos::Proxy to use fake data with a student that has profile data.... still I can't see it. I enabled the `cs_profile_emergency_contacts` feature flag, still no emergency contacts shown for me to test the styling with.

src/assets/templates/widgets/profile/emergency_contact.html
```diff
@@ -12,7 +12,7 @@
         <span data-ng-bind="phone.number"></span>
       </div>
       <div data-ng-bind="emergencyContact.email.emailAddress"></div>
-      <div class="cc-text-pre-line" data-ng-bind="emergencyContact.address.formattedAddress" class="cc-page-widget-profile-section-content-item-margin-top"></div>
+      <div class="cc-text-pre-line cc-page-widget-profile-margin-top" data-ng-bind="emergencyContact.address.formattedAddress"></div>
```
christianvuerings: @redconfetti thanks for this fix!

PR #4978, opened 2016-02-29
SISRP-15634, remove last slash from UC_FA_GET_T_C URL in verify script

johncrossman:
https://jira.berkeley.edu/browse/SISRP-15634

Strange but true: all CS FinAid-related endpoints have URL convention `.../get/?` except UC_FA_GET_T_C.


johncrossman:
We have one green Travis.

PR #4983, opened 2016-02-29
SISRP-15536, UID/SID lookup producing two identical results

johncrossman:
https://jira.berkeley.edu/browse/SISRP-15536

Use an enumerable type (Set) that forces uniqueness.


paulkerschen:
+1

PR #4984, opened 2016-02-29
SISRP-15673 - Change "..." to &hellip; entity in profile templates

dfkaye:
https://jira.berkeley.edu/browse/SISRP-15673


christianvuerings:
lgtm - @pfarestveit would be great if we could test this for accessibility (how the `...` is being read to the screen reader)

PR #4986, opened 2016-03-01
SISRP-15634, remove trailing slash from UC_DEPOSIT_AMT call

johncrossman:
https://jira.berkeley.edu/browse/SISRP-15634

Change default IDs and fix URL path on UC_DEPOSIT_AMT check.


raydavis:
I will ignore Travis's evil twin, Travis.

PR #4988, opened 2016-03-01
SISRP-12707 - UX DESIGN: Personalize Profile Avatar for Staff and Faculty

dfkaye:
https://jira.berkeley.edu/browse/SISRP-12707


src/assets/templates/widgets/gear_popover.html
```diff
@@ -4,9 +4,11 @@
   data-ng-class="{true:'cc-header-icon-selected'}[api.popover.status('cc-popover-menu')]"
   title="Settings">
     <div class="cc-flex cc-flex-align-center">
-      <span data-ng-if="api.user.profile.hasPhoto" class="cc-popover-gear-round-profile cc-popover-gear-round-profile-has-photo">
+      <span data-ng-if="!api.user.profile.roles.student && !api.user.profile.hasPhoto"
```
christianvuerings: @dfkaye why are we adding the `api.user.profile.roles.student` logic in here? Shouldn't that stay the same?

E.g. what happens when a delegate is seeing this?

dfkaye: @christianvuerings My understanding is:
+ Students viewing themselves see either their own photo or the gray profile image.
+ Non-students viewing themselves see either their own photo or the *bear* profile image.
+ A delegate sees *only* the gray profile image for the student.

christianvuerings: @dfkaye in this PR (just confirmed with Flint) we should replace the gray profile image (in the top navigation) for every case.

dfkaye:
@christianvuerings second pass ready for review

src/assets/templates/widgets/gear_popover.html
```diff
@@ -4,11 +4,8 @@
   data-ng-class="{true:'cc-header-icon-selected'}[api.popover.status('cc-popover-menu')]"
   title="Settings">
     <div class="cc-flex cc-flex-align-center">
-      <span data-ng-if="api.user.profile.hasPhoto" class="cc-popover-gear-round-profile cc-popover-gear-round-profile-has-photo">
-      </span>
-      <span data-ng-if="!api.user.profile.hasPhoto" class="cc-popover-gear-round-profile cc-popover-gear-round-profile-no-photo">
```
christianvuerings: @dfkaye are we still using `cc-popover-gear-round-profile-no-photo` anywhere? If not, can we remove it from the CSS?

dfkaye: @christianvuerings : looks like we're not using it anywhere ~ removing...

christianvuerings:
@dfkaye thanks for this change - definitely don't forget to tag me on the PR when you've updated it (will get you faster merges)

PR #4989, opened 2016-03-01
SISRP-15695 - CalCentral: Enrollment Card - section #6: Adjust - Default to Open

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-15695


johncrossman:
merge: true

PR #4991, opened 2016-03-01
SISRP-13855 Use Cal1card proxy for user and roster photos

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-13855

Moving away from database photo queries loses us the ability to check (via photo_bytes join) whether a photo exists for a given student before trying to fetch it. An Angular directive now swaps in the "photo unavailable" image in the event of a load error.


raydavis:
I'll let @christianvuerings take a whack at the directive, but all else looks right to me. Let's remember to QA against a couple of big bCourses sites as a performance sanity check.

PR #4994, opened 2016-03-02
SISRP-15727 - New Students Can't Get to FinAid until they SIR (CalCentral-DEV)

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-15727

### DIFF
--- a/app/models/user/api.rb
+++ b/app/models/user/api.rb
@@ -168,7 +168,7 @@ module User
 
     def has_financials_tab?(roles, view_as_privileges)
       return false if view_as_privileges && !view_as_privileges[:financial]
-      roles[:student] || roles[:exStudent]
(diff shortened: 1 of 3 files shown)


app/models/user/api.rb
```diff
@@ -168,7 +168,7 @@ def has_academics_tab?(roles, has_instructor_history, has_student_history, view_
 
     def has_financials_tab?(roles, view_as_privileges)
       return false if view_as_privileges && !view_as_privileges[:financial]
-      roles[:student] || roles[:exStudent]
+      !!(roles[:student] || roles[:exStudent] || roles[:applicant])
```
christianvuerings: FYI - we had to convert this to a `boolean` since in the case of staff (see tests) it was returning `nil`

PR #4998, opened 2016-03-02
SISRP-14732 - Add CAS blacklist / exclude for ?cmd=logout&redirect_url=...

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-14732

Fix the default URL - see jira for the latest info

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -409,7 +409,7 @@ advising_proxy:
 campus_solutions_proxy:
   fake: true
   base_url: 'https://bcs-web-dev-03.is.berkeley.edu:8443/PSIGW/RESTListeningConnector/PSFT_CS'
-  logout_url: 'https://bcs-web-dev-03.is.berkeley.edu:8443/uc_cust/data/portal/html/uc_splash.html?cmd=logout'
+  logout_url: 'https://bcs-web-dev-03.is.berkeley.edu:8443/psp/logout/EMPLOYEE/HRMS/?cmd=logout'
   username: secret
   password: secret
 


johncrossman:
@christianvuerings I logged in to bcs-web-dev and hit your `cmd=logout` URL but I remained logged in.  The URL specified in the Jira worked in logging me out. Am I missing something?

christianvuerings:
@johncrossman Which URL in the jira did you try out? I would recommend following the whole thread.
`https://bcs-web-dev-03.is.berkeley.edu:8443/psp/logout/EMPLOYEE/HRMS/?cmd=logout` is the last one we got from Shiva.

And since that URL won't hit the CAS logout, it won't work the way you tested it. This URL in this PR will only remove the CS cookies (both [HTTPOnly](https://www.owasp.org/index.php/HttpOnly) and the session ones)

Try the following instead:
* Log into https://calcentral-sis-dev.berkeley.edu in tab 1
* Log into https://bcs-web-dev-03.is.berkeley.edu:8443/psp/bcsdev/EMPLOYEE/HRMS/h/?tab=DEFAULT in tab2
* Go to any BCSDEV page - e.g . `My Personalizations` in tab 2
* Logout of CalCentral (tab 1)
* Refresh tab 2

You should be logged out now.

I'll also be very honest and say that this feels like one of the ugliest hacks I've ever programmed (the redirects). Ideally CS would be able to receive a back-end API call to log a user out.

johncrossman:
Thanks @christianvuerings – it works. Fyi, I did read through the Jira but your comments make it all clear. Merging...

PR #4999, opened 2016-03-02
SISRP-15682, filter /api/edos/ feeds per view-as type

johncrossman:
https://jira.berkeley.edu/browse/SISRP-15682

We don't want proxy classes to care about session variables so I've made the previously static `include_fields` list configurable via `options` hash. I also added `Cache::KeyGenerator.per_view_as_type` as needed.

### DIFF
--- a/src/assets/javascripts/angular/controllers/widgets/statusController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/statusController.js
@@ -6,7 +6,7 @@ var angular = require('angular');
 /**
  * Status controller
  */
-angular.module('calcentral.controllers').controller('StatusController', function(activityFactory, apiService, badgesFactory, financesFactory, holdsFactory, profileFactory, $http, $scope, $q) {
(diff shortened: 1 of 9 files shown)


src/assets/javascripts/angular/controllers/widgets/statusController.js
```diff
@@ -6,7 +6,7 @@ var angular = require('angular');
 /**
  * Status controller
  */
-angular.module('calcentral.controllers').controller('StatusController', function(activityFactory, apiService, badgesFactory, financesFactory, holdsFactory, profileFactory, $http, $scope, $q) {
+angular.module('calcentral.controllers').controller('StatusController', function(activityFactory, apiService, badgesFactory, financesFactory, holdsFactory, $http, $scope, $q) {
```
johncrossman: Fyi, `profileFactory` removed because 'tis not used.

christianvuerings: @johncrossman thanks!

raydavis:
Looks like that should do it -- on to QA!

PR #5001, opened 2016-03-02
SISRP-10777 - Front End Dev: Credentials Tab - Languages Card - Edit

dfkaye:
https://jira.berkeley.edu/browse/SISRP-10777

@christianvuerings PR fixes for:
+ the promise creation logic in `loadInformation()`
+ a sorting problem in `parsePerson()`
+ the filtering logic in `parseLanguages()`

### DIFF
--- a/src/assets/javascripts/angular/controllers/widgets/profile/languagesSectionController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/profile/languagesSectionController.js
@@ -6,7 +6,7 @@ var angular = require('angular');
 /**
  * Language section controller
  */
-angular.module('calcentral.controllers').controller('LanguagesSectionController', function(apiService, profileFactory, $scope, $q) {
+angular.module('calcentral.controllers').controller('LanguagesSectionController', function(apiService, profileFactory, $scope) {
   angular.extend($scope, {
     addingItem: false,
     currentObject: {},
@@ -148,9 +148,7 @@ angular.module('calcentral.controllers').controller('LanguagesSectionController'
     languages = _.map(languages, function(language) {
       if (!language.levels) {
         language.levels = _.filter(levelMapping, function(value, level) {
-          if (_.isBoolean(language[level])) {
-            return value;
-          }
+          return language[level] && value;
         });
       }
 
@@ -170,8 +168,11 @@ angular.module('calcentral.controllers').controller('LanguagesSectionController'
   };
 
   var parsePerson = function(data) {
-    var person = data.data.feed.student;
-    var languages = parseLanguages(person.languages);
+    var languages = parseLanguages(_.get(data, 'data.feed.student.languages'));
+
+    languages = _.sortBy(languages, function(o) {
+      return o.name;
+    });
 
     angular.extend($scope, {
       items: {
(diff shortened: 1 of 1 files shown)


src/assets/javascripts/angular/controllers/widgets/profile/languagesSectionController.js
```diff
@@ -215,15 +216,19 @@ angular.module('calcentral.controllers').controller('LanguagesSectionController'
...
     $scope.isLoading = false;
   };
 
   var loadInformation = function(options) {
+    // Create a fresh promise each time loadInformation is called because a
+    // resolved promise can't be re-used. Note that the `url` field must be
+    // specified in order for the underlying httpService.request() call to work.
+    var getPerson = profileFactory.getPerson({
+      refreshCache: true,
+      url: '/api/edos/student'
```
christianvuerings: @dfkaye URL's should be part of the factory, not in the controller.

src/assets/javascripts/angular/controllers/widgets/profile/languagesSectionController.js
```diff
@@ -215,15 +216,19 @@ angular.module('calcentral.controllers').controller('LanguagesSectionController'
...
-  }).then(parsePerson);
-
   var loadComplete = function() {
     $scope.isLoading = false;
   };
 
   var loadInformation = function(options) {
+    // Create a fresh promise each time loadInformation is called because a
+    // resolved promise can't be re-used. Note that the `url` field must be
+    // specified in order for the underlying httpService.request() call to work.
```
christianvuerings: @dfkaye interesting, let's try to find a better way to fix this.

src/assets/javascripts/angular/controllers/widgets/profile/languagesSectionController.js
dfkaye: @christianvuerings - just doublechecked - works without that field now

src/assets/javascripts/angular/controllers/widgets/profile/languagesSectionController.js
dfkaye: @christianvuerings what do you have in mind?

christianvuerings: @dfkaye something like how we do it in `profileAddressController`:

https://github.com/ets-berkeley-edu/calcentral/blob/2a7ef11400f1c8de1d88a18a7e2cf2b6b1c4434e/src/assets/javascripts/angular/controllers/widgets/profile/profileAddressController.js#L136-L160

So it's ok to define the `getPerson` method outside of it, we should just make sure we only call it within `loadInformation`.

christianvuerings: @dfkaye looking at that code some more, I think we should be fine with also getting rid of `if (_.get(options, 'fetchCodes') === true) {` since it won't make an extra Ajax call to fetch the code again since it's cached.

dfkaye: @christianvuerings yay, i have it working now ~ looks a little different, and less code to boot..

dfkaye:
@christianvuerings loadInfo changes pushed... awaiting travis...

dfkaye:
@christianvuerings `loadInformation()` *et alia*., ready for review&hellip;

dfkaye:
&amp;hellip; vs. &hellip;

christianvuerings:
@dfkaye thanks!

PR #5002, opened 2016-03-03
SISRP-13242 Incorporate Hub academic status data into My Academics profile

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-13242

As directed, we continue to use legacy sources for eight-digit IDs and switch to the Hub for ten-digit IDs. Loath as I am to create one more feature flag, it seems prudent to give us an emergency switchoff if needed.

Since our test data currently lacks variety, minors are not yet handled, and cases such as double majors are not yet tested.

The length-based determination of whether an ID counts as legacy is rehomed in the User::Student module. The module is also nudged toward more frequent caching of looked-up CS IDs, without a full-on refactor.


app/models/user/student.rb
```diff
@@ -16,6 +20,11 @@ def lookup_delegate_user_id
       CalnetCrosswalk::ByUid.new(user_id: @uid).lookup_delegate_user_id
     end
 
+    def legacy_campus_solutions_id?
```
johncrossman: @paulkerschen without the `?` I'd assume this a "getter" method. Would `has_legacy_campus_solutions_id?` be closer to conventional naming? (And more in line w/ `@edo_attributes` key?)

spec/models/my_academics/gpa_units_spec.rb
```diff
@@ -0,0 +1,68 @@
+describe 'MyAcademics::GpaUnits' do
+
```
johncrossman: Nice overall test coverage, @paulkerschen

johncrossman:
Looks good, @paulkerschen  Good to merge once you comment on the above.

app/models/user/student.rb
raydavis: I think the Ruby convention is that the "?" eliminates the need for a "has_" or "is_", although I'm often tempted to go against it.

raydavis: Maybe `legacy_student?` would convey the functional intention?

raydavis:
Aside from the evergreen topic of naming, this looks like a coherent-as-possible depiction of a well-muddied situation. Thanks for wading through it, @paulkerschen .

app/models/user/student.rb
paulkerschen: The only problem with `legacy_student?` as a method name is the same problem with `User::Student` as a module name (and "student EDO" as our data source...); that it applies to everyone, not just students. I foresee a future refactor trying to mitigate this.

Anyhow, I think @johncrossman's more explicit `has_legacy_campus_solutions_id?` is reasonable, and will amend.

paulkerschen: Or, more concisely, perhaps `legacy_user?` with a corresponding `:isLegacyUser` key cuts to the chase.

paulkerschen:
Thanks for conscientious review, @johncrossman and @raydavis. Travis has blessed an amendment with `legacy_user?` method name and `:is_legacy_user` attributes key.

raydavis:
Wouldn't `valued_customer?` be friendlier?

PR #5003, opened 2016-03-03
SISRP-15761 - Delegated Access: college scheduler link is available to delegate

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-15761

PR #5004, opened 2016-03-03
SISRP-15761, college scheduler link is available to delegate

johncrossman:
https://jira.berkeley.edu/browse/SISRP-15761

Filter array of sections based on view-as restrictions. And tell me if I'm wrong: the `data-ng-if` in `schedule.html` was not serving a purpose. The `toggleShow` is taken care of in `enrollment_card.html`. In my testing, that `data-ng-if` was ineffectual.


src/assets/javascripts/angular/controllers/widgets/enrollment/enrollmentCardController.js
```diff
@@ -49,6 +21,41 @@ angular.module('calcentral.controllers').controller('EnrollmentCardController',
...
+        title: 'Plan'
+      },
+      {
+        id: 'explore',
+        title: 'Explore'
+      },
+      {
+        id: 'schedule',
+        title: 'Schedule',
+        omit: !!apiService.user.profile.delegateActingAsUid
```
christianvuerings: @johncrossman although I do like this approach, it's different from what we've done before by hiding links within each section. E.g. I can definitely imagine a parent saying something like `have you completed section 5 of your enrollment already`?

src/assets/templates/widgets/enrollment/schedule.html
```diff
@@ -5,7 +5,7 @@
   data-title="section.title"
 ></div>
 
-<div class="cc-enrollment-card-section-content" data-ng-if="section.show">
+<div class="cc-enrollment-card-section-content">
```
christianvuerings: @johncrossman intesting, removing this does break the toggle functionality on `schedule` section.
`toggleShow` will only set the `.show` value to `true` or `false`.

FYI the reason why we couldn't handle this a level higher is because of:
https://github.com/ets-berkeley-edu/calcentral/blob/2a7ef11400f1c8de1d88a18a7e2cf2b6b1c4434e/src/assets/templates/widgets/enrollment/meet_advisor.html#L8-L18

https://github.com/ets-berkeley-edu/calcentral/blob/2a7ef11400f1c8de1d88a18a7e2cf2b6b1c4434e/src/assets/templates/widgets/enrollment/meet_advisor.html#L20

In the meet advisor, we always need to show them their `holds` status, even in collapsed state.

johncrossman:
Thanks @christianvuerings  I'm closing this in favor of #5003  I agree that showing parents "section 5", albeit stripped down, has value.

PR #5006, opened 2016-03-03
SISRP-15504 - Grade base link vs add/drop/swap

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-15504

**NOTE**: Only merge this evening around 5pm!


christianvuerings:
@raydavis @paulkerschen @johncrossman this is the 5pm merge for today.

Would be great if we can review it before that but only merge at 5pm (or after).

paulkerschen:
I think this is good to merge when the clock strikes, @christianvuerings.

christianvuerings:
Going to do this self merge so we don't forget about it.

PR #5010, opened 2016-03-03
SISRP-13891 Cache non-user-specific CS APIs globally

raydavis:
https://jira.berkeley.edu/browse/SISRP-13891

The primary task was to make the FERPA link feed a global value, with the added twist that the global value is currently only reachable with a CS EmplID.

While doing that, though, I noticed that many other CS global classes were wrongly being pinged from user-specific cache expirations, and so I added a GlobalCachedProxy superclass.


paulkerschen:
Thanks for the extra cleanup, @raydavis.

PR #5011, opened 2016-03-03
SISRP-15812, more robust auth for oec.admin

johncrossman:
https://jira.berkeley.edu/browse/SISRP-15812

Thank you @redconfetti for reporting this bug. Auth is strengthened and we log error if config is blank.


paulkerschen:
"If no one is the OEC administrator, everyone is the OEC administrator." Thanks @johncrossman

PR #5015, opened 2016-03-04
SISRP-13248 Wire up additional expiry behavior to JMS listener

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-13248

### DIFF
--- a/app/models/campus_solutions/user_api_updating_model.rb
+++ b/app/models/campus_solutions/user_api_updating_model.rb
@@ -3,7 +3,6 @@ module CampusSolutions
     def passthrough(model_name, params)
       proxy = model_name.new({user_id: @uid, params: params})
       result = proxy.get
-      PersonDataExpiry.expire @uid
       UserApiExpiry.expire @uid
       result
     end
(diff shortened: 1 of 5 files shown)


app/models/campus_solutions/user_api_updating_model.rb
```diff
@@ -3,7 +3,6 @@ module UserApiUpdatingModel
     def passthrough(model_name, params)
       proxy = model_name.new({user_id: @uid, params: params})
       result = proxy.get
-      PersonDataExpiry.expire @uid
```
paulkerschen: (This one is removed because adding HubEdos::MyStudent to UserApiExpiry above has the same effect.)

app/models/notifications/sis_expiry_processor.rb
```diff
@@ -41,7 +41,7 @@ def get_uid(event)
       'sis:student:ferpa' => nil,
       'sis:student:finaid' => CampusSolutions::FinancialAidExpiry,
       'sis:student:financials' => nil,
-      'sis:student:messages' => nil
+      'sis:student:messages' => MyActivities::Merged
```
johncrossman: As promised, we swap out the 'nil' values. Thanks, @paulkerschen

johncrossman:
image

PR #5016, opened 2016-03-04
SISRP-15849, oec auth policy w.r.t view-as mode

johncrossman:
https://jira.berkeley.edu/browse/SISRP-15849

One line tweak plus tidying in spec.

### DIFF
--- a/spec/policies/authentication_state_policy_spec.rb
+++ b/spec/policies/authentication_state_policy_spec.rb
@@ -1,8 +1,16 @@
 describe AuthenticationStatePolicy do
   let(:session_state) do
-    {'user_id' => user_id, SessionKey.original_user_id => original_user_id, 'lti_authenticated_only' => lti_authenticated_only}
+    {
+      'user_id' => user_id,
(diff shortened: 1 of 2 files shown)


spec/policies/authentication_state_policy_spec.rb
```diff
@@ -34,153 +42,167 @@
   describe '#access_google?' do
     context 'as self' do
       let(:user_id) {average_joe_uid}
-      its(:access_google?) { should be_truthy }
+      its(:access_google?) { is_expected.to be true }
```
paulkerschen: FWIW @johncrossman, I think this style of `should` is not deprecated because, unlike the `object.should be ...` pattern, it doesn't involve patching new methods onto objects. No harm in consistency though.

paulkerschen:
+1

spec/policies/authentication_state_policy_spec.rb
johncrossman: Good to know, @paulkerschen  Thx

PR #5018, opened 2016-03-04
CLC-6229 sync-canvas-new-users should check for "inactive-" bCourses accounts

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6229

Also brushed some dust off the specs.

### DIFF
--- a/app/models/canvas_csv/add_new_users.rb
+++ b/app/models/canvas_csv/add_new_users.rb
@@ -71,7 +71,9 @@ module CanvasCsv
       all_active_sis_user_uids = CampusOracle::Queries.get_all_active_people_uids.to_set
       all_current_canvas_uids = []
       CSV.foreach(get_canvas_user_report_file, headers: :first_row) do |canvas_user|
-        all_current_canvas_uids << canvas_user['login_id']
+        if (existing_ldap_uid = MaintainUsers.parse_login_id(canvas_user['login_id'])[:ldap_uid])
+          all_current_canvas_uids << existing_ldap_uid.to_s
+        end
       end
       all_active_sis_user_uids.subtract(all_current_canvas_uids).to_a
     end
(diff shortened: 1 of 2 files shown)


app/models/canvas_csv/add_new_users.rb
```diff
@@ -71,7 +71,9 @@ def new_active_user_uids
       all_active_sis_user_uids = CampusOracle::Queries.get_all_active_people_uids.to_set
       all_current_canvas_uids = []
       CSV.foreach(get_canvas_user_report_file, headers: :first_row) do |canvas_user|
-        all_current_canvas_uids << canvas_user['login_id']
+        if (existing_ldap_uid = MaintainUsers.parse_login_id(canvas_user['login_id'])[:ldap_uid])
```
johncrossman: @raydavis Just so I'm clear... We don't care about truthiness of `[:inactive_account]` here?

spec/models/canvas_csv/add_new_users_spec.rb
```diff
@@ -98,7 +96,7 @@
       expect(csv_array[0]).to eq %w(canvas_user_id user_id login_id first_name last_name full_name email status)
       expect(csv_array[1][2]).to eq '946123'
       expect(csv_array[2][2]).to eq '946124'
-      expect(csv_array[3][2]).to eq '946125'
+      expect(csv_array[3][2]).to eq 'inactive-946125'
```
johncrossman: This change makes `946125` inactive and yet the uid is still in list of `sis_active_uids`. All is correct?

app/models/canvas_csv/add_new_users.rb
raydavis: Right, @johncrossman . This script does not concern itself with inactivating or reactivating _existing_ Canvas accounts. Its only job is to create _new_ Canvas accounts for LDAP UIDs which don't already have one (regardless of inactive state).

johncrossman:
Makes sense. Thx, @raydavis

spec/models/canvas_csv/add_new_users_spec.rb
raydavis: It's kind of a dopy test, since it's just checking that our method to download a CSV file _from_ Canvas actually does read the downloaded CSV file. Since we've said that Canvas has a user account with Login ID 'inactive-946215', that Login ID should dutifully appear in the CSV export.

PR #5020, opened 2016-03-05
SISRP-15767, HigherOne GET URL API to optional delegate_uid

johncrossman:
https://jira.berkeley.edu/browse/SISRP-15767
https://jira.berkeley.edu/browse/SISRP-15658

Resulting URL from CS will be appropriate to user session.

### DIFF
--- a/app/controllers/campus_solutions/higher_one_url_controller.rb
+++ b/app/controllers/campus_solutions/higher_one_url_controller.rb
@@ -2,8 +2,15 @@ module CampusSolutions
   class HigherOneUrlController < CampusSolutionsController
 
     def get
-      model = CampusSolutions::MyHigherOneUrl.from_session(session)
+      model = CampusSolutions::MyHigherOneUrl.from_session proxy_args
       render json: model.get_feed_as_json
+    end
+
+    private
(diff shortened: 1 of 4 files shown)


christianvuerings:
@johncrossman lgtm

Also looks like this might be a duplicate of https://jira.berkeley.edu/browse/SISRP-15733. In there you'll see that this PR is only the first fix.

The second fix should be to make sure that we use the same logic as with the College Scheduler URL, which would make sure that the front-end no longer has to pull.

We would make a request to `/api/campus_solutions/higher_one_url` which would then, with no caching (the URL expires after 5min), fetch the CS API. That API contains the URL for HigherOne which we then redirect to.

FYI - talked to @paulkerschen about this before.

johncrossman:
Thanks, @christianvuerings I missed SISRP-15733. The timing of this PR is meant to coincide with the QAT migration, per release mgmt meeting. @paulkerschen can merge this if the work is headed in the right direction.

paulkerschen:
Looks good to me, @johncrossman, thanks! Fetching and redirecting on request can be done as a follow-up.

PR #5024, opened 2016-03-07
SISRP-15749 - Enrollment card: season and year are transposed

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-15749

### DIFF
--- a/src/assets/javascripts/angular/controllers/widgets/enrollment/enrollmentCardController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/enrollment/enrollmentCardController.js
@@ -69,6 +69,19 @@ angular.module('calcentral.controllers').controller('EnrollmentCardController', 
   };
 
   /**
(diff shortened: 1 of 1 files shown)


paulkerschen:
@christianvuerings: The fix makes sense as code, but do we really want the front end to be in the business of swapping around strings from the API? Seems like it would be far preferable to have the API return the right value; or, if there's really no way around reformatting it ourselves, do it on the back end and ensure it's always a year that comes last.

christianvuerings:
@paulkerschen tried the API way but they really don't want to change it since it impacts all the different functional modules. I do agree on the back-end approach and it would probably be good to do that as a follow up to this PR. Would you be ok with taking this on?

paulkerschen:
Sure, @christianvuerings, I'll take on the back-end fix.

christianvuerings:
Back-end PR at #5028

PR #5027, opened 2016-03-07
CLC-6230 Guard Bearfacts proxies against calls with CS IDs

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6230


johncrossman:
Pretty darn nice, @paulkerschen

PR #5000, opened 2016-03-02
SISRP-15754 Show applicants the same My Campus links as students

raydavis:
https://jira.berkeley.edu/browse/SISRP-15754

The campuslinks.json changes add a '"applicant": true,' line under every '"student": true,' line.

PR #5030, opened 2016-03-08
SISRP-15960, My Campus and Financial Resources links are not showing for ex-students

johncrossman:
https://jira.berkeley.edu/browse/SISRP-15960

Includes:
* a spec that tests only ex-student scenarios
* code cleanup

### DIFF
--- a/app/models/links/my_campus_links.rb
+++ b/app/models/links/my_campus_links.rb
@@ -2,7 +2,6 @@ module Links
   class MyCampusLinks
 
 =begin
-
   NOTES:
 
   - Navigation consists of Main Categories, Subcategories, and On-page categories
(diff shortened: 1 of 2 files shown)


johncrossman:
*Note:* This PR restores logic that was displaced by PR #5000

app/models/links/my_campus_links.rb
```diff
@@ -17,67 +16,66 @@ class MyCampusLinks
...
-        @section = {
-          "label" => cat.name,
-          "categories" => get_subsections_for_nav(cat)
-        }
-        navigation.push(@section)
-
+      Links::LinkCategory.where('root_level = ?', true).order('LOWER(name)').each do |category|
+        navigation.push({
+                          'label' => category.name,
+                          'categories' => get_subsections_for_nav(category)
```
paulkerschen: @johncrossman As long as we're fixing up the formatting, here and below, should the line-continuation hashes have only two-space indents?

paulkerschen:
Makes sense, @johncrossman, and thanks for the spec. One q on formatting.

app/models/links/my_campus_links.rb
johncrossman: Fixed. Thanks, @paulkerschen

PR #5035, opened 2016-03-08
CLC-15999 Point legacy_user? check to LDAP if CS ID is unavailable

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-15999 should have been addressed in #4973, but this gives us a fighting chance to verify outside of production.

Also ensure that Hub-sourced academic data includes a term name even if nothing else.

### DIFF
--- a/app/models/my_academics/college_and_level.rb
+++ b/app/models/my_academics/college_and_level.rb
@@ -41,6 +41,7 @@ module MyAcademics
         response[:termName] = parse_hub_term_name status
       else
         response[:empty] = true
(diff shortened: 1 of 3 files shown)


raydavis:
@paulkerschen , this looks like expert patching of this leaky vessel but Bamboo reports a suspiciously germane testext failure. Can you take a look?

raydavis:
Spoke too soon, @paulkerschen -- the Bamboo failure seems unrelated. Sorry for the false alarm!

PR #5036, opened 2016-03-08
SISRP-16031 Stop pretending we can predict CS FERPA URL paths

raydavis:
https://jira.berkeley.edu/browse/SISRP-16031

The RSpec change is backwards-compatible (just in case the CS API reverts to the old payload format).


raydavis:
Should fix Bamboo.

christianvuerings:
+1

johncrossman:
Thanks @raydavis  Bamboozled no more.

PR #5037, opened 2016-03-08
SISRP-14774 Identify Law students from Campus Solutions

raydavis:
https://jira.berkeley.edu/browse/SISRP-14774

Replacement for BearFacts-API-dependent logic.

For now, the Academics profile continues to rely on academic status feeds rather than affiliations.

### DIFF
--- a/spec/models/my_badges/student_info_spec.rb
+++ b/spec/models/my_badges/student_info_spec.rb
@@ -76,7 +76,7 @@ describe 'MyBadges::StudentInfo' do
     end
   end
 
-  context 'for Law student users' do
+  context 'for legacy Law student users' do
     before { Bearfacts::Proxy.any_instance.stub(:lookup_student_id).and_return(99999997) }
     let! (:law_proxy) { Bearfacts::Profile.new({user_id: '212381', fake: true}) }
     before do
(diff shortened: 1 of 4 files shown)


spec/models/my_badges/student_info_spec.rb
```diff
@@ -148,4 +148,56 @@
...
+                  {'academicCareer'=>{'code'=>'UGRD', 'description'=>'Undergraduate'},
+                    'fromDate'=>'2015-08-19'},
+                  'studentPlans' => []
+                }]
+              }
+            }
+        }
+      end
+      it 'does not set isLawStudent' do
+        subject[:isLawStudent].should be_falsey
```
paulkerschen: @raydavis FYI, RSpec has deprecated `should` as a stubbed method and we are slowly and painfully overhauling in favor of the `expect(subject[:isLawStudent]).to ...` syntax.

Also, might `eq true` and `eq false` be more precise than `be_truthy` and `be_falsey`, since we really are checking for boolean values?

paulkerschen:
Thanks, @raydavis, logic looks right to me. One query on house style in specs.

spec/models/my_badges/student_info_spec.rb
raydavis: D'oh! Copy-and-paste took mental precedence over the new spec style. I'll fix that and add true/false checks.

paulkerschen:
the-clash-i-fought-the-law-23169

PR #5039, opened 2016-03-09
SISRP-15830, proxy can have custom HTTP-timeout setting

johncrossman:
https://jira.berkeley.edu/browse/SISRP-15830

Notes:
* I've put default 30 second timeout on Hub, CS, Crosswalk. That value matches what was in use before: `application.outgoing_http_timeout`
* BaseProxy intercepts the call to `get_response()` and injects optional timeout. This seemed the most elegant way to avoid code duplication in subclasses. Notice that `Proxies::HttpClient` does not have a handle on the Settings object.
* In a moment I'll describe steps to verify in Jira above.


paulkerschen:
Thanks, @johncrossman; a nice addition to the toolkit.

PR #5040, opened 2016-03-09
SISRP-10786 - Front End Dev: Credentials Tab - Work Experience Card - Edit

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-10786

Also includes Delete functionality.  Apologies for the size!
Travis, please be kind.


app/models/campus_solutions/my_work_experience.rb
```diff
@@ -11,5 +13,14 @@ def delete(params = {})
       passthrough(CampusSolutions::WorkExperienceDelete, params)
     end
 
+    # Formats input date to Campus Solutions acceptable ISO 8601 (YYYY-MM-DD)
+    # Front-end date validator ensures date will be in '%m/%d/%Y' format
+    def cs_date_formatter(date)
+      if date == ''
+        return date
+      end
+      return Date.strptime(date, '%m/%d/%Y').strftime('%F')
```
paulkerschen: @davidlimcheng Looks like the Travis failure is caused by a mismatch between this date format (using slashes) and the strings provided in the controller spec (using dashes): https://github.com/ets-berkeley-edu/calcentral/blob/master/spec/controllers/campus_solutions/work_experience_controller_spec.rb#L29

Even with front-end validation, it's still usually a good idea to wrap date parsing in a `begin... rescue ArgumentError` for safety's sake. These passthrough models tend not to have specs of their own, but since this once now contains some logic, it ought to have a corresponding test class. I would have it exercise this method on a few different strings: blank, properly formatted date, improperly formatted date that somehow got past the front end.

Finally :), note that `return` on the last line of a Ruby method can (and almost always should) be omitted.

app/models/campus_solutions/my_work_experience.rb
```diff
@@ -11,5 +13,14 @@ def delete(params = {})
       passthrough(CampusSolutions::WorkExperienceDelete, params)
     end
 
+    # Formats input date to Campus Solutions acceptable ISO 8601 (YYYY-MM-DD)
+    # Front-end date validator ensures date will be in '%m/%d/%Y' format
+    def cs_date_formatter(date)
+      if date == ''
+        return date
+      end
```
paulkerschen: @davidlimcheng The Rails `.blank?` method is idiomatic for this sort of check: in addition to empty strings it catches nil, whitespace strings, empty arrays, etc. Also, it's standard to use a Perl-style appended `if` when only one line is affected: instead of a three-line block, you can simply write `return date if date.blank?`

app/models/campus_solutions/my_work_experience.rb
```diff
@@ -1,9 +1,11 @@
 module CampusSolutions
   class MyWorkExperience < UserSpecificModel
 
-    include PersonDataUpdatingModel
+    include WorkExperienceUpdatingModel
 
     def update(params = {})
+      params[:startDt] = cs_date_formatter (params[:startDt])
+      params[:endDt] = cs_date_formatter (params[:endDt])
```
paulkerschen: @davidlimcheng It's usual in Ruby to omit parentheses if there's only one method argument. In cases where parentheses are preferable, there shouldn't be a space between method name and open parenthesis, as this sometimes confuses the parser.

src/assets/javascripts/angular/controllers/widgets/profile/workExperienceSectionController.js
```diff
@@ -8,12 +8,50 @@ var _ = require('lodash');
...
+    currencies: [],
+    currentObject: {},
+    emptyObject: {
+      country: 'USA',
+      currencyType: 'USD',
+      payFrequency: 'M',
+      startDt: '',
+      endDt: ''
+    },
+    employFracValues: [
```
christianvuerings: @davidlimcheng not a blocker but could we open a ticket to investigate whether these are also XLAT values?

src/assets/javascripts/angular/controllers/widgets/profile/workExperienceSectionController.js
```diff
@@ -53,19 +91,211 @@ angular.module('calcentral.controllers').controller('WorkExperienceSectionContro
...
+    $scope.countries = _.sortBy(_.filter(_.get(data, 'data.feed.countries'), {
+      hasAddressFields: true
+    }), 'descr');
+  };
+
+  var parseAddressFields = function(data) {
+    /**
+     * We only want the City and State fields from the AddressFields
+     */
+    var returnCityState = function(fields) {
```
christianvuerings: @davidlimcheng can we put this outside of `parseAddressFields`? Usually, we try to avoid nesting functions inside of functions

src/assets/javascripts/angular/controllers/widgets/profile/workExperienceSectionController.js
```diff
@@ -53,19 +91,211 @@ angular.module('calcentral.controllers').controller('WorkExperienceSectionContro
...
+  var parseTypesPayFrequency = function(data) {
+    $scope.typesPayFrequency = _.get(data, 'data.feed.xlatvalues.values');
+  };
+
+  /**
+   * Removes previous address data, we need to do this every time you change the country
+   */
+  var removePreviousAddressData = function() {
+    $scope.currentObject.data = _.fromPairs(_.map($scope.currentObject.data, function(value, key) {
+      if (['sequenceNbr', 'employmentDescr', 'country', 'currencyType', 'payFrequency', 'phone', 'startDt', 'endDt', 'titleLong', 'employFrac', 'endingPayRate', 'hoursPerWeek'].indexOf(key) === -1) {
```
christianvuerings: @davidlimcheng could we alphabetize these? Makes it easier to read and find things.

src/assets/javascripts/angular/controllers/widgets/profile/workExperienceSectionController.js
```diff
@@ -53,19 +91,211 @@ angular.module('calcentral.controllers').controller('WorkExperienceSectionContro
...
+    if (!countryCode) {
+      return;
+    }
+    if (!initialEdit.load) {
+      removePreviousAddressData();
+      removeErrorMessage();
+    }
+    $scope.currentObject.stateFieldLoading = true;
+    initialEdit.load = false;
+    // $scope.currentObject.data = {};
```
christianvuerings: @davidlimcheng do we still need this comment?

src/assets/javascripts/angular/controllers/widgets/profile/workExperienceSectionController.js
```diff
@@ -53,19 +91,211 @@ angular.module('calcentral.controllers').controller('WorkExperienceSectionContro
...
+    .then(parseTypesPayFrequency)
     .then(function() {
       $scope.isLoading = false;
     });
   };
 
+  /**
+   * We need to map the address fields for the current country and the ones (depending on country) that the user has entered
+   */
+  var matchFields = function(fields, item) {
```
christianvuerings: @davidlimcheng looks like this method is a complete copy/paste from the `profileAddressController` - can we move it to the `profileService` instead?

Might be good to also look at all the other methods in this file, to see whether we can keep avoid the copy/pasting when possible.

src/assets/javascripts/angular/controllers/widgets/profile/workExperienceSectionController.js
```diff
@@ -53,19 +91,211 @@ angular.module('calcentral.controllers').controller('WorkExperienceSectionContro
...
+  };
+
+  $scope.showAdd = function() {
+    apiService.profile.showAdd($scope, $scope.emptyObject);
+  };
+
+  $scope.showEdit = function(item) {
+    apiService.profile.showEdit($scope, item);
+    $scope.currentObject.data = {
+      sequenceNbr: item.id,
```
christianvuerings: @davidlimcheng As discussed - could we open up a follow up ticket to clean this up?

src/assets/templates/widgets/profile/work_experience_editor.html
```diff
@@ -0,0 +1,185 @@
...
+    <form data-ng-submit="save(currentObject.data)" name="cc_page_widget_profile_work_experience">
+
+      <div>
+        <label for="cc-page-widget-profile-work-experience-employment-descr">
+            <strong>Employer</strong> <span class="cc-text-light">(Required)</span>
+        </label>
+        <input type="text" data-ng-model="currentObject.data.employmentDescr" id="cc-page-widget-profile-work-experience-employment-descr" required aria-required="true" />
+      </div>
+
+    <br/>
```
christianvuerings: @davidlimcheng Could we use something like `cc-page-widget-profile-margin-top` on the `div` instead of a `<br />`?

Btw - also check indentation everywhere in this file.

src/assets/templates/widgets/profile/work_experience_editor.html
```diff
@@ -0,0 +1,185 @@
...
+            <strong>Country</strong> <span class="cc-text-light">(Required)</span>
+          </label>
+        </div>
+        <div class="cc-select">
+          <select data-ng-options="country.country as country.descr for country in countries" data-ng-model="currentObject.data.country" id="cc-page-widget-profile-work-experience-country" required aria-required="true" />
+        </div>
+      </div>
+
+      <div data-cc-spinner-directive="currentObject.stateFieldLoading">
+        <div data-ng-repeat="field in currentObject.fields">
```
christianvuerings: @davidlimcheng is this exactly the same code as for the address html? If so, could we abstract it into a template and use `ng-include`?

src/assets/templates/widgets/profile/work_experience_editor.html
```diff
@@ -0,0 +1,185 @@
...
+            data-ng-model="currentObject.data.startDt"
+            name="cc_page_widget_profile_work_experience_start_date"
+            id="cc-page-widget-profile-work-experience-start-date"
+            placeholder="mm/dd/yyyy"
+            data-cc-date-validator-directive />
+          <p class="cc-alert cc-alert-text" role="alert" data-ng-show="cc_page_widget_profile_work_experience.cc_page_widget_profile_work_experience_start_date.$error.ccDateValidator">
+            Please use mm/dd/yyyy date format
+          </p>
+        </div>
+        <div class="medium-2 small-2 columns">
```
christianvuerings: @davidlimcheng did you also test this on every screen size, to make sure it always looks good?

src/assets/templates/widgets/profile/work_experience_editor.html
```diff
@@ -0,0 +1,185 @@
...
+          </button>
+        </div>
+      </div>
+
+      <div>
+        <label for="cc-page-widget-profile-work-experience-end-date">
+          <strong>End Date</strong>
+        </label>
+      </div>
+      <div class="row collapse">
```
christianvuerings: @davidlimcheng I know we used this before, but could we see whether we can use `cc-flex` instead?

src/assets/templates/widgets/profile/work_experience_editor.html
```diff
@@ -0,0 +1,185 @@
...
+          <select data-ng-options="payfreq.fieldvalue as payfreq.xlatshortname for payfreq in typesPayFrequency" data-ng-model="currentObject.data.payFrequency" id="cc-page-widget-profile-work-experience-pay-frequency" required aria-required="true" />
+        </div>
+      </div>
+
+    <br/>
+
+      <div>
+        <button class="cc-button cc-button-blue" type="submit" data-ng-disabled="cc_page_widget_profile_work_experience.$invalid || isSaving || currentObject.stateFieldLoading">
+          <span data-ng-if="!isSaving">Save</span>
+          <span data-ng-if="isSaving">Updating...</span>
```
christianvuerings: @davidlimcheng I think @dfkaye changed all of the `...` to `&hellip;`, we should probably do the same here.

src/assets/templates/widgets/profile/work_experience_editor.html
```diff
@@ -0,0 +1,185 @@
...
+
+  </form>
+
+  <form data-ng-submit="delete(currentObject.data)" data-ng-if="!currentObject.isAdding">
+    (<button class="cc-button-link" type="submit" data-ng-disabled="isDeleting">Delete work experience</button>)
+  </form>
+
+  </div>
+  <div data-ng-if="isDeleting">
+    <i class="fa fa-spinner fa-spin"></i> Deleting...
```
christianvuerings: @davidlimcheng see above `...`

src/assets/templates/widgets/profile/work_experience_section.html
```diff
@@ -3,19 +3,33 @@
...
-        <span data-ng-bind="item.payFrequency.toDate"></span>
+      <div data-ng-if="!item.isModifying">
+        <div>
+          <strong data-ng-bind="item.employer"></strong>
+        </div>
+        <div data-ng-if="item.payFrequency.fromDate || item.payFrequency.toDate">
+          <span data-ng-bind="item.payFrequency.fromDate"></span>
+          <span data-ng-if="!item.payFrequency.fromDate">No start date</span>
+          -
+          <span data-ng-bind="item.payFrequency.toDate"></span>
```
christianvuerings: @davidlimcheng do we always have a `toDate`? (If not, we should probably make the `-` conditional)

christianvuerings:
@davidlimcheng also try to be friends with Travis, he's a pretty cool guy.

```
Failures:

  1) CampusSolutions::WorkExperienceController updating work experience authenticated user should let an authenticated user post
     Failure/Error: post :post,
     ArgumentError:
       invalid date
     # ./app/models/campus_solutions/my_work_experience.rb:22:in `cs_date_formatter'
     # ./app/models/campus_solutions/my_work_experience.rb:7:in `update'
     # ./app/controllers/campus_solutions/campus_solutions_controller.rb:14:in `post_passthrough'
     # ./app/controllers/campus_solutions/work_experience_controller.rb:7:in `post'
     # ./spec/controllers/campus_solutions/work_experience_controller_spec.rb:17:in `(root)'

Finished in 2 minutes 59.8 seconds (files took 22.41 seconds to load)
```

app/models/campus_solutions/my_work_experience.rb
davidlimcheng: Thanks @paulkerschen!  I'll do a bit of research on those keywords & include the fix in the next push.

app/models/campus_solutions/my_work_experience.rb
davidlimcheng: @paulkerschen so much cleaner - will include in the next push.

app/models/campus_solutions/my_work_experience.rb
davidlimcheng: @paulkerschen Got it!

src/assets/javascripts/angular/controllers/widgets/profile/workExperienceSectionController.js
davidlimcheng: @christianvuerings not sure I'm following you - you mean the data coming back from the APIs?  Out of all of them, I believe only `typesPayFrequency` comes from an XLAT table, the rest were custom APIs.

src/assets/javascripts/angular/controllers/widgets/profile/workExperienceSectionController.js
davidlimcheng: @christianvuerings got it!  Will do so in next push.

src/assets/javascripts/angular/controllers/widgets/profile/workExperienceSectionController.js
davidlimcheng: @christianvuerings For sure, I'll also alphabetize other areas where I can to make it more readable, e.g. `initialState` declaration.

src/assets/javascripts/angular/controllers/widgets/profile/workExperienceSectionController.js
davidlimcheng: @christianvuerings Will remove!

src/assets/javascripts/angular/controllers/widgets/profile/workExperienceSectionController.js
davidlimcheng: @christianvuerings Sure thing.  I'll move these methods to the `profileService`, and start calling that - I'm guessing it would be best to change the `profileAddressController` methods to the Service as well, in a different PR?  If so I'll open a new JIRA for that.

src/assets/javascripts/angular/controllers/widgets/profile/workExperienceSectionController.js
davidlimcheng: @christianvuerings Sure thing.  Just to be clear, I will:

1. Move these methods to `profileService`, and change `workExperienceSectionController` to call from `profileService.
2. Open a JIRA ticket to change `profileAddressController` shared functions to call from `profileService` as well.
3. Make those fixes in a separate PR.

Does this sound OK to you?

src/assets/templates/widgets/profile/work_experience_editor.html
davidlimcheng: @christianvuerings ahh that makes sense - will do.

src/assets/templates/widgets/profile/work_experience_editor.html
davidlimcheng: @christianvuerings The `country` div is exactly the same, but the `fields` div unfortunately has a `<strong>` tag on the label that `contact_address_editor.html` does not.  Should I abstract just the `country` div onto a separate template?

src/assets/templates/widgets/profile/work_experience_editor.html
davidlimcheng: @christianvuerings I did not!  But I will do so before my next push.

src/assets/templates/widgets/profile/work_experience_editor.html
davidlimcheng: @christianvuerings Sure thing, I'll look into this.

src/assets/templates/widgets/profile/work_experience_editor.html
davidlimcheng: @christianvuerings sounds good.

src/assets/templates/widgets/profile/work_experience_section.html
davidlimcheng: @christianvuerings That's good stuff, will include in the next push.

src/assets/javascripts/angular/controllers/widgets/profile/workExperienceSectionController.js
christianvuerings: @davidlimcheng almost every single dropdown in CS is based on XLAT values. So we should investigate whether we should also get these values from the API instead of hardcoding them in CalCentral

src/assets/javascripts/angular/controllers/widgets/profile/workExperienceSectionController.js
christianvuerings: @davidlimcheng it's fine for me to also change it as part of this PR, so we don't forget this cleanup.

src/assets/javascripts/angular/controllers/widgets/profile/workExperienceSectionController.js
christianvuerings: @davidlimcheng could you be a bit more precise in what you mean with "these methods"?
The only ones that should be moved to `profileService` are the ones that are being used by multiple controllers.

The clean up I was talking about was the one we mentioned yesterday, to bring these names in sync.

src/assets/templates/widgets/profile/work_experience_editor.html
christianvuerings: @davidlimcheng or, even better, we can make that conditional. If it's loaded in the regular address, no `<strong>`, if it's loaded within work experience, do use a `<strong>`

src/assets/javascripts/angular/controllers/widgets/profile/workExperienceSectionController.js
davidlimcheng: Ahh you're right, it was just `matchFields()` we talked about.  My bad, and yes, I'll include it as part of this PR.

src/assets/javascripts/angular/controllers/widgets/profile/workExperienceSectionController.js
davidlimcheng: @christianvuerings I see what you mean.  So, instead of `country: 'USA'`, it'd be 
`country: $scope.countries[0].country` or something like that?

christianvuerings: @davidlimcheng not really, I'm talking about the `employFracValues` - e.g. `Full Time` & `Half Time` - those values, we should probably get back from an XLAT API.

davidlimcheng: @christianvuerings OH, oh lord, OK, now I'm on the same page hahaha.

The `Full Time` etc. options were actually purely born out of the design.  In CS, it's not a dropdown but a text input box:  
employfrac

So in CS you can actually put anything you want in there, as long as it's under the length limit.

christianvuerings: @davidlimcheng thanks - didn't know about this. Would be interesting to test what happens if you add a value that doesn't exist in our dropdown in CS.

davidlimcheng:
@christianvuerings , @paulkerschen , I wasn't able to complete the push today, and I'm off the next two days for a wedding, so I'm going to close this PR and open a new one later instead of letting it sit here for a couple of days.  Thanks a bunch for the direction, will have the new push up ASAP!

PR #5042, opened 2016-03-10
SISRP-16120 Have RelatedCacheKeyTracker register keys after creation, not before

paulkerschen:
I wouldn't call this a long-term fix for https://jira.berkeley.edu/browse/SISRP-16120, but I would like to see if it reduces blockage for QA while we consider next steps.

Also remove the ucUpdateCache query-param check for finaid and enrollment, since that's one bit of redundancy we have power to skip.


raydavis:
Thanks, @paulkerschen . I'll do my best to get the change through Bamboo this morning.

PR #5045, opened 2016-03-10
CLC-6226 OEC: tighten data integrity checks prior to Blue push

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6226 revealed a need for more explicit referential integrity checks between worksheets prior to sending them Blue-ward. 

Validate that certain ID sets are identical between worksheets, as specified.


johncrossman:
I have validated_integrity of this PR.

PR #5051, opened 2016-03-11
SISRP-16199 - Better messaging for continuing students in - Finaid

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-16199

### DIFF
--- a/src/assets/templates/widgets/finaid_summary.html
+++ b/src/assets/templates/widgets/finaid_summary.html
@@ -1,6 +1,6 @@
 <div class="cc-widget cc-widget-finaid-summary" data-ng-controller="FinaidSummaryController">
   <div class="cc-widget-title">
-    <h2 class="cc-left">Financial Aid and Scholarships</h2>
+    <h2 class="cc-left">Financial Aid and Scholarships 2016-17</h2>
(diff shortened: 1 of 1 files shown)


src/assets/templates/widgets/finaid_summary.html
```diff
@@ -1,6 +1,6 @@
 <div class="cc-widget cc-widget-finaid-summary" data-ng-controller="FinaidSummaryController">
   <div class="cc-widget-title">
-    <h2 class="cc-left">Financial Aid and Scholarships</h2>
+    <h2 class="cc-left">Financial Aid and Scholarships 2016-17</h2>
```
raydavis: I'm sorry, @christianvuerings , but hardcoding a year seems like big trouble to me. CalCentral backend code already has some support for FinAid dates. Could we add the years-of-interest to the API payload?

christianvuerings: @raydavis fyi we would be removing this together with the current hardcoded `2015-2016` (see screenshot in jira) as soon as that time comes. Would you prefer me to reassign that ticket to you?

raydavis: Yuck, I didn't notice the other hardcoded year-span. Thanks for the pointer.

Considering the gross circumstances, I'll merge the PR. But please ping me on the follow-up ticket. I have some nostalgia around the earlier incarnation of this issue. :)

PR #5052, opened 2016-03-11
SISRP-16120 Restore ucUpdateCache param check for finaid, enrollment

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-16120

Partial revert of #5042.

### DIFF
--- a/app/controllers/bootstrap_controller.rb
+++ b/app/controllers/bootstrap_controller.rb
@@ -45,8 +45,12 @@ class BootstrapController < ApplicationController
     # a Fluid activity that may have changed some data.
     flag = params['ucUpdateCache']
     case flag
+      when 'finaid'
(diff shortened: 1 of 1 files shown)


app/controllers/bootstrap_controller.rb
```diff
@@ -45,8 +45,12 @@ def check_cache_clear_flag
     # a Fluid activity that may have changed some data.
     flag = params['ucUpdateCache']
     case flag
+      when 'finaid'
```
johncrossman: No concern about case-sensitivity?

paulkerschen: We set the flag ourselves in the front end and pass it over to CS, so I think straight equality is safe.

johncrossman:
Thanks, @paulkerschen – looks good. 
As a courtesy to others, CalCentral will always flush.

PR #5056, opened 2016-03-13
QA CLC-6238 Display CalNet Directory first/last names as determined by user

raydavis:
QA version of https://github.com/ets-berkeley-edu/calcentral/pull/5055 .

Please do not merge until we get an OK from @johncrossman and @pfarestveit .


johncrossman:
@raydavis You have my vote. I think we're ready to proceed based on @pfarestveit comment in https://jira.ets.berkeley.edu/jira/browse/CLC-6235

PR #5059, opened 2016-03-14
SISRP-10786 - Front End Dev: Credentials Tab - Work Experience Card - Edit

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-10786

Includes delete functionality, follow-up PR to #4060.

Apologies, I realized too late that I shouldn't have rebased and instead just reopened the PR with the original branch.  Will keep that in my pocket for future PRs.  Hope it doesn't cause too much trouble.


src/assets/javascripts/angular/controllers/widgets/profile/workExperienceSectionController.js
```diff
@@ -53,19 +132,150 @@ angular.module('calcentral.controllers').controller('WorkExperienceSectionContro
...
+  };
+
+  $scope.showAdd = function() {
+    apiService.profile.showAdd($scope, $scope.emptyObject);
+  };
+
+  $scope.showEdit = function(item) {
+    apiService.profile.showEdit($scope, item);
+    $scope.currentObject.data = {
+      sequenceNbr: item.id,
```
christianvuerings: @davidlimcheng do you remember the clean-up jira number for the cleanup?

christianvuerings:
Front-end lgtm - @paulkerschen @johncrossman @raydavis  would you like a last pass at this?

johncrossman:
Thanks, @davidlimcheng! I'll review it this eve.

app/models/campus_solutions/my_work_experience.rb
```diff
@@ -11,5 +13,15 @@ def delete(params = {})
       passthrough(CampusSolutions::WorkExperienceDelete, params)
     end
 
+    # Formats input date to Campus Solutions acceptable ISO 8601 (YYYY-MM-DD)
+    # Front-end date validator ensures date will be in '%m/%d/%Y' format
+    def cs_date_formatter(date)
+      begin
+        return Date.strptime(date, '%m/%d/%Y').strftime('%F')
+      rescue ArgumentError
+        return date
```
johncrossman: @davidlimcheng Why rescue the ArgumentError? I'm a bit concerned that we post bad data to CS and get unpredictable results.

app/models/campus_solutions/my_work_experience.rb
```diff
@@ -11,5 +13,15 @@ def delete(params = {})
       passthrough(CampusSolutions::WorkExperienceDelete, params)
     end
 
+    # Formats input date to Campus Solutions acceptable ISO 8601 (YYYY-MM-DD)
+    # Front-end date validator ensures date will be in '%m/%d/%Y' format
```
johncrossman: @davidlimcheng Minor point: I tend not to put implementation details (e.g., incoming date format) in comments. The risk is that business logic changes and the comment is not updated. The comment is then obsolete / wrong.

app/models/campus_solutions/my_work_experience.rb
paulkerschen: Rescuing the ArgumentError was my suggestion; I mentioned that we usually do so in cases of date formatting, because one never knows about the input. But @johncrossman is right that we shouldn't forward data on to CS willy-nilly.

@davidlimcheng, I think the behavior we want is that if `date` is blank, `cs_date_formatter` returns the empty string that CS expects; anything else that raises an error should trigger a `logger.error` and send a polite error response back to the front end.

davidlimcheng: @paulkerschen @johncrossman ahh I see. OK, I'll make the changes and push.  My original thinking was that if we send whatever the bad data was over to CS, it would throw back the corresponding error message from CS.  But this makes more sense!

app/models/campus_solutions/my_work_experience.rb
davidlimcheng: @johncrossman Sounds good, will get rid of the comments.

src/assets/javascripts/angular/controllers/widgets/profile/workExperienceSectionController.js
davidlimcheng: @christianvuerings Just created - [SISRP-16336](https://jira.berkeley.edu/browse/SISRP-16336)

app/models/campus_solutions/my_work_experience.rb
```diff
@@ -1,9 +1,15 @@
...
-    include PersonDataUpdatingModel
+    include ClassLogger
+    include WorkExperienceUpdatingModel
 
     def update(params = {})
+      params[:startDt] = cs_date_formatter params[:startDt]
+      params[:endDt] = cs_date_formatter params[:endDt]
+      if !params[:startDt] or !params[:endDt]
+        return err_msg
+      end
```
davidlimcheng: @paulkerschen - I think I finally got it right, or am close at least.  I did base the conditionals off of a `false` returned by `cs_date_formatter`, do you think that's acceptable here?

davidlimcheng:
Looks like the Travis failed because `work_experience_controller_spec.rb` is still testing with the outdated fields and format.  I'll update those now.

app/models/campus_solutions/my_work_experience.rb
```diff
@@ -11,5 +17,25 @@ def delete(params = {})
       passthrough(CampusSolutions::WorkExperienceDelete, params)
     end
 
+    def cs_date_formatter(date)
+      begin
+        return date if date.blank?
```
paulkerschen: @davidlimcheng perhaps `return '' if date.blank?` is better; that way you're sure that nil or whitespace values will be coerced to the empty string you want.

app/models/campus_solutions/my_work_experience.rb
```diff
@@ -11,5 +17,25 @@ def delete(params = {})
...
+        return Date.strptime(date, '%m/%d/%Y').strftime('%F')
```
paulkerschen: @davidlimcheng No `return` needed here

app/models/campus_solutions/my_work_experience.rb
```diff
@@ -11,5 +17,25 @@ def delete(params = {})
...
+      rescue ArgumentError
+        return false
```
paulkerschen: @davidlimcheng No `return` needed here

app/models/campus_solutions/my_work_experience.rb
```diff
@@ -11,5 +17,25 @@ def delete(params = {})
...
+      end
+    end
+
+    def err_msg
+      logger.error "Error reported in back-end validation."
```
paulkerschen: @davidlimcheng To help debugging I'd make this a logger message a little more descriptive, and would move it up to immediately after `rescue ArgumentError`. That way you have access to the the bad date string and can log something like: `"Could not parse date string #{date}"`.

app/models/campus_solutions/my_work_experience.rb
paulkerschen: @davidlimcheng Yep, returning false on error for a conditional check works just fine.

paulkerschen:
Thanks for all the diligent work on this, @davidlimcheng! I have a last round of (mostly stylistic) comments; otherwise the back end looks good to me.

davidlimcheng:
@paulkerschen Made the changes, thanks a bunch for all of the direction!

PR #5060, opened 2016-03-14
SISRP-16301 - Academic Profile: "Earned Units" label showing when there are no units yet

dfkaye:
https://jira.berkeley.edu/browse/SISRP-16301

Main issue here is the strict-equality check, `value !== null`.  If value is `0` or `undefined` or the empty string, this check will still return `true`.  

<del>Better to set the property to a Number explicitly in the controller using `parseFloat` and check in the template that `value > 0`</del>

@christianvuerings `parseFloat()` returns `NaN` when the input does not contain at least one digit.  Also,  the `value > 0` checks should be corrected to `value >= 0`

### DIFF
--- a/src/assets/javascripts/angular/controllers/pages/academicsController.js
+++ b/src/assets/javascripts/angular/controllers/pages/academicsController.js
@@ -153,8 +153,11 @@ angular.module('calcentral.controllers').controller('AcademicsController', funct
         $scope.widgetSemesterName = $scope.selectedTeachingSemester.name;
       }
     }
-    $scope.gpaUnits.cumulativeGpaFloat = $scope.gpaUnits.cumulativeGpa; // cumulativeGpa is passed as a string to maintain two significant digits
-    $scope.gpaUnits.cumulativeGpa = parseFloat($scope.gpaUnits.cumulativeGpa); // converted to Float to be processed regularly
+    // cumulativeGpa is passed as a string to maintain two significant digits
+    $scope.gpaUnits.cumulativeGpaFloat = $scope.gpaUnits.cumulativeGpa;
+    // Convert these to Number types to be processed regularly. `parseFloat` returns NaN if the input value does not contain at least one digit.
+    $scope.gpaUnits.cumulativeGpa = parseFloat($scope.gpaUnits.cumulativeGpa);
+    $scope.gpaUnits.totalUnits = parseFloat($scope.gpaUnits.totalUnits);
   };
 
   var filterWidgets = function() {
--- a/src/assets/templates/academics_student_profile.html
+++ b/src/assets/templates/academics_student_profile.html
@@ -166,7 +166,7 @@
       </table>
     </div>
 
-    <div class="cc-table cc-table-top-border cc-clearfix" data-ng-if="gpaUnits.totalUnits !== null">
+    <div class="cc-table cc-table-top-border cc-clearfix" data-ng-if="gpaUnits.totalUnits > 0">
       <table>
         <tbody>
           <tr>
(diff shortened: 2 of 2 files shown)


src/assets/javascripts/angular/controllers/pages/academicsController.js
```diff
@@ -153,8 +153,11 @@ angular.module('calcentral.controllers').controller('AcademicsController', funct
         $scope.widgetSemesterName = $scope.selectedTeachingSemester.name;
       }
     }
-    $scope.gpaUnits.cumulativeGpaFloat = $scope.gpaUnits.cumulativeGpa; // cumulativeGpa is passed as a string to maintain two significant digits
-    $scope.gpaUnits.cumulativeGpa = parseFloat($scope.gpaUnits.cumulativeGpa); // converted to Float to be processed regularly
+    // cumulativeGpa is passed as a string to maintain two significant digits
+    $scope.gpaUnits.cumulativeGpaFloat = $scope.gpaUnits.cumulativeGpa;
+    // convert to Float to be processed regularly
+    $scope.gpaUnits.cumulativeGpa = parseFloat($scope.gpaUnits.cumulativeGpa); // convert to Float to be processed regularly
+    $scope.gpaUnits.totalUnits = parseFloat($scope.gpaUnits.totalUnits);
```
christianvuerings: @dfkaye is this actually ever going to be `null`or `undefined`? If so, won't this result in a JS error?
Might be good to have test data before making the fix.

src/assets/javascripts/angular/controllers/pages/academicsController.js
```diff
@@ -153,8 +153,11 @@ angular.module('calcentral.controllers').controller('AcademicsController', funct
         $scope.widgetSemesterName = $scope.selectedTeachingSemester.name;
       }
     }
-    $scope.gpaUnits.cumulativeGpaFloat = $scope.gpaUnits.cumulativeGpa; // cumulativeGpa is passed as a string to maintain two significant digits
-    $scope.gpaUnits.cumulativeGpa = parseFloat($scope.gpaUnits.cumulativeGpa); // converted to Float to be processed regularly
+    // cumulativeGpa is passed as a string to maintain two significant digits
+    $scope.gpaUnits.cumulativeGpaFloat = $scope.gpaUnits.cumulativeGpa;
+    // convert to Float to be processed regularly
+    $scope.gpaUnits.cumulativeGpa = parseFloat($scope.gpaUnits.cumulativeGpa); // convert to Float to be processed regularly
```
christianvuerings: @dfkaye seems like we have the `// convert to Float to be processed regularly` comment twice, once above the line and once next to it.

src/assets/templates/academics_student_profile.html
```diff
@@ -177,7 +177,7 @@ <h3 data-ng-if="!isProfileCurrent && api.user.profile.roles.student && collegeAn
       </table>
     </div>
 
-    <div class="cc-table cc-table-top-border cc-clearfix" data-ng-if="gpaUnits.cumulativeGpa !== null && gpaUnits.cumulativeGpa > 0 && !studentInfo.isLawStudent">
+    <div class="cc-table cc-table-top-border cc-clearfix" data-ng-if="gpaUnits.cumulativeGpa > 0 && !studentInfo.isLawStudent">
```
christianvuerings: @dfkaye don't we actually want to show it, even though it's `0`? That's probably why we have the `null` check in there.

paulkerschen:
@dfkaye @christianvuerings This looks good to me, except I think we actually do want `value > 0` and not `value >= 0`; CS tends to return a numeric 0 when no data is available.

We definitely don't want to show zero for `cumulativeGPA`. In the case of `totalUnits`, I suppose it's worth validating with a functional owner whether they want zero-unit totals to show or not.

I think those misleading null checks were not put in to surface zero values, but were a mistake due to coding the template before test data was available.

paulkerschen:
Thanks @dfkaye.

PR #5061, opened 2016-03-15
SISRP-16318 - Checklist not immediately showing in FA Comms after signing T&Cs

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-16318


johncrossman:
Looks good @christianvuerings   QA PR once fix is verified on master?

PR #5064, opened 2016-03-15
SISRP-16284, new API calls to verify script plus alpha order

johncrossman:
https://jira.berkeley.edu/browse/SISRP-16284

As we troubleshoot poor performance from Hub I've cleaned up script and added new endpoints. E.g., FERPA


raydavis:
TIDY!

PR #5067, opened 2016-03-15
SISRP-16369, remove two unused queries from CampusOracle::Queries

johncrossman:
https://jira.berkeley.edu/browse/SISRP-16369

@raydavis should review/merge this one because I think he is author of `get_student_term_info(person_id)` which is documented with 'currently used only for support and research, not by deployed code'.  If its usefulness has expired then this PR can be merged.

`is_previous_ugrad?(ldap_uid)` was only used by the spec.

### DIFF
--- a/app/models/campus_oracle/queries.rb
+++ b/app/models/campus_oracle/queries.rb
@@ -166,23 +166,6 @@ module CampusOracle
       end
     end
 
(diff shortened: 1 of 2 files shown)


raydavis:
Always happy to see my code disappear, @johncrossman -- thank you for the cleanup.

PR #5068, opened 2016-03-15
SISRP-16369, oec_legacy is headed to that big hackfest in the sky, free of earthly burdens.

johncrossman:
https://jira.berkeley.edu/browse/SISRP-16369

Put the code in a jar of formaldehyde, up on the shelf. We're done.


paulkerschen:
Ahhh, that's a nice feeling.

PR #5069, opened 2016-03-15
NOJIRA, HTTP response time to stdout in case of further CS/Hub performance problems

johncrossman:
https://jira.berkeley.edu/browse/SISRP-16284

Having HTTP response time in stdout was helpful during recent Hub problems.


paulkerschen:
Thanks @johncrossman.

knowing-is-half-the-battlejpg1425907458645

PR #5075, opened 2016-03-16
CLC-6243 Handle special characters from Calnet LDAP

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6243


raydavis:
Wot, no Cyrillic?

Nevertheless, if Travis is happy, I'm very happy to merge.

PR #5079, opened 2016-03-16
SISRP-16309 - Create Feature Flag and Back-End for CS Billing

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-16309

*fingers crossed*

I did default the feature flags to `true` in all of the `.yml` files except for `settings.yml` - not sure if this is the convention I should be following or not.  Let me know if otherwise, and I'll push a new commit up with correct flags.


davidlimcheng:
Hmm looks like there are conflicts.  Let me rebase and open a new PR.

PR #5083, opened 2016-03-17
SISRP-16500 - Retrofit CARS Activity Front-end for transition, incorporating feature flag

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-16500

### DIFF
--- a/src/assets/stylesheets/_myfinances.scss
+++ b/src/assets/stylesheets/_myfinances.scss
@@ -20,6 +20,9 @@
       margin-right: 10px;
     }
   }
+  .cc-page-myfinances-account-summary-alert {
+    margin: 15px 20px 0 15px;
+  }
   .cc-page-myfinances-account-summary-buttons {
     margin: 10px 20px 10px 15px;
   }
--- a/src/assets/templates/cars_summary.html
+++ b/src/assets/templates/cars_summary.html
@@ -1,12 +1,19 @@
 <div class="cc-widget" data-ng-controller="CarsController">
   <div class="cc-widget-title">
-    <h2 class="cc-left">Billing Summary</h2>
+    <h2 class="cc-left">
+      Billing Summary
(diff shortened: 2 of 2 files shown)


src/assets/stylesheets/_myfinances.scss
```diff
@@ -20,6 +20,9 @@
       margin-right: 10px;
     }
   }
+  .cc-page-myfinances-account-summary-alert {
+    margin: 14px 20px 0 15px;
```
christianvuerings: @davidlimcheng could we make this `margin: 15px 20px 0 15px`? Or is there a specific reason for the `14px`?

src/assets/templates/cars_summary.html
```diff
@@ -1,12 +1,19 @@
 <div class="cc-widget" data-ng-controller="CarsController">
   <div class="cc-widget-title">
-    <h2 class="cc-left">Billing Summary</h2>
+    <h2 class="cc-left">
+      Billing Summary
+      <span data-ng-if="api.user.profile.features.csBilling" class="cc-widget-title-sub">PRIOR TO FALL 2016</span>
```
christianvuerings: @davidlimcheng Could we use the class `cc-text-uppercase` on the uppercase text and use normal case in the html?

christianvuerings:
Lgtm as soon as comments are addressed.

src/assets/stylesheets/_myfinances.scss
davidlimcheng: @christianvuerings No reason, I set it to `8px` (same as `cc-page-myfinances-account-summary-item`) and just eyeballed it from there to make it similar to the design.  Will set it to 15px.

src/assets/templates/cars_summary.html
davidlimcheng: @christianvuerings Sure thing, and I'll be sure to follow suit on future uppercased texts.

davidlimcheng:
Changes have been made, thanks @christianvuerings !

PR #5085, opened 2016-03-17
CLC-6245 New CalCentral introductory video

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6245


johncrossman:
I tube, YouTube, we tube. Thx @raydavis

PR #5089, opened 2016-03-19
SISRP-14768 - Retrofit preparation

redconfetti:
https://jira.berkeley.edu/browse/SISRP-14768

* Establishes OracleBase to share logic between Oracle modules
* Introduces EdoOracle classes
* Migrates shared logic to OracleBase from CampusOracle::Connection
* Adds EDO Term conversion methods to Berkeley::TermCodes

### DIFF
--- a/app/models/berkeley/term_codes.rb
+++ b/app/models/berkeley/term_codes.rb
@@ -17,6 +17,18 @@ module Berkeley
       @names ||= init_names
     end
 
+    def to_edo_id(term_yr, term_cd)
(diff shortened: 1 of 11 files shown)


redconfetti:
I'm not sure if we should create some sort of fake data mechanism for EDO Oracle, or just continue to use H2. I know that the H2 data loading really slows down the initialization of the application.

I still need to apply a test, as well as test ext test, to `EdoOracle::Queries.get_sections_from_section_ids`, but that can come later.

paulkerschen:
And we're on our way... thanks @redconfetti.

PR #5091, opened 2016-03-21
SISRP-16579, SIS API endpoint must know about new Hub API error format

johncrossman:
https://jira.berkeley.edu/browse/SISRP-16579

I've moved `API_ERROR_INDICATORS` to top of script, for the sake of prominence, and added to its list per Jira above.


paulkerschen:
There's always room for one more error format. Thanks @johncrossman.

PR #5092, opened 2016-03-21
SISRP-14768 Per-environment edodb settings on campusdb pattern

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-14768

EDO DB settings should be configured per environment on the `campusdb` convention, and called `edodb` to match the database.yml entry.


redconfetti:
Thanks @paulkerschen

PR #5093, opened 2016-03-21
SISRP-16476 - Create Billing Summary Front-End

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-16476

### DIFF
--- a/src/assets/javascripts/angular/controllers/widgets/billingController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/billingController.js
@@ -1,0 +1,38 @@
+'use strict';
+
+var angular = require('angular');
+var _ = require('lodash');
+
+angular.module('calcentral.controllers').controller('BillingController', function(apiService, financesFactory, $scope) {
+  $scope.billing = {
+    data: {},
+    isLoading: true
+  };
+
+  var parseAmount = function(value) {
+    if (_.isNumber(value)) {
+      return value.toFixed(2);
+    }
+  };
+
+  var parseBillingInfo = function(data) {
+    var billing = _.get(data, 'data.feed.ucSfActivity');
+
+    billing.summary = _.mapValues(billing.summary, function(value) {
+      value = parseAmount(value);
+      return value;
+    });
+
+    $scope.billing.data = billing;
+  };
+
+  var loadBillingInfo = function() {
+    financesFactory.getCsFinances()
+      .then(parseBillingInfo)
+      .then(function() {
+        $scope.billing.isLoading = false;
+      });
+  };
+
+  loadBillingInfo();
+});
--- a/src/assets/templates/billing_details.html
+++ b/src/assets/templates/billing_details.html
@@ -1,9 +1,9 @@
-<div class="cc-page-myfinances">
+<div class="cc-page-myfinances" data-ng-if="api.user.profile.features.csBilling">
   <div class="row collapse">
 
     <div class="column">
       <h1 class="cc-heading-page-title cc-print-hide">
-        <a href="/finances">My Finances</a> &raquo; Details <span data-ng-if="api.user.profile.features.csBilling">(Prior to Fall 2016)</span>
+        <a href="/finances">My Finances</a> &raquo; Details
       </h1>
       <h1 class="cc-heading-page-title cc-hidden cc-print-show">
         My Finances for <span data-ng-bind="api.user.profile.fullName"></span>
@@ -11,11 +11,12 @@
     </div>
 
     <div class="medium-4 columns">
-      <div data-ng-include="'cars_summary.html'" data-onload="carsDetails=true"></div>
+      <div data-ng-include="'widgets/billing_summary.html'" data-onload="billingDetails=true"></div>
     </div>
     <div class="medium-8 columns cc-column-last">
-      <div data-ng-include="'cars_activity.html'"></div>
+
     </div>
 
   </div>
 </div>
+<div data-ng-include="'404.html'" data-ng-if="!api.user.profile.features.csBilling"></div>
--- a/src/assets/templates/widgets/billing_summary.html
+++ b/src/assets/templates/widgets/billing_summary.html
@@ -1,0 +1,48 @@
+<div class="cc-widget" data-ng-controller="BillingController">
+  <div class="cc-widget-title">
+    <h2 class="cc-left">Billing Summary</h2>
+    <a data-ng-if="!billingDetails && billing.data.activity" class="cc-right cc-widget-title-link" href="/billing/details">
+      <strong>Details</strong>
+    </a>
+  </div>
+
+  <div data-cc-spinner-directive="billing.isLoading">
+    <ul class="cc-widget-list cc-page-myfinances-account-summary-list" data-ng-show="billing.data.summary">
+      <li class="cc-page-myfinances-account-summary-item cc-clearfix-container">
+        <div class="small-8 columns cc-print-width-auto">
+          <strong data-ng-if="billing.data.summary.amountDueNow >= 0">
+            Amount Due Now
+          </strong>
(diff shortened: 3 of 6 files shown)


src/assets/templates/widgets/billing_summary.html
```diff
@@ -0,0 +1,51 @@
...
+        <div class="small-8 columns cc-print-width-auto">Account Balance
+          <div class="cc-page-myfinances-account-summary-subtitle cc-page-myfinances-light">Includes charges not yet due</div>
+        </div>
+        <div class="small-4 columns cc-print-width-auto cc-print-right">
+          <div class="cc-page-myfinances-amount" data-cc-amount-directive="billing.summary.accountBalance"></div>
+        </div>
+      </li>
+      <li data-ng-if="myfinances.summary.accountBalanceFloat !== 0">
+        <div class="cc-page-myfinances-account-summary-buttons cc-print-hide">
+          <a class="cc-button" href="">Make Payment for Fall 2016</a>
```
davidlimcheng: Dead button here for now, but will change to SSO to HigherOne in a future pull request.

src/assets/templates/billing_details.html
```diff
@@ -0,0 +1,21 @@
...
+      <h1 class="cc-heading-page-title cc-hidden cc-print-show">
+        My Finances for <span data-ng-bind="api.user.profile.fullName"></span>
+      </h1>
+    </div>
+
+    <div class="medium-4 columns">
+      <div data-ng-include="'widgets/billing_summary.html'" data-onload="billingDetails=true"></div>
+    </div>
+    <div class="medium-8 columns cc-column-last">
+
```
davidlimcheng: Will add `billing_activity.html` card in a future pull request.

src/assets/javascripts/angular/controllers/widgets/billingController.js
```diff
@@ -0,0 +1,36 @@
+'use strict';
+
+var angular = require('angular');
+var _ = require('lodash');
+
+angular.module('calcentral.controllers').controller('BillingController', function(apiService, financesFactory, $scope) {
+  $scope.billing = {};
+  $scope.isLoading = true;
```
christianvuerings: @davidlimcheng we've changed this a bit over time, would we be able to make it part of `billing`?

Something like
```
$scope.billing = {
  isLoading: true,
  data: {}
}
```

That way we always have a `.` in our model ...
http://stackoverflow.com/questions/17606936/angularjs-dot-in-ng-model

src/assets/templates/billing_details.html
```diff
@@ -0,0 +1,21 @@
+<div class="cc-page-myfinances">
```
christianvuerings: @davidlimcheng could we also add the feature flag here? (might we worthwhile checking that they actually do get a 404 when it's off)

src/assets/templates/billing_details.html
```diff
@@ -0,0 +1,21 @@
...
+  <div class="row collapse">
+
+    <div class="column">
+      <h1 class="cc-heading-page-title cc-print-hide">
+        <a href="/finances">My Finances</a> &raquo; Details (Fall 2016)
+      </h1>
+      <h1 class="cc-heading-page-title cc-hidden cc-print-show">
+        My Finances for <span data-ng-bind="api.user.profile.fullName"></span>
```
christianvuerings: @davidlimcheng could you look at the "printed" version and see whether it looks good?

src/assets/templates/widgets/billing_summary.html
```diff
@@ -0,0 +1,51 @@
+<div class="cc-widget" data-ng-controller="BillingController">
+  <div class="cc-widget-title">
+    <h2 class="cc-left">
+      Billing Summary
+      <span class="cc-widget-title-sub cc-text-uppercase">Fall 2016</span>
```
christianvuerings: @davidlimcheng this should probably not be hardcoded - do we get this from the API?

src/assets/javascripts/angular/controllers/widgets/billingController.js
davidlimcheng: Sure thing, will add in the next push.

src/assets/templates/billing_details.html
davidlimcheng: Will do.

src/assets/templates/widgets/billing_summary.html
davidlimcheng: For this one, I had to discuss with Adrianne.  The API does not give us the current term, and actually we do not want to show the current term here.  According to Adrianne, this is a temporary subheading that will only be present throughout the transition period, and we'd remove it once the transition period is over.  I do agree that we shouldn't hardcode this - created [SISRP-16662] (https://jira.berkeley.edu/browse/SISRP-16662) to track and will remove in the following push.

src/assets/templates/billing_details.html
davidlimcheng: @christianvuerings I added a feature flag here, but visiting `/billing/details.html` took me to a page with a header and footer, with an empty body - so no 404.  

What about adding a div with `404.html` under something like `data-ng-show="!api.user.profile.features.csBilling`?  
I also ended up at [this](http://code.ciphertrick.com/2014/12/14/check-condition-before-loading-route-in-angular-js/) interesting read trying to research if we could have conditional routing with $routeProvider.  

What do you think?

src/assets/templates/billing_details.html
davidlimcheng: @christianvuerings Checked it out and it looks good - shows the full name as expected.

src/assets/templates/billing_details.html
christianvuerings: @davidlimcheng if we replace the `ng-show` with `ng-if`, that should be fine.

As for the `resolve` solution for `$routeProvider`, it would require us to make an Ajax call at that time, which would be too expensive.

fyi have a look at
```
<div data-ng-include="'404.html'" data-ng-if="!canViewAcademics"></div>
```

davidlimcheng: @christianvuerings Sounds good!  Will include.

davidlimcheng:
Excuse the flurry of amended commits, but should be good now.  All discussed changes have been made.

src/assets/templates/billing_details.html
```diff
@@ -0,0 +1,22 @@
+<div class="cc-page-myfinances" data-ng-if="api.user.profile.features.csBilling">
+  <div class="row collapse">
+
+    <div class="column">
+      <h1 class="cc-heading-page-title cc-print-hide">
+        <a href="/finances">My Finances</a> &raquo; Details (Fall 2016)
```
davidlimcheng: @christianvuerings Actually should I remove the "Fall 2016" from here too?

src/assets/templates/widgets/billing_summary.html
```diff
@@ -0,0 +1,48 @@
...
+        <div class="small-8 columns cc-print-width-auto">Account Balance
+          <div class="cc-page-myfinances-account-summary-subtitle cc-page-myfinances-light">Includes charges not yet due</div>
+        </div>
+        <div class="small-4 columns cc-print-width-auto cc-print-right">
+          <div class="cc-page-myfinances-amount" data-cc-amount-directive="billing.data.summary.accountBalance"></div>
+        </div>
+      </li>
+      <li data-ng-if="billing.data.summary.accountBalance !== 0">
+        <div class="cc-page-myfinances-account-summary-buttons cc-print-hide">
+          <a class="cc-button" href="">Make Payment for Fall 2016</a>
```
davidlimcheng: @christianvuerings Here as well, "Fall 2016"

christianvuerings: @davidlimcheng yes to both

davidlimcheng:
@christianvuerings pesky `Fall 2016`'s removed, we should be good to go now.

PR #5095, opened 2016-03-21
SISRP-15857 EdoDb: Get instructing sections for UID and terms

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-15857

No spec yet; a sanity check under testext is coming.

Term-code converter also takes more flexible input.

### DIFF
--- a/app/models/berkeley/term_codes.rb
+++ b/app/models/berkeley/term_codes.rb
@@ -18,12 +18,14 @@ module Berkeley
     end
 
     def to_edo_id(term_yr, term_cd)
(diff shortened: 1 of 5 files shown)


app/models/berkeley/term_codes.rb
```diff
@@ -18,12 +18,14 @@ def names
     end
 
     def to_edo_id(term_yr, term_cd)
+      term_yr = term_yr.to_s
```
redconfetti: Oh yeah, good idea

PR #5097, opened 2016-03-22
SISRP-16631, Afrikaans removed from CampusSolutions:/UC_CC_LANGUAGES

johncrossman:
https://jira.berkeley.edu/browse/SISRP-16631

Afrikaans was deliberately removed so I've updated the spec. I dropped the latest prod XML into fixture, too. Quarantined test on Bamboo has been unleashed.


christianvuerings:
@johncrossman Dankie vir die maak van hierdie oplossing!

PR #5106, opened 2016-03-23
SISRP-16691 - Create Front-end for CS Billing Activity

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-16691

Also includes some edits to the fixture XML, to test some cases.
Should be adding a search filter and more presentation logic in a future PR.


src/assets/javascripts/angular/controllers/widgets/billingController.js
```diff
@@ -4,25 +4,40 @@ var angular = require('angular');
 var _ = require('lodash');
 
 angular.module('calcentral.controllers').controller('BillingController', function(apiService, financesFactory, $scope) {
+  $scope.activityIncrement = 50;
+  $scope.activityLimit = 100;
   $scope.billing = {
     data: {},
     isLoading: true
   };
+  $scope.sort = {
```
christianvuerings: @davidlimcheng can we make this part of `$scope.billing`?

src/assets/javascripts/angular/controllers/widgets/billingController.js
```diff
@@ -4,25 +4,40 @@ var angular = require('angular');
 var _ = require('lodash');
 
 angular.module('calcentral.controllers').controller('BillingController', function(apiService, financesFactory, $scope) {
+  $scope.activityIncrement = 50;
+  $scope.activityLimit = 100;
```
christianvuerings: @davidlimcheng can we make these part of `$scope.billing`?

src/assets/javascripts/angular/controllers/widgets/billingController.js
```diff
@@ -34,5 +49,10 @@ angular.module('calcentral.controllers').controller('BillingController', functio
       });
   };
 
+  $scope.printPage = function() {
```
christianvuerings: @davidlimcheng should this be in a service?

src/assets/templates/widgets/billing_activity.html
```diff
@@ -0,0 +1,104 @@
+<div class="cc-widget cc-page-myfinances-account-activity" data-ng-show="billing.data.activity.length" data-ng-controller="BillingController">
+  <div class="cc-widget-title">
+    <h2 class="cc-left">Activity</h2>
+    <button class="cc-button cc-right cc-widget-title-button" data-ng-click="printPage()">
+      <i class="fa fa-print"></i> Print
+    </button>
+  </div>
+
+    <div class="cc-table cc-page-myfinances-table cc-table-finances">
```
christianvuerings: @davidlimcheng indentation - ... so much fun!

src/assets/templates/widgets/billing_activity.html
```diff
@@ -0,0 +1,104 @@
+<div class="cc-widget cc-page-myfinances-account-activity" data-ng-show="billing.data.activity.length" data-ng-controller="BillingController">
```
christianvuerings: @davidlimcheng can we see whether an `ng-if` would work here instead? Try to look at all the `ng-show` / `ng-hide` in this file (and future files)

src/assets/templates/widgets/billing_activity.html
```diff
@@ -0,0 +1,104 @@
...
+            <th width="30%" class="cc-table-right cc-table-right-spacing" scope="col">Amount</th>
+            <th width="10%" scope="col">Type</th>
+            <th width="15%" class="cc-table-center" scope="col">Due</th>
+          </tr>
+        </thead>
+        <tbody
+          class="cc-widget-table-hover"
+          data-cc-accessible-focus-directive
+          data-ng-class="{'cc-widget-table-hover-opened cc-widget-list-selected':(item.show)}"
+          data-ng-click="api.widget.toggleShow($event, billing.data.activity, item, 'My Finances - Transaction')"
```
christianvuerings: @davidlimcheng might be good to replace `My Finances` with `Billing` so we can see in Google Analytics which one is used.

src/assets/templates/widgets/billing_activity.html
```diff
@@ -0,0 +1,104 @@
...
+            <td class="cc-table-center">
+              <i class="fa" data-ng-class="{'fa-exclamation-circle cc-icon-red':(item.itemDueFlag === 'OVER_DUE'),'fa-exclamation':(item.itemDueFlag === 'DUE_NOW'),'fa-arrow-right':(item.itemDueFlag === 'FUTURE_DUE')}" data-ng-if="item.itemDueFlag">
+            </td>
+          </tr>
+          <tr data-ng-if="item.show">
+            <td></td>
+            <td colspan="4">
+              <div class="row collapse">
+                <div data-ng-if="item.itemReferenceDescription">
+                  <strong>Transaction #:</strong>
```
christianvuerings: @davidlimcheng we should probably `aria-hidden` the `#` and add a `visually-hidden` `span` for `Number`.

christianvuerings:
Some comments inline - would be so good to run over every single `if` in here with Jeff and see whether it all still makes sense

src/assets/javascripts/angular/controllers/widgets/billingController.js
davidlimcheng: @christianvuerings Hmm, so when I tried to add this to its own `apiService.financials.printPage()`, I get a circular dependency error:
```Error: [$injector:cdep] Circular dependency found: apiService <- financialsService <- apiService```
Makes sense since I'm sticking `apiService` as an argument in my `financialsService.js` as it does need to call on `apiService.analytics.sendEvent()` within `printPage()` - where am I going wrong?

davidlimcheng: @christianvuerings Should I bypass `apiService` and directly call `analyticsService`?

christianvuerings: @davidlimcheng that should be fine

src/assets/templates/widgets/billing_activity.html
davidlimcheng: @christianvuerings This one gave me some trouble, but I found some good information here:
http://stackoverflow.com/questions/33151920/accessing-angularjs-controller-properties-from-the-root-node-ng-if

It looks like changing it to an `ng-if` will work if I use it in a child `div` that holds all of the content, separate from the parent `div` that contains `BillingController`.  Would that be preferable to using `data-ng-show` on the parent `div`?

christianvuerings: @davidlimcheng indeed - child `div` it should be. Already thought it looked strange just couldn't put my finger on it why.

davidlimcheng: @christianvuerings Sounds good, I'll do the same thing to `cars_activity.html` as well.

christianvuerings: @davidlimcheng would suggest leaving `cars_activity.html` since it'll make `git blame` / history way harder.

davidlimcheng: @christianvuerings Interestingly, doing it to `cars_activity.html` causes an infinitely spinning spinner - I suspect there's some kind of conflict with the filter.  In any case I opened [SISRP-16841](https://jira.berkeley.edu/browse/SISRP-16841) for further research and to not hold up this PR.

davidlimcheng: Whoops!  Didn't see your last comment, disregard!

src/assets/javascripts/angular/controllers/widgets/billingController.js
davidlimcheng: @christianvuerings Interestingly, when I made these part of `$scope.billing`, it broke the `showMoredirective`.  After a lot of research I'm still having trouble figuring out why, so to avoid holding up this PR more I created [SISRP-16860](https://jira.berkeley.edu/browse/SISRP-16860) for further research.

davidlimcheng:
@christianvuerings Thanks for the direction!  I was able to resolve most of the comments except for one.  I'm hoping that one is minor enough that we can investigate that further later on.  

And yes, I totally agree about running through the `if` statements with Jeff.  Created [SISRP-16861](https://jira.berkeley.edu/browse/SISRP-16861) to keep track of that.

christianvuerings:
Looks like travis is failing on this, just restarted the builds.

davidlimcheng:
Looks like it was failing because of a line break - should be good now.

PR #5109, opened 2016-03-23
SISRP-16714 Add comments and specs for instructing and enrolled section queries

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-16714


raydavis:
Education can be fun!

PR #5112, opened 2016-03-23
SISRP-16719 - Status Holds and Blocks Card - Residency update

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-16719


raydavis:
Vagueness is always a safe policy. Thanks, @christianvuerings

PR #5096, opened 2016-03-21
CLC-6250 Centralize delegate logic

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6250

PR #5113, opened 2016-03-23
CLC-6250 Centralize delegate logic

raydavis:
Master-branch version of https://github.com/ets-berkeley-edu/calcentral/pull/5096

Since the change is in production, we might as well keep it around...


johncrossman:
Yes, we like this code. Let's keep it. Thx, @raydavis

PR #5114, opened 2016-03-24
SISRP-15118, student summary card needs a standalone, split-brain getUserAttributes feed

johncrossman:

The student-overview page for Advisors needs an easy way to pull student attributes using our split-brain logic. This PR extracts logic of EDO/LDAP/Legacy precedence from User::Api and makes it available as a standalone for future use in student_overview_controller. 

**Note:** No change to `api_spec.rb`.


paulkerschen:
While I'm disappointed that the new model isn't called `BrainSplitter`, we can't have everything.

This is all much more readable, @johncrossman; thanks.

PR #5115, opened 2016-03-24
SISRP-16530 - stop nightly run of QA profile tests

pfarestveit:
https://jira.berkeley.edu/browse/SISRP-16530

Oski is no longer reliable as a test user, and I don't want to use real student data for test runs on a server.


johncrossman:
Looks good, @pfarestveit  Waiting on T

PR #5116, opened 2016-03-24
SISRP-16710 - Adds EdoOracle::Queries.get_section_meetings

redconfetti:
https://jira.berkeley.edu/browse/SISRP-16710

Adds EdoOracle::Queries.get_section_meetings method as equivalent of CampusOracle::Queries.get_section_schedules


paulkerschen:
Thanks for documentation @redconfetti.

PR #5117, opened 2016-03-24
SISRP-16877 Add cutoff in YAML to specify per-term academic data source

paulkerschen:
First bit of support for https://jira.berkeley.edu/browse/SISRP-16877.

A YAML cutoff is handy for two reasons:
- Moving it into the future is equivalent to switching off an EdoOracle feature flag entirely.
- Moving it into the past allows devs and QA to do limited verification of EdoOracle against converted legacy data.

### DIFF
--- a/app/models/berkeley/term.rb
+++ b/app/models/berkeley/term.rb
@@ -1,6 +1,7 @@
 module Berkeley
   class Term
     include ActiveAttrModel, ClassLogger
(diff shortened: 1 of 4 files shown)


raydavis:
Whenever I see a hard-coded date, I reach for my YAML. Thank you, @paulkerschen

PR #5119, opened 2016-03-24
SISRP-16711 - Adds EdoOracle::Queries.get_section_instructors

redconfetti:
https://jira.berkeley.edu/browse/SISRP-16711

### DIFF
--- a/app/models/edo_oracle/queries.rb
+++ b/app/models/edo_oracle/queries.rb
@@ -151,5 +151,47 @@ module EdoOracle
       stringify_ints!(result)
     end
 
+    # EDO equivalent of CampusOracle::Queries.get_section_instructors
+    # Changes:
+    #   - 'ccn' replaced by 'section_id' argument
+    #   - 'term_yr' and 'term_yr' replaced by 'term_id'
+    #   - 'instructor_func' has become represented by 'role_code' and 'role_description'
+    #   - Does not provide all user profile fields ('email_address', 'student_id', 'affiliations').
+    #     This will require a programatic join at a higher level.
+    #     See CLC-6239 for implementation of batch LDAP profile requests.
+    #
+    # TODO: Update dependencies in CampusOracle::CourseSections and CanvasCsv::SiteMembershipsMaintainer
+    #   to merge user profile data into this feed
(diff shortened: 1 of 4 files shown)


app/models/edo_oracle/queries.rb
```diff
@@ -151,5 +151,47 @@ def self.get_sections_by_ids(term_id, section_ids)
...
     end
 
+    # EDO equivalent of CampusOracle::Queries.get_section_instructors
+    # Changes:
+    #   - 'ccn' replaced by 'section_id' argument
+    #   - 'term_yr' and 'term_yr' replaced by 'term_id'
+    #   - 'instructor_func' has become represented by 'role_code' and 'role_description'
+    #   - Does not provide all user profile fields ('email_address', 'student_id', 'affiliations').
+    #     This will require a programatic join at a higher level.
+    #     See SISRP-XXXXX for implementation of batch LDAP profile requests.
```
redconfetti: I need to replace this with Ray's Jira that he referenced - CLC-6239

spec/support/custom_matchers.rb
```diff
@@ -0,0 +1,30 @@
...
+    end
+
+    def failure_message_when_negated
+      failure_message
+    end
+  end
+
+  def have_keys(expected_keys)
+    HaveKeys.new(expected_keys)
+  end
```
paulkerschen: @redconfetti This custom matcher looks super handy, but is there any reason you're creating the class from scratch rather than using the `RSpec::Matchers.define` shorthand?

redconfetti: Yes. I don't think it's possible to share the contents of @missing_columns or @target with that method like you can with a class. There is a lot more flexibility with the class. I wonder if I should make it so that this can be used with arrays, checking to ensure each object inside of it (each row), contains the keys specified.... What do you think?

paulkerschen: I accept your reasoning, @redconfetti!

redconfetti: Decided against updating it to support arrays. It would have to track and report which elements are missing which columns. Such an occurrence is not likely, so it's not worth the dev effort.

PR #5121, opened 2016-03-24
SISRP-16877 EdoOracle::UserCourses models and tests

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-16715
https://jira.berkeley.edu/browse/SISRP-16716
https://jira.berkeley.edu/browse/SISRP-16877
https://jira.berkeley.edu/browse/SISRP-16878

EdoOracle::UserCourses contains translator classes on the pattern of CampusOracle::UserCourses. There is some oddness and redundancy in the constructed feeds because that's what the higher-level classes are expecting.

Cross-listings can now be determined from the cs-course-id column; implicitly taught sections can be determined from the primaryAssociatedSection column. Both are tracked while building the UserCourses feed for instructors.

Specs are present but will need better data to exercise a wider range of cases.

Legacy term cutoff logic is moved into Berkeley::Terms initialization so as to avoid a bunch of unnecessary cache hits.

### DIFF
--- a/app/models/berkeley/term.rb
+++ b/app/models/berkeley/term.rb
@@ -62,7 +62,11 @@ module Berkeley
     end
 
     def legacy?
-      self <= Berkeley::Terms.fetch.legacy_cutoff
+      @is_legacy || false
+    end
+
+    def set_as_legacy
+      @is_legacy = true
     end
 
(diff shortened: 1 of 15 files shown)


raydavis:
Looks like it was a good Friday!

PR #5122, opened 2016-03-26
SISRP-15118, advisor_student_overview page and initial student summary widget

johncrossman:

Includes:
* groundwork for Advisor's student_overview page
* `aggregated_attributes` becomes a proper feed with camelCase keys

### DIFF
--- a/src/assets/templates/advisor_student_overview.html
+++ b/src/assets/templates/advisor_student_overview.html
@@ -1,0 +1,15 @@
+<div data-ng-if="api.user.profile.features.csAdvisorStudentLookup && (api.user.profile.isViewer || api.user.profile.isSuperuser || api.user.profile.roles.advisor)">
+  <div class="row collapse">
+    <div class="column">
+      <h1 class="cc-heading-page-title"><a href="/">My Dashboard</a> &raquo; Student Overview</h1>
+    </div>
+    <div class="medium-4 columns">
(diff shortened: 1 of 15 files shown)


src/assets/templates/advisor_student_overview.html
```diff
@@ -0,0 +1,14 @@
+<div>
```
christianvuerings: @johncrossman should this be behind a feature flag?
Also, can we make sure we see a 404 page when the user doesn't have the `advisor` role?

johncrossman: Thanks @christianvuerings 
* I'll drop in a new feature flag, if necessary: https://jira.berkeley.edu/browse/SISRP-16963
* I've put in the 404 logic and it matches the restriction we have in `StudentOverviewController: can_view_as_for_all_uids?`

christianvuerings:
@johncrossman thanks for making the 404 update. I would suggest though that we add the feature flag before merging.

johncrossman:
@christianvuerings : added feature flag, spec and controller-level check.

christianvuerings:
@johncrossman Thanks - lgtm as soon as Travis failure (`StudentLookupFeatureFlagged not found`) is resolved.

johncrossman:
Whoops, forgot to git-add. All in now. Thx, @christianvuerings

PR #5123, opened 2016-03-27
SISRP-16726 Filter FinAid Communications for delegates

raydavis:
https://jira.berkeley.edu/browse/SISRP-16726


johncrossman:
Looks right to me, @raydavis  Thank you.

PR #5124, opened 2016-03-28
CLC-6228, defaults for advisor during view-as session

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-6228

`accessible_during_advisor_view_as?` is true by default in all controllers. This PR puts override in Canvas-related controllers and `OecTasksController`. Test coverage on high-risk endpoints.


app/controllers/application_controller.rb
```diff
@@ -54,9 +54,15 @@ def check_reauthentication
   def accessible_by_delegate?
     false
   end
+  # The majority of API feeds are available to advisors during view-as mode. Remove advisor access by overriding this method.
+  def accessible_during_advisor_view_as?
```
raydavis: Could we use the same method naming convention as for delegate? You can change the delegate version if you think the new name is better.

app/controllers/concerns/unavailable_when_advisor_view_as.rb
```diff
@@ -0,0 +1,7 @@
+module UnavailableWhenAdvisorViewAs
```
raydavis: Again, I think the code might be easier to maintain if we align the names.

app/controllers/application_controller.rb
johncrossman: Sure thing, @raydavis  I did not use `accessible_by_advisor?` for risk of confusion. I'll align the names.

raydavis:
Aside from the naming nitpick, this looks spot on.

johncrossman:
@raydavis My renaming included `DelegateAccessible` to `AllowDelegateViewAs` thus many files touched. Travis is green.

johncrossman:
@raydavis  Please don't merge this until #5123 is merged and I've rebased to resolved conflicts on this PR.

PR #5128, opened 2016-03-28
NOJIRA, add cs_billing endpoint to verify_sis_api_endpoints script

johncrossman:
New endpoint: `/BILLING_API/get?EMPLID=`

`echo_usage()` moved up to serve as both function and script documentation.

### DIFF
--- a/script/sis/verify_sis_api_endpoints.sh
+++ b/script/sis/verify_sis_api_endpoints.sh
@@ -1,12 +1,43 @@
 #!/bin/bash
 
-# cd to 'calcentral' directory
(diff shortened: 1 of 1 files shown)


johncrossman:
Note: `/${CAMPUS_SOLUTIONS_ID}?id-type=uid` was deliberately removed. It is not used in our codebase.

paulkerschen:
Thanks @johncrossman

christianvuerings:
@davidlimcheng FYI - we might have to update this script once we have the actual URL in CS.

christianvuerings:
@davidlimcheng FYI - we might have to update this script once we have the actual URL in CS.

PR #5129, opened 2016-03-29
SISRP-16063, Advisor student overview, university requirements widget

johncrossman:
https://jira.berkeley.edu/browse/SISRP-16063

Advisors get a dedicated `student_university_requirements` card. The `/api/student/${uid}` feed is expanded to include academics.


src/assets/templates/widgets/advisor/student_university_requirements.html
```diff
@@ -0,0 +1,32 @@
+<div class="cc-widget cc-widget-university-requirements">
```
christianvuerings: @johncrossman this looks like close a copy / paste of the `academics_university_requirements`. Is there no way we can at least re-use part of it?

johncrossman: @christianvuerings Yes, essentially copy/paste. I was uneasy w/ having advisors and students shared widget code. However, some reuse is possible if we put in comments of caution. Stay tuned...

johncrossman:
I'll re-submit this PR shortly with university-requirement widget shared between use cases. cc: @christianvuerings

PR #5137, opened 2016-03-29
SISRP-16880 EDO DB knows no Booleans

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-16880

### DIFF
--- a/app/models/edo_oracle/user_courses/base.rb
+++ b/app/models/edo_oracle/user_courses/base.rb
@@ -119,17 +119,20 @@ module EdoOracle
         section_data = {
           ccn: row['section_id'].to_s,
           instruction_format: row['instruction_format'],
-          is_primary_section: row['primary'],
+          is_primary_section: to_boolean(row['primary']),
           section_label: "#{row['instruction_format']} #{row['section_num']}",
           section_number: row['section_num']
         }
-        section_data[:units] = row['units'] if row['primary']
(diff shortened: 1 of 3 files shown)


app/models/edo_oracle/user_courses/base.rb
```diff
@@ -119,17 +119,17 @@ def row_to_section_data(row, cross_listing_tracker=nil)
         section_data = {
           ccn: row['section_id'].to_s,
           instruction_format: row['instruction_format'],
-          is_primary_section: row['primary'],
+          is_primary_section: (row['primary'] == 'true'),
```
johncrossman: @paulkerschen : how about a concise util. E.g., `to_b(string)`?  (No booleans in the db brings back memories of Matterhorn. Ouch.)

paulkerschen:
Thanks, @johncrossman, util method seems warranted.

paulkerschen:
Amended to use a utility method, and to rely on the stored boolean instead of converting over and over.

app/models/edo_oracle/user_courses/base.rb
```diff
@@ -166,6 +169,10 @@ def merge_detailed_section_data(campus_classes)
         end
       end
 
+      def to_boolean(string)
```
redconfetti: @paulkerschen `true` dat!

PR #5140, opened 2016-03-30
SISRP-16880 EdoOracle::Queries: 'true' > 'false'

paulkerschen:
Follow-up to #5137 (fun with booleans that ain't booleans). We want primary sections ordered first.

https://jira.berkeley.edu/browse/SISRP-16880


raydavis:
So 'true'!

PR #5144, opened 2016-03-30
SISRP-16962, CalCentral must consider isPrimaryForIdentifierType flag in Crosswalk response

johncrossman:
https://jira.berkeley.edu/browse/SISRP-16962

Note:
* We do a conservative check on `isPrimaryForIdentifierType`: evaluate to false if and only if element is present and equal to false.
* The `isActive` attribute has been unreliable in production so we give it no import in our spec / fixture.


johncrossman:
Sign off prior to merge should include @raydavis  I made a rogue decision to prioritize the issue based on frustrations seen in Perf Testing yesterday. IMO, this is a must fix for GL 5.2

raydavis:
Who am I to interfere with roguishness?

PR #5147, opened 2016-03-30
SISRP-17144 Do not create unusable link to Schedule Planner

raydavis:
https://jira.berkeley.edu/browse/SISRP-17144

### DIFF
--- a/app/controllers/campus_solutions/campus_solutions_controller.rb
+++ b/app/controllers/campus_solutions/campus_solutions_controller.rb
@@ -22,7 +22,7 @@ module CampusSolutions
     def exclude_acting_as_users
       return unless Settings.features.prevent_acting_as_users_from_posting
       unless current_user.directly_authenticated?
-        logger.warn "ACT-AS: User #{current_user.original_user_id} attempted access to an endpoint which is forbidden while acting-as user #{current_user.user_id}"
+        logger.warn "ACT-AS: User #{current_user.real_user_id} attempted access to an endpoint which is forbidden while acting-as user #{current_user.user_id}"
         render :nothing => true, :status => 403
       end
     end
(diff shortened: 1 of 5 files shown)


app/controllers/campus_solutions/campus_solutions_controller.rb
```diff
@@ -22,7 +22,7 @@ def delete_passthrough(classname)
     def exclude_acting_as_users
       return unless Settings.features.prevent_acting_as_users_from_posting
       unless current_user.directly_authenticated?
-        logger.warn "ACT-AS: User #{current_user.original_user_id} attempted access to an endpoint which is forbidden while acting-as user #{current_user.user_id}"
+        logger.warn "ACT-AS: User #{current_user.real_user_id} attempted access to an endpoint which is forbidden while acting-as user #{current_user.user_id}"
```
raydavis: The old code would not have shown Advisor or Delegate UIDs.

christianvuerings:
Change lgtm - I'll also let @paulkerschen  or @johncrossman have a look at this one.

src/assets/templates/widgets/enrollment/schedule.html
```diff
@@ -16,7 +16,7 @@
   </div>
 
   <div
-    data-ng-if="!api.user.profile.actAsOptions.canSeeCSLinks"
+    data-ng-if="!api.user.profile.actAsOptions.isDirectlyAuthenticated"
     data-ng-include="'widgets/enrollment/enrollment_hide_links.html'">
```
johncrossman: Do we need same treatment for decide.html, explore.html and plan.html in directory `src/assets/templates/widgets/enrollment`?

johncrossman: After a closer look, I think the answer is no. I'll let @raydavis confirm.

raydavis: I've asked for the folks what should know to tell me what other magic URLs need more protection. I haven't heard of any yet and I don't think we should wait to resolve the blocker. In other words, confirmed. :)

johncrossman:
Thanks, @raydavis  The known knowns are verified. Unknown unknowns are still out there.

PR #5148, opened 2016-03-30
SISRP-17167 - Retrofit Canvas::CurrentTeacher#user_currently_teaching?

redconfetti:
https://jira.berkeley.edu/browse/SISRP-17167

* Introduces Berkeley::Terms.legacy_group to segregate terms into a hash with `:legacy` and `:sisedo` arrays.
* Updated EdoOracle::Queries.has_instructor_history? so that it returns `false` when terms array is empty
* Ensures CampusOracle::Queries.has_instructor_history? returns false when terms array is empty
* Retrofits Canvas::CurrentTeacher to use User::AcademicHistory.has_instructor_history? with support for legacy and SISEDO sources

### DIFF
--- a/app/models/edo_oracle/queries.rb
+++ b/app/models/edo_oracle/queries.rb
@@ -296,8 +296,11 @@ module EdoOracle
     # EDO equivalent of CampusOracle::Queries.has_instructor_history?
     def self.has_instructor_history?(ldap_uid, instructor_terms = nil)
       result = {}
-      return result if fake?
-      terms_list = terms_query_list(instructor_terms)
+      return false if fake?
+      if instructor_terms.to_a.any?
+        terms_list = terms_query_list(instructor_terms.to_a)
+        instructor_term_clause = "AND instr.\"term-id\" IN (#{terms_list})"
+      end
       use_pooled_connection {
         sql = <<-SQL
         SELECT
@@ -306,8 +309,8 @@ module EdoOracle
           SISEDO.ASSIGNEDINSTRUCTORV00_VW instr
         WHERE
           instr."campus-uid" = '#{ldap_uid}' AND
-          rownum < 2 AND
-          instr."term-id" IN (#{terms_list})
+          rownum < 2
+          #{instructor_term_clause}
         SQL
         result = connection.select_one(sql)
       }
@@ -317,8 +320,11 @@ module EdoOracle
 
     def self.has_student_history?(ldap_uid, student_terms = nil)
       result = {}
-      return result if fake?
-      terms_list = terms_query_list(student_terms)
+      return false if fake?
+      if student_terms.to_a.any?
+        terms_list = terms_query_list(student_terms)
+        student_term_clause = "AND enroll.\"TERM_ID\" IN (#{terms_list})"
+      end
       use_pooled_connection {
         sql = <<-SQL
         SELECT
@@ -327,8 +333,8 @@ module EdoOracle
           SISEDO.ENROLLMENTV00_VW enroll
         WHERE
           enroll."CAMPUS_UID" = '#{ldap_uid.to_i}' AND
-          rownum < 2 AND
-          enroll."TERM_ID" IN (#{terms_list})
+          rownum < 2
+          #{student_term_clause}
         SQL
         result = connection.select_one(sql)
       }
--- a/app/models/user/academic_history.rb
+++ b/app/models/user/academic_history.rb
@@ -1,0 +1,10 @@
+module User
+  class AcademicHistory < UserSpecificModel
+    def has_instructor_history?(current_terms)
+      grouped_terms = Berkeley::Terms.legacy_group(current_terms)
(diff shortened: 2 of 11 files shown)


raydavis:
RSpec failures in Travis.

app/models/edo_oracle/queries.rb
```diff
@@ -296,8 +296,11 @@ def self.get_enrolled_students(section_id, term_id)
     # EDO equivalent of CampusOracle::Queries.has_instructor_history?
     def self.has_instructor_history?(ldap_uid, instructor_terms = nil)
       result = {}
-      return result if fake?
-      terms_list = terms_query_list(instructor_terms)
+      return false if fake?
+      if instructor_terms.to_a.any?
```
redconfetti: I just discovered from the last failing test that the CampusOracle::Queries methods would check all current terms if no terms were provided. Somehow I had gotten the impression that they would return `false` if no terms were provided. This update ensures that the behavior is the same for EdoOracle::Queries with these methods.

app/models/user/academic_history.rb
```diff
@@ -0,0 +1,10 @@
+module User
+  class AcademicHistory < UserSpecificModel
+    def has_instructor_history?(current_terms)
+      grouped_terms = Berkeley::Terms.legacy_group(current_terms)
+      has_legacy_instructor_history = CampusOracle::Queries.has_instructor_history?(@uid, grouped_terms[:legacy])
+      has_sisedo_instructor_history = EdoOracle::Queries.has_instructor_history?(@uid, grouped_terms[:sisedo])
+      return has_legacy_instructor_history || has_sisedo_instructor_history
```
paulkerschen: @redconfetti Extra `return` here.

Also, what would you think about not storing the two booleans in variable, but just returning the expression `CampusOracle::Queries.has_instructor_history?(@uid, grouped_terms[:legacy]) || EdoOracle::Queries.has_instructor_history?(@uid, grouped_terms[:sisedo])`. The advantage is that if the CampusOracle part of the query returns true, it will never have to call EdoOracle, saving load on that system.

redconfetti: @paulkerschen Good call. I did it this way for easy readability, but the performance note overrides that.

app/models/user/academic_history.rb
```diff
@@ -0,0 +1,10 @@
+module User
+  class AcademicHistory < UserSpecificModel
+    def has_instructor_history?(current_terms)
+      grouped_terms = Berkeley::Terms.legacy_group(current_terms)
+      has_legacy_instructor_history = Proc.new { CampusOracle::Queries.has_instructor_history?(@uid, grouped_terms[:legacy]) }
+      has_sisedo_instructor_history = Proc.new { EdoOracle::Queries.has_instructor_history?(@uid, grouped_terms[:sisedo]) }
+      return has_legacy_instructor_history.call || has_sisedo_instructor_history.call
```
redconfetti: @paulkerschen What do you think about this proc approach? Better performance, more readable code.

paulkerschen: Works for me, @redconfetti, thanks! Will merge post Travis.

redconfetti:
One of these is failing due to an unrelated test:
```
Pending:
  CampusSolutions::Deposit SISRP-13071
    # No reason given
    # ./spec/models/campus_solutions/deposit_spec.rb:27

Failures:

  1) Textbooks::Proxy#get fake proxy feed including malformed item logs error and skips bad entry
     Failure/Error: subject { fake_proxy.get }
       #<Log4r::Logger:0x7d3772e0> received :error with unexpected arguments
         expected: (/invalid ISBN/)
              got: ("SLOW PROXY duration: 30095ms  {:url=>\"https://www.googleapis.com/books/v1/volumes?q=isbn:9780205779888\", :class=>Textbooks::Proxy, :exception=>[\"Errors::ProxyError\", \"Timeout error; url: https://www.googleapis.com/books/v1/volumes?q=isbn:9780205779888\"]} ")
     # ./config/initializers/proxy_notifications.rb:6:in `(root)'
     # ./lib/proxies/http_client.rb:6:in `get_response'
     # ./lib/proxies/base_proxy.rb:20:in `get_response'
     # ./app/models/textbooks/proxy.rb:25:in `google_book'
     # ./app/models/textbooks/proxy.rb:45:in `process_material'
     # ./app/models/textbooks/proxy.rb:89:in `process_response'
     # ./app/models/textbooks/proxy.rb:88:in `process_response'
     # ./app/models/textbooks/proxy.rb:86:in `process_response'
     # ./app/models/textbooks/proxy.rb:115:in `get'
     # ./spec/models/textbooks/proxy_spec.rb:72:in `subject'
     # ./spec/models/textbooks/proxy_spec.rb:6:in `it_is_a_normal_server_response'
     # ./spec/models/textbooks/proxy_spec.rb:96:in `(root)'

Finished in 4 minutes 15.9 seconds (files took 22.97 seconds to load)
2945 examples, 1 failure, 1 pending

Failed examples:

rspec ./spec/models/textbooks/proxy_spec.rb:94 # Textbooks::Proxy#get fake proxy feed including malformed item logs error and skips bad entry
```

PR #5149, opened 2016-03-30
SISRP-16961 User::BasicAttributes module and spec

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-16961


raydavis:
Neat as a pin. Thanks!

PR #5152, opened 2016-03-30
SISRP-16877 Handle sparse results from EDO DB

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-16877

The nature of the course-section join in the EDO DB is such that the "course" half of the row might come back null. We must adapt or perish.


johncrossman:
Doing the best we can with what we get. Thx, @paulkerschen

PR #5154, opened 2016-03-31
SISRP-17174 Implement basic fake mode for EDO DB

raydavis:
https://jira.berkeley.edu/browse/SISRP-17174

WARNING: A no-op "fake: true" development.yml configuration setting (along with misleading default "edodb" values in settings.yml) wrongly slipped into an earlier commit. This default setting will now have an effect, and developers who want real EDO DB access must update their development.local.yml files.


redconfetti:
Thanks @raydavis

image

PR #5155, opened 2016-03-31
SISRP-15857 EDO DB: Fix instructor and secondary-section queries

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-15857


app/models/edo_oracle/queries.rb
```diff
@@ -210,7 +210,7 @@ def self.get_section_instructors(term_id, section_id)
             instr."term-id" = sec."term-id" AND
             instr."session-id" = sec."session-id" AND
             instr."offeringNumber" = sec."offeringNumber" AND
-            instr."number" = sec."number"
+            instr."number" = sec."sectionNumber"
```
redconfetti: image

PR #5158, opened 2016-03-31
CLC-6239 Switch bCourses user account data source to Oracle-LDAP merge

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6239

Most of this PR teaches bCourses to accept the slightly more abstracted User::BasicAttributes structure in lieu of an Oracle row. 

The only (intended) change to logic is removal of the code path that handles Oracle-LDAP disjuncture.

### DIFF
--- a/app/models/canvas_csv/maintain_users.rb
+++ b/app/models/canvas_csv/maintain_users.rb
@@ -134,7 +134,7 @@ module CanvasCsv
       end
     end
 
-    def categorize_user_account(existing_account, campus_user_rows)
+    def categorize_user_account(existing_account, campus_user_attributes)
       # Convert from CSV::Row for easier manipulation.
       old_account_data = existing_account.to_hash
       parsed_login_id = self.class.parse_login_id old_account_data['login_id']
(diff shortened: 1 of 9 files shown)


app/models/canvas_csv/maintain_users.rb
```diff
@@ -142,18 +142,12 @@ def categorize_user_account(existing_account, campus_user_rows)
...
-          new_account_data = canvas_user_from_campus_row(campus_row)
+          new_account_data = canvas_user_from_campus_attributes campus_user
         else
           return unless Settings.canvas_proxy.inactivate_expired_users
-          if (ldap_result = CalnetLdap::Client.new.search_by_uid(ldap_uid)).present?
-            # Our LDAP bind does not provide enough data to fully update the Canvas account, and so
-            # all we can do is log the disconnect between LDAP and our campus DB.
-            logger.error "UID #{ldap_uid} is not in DB but LDAP reports #{ldap_result.inspect}"
-            return
-          end
```
paulkerschen: My assumption is that this block of code no longer gets exercised with a merged Oracle-LDAP source, but inactivation expert @raydavis should confirm.

raydavis: Right, @paulkerschen . The comment is obsolete (as all comments eventually become), and now not only _can_ we update the Canvas account from LDAP, we _must_ do so.

PR #5161, opened 2016-03-31
SISRP-16536 - QA test for Final Grades, GPA Calc cards

pfarestveit:
https://jira.berkeley.edu/browse/SISRP-16536

New Selenium test to verify presentation on the Final Grades and GPA Calculator cards and to support academics retrofit testing.  Does not include functional testing of the GPA Calc card, which will be added when there's time to do it.


pfarestveit:
@paulkerschen - thanks for the tip!  I fixed the relevant methods, and I'll fix up other test code as time permits.

PR #5163, opened 2016-04-01
SISRP-15118, advising student summary card with filtered set of links

johncrossman:

Includes:
* standardize naming and URL routes (e.g., advisor_student_overview.html -> advising_student.html)
* filtered set of CS links for Advisors; filtering is server-side
* reuse of some existing widget templates via data-ng-includes

*Note:* I did a `git mv` in order to rename files and then continued modifying code. This seemed to confuse Git because the renaming shows as a delete + add. This is unreleased code so I don't think retaining Git history is vital yet.


src/assets/javascripts/angular/factories/advisorStudentOverviewFactory.js
```diff
@@ -1,16 +0,0 @@
...
- */
-angular.module('calcentral.factories').factory('advisorStudentOverviewFactory', function($http) {
-  var getAcademics = function(options) {
-    return $http.get('/api/advising/academics/' + options.uid);
-  };
-
-  return {
-    getAcademics: getAcademics
-  };
-});
```
johncrossman: This js factory was tossed out in favor of `advisingFactory.js`

src/assets/templates/widgets/advisor/student_profile.html
```diff
@@ -3,6 +3,22 @@
...
+    Terms in Attendance <div data-ng-bind="student.attributes.defaultName"></div>
+
+    <hr/>
+
+    Contact Information
+
+    <hr/>
+
+    <div data-ng-include="'widgets/advisor/advising_links.html'"></div>
+
```
johncrossman: A work in progress. Do not consider this done.

raydavis:
The back-end name change sounds reasonable.

paulkerschen:
Standardization and use of includes makes sense to me. Thanks @johncrossman

PR #5176, opened 2016-04-05
SISRP-16006 - FERPA display message text needs confirmation

dfkaye:
https://jira.berkeley.edu/browse/SISRP-16006

### DIFF
--- a/src/assets/templates/widgets/profile/information_disclosure.html
+++ b/src/assets/templates/widgets/profile/information_disclosure.html
@@ -1,10 +1,10 @@
 <div class="cc-widget-text" data-ng-controller="InformationDisclosureController">
   <p>
-    The university's internal directory information includes personal information such as your address, email, name, phone, and portions of your academic record.
+    The university's directory information includes personal information such as your address, email, name, phone, and portions of your academic record.
   </p>
   <div data-cc-spinner-directive="ferpa.isLoading">
     <p>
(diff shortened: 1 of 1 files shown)


src/assets/templates/widgets/profile/information_disclosure.html
```diff
@@ -17,4 +17,7 @@
   <p>
     To learn more, view <a href="http://registrar.berkeley.edu/GeneralInfo/ferpa.html">UC Berkeley's policy governing disclosure of information from student records</a>.
   </p>
+  <p>
+    Public directory information can be updated at <a href="https://calnet.berkeley.edu/directory/update/index.pl">https://calnet.berkeley.edu/directory/update/index.pl</a>.
+  </p>
```
johncrossman: @dfkaye I would check with BenH or @raydavis about hiding `ferpa.deeplink.url` and CalNet URL from delegates and/or advisors.

raydavis: Thanks for the ping, @johncrossman . Neither of these links are "magic URLs", so if delegates and/or advisors are shown "profile/information_disclosure.html" at all, there's no reason to mask these particular elements. That said, I don't know why we'd show that part of the Profile to delegates or advisors. :)

johncrossman: Thanks @raydavis @dfkaye

PR #5178, opened 2016-04-05
SISRP-17477 Do not translate GRAD DIV affiliation for profile

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-17477


paulkerschen:
Updated to remove rogue extra commit.

PR #5181, opened 2016-04-05
SISRP-17479 Suppress CampusOracle-derived residency status and messaging

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-17479

Residency status is no longer being kept current in DB2, and must be suppressed in CalCentral until we switch to Campus Solutions as source of record. We've been told to suppress the accompanying message and link as well.

### DIFF
--- a/app/models/campus_oracle/user_attributes.rb
+++ b/app/models/campus_oracle/user_attributes.rb
@@ -25,12 +25,10 @@ module CampusOracle
         result[:roles] = roles_from_campus_row result
         result.merge! Berkeley::SpecialRegistrationProgram.attributes_from_code(result['reg_special_pgm_cd'])
 
-        if term_transition?
-          result[:california_residency] = nil
-          result[:reg_status][:transitionTerm] = true
-        else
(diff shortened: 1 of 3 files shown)


app/models/campus_oracle/user_attributes.rb
```diff
@@ -25,12 +25,10 @@ def get_feed_internal
...
 
-        if term_transition?
-          result[:california_residency] = nil
-          result[:reg_status][:transitionTerm] = true
-        else
-          result[:california_residency] = Berkeley::CalResidency.california_residency_from_campus_row result
-        end
+        result[:reg_status][:transitionTerm] = true if term_transition?
+
+        # SISRP-17479 CampusOracle residency information is no longer current and must be suppressed.
```
johncrossman: @paulkerschen Not a big deal but... My personal opinion is that code should only have Jira numbers when it's either:
* a pending specs (i.e., "fix me" items)
* a TODO

johncrossman:
Looks good, @paulkerschen My comment above can be addressed later, if at all. This is ready to merge.

PR #5182, opened 2016-04-06
SISRP-16933, advisors can only view-as students and applicants

johncrossman:
https://jira.berkeley.edu/browse/SISRP-16933

This PR does not include additional user attribute based on LDAP: STUDENT-STATUS-EXPIRED status. My next PR will have that portion.

Advisors had been riding the coattails of `policy.can_view_as_for_all_uids?` but no more. With this PR, advisors can only view-as students and applicants. The level of view-as power is relevant when querying stored_users, initiating view-as session and viewing student profile. It is those controllers which now reference filters in `AdvisorAuthorization` and `ViewAsAuthorization`.


paulkerschen:
Thanks for extracting the advisor authorization, @johncrossman; clearly represented.

PR #5183, opened 2016-04-06
SISRP-17104 Cache fixes for CampusSolutions, HubEdos feeds

raydavis:
https://jira.berkeley.edu/browse/SISRP-17104

Last batch for now. Similar problems doubtlessly lurk in other classes, but the changes thus far cover the feeds directly involved in SISRP-16997 and make a pretty big QA footprint.

* LiveUpdatesEnabled with FreshenOnWarm bypasses the feed's configured lifespan while the user is logged in by forcing a feed refresh every 5 minutes. We don't want to hit external APIs that often, so this should only be done if the feed is made up of other source feeds all of which are themselves cached, the idea being to catch them not long after they expire.
* All server-side feeds marked as LiveUpdatesEnabled will react to the 5-minute ping. But the only ones which get checked by the browser are the 6 listed in "updatedFeedsService.js". The others won't be noticed on the front-end until the user refreshes their page, just as if LiveUpdatesEnabled wasn't there, and so the server-side CPU activity is pointless. (Selenium-script warning: removing LiveUpdatesEnabled also removes  the feed's injected "lastModified" and "feedName" properties.)
* If a feed is only called by a controller, JsonifiedFeed can save CPU and memory by caching only the JSON version. (JsonifiedFeed is incompatible with LiveUpdatesEnabled and FilteredViewAsFeed.)
* UserCacheExpiry tells HotPlate to force a refresh when it reaches that UID. It can help us out with production issues via "/api/cache/warm/:uid", which starts a single-person HotPlate run.

### DIFF
--- a/app/models/campus_solutions/my_academic_plan.rb
+++ b/app/models/campus_solutions/my_academic_plan.rb
@@ -1,8 +1,9 @@
 module CampusSolutions
   class MyAcademicPlan < UserSpecificModel
 
-    include Cache::LiveUpdatesEnabled
-    include Cache::JsonAddedCacher
+    include Cache::CachedFeed
+    include Cache::JsonifiedFeed
+    include Cache::UserCacheExpiry
     include EnrollmentCardFeatureFlagged
 
     attr_accessor :term_id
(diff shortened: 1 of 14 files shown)


raydavis:
Pinging @pfarestveit , @johncrossman , @paulkerschen .

paulkerschen:
Thank you, @raydavis. My impression is that adding Live Updates to all those `my_` models was a small initial oversight that spread like kudzu once established. This looks like the sane fix to me.

johncrossman:
Looks right to me, @raydavis  Thank you.

PR #5185, opened 2016-04-06
SISRP-16933, surface 403 error message on View As widget

johncrossman:
https://jira.berkeley.edu/browse/SISRP-16933

The view-as related controllers, which include *AdvisorAuthorization*, override new `render_403(error)` in *ApplicationController*. This gives Advisors an explanation for denied access to non-student UIDs.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -113,6 +113,11 @@ class ApplicationController < ActionController::Base
 
   def user_not_authorized(error)
     Rails.logger.warn "Unauthorized request made by UID: #{session['user_id']} to #{controller_name}\##{action_name}: #{error.message}"
+    render_403 error
+  end
+
+  def render_403(error)
+    # Subclasses might render JSON including error message.
     render :nothing => true, :status => 403
   end
 
--- a/app/controllers/act_as_controller.rb
+++ b/app/controllers/act_as_controller.rb
@@ -1,5 +1,5 @@
 class ActAsController < ApplicationController
-
+  include ViewAsAuthorization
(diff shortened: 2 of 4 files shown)


app/controllers/application_controller.rb
```diff
@@ -113,6 +113,11 @@ def check_google_access
 
   def user_not_authorized(error)
     Rails.logger.warn "Unauthorized request made by UID: #{session['user_id']} to #{controller_name}\##{action_name}: #{error.message}"
+    render_403 error
+  end
+
+  def render_403(error)
+    # Subclasses might put error message to the response.
     render :nothing => true, :status => 403
   end
```
johncrossman: Fyi, I did not change the logic in `application_controller` because that would require updating a dozen or more specs.

app/controllers/application_controller.rb
```diff
@@ -113,6 +113,11 @@ def check_google_access
 
   def user_not_authorized(error)
     Rails.logger.warn "Unauthorized request made by UID: #{session['user_id']} to #{controller_name}\##{action_name}: #{error.message}"
+    render_403 error
+  end
+
+  def render_403(error)
+    # Subclasses might put error message to the response.
```
paulkerschen: @johncrossman This sentence confused me for a moment as I thought it had something to do with the PUT HTTP method. Is there perhaps a clearer phrasing?

johncrossman: I agree, @paulkerschen   Amendment pushed.

paulkerschen:
Bueno @johncrossman

PR #5186, opened 2016-04-06
SISRP-17554 Return empty User::BasicAttributes for empty enrollments

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-17554

Avoid generating invalid SQL when passed an empty array.

Also make the rosters front end do the right thing when handed an empty array of students, and fix the longstanding annoyance where "Students have not yet signed up for this class" always appears beneath the "You must be a teacher in this bCourses course" auth failure message.

### DIFF
--- a/app/models/user/basic_attributes.rb
+++ b/app/models/user/basic_attributes.rb
@@ -3,6 +3,7 @@ module User
     extend self
 
     def attributes_for_uids(uids)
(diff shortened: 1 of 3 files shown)


johncrossman:
Say "no" to invalid SQL. Thanks @paulkerschen

PR #5187, opened 2016-04-07
SISRP-17560 Correctly sort sparse EdoOracle course rows

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-17560

Follow-up to #5152 leverages that PR's `course_ids_from_row` implementation to ensure that rows are correctly sorted by course code even when `dept_name` and `catalog_id` are missing.

### DIFF
--- a/app/models/edo_oracle/user_courses/base.rb
+++ b/app/models/edo_oracle/user_courses/base.rb
@@ -67,7 +67,8 @@ module EdoOracle
       end
 
       def row_to_feed_item(row, previous_item, cross_listing_tracker=nil)
-        unless (course_item = new_course_item(row, previous_item))
+        course_item = course_ids_from_row row
+        if course_item[:id] == previous_item[:id]
           previous_item[:sections] << row_to_section_data(row, cross_listing_tracker)
           nil
         else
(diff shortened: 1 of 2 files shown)


app/models/edo_oracle/user_courses/base.rb
```diff
@@ -90,9 +90,8 @@ def row_to_feed_item(row, previous_item, cross_listing_tracker=nil)
       end
 
       def new_course_item(row, previous_item)
-        if row.values_at('dept_name', 'catalog_id', 'term_id') != previous_item.values_at(:dept, :catid, :term_id)
-          course_ids_from_row row
-        end
+        course_item = course_ids_from_row row
+        course_item unless course_item[:id] == previous_item[:id]
```
johncrossman: @paulkerschen Logic looks good. For the sake of readability (e.g., fewer `unless`), what do you think about ditching `new_course_item` method and moving the two lines above to the start of `row_to_feed_item`? I.e.,
```
def row_to_feed_item(...)
  course_item = course_ids_from_row row
  if (course_item[:id] == previous_item[:id])
    # Handle cross-listing
    ...
  else
    ...
```

paulkerschen: Good call, @johncrossman; much easier to understand. Amended.

johncrossman:
Nicely done. Thx @paulkerschen

PR #5190, opened 2016-04-07
SISRP-17446, explain why 403 on /advising/student page; clickable email

johncrossman:
https://jira.berkeley.edu/browse/SISRP-17446

User gets:
* error summary
* clickable email address 
* spinner stop via `finally`

### DIFF
--- a/src/assets/javascripts/angular/controllers/pages/advisingStudentController.js
+++ b/src/assets/javascripts/angular/controllers/pages/advisingStudentController.js
@@ -11,16 +11,32 @@ angular.module('calcentral.controllers').controller('AdvisingStudentController',
     isLoading: true,
     excludeLinksToRegistrar: true
   };
+  $scope.student = {
+    isLoading: true
+  };
+
+  var defaultErrorDescription = function(status) {
+    return status === 403 ?
+      'You are not authorized to view this user\'s data.' :
+      'Sorry, there was a problem fetching this user\'s data. Contact CalCentral support if the error persists.';
+  };
+
+  var errorReport = function(status, errorDescription) {
+    return {
+      summary: status === 403 ? 'Access Denied' : 'Unexpected Error',
+      description: errorDescription || defaultErrorDescription(status)
+    };
+  };
 
   var loadProfile = function() {
     advisingFactory.getStudent({
       uid: $routeParams.uid
-    }).then(function(data) {
-      $scope.student = _.get(data, 'data.attributes');
+    }).success(function(data) {
+      angular.extend($scope.student, _.get(data, 'attributes'));
       $scope.student.uid = $routeParams.uid;
-      $scope.student.addresses = apiService.profile.fixFormattedAddresses(_.get(data, 'data.contacts.feed.student.addresses'));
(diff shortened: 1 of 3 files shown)


src/assets/javascripts/angular/controllers/pages/advisingStudentController.js
```diff
@@ -29,20 +45,26 @@ angular.module('calcentral.controllers').controller('AdvisingStudentController',
...
     });
   };
 
   var loadAcademics = function() {
     advisingFactory.getStudentAcademics({
       uid: $routeParams.uid
-    }).then(function(data) {
-      $scope.academics = _.get(data, 'data.academics');
+    }).success(function(data) {
+      $scope.academics = _.get(data, 'academics');
```
paulkerschen: @johncrossman Is `_get` needed here, now that the intervening `.data` object is gone?

johncrossman: @paulkerschen I had same question. We have similar usages in *enrollmentCardController.js*, *profilePhoneController.js*, etc. but I doubt lodash is needed here. I'll test.

johncrossman:
Interesting lesson learned, @paulkerschen   I actually need `_.get()` in these cases but for an unexpected reason. I need a default of `{}` when *data.attributes* is nil because subsequent lines add elements to the hash. Amendment is in.

johncrossman:
Lo and behold, scrutinizing the `_get()` led me to a related problem: `student.isLoading` status was lost when `$scope.student` was set. This has also been amended (see `studentProfile.isLoading`).  cc: @paulkerschen

johncrossman:
@paulkerschen I'm going to use `angular.extend` instead.  Please wait for one more amendment.

src/assets/templates/widgets/advisor/student_profile.html
```diff
@@ -2,82 +2,90 @@
...
-            <br/>
-            <div data-ng-repeat="phone in student.phones">
-              <div>
-                <strong data-ng-bind-template="{{phone.type.description}} Phone"></strong>
-                <span data-ng-if="phone.primary">(preferred)</span>
+              <br/>
+              <div data-ng-repeat="email in student.emails">
+                <div>
+                  <strong data-ng-bind-template="{{email.type.description}} Email"></strong>
+                  <span data-ng-if="email.primary">(preferred)</span>
```
paulkerschen: @johncrossman I see this was in the existing code, but just checking: on email and phone, "primary" is synonymous with "preferred"? I believe they mean different things when translating back-end CS feeds.

johncrossman: @paulkerschen  Yes, w.r.t email and phone, 'primary' is synonymous with 'preferred'. It's Hub naming translated for CalCentral user experience. The confusion continues with the Hub's "names" schema in which a name item has both:
* `type.code = [PRI | PRF]`
* `preferred: [true | false]`

paulkerschen: Thanks, @johncrossman, it's that name schema I was thinking of.

PR #5191, opened 2016-04-07
SISRP-17510 Force canonical course-code sorting on EdoOracle::UserCourses

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-17510

EdoOracle doesn't guarantee the database-level sorting of course codes that we got from CampusOracle, so we must run feeds through our own Ruby sorter at a higher level.

### DIFF
--- a/app/models/berkeley/course_codes.rb
+++ b/app/models/berkeley/course_codes.rb
@@ -3,8 +3,12 @@ module Berkeley
     extend self
 
     def comparable_course_code(course)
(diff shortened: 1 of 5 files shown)


johncrossman:
Looks good to me, @paulkerschen  but I recommend a second pair of eyes. Ping @raydavis or @redconfetti

raydavis:
Looks right to me, too.

PR #5193, opened 2016-04-07
SISRP-17255 - Retrofit CanvasLti::Egrades

redconfetti:
https://jira.berkeley.edu/browse/SISRP-17255

### DIFF
--- a/spec/models/canvas_lti/egrades_spec.rb
+++ b/spec/models/canvas_lti/egrades_spec.rb
@@ -180,14 +180,14 @@ describe CanvasLti::Egrades do
   context 'when providing canvas course student grades' do
     let(:course_users_array) do
       [
-        ['4000123', 58, 'F'],
-        ['4000309', 93.75, 'A-'],
-        ['4000199', 81.25, 'B-']
+        ['4000123', 58, 58, 'F', 'F'],
+        ['4000309', 93.75, 93.75, 'A-', 'A-'],
+        ['4000199', 81.25, 87.5, 'B-', 'B+']
       ].collect do |u|
         {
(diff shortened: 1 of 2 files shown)


spec/models/canvas_lti/egrades_spec.rb
```diff
@@ -178,41 +178,42 @@
...
+      expect(result[1][:current_grade]).to eq 'A-'
 
-      # Student with Grade
-      expect(result[5][:sis_login_id]).to eq '4000272'
-      expect(result[5][:final_grade]).to eq 'F-'
-      expect(result[5][:current_grade]).to eq 'F-'
+      expect(result[2][:sis_login_id]).to eq '4000199'
+      expect(result[2][:final_grade]).to eq 'B-'
+      expect(result[2][:current_grade]).to eq 'B-'
     end
```
paulkerschen: @redconfetti I don't know this tool well, but is it still possible to have a case where `final_grade` differs from `current_grade`? If so, would it be worth throwing in one such example?

paulkerschen:
Looks good; left one question about test cases.

spec/models/canvas_lti/egrades_spec.rb
redconfetti: @paulkerschen Sure. I'll expand it to have that now.

redconfetti: @paulkerschen There actually is a scenario where the current grade differs from the final grade. One of them (current or final) does not factor in ungraded assignments into the calculation. I forget which one.

paulkerschen:
Thanks @redconfetti!

PR #5199, opened 2016-04-08
SISRP-17445, restrictions in template of /advising/student

johncrossman:
https://jira.berkeley.edu/browse/SISRP-17445

No compelling reason for users with `isViewer` and `isSuperuser` to access `/advising/student`.


johncrossman:
@raydavis reminded me that traditional view-as access might be important during production bug diagnosis, etc. Closing the PR.

PR #5200, opened 2016-04-08
CLC-6261, var referenced in error msg does not exist (OEC/Google related)

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-6261

Error message should report on `item.id`.


paulkerschen:
Thanks @johncrossman. I wish we could train Github in our convention that half a Travis is good enough on our bleeding edge of academic innovation.

PR #5202, opened 2016-04-08
SISRP-16949 Current-student affiliation now trumps expired-student

raydavis:
... no matter what the expiry date.

https://jira.berkeley.edu/browse/SISRP-16949


johncrossman:
One successful Travis is all we need. The change looks right. Thx, @raydavis

PR #5206, opened 2016-04-08
SISRP-17620 Replace "0.0" units with "--" in Class Enrollment

raydavis:
https://jira.berkeley.edu/browse/SISRP-17620

Blind PR, since a sticky SSLError keeps me from fully testing locally.


paulkerschen:
Thanks @raydavis. There's only one way forward.

PR #5207, opened 2016-04-11
SISRP-17622 De-duplicate repeated secondary sections from EDODb queries

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-17622

PR #5210, opened 2016-04-11
[QA] CLC-6260 SISRP-17617 SISRP-17622 EDODb bugfixes

paulkerschen:
QA version of #5207, #5209, #5211.

### DIFF
--- a/app/models/edo_oracle/queries.rb
+++ b/app/models/edo_oracle/queries.rb
@@ -15,7 +15,7 @@ module EdoOracle
       sec."term-id" AS term_id,
       TRIM(crs."title") AS course_title,
       TRIM(crs."transcriptTitle") AS course_title_short,
-      crs."academicDepartment-code" AS dept_name,
+      crs."subjectArea" AS dept_name,
       sec."primary" AS primary,
       sec."sectionNumber" AS section_num,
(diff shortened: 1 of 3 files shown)


johncrossman:
@redconfetti please merge this when ready. I'd like a second pair of eyes on edo db related changes.  Thx

app/models/edo_oracle/queries.rb
```diff
@@ -159,8 +159,6 @@ def self.get_section_meetings(term_id, section_id)
           sec."term-id" = '#{term_id}' AND
           sec."id" = '#{section_id}' AND
           mtg."location-code" IS NOT NULL
-        ORDER BY
```
redconfetti: In the [CampusOracle equivalent](https://github.com/ets-berkeley-edu/calcentral/blob/c74eb7f50/app/models/campus_oracle/queries.rb#L326) of this query we ordered on `print_cd`.

We also used these codes to determine which ones got displayed. I guess this isn't necessary now?

paulkerschen: No, I haven't seen an equivalent. This ORDER BY had to come out because of the SELECT DISTINCT; it causes an error because mtg."sectionNumber" isn't actually one of the selected columns.

PR #5201, opened 2016-04-08
CLC-6260 Use correct department code column in EdoOracle::Queries

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6260

PR #5211, opened 2016-04-11
CLC-6260 Use actual real correct EDODb column for department code

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6260

#5201 was an honest try but wrong. The real equivalent to our legacy `dept_name` is not `academicDepartment-code` but `subjectArea` (e.g. "CATALAN" not "SPANPOR").


johncrossman:
This PR is truthier than 5201.  +1 @paulkerschen

PR #5213, opened 2016-04-12
SISRP-17735, delegated access: new and improved links for parents

johncrossman:
https://jira.berkeley.edu/browse/SISRP-17735


paulkerschen:
Thanks @johncrossman; I'm relieved that "Jobs & Careers" is now a secure link.

PR #5217, opened 2016-04-12
SISRP-17647 User tracker for Canvas maintenance should be hash, not array

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-17647

### DIFF
--- a/app/models/canvas_csv/site_memberships_maintainer.rb
+++ b/app/models/canvas_csv/site_memberships_maintainer.rb
@@ -27,7 +27,7 @@ module CanvasCsv
     def self.import_memberships(sis_course_id, sis_section_ids, enrollments_csv_filename, into_canvas_course_id = nil)
       enrollments_rows = []
       users_rows = []
-      known_users = []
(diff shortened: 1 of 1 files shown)


app/models/canvas_csv/site_memberships_maintainer.rb
```diff
@@ -27,7 +27,7 @@ def self.process(sis_course_id, sis_section_ids, enrollments_csv_output, users_c
     def self.import_memberships(sis_course_id, sis_section_ids, enrollments_csv_filename, into_canvas_course_id = nil)
       enrollments_rows = []
       users_rows = []
-      known_users = []
+      known_users = {}
```
redconfetti: Wow @paulkerschen, I wonder how we missed that.

paulkerschen: @redconfetti: As I noted in the JIRA, it's exactly the sort of typo that no one would catch until an actual test run was possible. :)

PR #5220, opened 2016-04-13
SISRP-17616 Use translation table when joining class and section by displayName

paulkerschen:
Under the wide umbrella of https://jira.berkeley.edu/browse/SISRP-17616, change course-section join to run through a display-name translation table.

Also remove unnecessary LEFT OUTER qualifiers from instructor, enrollment and meeting joins.


johncrossman:
Looks right to me, @paulkerschen  I think @redconfetti should review and merge

PR #5222, opened 2016-04-13
SISRP-17647 - Add fault tolerance for enrollment processing when UID not present

redconfetti:
https://jira.berkeley.edu/browse/SISRP-17647

Adds fault tolerance for enrollment processing when UID not present in campus data


paulkerschen:
Thank you @redconfetti! Waiting for Travis....

PR #5226, opened 2016-04-13
SISRP-17462, LDAP search by name with option to filter by role

johncrossman:
https://jira.berkeley.edu/browse/SISRP-17462 and its parent https://jira.berkeley.edu/browse/SISRP-16693

The following searches are equivalent: "John Doe", "Doe, John", "Mr. John Doe", "John Doe" and "Doe Jr., John". This code tokenizes the incoming string and tosses out "Jr.", etc. In this example, we'd perform two LDAP searches on `displayname`:  'John* Doe*' and 'Doe* John*'.  In my own testing, this is what gets near feature parity with _berkeley.edu/directory_.  Note: these searches are not cached – whether they should be is debatable.

### DIFF
--- a/app/models/calnet_ldap/client.rb
+++ b/app/models/calnet_ldap/client.rb
@@ -30,6 +30,18 @@ module CalnetLdap
       search(base: GUEST_DN, filter: modified_timestamp_filter)
     end
 
+    def search_by_name(name, include_guest_users=false)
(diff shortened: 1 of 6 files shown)


paulkerschen:
Very nice expansion of existing CalnetLdap code, @johncrossman - thanks.

PR #5227, opened 2016-04-13
SISRP-5136 - UX Dev: Repackage Compare Card

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-5136

### DIFF
--- a/src/assets/templates/widgets/finaid_awards_compare_current_prior.html
+++ b/src/assets/templates/widgets/finaid_awards_compare_current_prior.html
@@ -10,25 +10,8 @@
 </div>
 
 <div class="cc-widget-text" data-cc-spinner-directive="finaidAwardsCompare.isLoadingCurrentAndPrior">
-  <div class="cc-flex cc-flex-space-between-horizontal-5 cc-flex-align-base">
-    <div>
-      <h3>Summary Information</h3>
-    </div>
-    <div>
-      <button
-        class="cc-button-link"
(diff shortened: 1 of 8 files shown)


src/assets/templates/widgets/finaid_awards_compare_current_prior.html
```diff
@@ -10,25 +10,8 @@
...
-  <div data-ng-if="finaidAwardsCompare.toggle.summary.show" class="cc-flex cc-finaid-compare-awards-current-prior-columns">
-    <div
-      data-ng-repeat="type in finaidAwardsCompare.types"
-      data-ng-include="'widgets/finaid_awards_compare_summary.html'"
-    ></div>
-  </div>
+  <div
+    data-ng-repeat="section in finaidAwardsCompare.sections"
+    data-ng-include="'widgets/finaid_awards_compare_section.html'"
+  ></div>
```
redconfetti: I don't like this floating greater than sign, but I guess it's necessary to keep the number of columns per each line down.

christianvuerings: @redconfetti indeed, and it also makes diffs a bit cleaner.

If we do the following, and add a new line to the end, it will create highlight 2 lines instead of 1.
```
<div
    data-ng-repeat="section in finaidAwardsCompare.sections"
    data-ng-include="'widgets/finaid_awards_compare_section.html'"></div>
```

PR #5228, opened 2016-04-13
SISRP-17531 Do not allow inactive EDODb courses to sneak through as null rows

paulkerschen:
In investigating https://jira.berkeley.edu/browse/SISRP-17531, I regrettably learned that a partial revert of #5224 is necessary. 

The aim was to allow withdrawn course descriptions to filter through for past data; but I opened the door too wide and started allowing a mix of active and withdrawn descriptions, with the withdrawn rows partially nulled out.

Also switch our translation table to the materialized MVW version.


johncrossman:
+1  @paulkerschen

PR #5235, opened 2016-04-14
[QA] SISRP-16006: hide paragraphs with ferpa and calnet directory links

dfkaye:
QA PR for https://jira.berkeley.edu/browse/SISRP-16006

### DIFF
--- a/src/assets/templates/widgets/profile/information_disclosure.html
+++ b/src/assets/templates/widgets/profile/information_disclosure.html
@@ -2,7 +2,7 @@
   <p>
     The university's directory information includes personal information such as your address, email, name, phone, and portions of your academic record.
   </p>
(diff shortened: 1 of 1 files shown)


src/assets/templates/widgets/profile/information_disclosure.html
```diff
@@ -17,7 +17,7 @@
   <p>
     To learn more, view <a href="http://registrar.berkeley.edu/GeneralInfo/ferpa.html">UC Berkeley's policy governing disclosure of information from student records</a>.
   </p>
-  <p>
+  <p data-ng-if="api.user.profile.actAsOptions.canSeeCSLinks">
```
christianvuerings: @dfkaye fyi - we probably don't need this here since it looks like this is only a public URL, not a CS URL

PR #5237, opened 2016-04-14
SISRP-17462, Bamboo failure on static method calling module mixin

johncrossman:
https://jira.berkeley.edu/browse/SISRP-16693
Fix Bamboo: https://bamboo.ets.berkeley.edu/bamboo/browse/MYB-CLCMVPMASTER-4071

### DIFF
--- a/app/models/calnet_ldap/user_attributes.rb
+++ b/app/models/calnet_ldap/user_attributes.rb
@@ -1,7 +1,7 @@
 module CalnetLdap
   class UserAttributes < BaseProxy
+    extend Parser
 
-    include Parser
     include Cache::UserCacheExpiry
 
     def initialize(options = {})
@@ -16,7 +16,7 @@ module CalnetLdap
 
     def get_feed_internal
       if (result = CalnetLdap::Client.new.search_by_uid @uid)
-        parse result
+        self.class.parse result
       else
         {}
       end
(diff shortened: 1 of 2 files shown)


spec/models/calnet_ldap/client_spec.rb
```diff
@@ -126,11 +126,6 @@
       expect(results_with_comma).to_not be_empty
       expect(subject.search_by_name 'Jo  MAN M.A.').to have(results_with_comma.length).items
     end
-
-    it 'should limit number of search results' do
-      results = subject.search_by_name 'Joe', limit: 3
```
johncrossman: Note: the 'limit' arg is no longer on `CalnetLdap::Client` so spec is obsolete. It was moved up to `User::SearchUsersByName`

johncrossman:
Bamboo build running on branch of this PR: 
https://bamboo.ets.berkeley.edu/bamboo/browse/CPB-CJT79-1

app/models/calnet_ldap/user_attributes.rb
```diff
@@ -1,5 +1,6 @@
 module CalnetLdap
   class UserAttributes < BaseProxy
+    extend Parser
 
     include Parser
```
paulkerschen: @johncrossman It seems a little odd to have to both extend and include the module... Since the module has no actual instance-level logic, I wonder if it would be cleaner to use only `extend` here and in User::SearchUsersByName, and call `self.class.parse` when inside an instance method?

johncrossman: @paulkerschen Yes, odd to me, too. Amended per your good advice.

paulkerschen:
Thanks @johncrossman!

PR #5247, opened 2016-04-18
SISRP-17984, update verify_sis_api_endpoints script post GL 5.2

johncrossman:
https://jira.berkeley.edu/browse/SISRP-17984

Latest endpoints, cleanup.


paulkerschen:
I guess those transitive dependencies were also transient. Thanks @johncrossman.

PR #5249, opened 2016-04-18
SISRP-17963, advisor auth check in view-as must know diff between UID and SID

johncrossman:
https://jira.berkeley.edu/browse/SISRP-17963

Notes:
* Because the incoming `id` might be UID or SID, we must move the advisor-authorization check *after* the SearchUsers call. At that point, we use `user['ldap_uid']` and remove all doubt concerning id type.
* I had to improve the error message because a search on SID would return an error message containing only the corresponding UID. That would certainly confuse an Advisor. We now have both in the message.


raydavis:
Thank you, @johncrossman

PR #5251, opened 2016-04-18
[QA] SISRP-17963, advisor auth check in view-as must know diff between UID and SID

johncrossman:
https://jira.berkeley.edu/browse/SISRP-17963

See notes in corresponding master PR #5249

### DIFF
--- a/app/controllers/advising_student_controller.rb
+++ b/app/controllers/advising_student_controller.rb
@@ -12,7 +12,7 @@ class AdvisingStudentController < ApplicationController
   def profile
     student_uid = student_uid_param
     render json: {
-      attributes: @attributes,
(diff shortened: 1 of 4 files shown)


app/controllers/advising_student_controller.rb
```diff
@@ -12,7 +12,7 @@ class AdvisingStudentController < ApplicationController
   def profile
     student_uid = student_uid_param
     render json: {
-      attributes: @attributes,
+      attributes: User::AggregatedAttributes.new(student_uid).get_feed,
```
johncrossman: This one line differs from master branch PR. The setting of `@attributes` was dropped from advisor_authorization.

PR #5252, opened 2016-04-18
SISRP-17759 - 5.3 - My Academics (Student) - Status, Blocks and Holds

dfkaye:
https://jira.berkeley.edu/browse/SISRP-17759

This fixes internal ruby class names, from SLRDeeplink to SlrDeeplink

### DIFF
--- a/app/controllers/campus_solutions/slr_deeplink_controller.rb
+++ b/app/controllers/campus_solutions/slr_deeplink_controller.rb
@@ -1,8 +1,11 @@
 module CampusSolutions
-  class SLRDeeplinkController < CampusSolutionsController
+  class SlrDeeplinkController < CampusSolutionsController
+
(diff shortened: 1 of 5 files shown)


app/controllers/campus_solutions/slr_deeplink_controller.rb
```diff
@@ -1,8 +1,8 @@
 module CampusSolutions
-  class SLRDeeplinkController < CampusSolutionsController
+  class SlrDeeplinkController < CampusSolutionsController
```
johncrossman: @dfkaye I think we should add the following to this controller:
```
    include DisallowAdvisorViewAs
    include DisallowClassicViewAs
```
As seen in HigherOneUrlController.

johncrossman:
Thanks @dfkaye

PR #5253, opened 2016-04-19
SISRP-17974 - Update front-end dependencies

christianvuerings:
https://jira.berkeley.edu/browse/SISRP-17974


christianvuerings:
FYI Travis builds are failing since we need a newer version of node (`>4.0`)
e.g. https://github.com/klei/gulp-inject/blob/667ff6348d60c2a2a816fb8c0d5011a1dc9601cb/src/inject/index.js#L127

.jscsrc
```diff
@@ -1,14 +1,17 @@
 {
   "disallowDanglingUnderscores": true,
   "disallowEmptyBlocks": true,
+  "disallowFunctionDeclarations": true,
```
redconfetti: @christianvuerings - This is because the best practice is to assign function to variables, preferably within a scope such as an anonymous function call, rather than global?

script/front-end-test.sh
```diff
@@ -1,3 +1,5 @@
+. $HOME/.nvm/nvm.sh
+nvm install node
```
redconfetti: 👍

src/assets/javascripts/angular/controllers/widgets/profile/languagesSectionController.js
```diff
@@ -59,7 +59,7 @@ angular.module('calcentral.controllers').controller('LanguagesSectionController'
   /**
    * Function maps a language level (native, teach, translate) boolean value to
    * either 'Y' or 'N', the only values accepted by the POST endpoint.
```
redconfetti: @christianvuerings - I've started to notice that we're using this documentation for methods. Do we use any specific for generating HTML formatted documentation based on this?

.jscsrc
christianvuerings: @redconfetti indeed, most of these rules are to make sure we keep being consistent in our code.

http://jscs.info/rule/disallowFunctionDeclarations

src/assets/javascripts/angular/controllers/widgets/profile/languagesSectionController.js
christianvuerings: @redconfetti we could, currently we're trying to write everything in JSDoc style - http://usejsdoc.org/

It shouldn't be too hard to build these into html formatted docs - would definitely recommend checking out that site.

PR #5255, opened 2016-04-19
SISRP-17803 Filter out blank-ID instructor assignments from EDODb queries

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-17803

Thanks to CalCentral Bugfinder Squad, we now know that a database row with instructor name and blank instructor ID should not be interpreted as "valid instructor assignment with missing ID," but as "missing instructor assignment with erroneous name."

### DIFF
--- a/app/models/edo_oracle/queries.rb
+++ b/app/models/edo_oracle/queries.rb
@@ -228,7 +228,8 @@ module EdoOracle
           )
           WHERE
             sec."id" = '#{section_id.to_s}' AND
-            sec."term-id" = '#{term_id.to_s}'
+            sec."term-id" = '#{term_id.to_s}' AND
+            TRIM(instr."instructor-id") IS NOT NULL
           ORDER BY
             role_code
         SQL


app/models/edo_oracle/queries.rb
```diff
@@ -228,7 +228,8 @@ def self.get_section_instructors(term_id, section_id)
           )
           WHERE
             sec."id" = '#{section_id.to_s}' AND
-            sec."term-id" = '#{term_id.to_s}'
+            sec."term-id" = '#{term_id.to_s}' AND
+            TRIM(instr."instructor-id") IS NOT NULL
```
johncrossman: @paulkerschen is performance a concern in having TRIM in the where clause?  I came across https://perfchron.com/2015/01/09/red-flags-how-a-data-quality-issue-becomes-a-performance-issue/
I assume Oracle will optimize the query to avoid a table scan. Agree? An alternative would be to post-process the results in Ruby.

paulkerschen: @johncrossman My belief and hope was that it will filter on `section_id` and `term_id` first, and will only have a handful of rows to check for blank `instructor-id`.

Testing a sample query in BCSDEV (run a few times because it seems to get faster with repeats):
With TRIM: 144 ms, 29 ms, 23 ms
Without TRIM: 235 ms, 36 ms, 40 ms, 24 ms.

I assume it's because of earlier caching or random noise that the TRIM version actually ran faster on first try.

johncrossman:
I agree with your beliefs and hopes. Thx @paulkerschen

PR #5257, opened 2016-04-19
SISRP-16662 - Add Transition Links to Billing Summary and Billing Details Pages

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-16662

Adding links to both the CARS and CS finances details pages, to each other, for the transition.
Also added `currentTerm` subheadings for certain places designated within the designs.

### DIFF
--- a/src/assets/javascripts/angular/controllers/widgets/billingController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/billingController.js
@@ -66,10 +66,7 @@ angular.module('calcentral.controllers').controller('BillingController', functio
   };
 
   var hasCarsActivity = function(data) {
-    var carsActivity = _.get(data, 'data.activity');
-    if (carsActivity && carsActivity.length) {
-      $scope.billing.hasCarsActivity = true;
-    }
+    $scope.billing.hasCarsActivity = !!_.get(data, 'data.activity.length');
   };
 
   var loadCarsInfo = function() {
--- a/src/assets/javascripts/angular/controllers/widgets/billingController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/billingController.js
@@ -8,7 +8,9 @@ angular.module('calcentral.controllers').controller('BillingController', functio
   $scope.activityLimit = 100;
   $scope.billing = {
     data: {},
+    hasCarsActivity: false,
     isLoading: true,
+    isLoadingCars: true,
     sort: {
       column: 'itemEffectiveDate',
       descending: true
@@ -61,6 +63,21 @@ angular.module('calcentral.controllers').controller('BillingController', functio
         itemTermId: billingItem.itemTermId
       });
     }
+  };
+
+  var hasCarsActivity = function(data) {
+    var carsActivity = _.get(data, 'data.activity');
+    if (carsActivity && carsActivity.length) {
+      $scope.billing.hasCarsActivity = true;
+    }
+  };
+
+  var loadCarsInfo = function() {
+    financesFactory.getFinances()
+      .then(hasCarsActivity)
+      .then(function() {
+        $scope.billing.isLoadingCars = false;
+      });
   };
 
   /**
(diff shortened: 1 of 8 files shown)


redconfetti:
@davidlimcheng Looks good to me. @christianvuerings, am I missing anything?

src/assets/javascripts/angular/controllers/widgets/billingController.js
```diff
@@ -63,6 +65,21 @@ angular.module('calcentral.controllers').controller('BillingController', functio
     }
   };
 
+  var hasCarsActivity = function(data) {
+    var carsActivity = _.get(data, 'data.activity');
```
christianvuerings: @davidlimcheng shorter way to write this function:

```
var carsActivity = !!_.get(data, 'data.activity.length');
```
(test whether it works though)

src/assets/javascripts/angular/controllers/widgets/carsController.js
```diff
@@ -267,6 +274,16 @@ angular.module('calcentral.controllers').controller('CarsController', function(a
     });
   };
 
+  var loadCsInfo = function() {
+    financesFactory.getCsFinances().success(function(data) {
+      if (data.feed.activity && data.feed.activity.length) {
+        $scope.csActivity.hasCsActivity = true;
```
christianvuerings: @davidlimcheng see above

src/assets/stylesheets/_myfinances.scss
```diff
@@ -44,6 +44,10 @@
   .cc-page-myfinances-amount {
     text-align: right;
   }
+  .cc-page-myfinances-billing-summary-footnote {
+    font-size: 11px;
+    margin: 8px 0 12px;
```
christianvuerings: @davidlimcheng seem like magic numbers - does `10px 0` also look good to you?

src/assets/templates/billing_details.html
```diff
@@ -1,9 +1,9 @@
-<div class="cc-page-myfinances" data-ng-if="api.user.profile.features.csBilling">
+<div class="cc-page-myfinances" data-ng-controller="BillingController" data-ng-if="api.user.profile.features.csBilling">
```
christianvuerings: @davidlimcheng could we check whether this invokes the `BillingController` twice?

src/assets/javascripts/angular/controllers/widgets/carsController.js
```diff
@@ -250,6 +256,7 @@ angular.module('calcentral.controllers').controller('CarsController', function(a
     // Data contains all the financial information for the current student
     financesFactory.getFinances().success(function(data) {
       angular.extend($scope, data);
+      loadCsInfo();
```
christianvuerings: @davidlimcheng we should only load this when the feature flag is enabled.

src/assets/stylesheets/_myfinances.scss
redconfetti: @christianvuerings - What makes 10px any less magic than 8px? Are we trying to conform to common units, like increments of 5px (5, 10, 15, 20, etc), instead of having the various measurements in between that might work for one browser, but not look exactly the same for another?

src/assets/templates/billing_details.html
davidlimcheng: @christianvuerings This does invoke the `BillingController` twice.  I'll push a new commit with a separate "BillingDetailsController" that will be used to grab the `currentTerm` from CS.  This will also replace `MyFinancesController` for the page and set the title as well.  

Thanks!

src/assets/stylesheets/_myfinances.scss
davidlimcheng: @christianvuerings  `10px 0` looks good to me.  But, same question as Jason - are incremental numbers preferable?

davidlimcheng:
@christianvuerings @redconfetti Pushed a new commit with the proposed changes, as well as moving the `data-cc-spinner-directive` out of the same `div` as the `data-ng-if` since that was causing the spinner to hide.

src/assets/stylesheets/_myfinances.scss
christianvuerings: @davidlimcheng @redconfetti indeed - we've been trying to use increments by `5px` for most things. Also https://css-tricks.com/magic-numbers-in-css/

PR #5259, opened 2016-04-19
SISRP-17934 Filter bCourses items from advisor feeds

raydavis:
https://jira.berkeley.edu/browse/SISRP-17934

(I was unable to test until updating the Signet Gem. There have been some solved issues with JRuby on Mac OS X lately.)


paulkerschen:
Webcast test failures seeming to involve real proxies: another case of testext masquerading as test?

```
Failures:

  1) Webcast::Rooms a real, non-fake proxy on remote server errors should return the fetch error message
     Failure/Error: expect(buildings[:proxyErrorMessage]).to include 'There was a problem'
       expected nil to include "There was a problem", but it does not respond to `include?`
     # ./spec/models/webcast/rooms_spec.rb:69:in `(root)'

  2) Webcast::Rooms a real, non-fake proxy when json formatting fails should return the fetch error message
     Failure/Error: expect(buildings[:proxyErrorMessage]).to include 'There was a problem'
       expected nil to include "There was a problem", but it does not respond to `include?`
     # ./spec/models/webcast/rooms_spec.rb:80:in `(root)'
```

spec/controllers/my_academics_controller_spec.rb
```diff
@@ -23,20 +23,28 @@
...
+        expect(subject['semesters']).to have(24).items
+        expect(subject['semesters'][0]['slug']).to be_present
+        expect(subject['semesters'][1]['classes'][0]['transcript'][0]['grade']).to be_present
+        expect(subject['transitionTerm']).to be_present
+      end
+    end
+    context 'advisor view-as' do
+      include_context 'advisor view-as'
+      it 'filters bCourses sites' do
+        expect(subject['otherSiteMemberships']).to be_blank
```
paulkerschen: @raydavis Worth testing for one or two items that we expect to get through the filter, to catch empty responses?

raydavis:
It looks like some side-effect of the Signet update, @paulkerschen -- I'll have to check into it later.

config/settings/test.yml
```diff
@@ -27,6 +27,9 @@ financials_proxy:
 textbooks_proxy:
   fake: true
 
+webcast_proxy:
+  fake: true
```
paulkerschen: @raydavis Ah, this is probably what's making Travis unhappy; perhaps those failing tests should be marked testext?

redconfetti:
Thanks for providing that Signet fix @raydavis

config/settings/test.yml
raydavis: Of course. Safest to just take that back out. I'll amend the PR ASAP.

raydavis:
PR amended.

paulkerschen:
Thanks @raydavis; it satisfies.

PR #5261, opened 2016-04-20
SISRP-16088, user search widget used by Advisors on Dashboard

johncrossman:

User search widget to be used by Advisors. The Saved and Recent sections will be constructed in future PR.

### DIFF
--- a/src/assets/templates/widgets/toolbox/user_search.html
+++ b/src/assets/templates/widgets/toolbox/user_search.html
@@ -1,0 +1,68 @@
+<div class="cc-widget cc-user-search">
+  <div class="cc-widget-title">
+    <h2 data-ng-bind="search.title"></h2>
+  </div>
+  <div class="cc-widget-text" data-cc-spinner-directive="search.isLoading">
+    <div data-ng-switch data-on="search.currentTab">
+      <div class="row collapse">
+        <div class="medium-10 medium-centered columns">
+          <ul class="cc-button-group cc-even-3" role="tablist">
+            <li data-ng-repeat="tab in search.tabs">
+              <button class="cc-button"
+                      data-ng-click="search.goToTab(tab)"
+                      data-ng-class="{'cc-button-selected':(search.currentTab == tab)}"
+                      aria-selected="{{search.currentTab == tab}}"
+                      role="tab"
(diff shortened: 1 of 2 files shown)


src/assets/templates/widgets/toolbox/user_search.html
```diff
@@ -0,0 +1,68 @@
+<div class="cc-widget cc-user-search">
```
christianvuerings: @johncrossman might be a good idea to combine this widget with the one we currently have on the admin page. But I'll leave that up to you and the designers.

johncrossman: @christianvuerings IMO, we should deprecate the existing view-as widget. Once this widget is well tested and stable in production we can swap it on the Toolbox page, replacing what we have now.  cc: @raydavis

src/assets/templates/widgets/toolbox/user_search.html
```diff
@@ -0,0 +1,68 @@
...
+          <div data-cc-spinner-directive="search.queryInProgress">
+            <ul class="cc-widget-list">
+              <li data-ng-repeat="user in search.users" data-ng-class="{'cc-widget-list-selected': user.show}">
+                <span class="cc-visuallyhidden" data-ng-bind-template="Show {{user.show ? 'less' : 'more'}} information about {{user.person_name}}"></span>
+                <div
+                  class="cc-widget-list-hover"
+                  data-ng-class="{'cc-widget-list-hover-opened':(user.show)}"
+                  data-cc-accessible-focus-directive
+                  data-ng-click="api.widget.toggleShow($event, search.users, user, 'Up Next')">
+                  <div class="cc-flex cc-flex-space-between">
```
christianvuerings: @johncrossman just as a FYI - this might not always work on inline elements (like `span` - `strong`) - http://stackoverflow.com/questions/22008135/internet-explorer-10-does-not-apply-flexbox-on-inline-elements

johncrossman: @christianvuerings I converted to div. I'll drop in necessary styling soon.

christianvuerings:
Lgtm

PR #5264, opened 2016-04-20
[QA] SISRP-17803 Filter out blank-ID instructor assignments from EDODb queries

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-17803

QA version of #5255.


johncrossman:
As before, looks good. Thx @paulkerschen

PR #5265, opened 2016-04-20
NOJIRA, base_proxy_spec no longer relies on real HTTP call

johncrossman:
The Bamboo QA build just failed on `base_proxy_spec` because http://www.url.com was down.  The spec has nothing to do with actual HTTP call so we stub it out.


johncrossman:
We have one happy Travis.

paulkerschen:
Thank you @johncrossman!

PR #5272, opened 2016-04-21
SISRP-18145 Fix more misaligned cache configurations

raydavis:
https://jira.berkeley.edu/browse/SISRP-18145

Mostly feeds marked for server-side LiveUpdates processing but ignored by client-side logic.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -227,7 +227,6 @@ cache:
     MyBadges::Merged: NEXT_00_00
     MyCampusLinksController: NEXT_08_00
     MyClasses::Merged: NEXT_08_00
-    Finaid::MyFinAid: NEXT_08_00
     MyGroups::Merged: NEXT_08_00
     MyTasks::Merged: <%= 10.minutes %>
     UpNext::MyUpNext: NEXT_00_00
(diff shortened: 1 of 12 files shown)


config/settings.yml
```diff
@@ -227,7 +227,6 @@ cache:
     MyBadges::Merged: NEXT_00_00
     MyCampusLinksController: NEXT_08_00
     MyClasses::Merged: NEXT_08_00
-    Finaid::MyFinAid: NEXT_08_00
```
raydavis: This feed is not currently being cached.

paulkerschen: @raydavis Was this perhaps a mistake for Finaid::Merged, which wraps Finaid::MyFinAid and nothing else, and is cached (presumably at the default 35 minutes)? Or do we want Finaid::Merged to have a shorter cache time since it doesn't participate in Live Updates?

raydavis: @paulkerschen :

"Was this perhaps a mistake for Finaid::Merged...?" That was my guess as well.

"do we want Finaid::Merged to have a shorter cache time...?" Since users have been living with 35 minute lifespans in production for a while now, I figured we might as well maintain the status quo until we heard differently.

paulkerschen:
Thanks for minding the shop @raydavis.

PR #5276, opened 2016-04-22
SISRP-16088, saved and recent tab on widget; rename files 'student' to generic 'user'

johncrossman:

Highlights:
* Rename `/advising/student/:uid` to more generic `/user/overview/:uid`.
* For advisors, dashboard gets search-users widget.
* Reusable `user_search_results.html` for search results, stored and recent lists.

### DIFF
--- a/src/assets/javascripts/angular/controllers/pages/userOverviewController.js
+++ b/src/assets/javascripts/angular/controllers/pages/userOverviewController.js
@@ -1,17 +1,18 @@
+/* jshint camelcase: false */
 'use strict';
 
 var angular = require('angular');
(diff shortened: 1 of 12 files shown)


src/assets/javascripts/angular/controllers/pages/userOverviewController.js
```diff
@@ -1,17 +1,18 @@
+/* jshint camelcase: false */
```
johncrossman: See below for `targetUser.ldap_uid`. A necessary evil because `targetUser` object eventually makes its way to `adminController` which has significant logic expecting `user.ldap_uid`.

PR #5277, opened 2016-04-22
NOJIRA - Renamed controller using proper Rails file naming convention

redconfetti:
Renamed controller using proper Rails file naming convention

Was `financial_aid_funding_sources_controller_term.rb`, renamed as `financial_aid_funding_sources_term_controller.rb`


raydavis:
Another outlet for self-expression squashed by conformity.

PR #5278, opened 2016-04-22
SISRP-17947 Postgres notifications query requires string UID

paulkerschen:
Fix for the very occasional https://jira.berkeley.edu/browse/SISRP-17947.


raydavis:
Speaking of nonconformists, "I am not a number! I am a numeric string!"

PR #5280, opened 2016-04-22
SISRP-5504 Plug leak to live Google in mocked Textbooks test

raydavis:
https://jira.berkeley.edu/browse/SISRP-5504

Hoping this will eliminate an old time-waster...


paulkerschen:
I think we've heard quite enough from Travis on this particular test. Don't you agree, Travis?

PR #5281, opened 2016-04-22
SISRP-18304 Preserve course titles when merging EDODb cross-listings

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-18304


raydavis:
I remember when we were looking forward to the perfectly cleaned-up data which would furnish our shiny new home...

PR #5282, opened 2016-04-22
SISRP-16088, all tabs are wired to save, delete and view-as user

johncrossman:

All three tabs (search, saved and recent) will shuffle user list and available links based on latest stored_users in db. To avoid complexity in the js I added a 'saved' property to server-side user searches. E.g., each user in search results has the flag.

A little cleanup:
* symbolize keys in stored_users 
* fewer js files offending the camelcase rule


raydavis:
Thanks, @johncrossman

PR #5283, opened 2016-04-22
SISRP-18293 - My Finances account summary - infinite spinner beneath 'Make Payment'

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-18293

Removing infinite spinner, + wrapped link to new CS page in a feature-flagged `div`.

### DIFF
--- a/src/assets/javascripts/angular/controllers/widgets/carsController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/carsController.js
@@ -258,6 +258,8 @@ angular.module('calcentral.controllers').controller('CarsController', function(a
       angular.extend($scope, data);
       if (userService.profile.features.csBilling) {
         loadCsInfo();
+      } else {
+        $scope.csActivity.isLoadingCs = false;
       }
 
       if (data && data.summary && data.activity) {
@@ -273,6 +275,7 @@ angular.module('calcentral.controllers').controller('CarsController', function(a
       }
     }).error(function(data) {
       angular.extend($scope, data);
+      $scope.csActivity.isLoadingCs = false;
     });
   };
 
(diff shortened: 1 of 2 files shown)


davidlimcheng:
Whoops!  There's an extra space in the HTML.  Amending now.

davidlimcheng:
OK, good to go.

src/assets/javascripts/angular/controllers/widgets/carsController.js
```diff
@@ -258,6 +258,8 @@ angular.module('calcentral.controllers').controller('CarsController', function(a
       angular.extend($scope, data);
       if (userService.profile.features.csBilling) {
         loadCsInfo();
+      } else {
+        $scope.csActivity.isLoadingCs = false;
```
johncrossman: @davidlimcheng Would it make sense for `isLoadingCs = false` to live in a `finally(...)` block? E.g.,
```
}).error(function(data) {
  ...
}).finally(function(data) {
  $scope.csActivity.isLoadingCs = false;
});;
```
This guarantees the spinner will stop, even when errors happen.

davidlimcheng: @johncrossman That makes sense, and I agree that it'll catch errors better than what's currently here.  But, would you recommend also leaving it here, as well?  Otherwise, there are a lot of functions that `getCarsInfo()` would have to run through before that spinner goes away.

johncrossman: @davidlimcheng, that makes sense. In that case, you don't need `finally()`. You can just put it in the `error()' block.

davidlimcheng: Amended!  Thanks @johncrossman.

PR #5292, opened 2016-04-25
SISRP-18341 Streamline Crosswalk cache expiration

raydavis:
https://jira.berkeley.edu/browse/SISRP-18341


johncrossman:
Looks right. Thanks, @raydavis

PR #5295, opened 2016-04-25
SISRP-14416 - 5.4 - Delegated Access - Dashboard (delegate) - Academic Dates and Deadlines

dfkaye:
https://jira.berkeley.edu/browse/SISRP-14416

This fixes assignment made on a string to an item instead.


dfkaye:
FYI, the original assignment was a boolean property made as `item.title.isArray` but should just be `item.isArray` - or better as `item.isList` - because `item.title` is a primitive string whereas `item` is an object

paulkerschen:
Makes sense, @dfkaye.

PR #5296, opened 2016-04-25
SISRP-15878 Transfer credits front end and dummy JSON

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-15878

### DIFF
--- a/src/assets/javascripts/angular/controllers/widgets/transferCreditController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/transferCreditController.js
@@ -1,0 +1,27 @@
+'use strict';
+
+var angular = require('angular');
+var _ = require('lodash');
+
+/**
+ * Transfer Credit Controller
+ */
+angular.module('calcentral.controllers').controller('TransferCreditController', function(apiService, $scope, transferCreditFactory) {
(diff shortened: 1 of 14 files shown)


johncrossman:
Mi-tee-fine. Thanks @paulkerschen

src/assets/javascripts/angular/controllers/widgets/transferCreditController.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+
+var angular = require('angular');
+var _ = require('lodash');
+
+/**
+ * Transfer Credit Controller
+ */
+angular.module('calcentral.controllers').controller('TransferCreditController', function(apiService, $scope, transferCreditFactory) {
```
redconfetti: @paulkerschen I'm surprised the mecha-@christianvuerings didn't catch this unused apiService injection. Do you think we could remove this with a NOJIRA PR? Or would we have to do a silly Jira with QA to remove this?

paulkerschen: True enough, @redconfetti. I think a NOJIRA PR for code cleanup is fine here.

PR #5298, opened 2016-04-26
SISRP-18343, Campus Solutions proxy for GET User Search

johncrossman:
https://jira.berkeley.edu/browse/SISRP-18343

See 'GET User Search' in doc https://docs.google.com/document/d/1fILoxlPJhh22ngJbshCU7AIHUGqco07KbbwpZFuG1q4/edit#

### DIFF
--- a/app/models/campus_solutions/user_search.rb
+++ b/app/models/campus_solutions/user_search.rb
@@ -1,0 +1,60 @@
+module CampusSolutions
+  class UserSearch < Proxy
+    include StudentLookupFeatureFlagged
+
+    attr_accessor :url
+
+    def initialize(options = {})
+      raise ArgumentError, 'Arg :name_1 is required in user search' if options[:name_1].blank?
+      raise ArgumentError, 'Arg :affiliations is required and must be an array' unless options[:affiliations].is_a? Array
+      super options
+      params = {
+        NAME1: options[:name_1],
+        AFFILIATIONS: options[:affiliations]
+      }
+      params[:NAME2] = options[:name_2] unless options[:name_2].blank?
+      @url = "#{@settings.base_url}/UC_CC_USER_LOOKUP.v1/lookup?#{params.to_query}"
+    end
+
+    def build_feed(response)
+      feed = { users: [] }
+      unless (xml = response.parsed_response).blank?
+        users = (root = xml['UC_AA_USER_SEARCH']) && (results = root['SEARCH_RESULTS']) && results['USER_DATA']
+        users.each do |user|
+          feed[:users] << {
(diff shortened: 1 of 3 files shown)


app/models/campus_solutions/user_search.rb
```diff
@@ -0,0 +1,65 @@
...
+    end
+
+    def xml_filename
+      'user_search.xml'
+    end
+
+    def url
+      url = "#{@settings.base_url}/UC_CC_USER_LOOKUP.v1/lookup?NAME1=#{@name_1}"
+      url << "&NAME2=#{@name_2}" unless @name_2.blank?
+      url << "&AFFILIATIONS=#{@affiliations.join(',')}"
```
paulkerschen: @johncrossman FYI, Rails lets you call `to_query` on a Hash, which is handy for building URLs as these.

Example https://github.com/ets-berkeley-edu/calcentral/blob/master/app/models/mailing_lists/site_mailing_list.rb#L110

johncrossman: Thanks @paulkerschen I'll see if my rspec approves of the change.

paulkerschen: @johncrossman `to_query` might also do the work of the `encode_query_arg` method below for free... at least it's supposed to. I would be curious to see if it works!

app/models/campus_solutions/user_search.rb
```diff
@@ -0,0 +1,60 @@
...
+    def initialize(options = {})
+      raise ArgumentError, 'Arg :name_1 is required in user search' if options[:name_1].blank?
+      raise ArgumentError, 'Arg :affiliations is required and must be an array' unless options[:affiliations].is_a? Array
+      super options
+      params = {
+        NAME1: options[:name_1],
+        AFFILIATIONS: options[:affiliations]
+      }
+      params[:NAME2] = options[:name_2] unless options[:name_2].blank?
+      @url = "#{@settings.base_url}/UC_CC_USER_LOOKUP.v1/lookup?#{params.to_query}"
```
johncrossman: @paulkerschen So... Your suggestion makes the code cleaner (thank you!). And `params.to_query` converts array to: `foo[]=value1&foo[]=value2&foo[]=value3` which matches the weak-typed convention described in http://stackoverflow.com/questions/3061273/send-an-array-with-an-http-get  I talked w/ Frankie (CS dev) and we're going to give this API signature a shot. The spec has been modified accordingly.  I'll submit a future PR if CS needs an alternate solution.

paulkerschen:
Great, let's see how CS takes it. I'll merge once Travis is done with its artisan build.

PR #5300, opened 2016-04-26
SISRP-18057 Add HubEdos::Registrations proxy

raydavis:
https://jira.berkeley.edu/browse/SISRP-18057

This will be needed to check CS for academic status in non-"current" terms, thus allowing us to (for example) know whether someone should be treated as a Law student in the Class Enrollment card for a future term.

(The PR also scrapes GoLive 4 API formats into the dustbin of history.)


paulkerschen:
Thanks for this very necessary puzzle piece, @raydavis.

PR #5301, opened 2016-04-26
SISRP-17895 Determine Law status in Class Enrollment from EnrollmentTerms

raydavis:

No need to check a feed other than the CS EnrollmentTerms for the academic career associated with the term.


paulkerschen:
This is a satisfying bit of cleanup, @raydavis. We can probably also lose the injected `badgesFactory` at the top of enrollmentCardController.js?

raydavis:
Good eye, @paulkerschen !

PR #5303, opened 2016-04-27
SISRP-17831 - Add SSO to HigherOne for Students and Delegates for Billing Summary

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-17831

I left the "old" methods of polling for the HigherOne URL we use on the SIR card there.  I believe the plan was to QA both methods, and if this way was preferable, then we'd replace the functionality on the SIR card with this method?  cc @pfarestveit 

Let me know if I can make this better - it feels pretty un-DRY, but since these methods are so authentication-state-specific, I wasn't sure if I could make some of these shared variables between the two methods more global.

I also wasn't able to test the delegate part of this yet, I've created [SISRP-18446] (https://jira.berkeley.edu/browse/SISRP-18446) to address this.

Thanks!

### DIFF
--- a/app/controllers/campus_solutions/higher_one_url_controller.rb
+++ b/app/controllers/campus_solutions/higher_one_url_controller.rb
@@ -6,10 +6,25 @@ module CampusSolutions
     before_filter :authorize_for_financial
 
     def get
+      model = model_from_session
+      render json: model.get_feed_as_json
+    end
+
+    def redirect
+      model = model_from_session
+      if (higher_one_url = model.get_higher_one_url)
+        redirect_to higher_one_url
+      else
+        redirect_to url_for_path '/404'
+      end
+    end
+
+    private
+
+    def model_from_session
       options = {}
       options[:delegate_uid] = current_user.original_delegate_user_id if current_user.authenticated_as_delegate?
-      model = CampusSolutions::MyHigherOneUrl.from_session(session, options)
-      render json: model.get_feed_as_json
+      CampusSolutions::MyHigherOneUrl.from_session(session, options)
     end
 
   end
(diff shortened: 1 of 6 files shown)


app/controllers/campus_solutions/higher_one_url_controller.rb
```diff
@@ -12,5 +12,16 @@ def get
       render json: model.get_feed_as_json
     end
 
+    def redirect
+      options = {}
+      options[:delegate_uid] = current_user.original_delegate_user_id if current_user.authenticated_as_delegate?
+      model = CampusSolutions::MyHigherOneUrl.from_session(session, options)
```
paulkerschen: @davidlimcheng One way to DRY this up would be to extract the three lines above into a private method in this class, called `model_from_session` or something of the sort. Then you can start both public methods with `model = model_from_session`.

app/models/campus_solutions/my_higher_one_url.rb
```diff
@@ -16,5 +16,14 @@ def get_feed_internal
       CampusSolutions::HigherOneUrl.new(proxy_args).get
     end
 
+    def get_higher_one_url
+      return {} unless is_feature_enabled
+      proxy_args = {
+        user_id: @uid,
+        delegate_uid: @options[:delegate_uid]
+      }
+      CampusSolutions::HigherOneUrl.new(proxy_args).build_url.strip
```
paulkerschen: @davidlimcheng Here, too, you can extract duplicated code into a private method. The first line `return {} unless is_feature_enabled` should probably be left alone, but the code below that creates a new proxy from args could go into a private `proxy` method. Then you just call `proxy.build_url.strip` in this method, and `proxy.get` in the method above.

davidlimcheng:
Awesome feedback, thanks @paulkerschen!  Commit has been amended.

paulkerschen:
Looks good, @davidlimcheng, thanks!

PR #5304, opened 2016-04-27
SISRP-18419 Bulletproof CalCentral against EDO DB failure

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-18419

Separate out query caching from exception handling, and put the exception handling in a common query wrapper to guard against leaks.


raydavis:
Exceptions to the left of them, exceptions to the right of them, into the valley of death...

PR #5306, opened 2016-04-27
SISRP-18358 - CC-SIS-DEV: Deposit Payment not displaying in All Transactions

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-18358

AngularJS clears filters if the search term resolves to `undefined`.  The previous filter wasn't working, because I had set the search property to `itemTermId: ''`  This worked for most items, but when an item was sent over with `itemTermId: null`, it would filter it out.  I've now correctly set the `searchTermId.itemTermId` to default to undefined, to ensure we're not incorrectly filtering out any items.


paulkerschen:
Makes sense to me.

PR #5309, opened 2016-04-28
SISRP-18339 - 5.4 - Advisor Dashboard, User Lookup: Star icons to save / delete users

dfkaye:
https://jira.berkeley.edu/browse/SISRP-18339

Ready for review.

- This includes an incidental fix to `tasksController.js` which was causing some issues during other testing.
- Star icons are used as content inside of buttons so I put all css for those in `_buttons.css`.
- There is some discrepancy between searched users and saved users, namely the saved state on searched users.  There's a `updateSearchedUserSavedStates()` function in `userSearchController.js` to synchronize the `saved` state in the search tab.

### DIFF
--- a/src/assets/javascripts/angular/controllers/widgets/userSearchController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/userSearchController.js
@@ -36,7 +36,13 @@ angular.module('calcentral.controllers').controller('UserSearchController', func
   };
 
   var decorate = function(users) {
+    var firstName = 'first_name';
+    var lastName = 'last_name';
+
     angular.forEach(users, function(user) {
+      // Normalize user's person name for the UI.
+      user.name = user.name || (user[firstName] || '').concat(' ', user[lastName] || '');
+
       user.actAs = function() {
         adminService.actAs(user);
       };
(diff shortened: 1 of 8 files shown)


paulkerschen:
@dfkaye, looks like to make JSHint pass, either the property names in userSearchController.js need to be changed to camel case (check with @johncrossman as to whether the back end is providing camel-case versions); or else the controller needs a magic comment at top to turn off that check, as in pages/academicsController.js.

Code looks right to me otherwise; but I'll ask @johncrossman to take a quick look too since he's been recently working on this feature.

One small request: could you amend your commit title from "rebase fix" to something more in line with the PR title? Makes things more readable when browsing the commit history.

dfkaye:
@paulkerschen I've fixed the commit message.  
I noticed the editor complaints about snake_case in JavaScript files.  In the controller, there are 3 of them:

- `first_name`
- `last_name`
- `ldap_uid`

One of the templates uses `ldap_uid` and `student_id`

Should I modify those in the Ruby model or controller?

src/assets/javascripts/angular/controllers/widgets/userSearchController.js
```diff
@@ -28,6 +28,10 @@ angular.module('calcentral.controllers').controller('UserSearchController', func
     }
   };
 
+  var firstName = 'first_name';
+  var lastName = 'last_name';
+  var ldapUid = 'ldap_uid';
```
johncrossman: @dfkaye In a follow up Jira we will do a minor cleanup of `ldap_uid` in the templates / js. It's here because our search logic partially relies on `User::SearchUsers`, legacy view-as code to search-by-id.  I don't want to do an across the board refactor yet.

dfkaye:
I've modified the controller to use the following style for the snake_case properties:

    var ldapUid = 'ldap_uid';

    if (target[ldapUid] === source[ldapUid]) {

PR #5311, opened 2016-04-28
SISRP-18518, comment lists the possible CS affiliation status codes

johncrossman:
https://jira.berkeley.edu/browse/SISRP-18518

Code comment for future generations. Possible CS affiliation status codes:
* 'ACT' (active)
* 'INA' (inactive)
* 'ERR' (bad data which CalCentral should ignore)


paulkerschen:
Thanks for laying down the knowledge, @johncrossman.

PR #5314, opened 2016-04-29
SISRP-18106 - Updates University Requirements card to support Exempt status

redconfetti:
https://jira.berkeley.edu/browse/SISRP-18106

### DIFF
--- a/app/models/my_academics/requirements.rb
+++ b/app/models/my_academics/requirements.rb
@@ -13,13 +13,18 @@ module MyAcademics
         req_name, req_value = requirement.to_a
 
         display_name = case req_name.upcase
-                 when 'SUBJECTA' then 'UC Entry Level Writing'
-                 when 'AMERICANHISTORY' then 'American History'
-                 when 'AMERICANINSTITUTIONS' then 'American Institutions'
-                 when 'AMERICANCULTURES' then 'American Cultures'
-                 else req_name
-               end
-        display_status = req_value.upcase == 'REQT SATISFIED' ? 'met' : ''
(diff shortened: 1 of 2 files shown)


app/models/my_academics/requirements.rb
```diff
@@ -13,13 +13,18 @@ def merge(data)
...
-                 when 'AMERICANINSTITUTIONS' then 'American Institutions'
-                 when 'AMERICANCULTURES' then 'American Cultures'
-                 else req_name
-               end
-        display_status = req_value.upcase == 'REQT SATISFIED' ? 'met' : ''
+          when 'SUBJECTA' then 'UC Entry Level Writing'
+          when 'AMERICANHISTORY' then 'American History'
+          when 'AMERICANINSTITUTIONS' then 'American Institutions'
+          when 'AMERICANCULTURES' then 'American Cultures'
+          else req_name
```
johncrossman: @redconfetti Should this be `req_name.upcase`?

johncrossman:
I put one question, @redconfetti  Otherwise, looks good.

app/models/my_academics/requirements.rb
redconfetti: @johncrossman Good question. No, I don't think so. This method is converting the uppercase representations into their lowercase equivalents. I could convert it to lowercase, but there isn't any way to properly space between the words in whatever value that might be.

The front-end code (for this card/widget) is only configured to respect these configured keys, so it wouldn't matter much anyway.

johncrossman:
Thanks @redconfetti

PR #5323, opened 2016-05-02
SISRP-17871 - Implement responsive Due Dates for Billing Activity

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-17871

Responsive due dates that hide when view is "small".


paulkerschen:
Thanks @davidlimcheng!

PR #5325, opened 2016-05-02
SISRP-18651 Add grade option, units, email address to campus roster feed

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-18651

- Add grade option and units to enrollment query and roster feed, making sure the graded component is the section used;
- Stop filtering out email addresses;
- Clear up some existing confusion in the enrollment query between legacy 'pnp_flag' and new 'grading_basis.'


raydavis:
Looks good. I checked the original email filtering commit and it looks like we're OK with losing it.

PR #5326, opened 2016-05-02
SISRP-18395 - Enrollment card: section arrows overlap dates

redconfetti:
https://jira.berkeley.edu/browse/SISRP-18395


paulkerschen:
OK with Tra; vis is overruled.

PR #5328, opened 2016-05-02
SISRP-18670 - Student Overview Personal Summary 'View As' text not inline with Student Name

dfkaye:
https://jira.berkeley.edu/browse/SISRP-18670

Change proposed here is a CSS class we could reuse, unless there are objections...


paulkerschen:
Seems _useful to me. Thanks @dfkaye.

PR #5332, opened 2016-05-03
SISRP-18730 - Create Back-End for Get FPP Enrollment API using Mock Fixture

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-18730

Question regarding `CampusSolutions::GlobalCachedProxy`:
I can't think of a use case in which an invalid ID-less call would be made on this endpoint, especially since we are disallowing delegate and act-as access to this endpoint.  Can someone enlighten me as to the use of this passthrough in this specific case?

### DIFF
--- a/app/controllers/campus_solutions/fpp_enrollment_controller.rb
+++ b/app/controllers/campus_solutions/fpp_enrollment_controller.rb
@@ -1,0 +1,12 @@
+module CampusSolutions
+  class FppEnrollmentController < CampusSolutionsController
+
+    include DisallowAdvisorViewAs
+    include DisallowClassicViewAs
+
+    def get
+      json_passthrough CampusSolutions::FppEnrollment
+    end
+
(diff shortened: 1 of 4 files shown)


paulkerschen:
@davidlimcheng: To answer your question about `CampusSolutions::GlobalCachedProxy`, the safety check for ID-less calls is there because one of the subclasses, `CampusSolutions::FerpaDeeplink`, connects to an API that senselessly requires a working EMPLID even though it doesn't do anything with it. In most cases, including your new class, that logic would just be skipped.

The value you get by inheriting from `GlobalCachedProxy` is 1) standardized error handling inherited from the `Proxy` superclass, and 2) caching a single feed instance for everyone instead of the more usual ID-specific feeds. (This is done by setting `instance_key` to nil, which overrides the superclass behavior of setting `instance_key` to the UID.)

PR #5333, opened 2016-05-03
SISRP-17895 Adapt MyAcademics::GpaUnits to end of Bear Facts data

raydavis:

Another one down. I also modified the CS/Hub logic to obey the existing GPA-as-string API contract.


paulkerschen:
Quite so, @raydavis.

PR #5334, opened 2016-05-03
CLC-6224 - Fixed CSS priority issue with primary buttons that are anchors

nicolaasmatthijs:
Fixed CSS priority issue with primary buttons that are anchors


paulkerschen:
Thanks @nicolaasmatthijs! FYI we'd usually stick an explicit link to https://jira.ets.berkeley.edu/jira/browse/CLC-6224 in the PR description, just to make it easier on future generations.

PR #5335, opened 2016-05-03
SISRP-18743 - Create Back-End for Get EFT Enrollment API using Mock Fixture

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-18743

Includes some configuration for an EFT proxy that has not yet been developed.

### DIFF
--- a/app/controllers/my_eft_enrollment_controller.rb
+++ b/app/controllers/my_eft_enrollment_controller.rb
@@ -1,0 +1,7 @@
+class MyEftEnrollmentController < ApplicationController
+
+  def get_feed
+    render json: Eft::MyEftEnrollment.from_session(session).get_feed_as_json
+  end
+
+end
--- a/app/models/eft/my_eft_enrollment.rb
+++ b/app/models/eft/my_eft_enrollment.rb
@@ -1,0 +1,50 @@
+module Eft
+  class MyEftEnrollment < UserSpecificModel
+    include Cache::CachedFeed
+    include Cache::JsonifiedFeed
+    include Cache::FeedExceptionsHandled
+    include Cache::UserCacheExpiry
+    include Proxies::HttpClient
+    include Proxies::Mockable
+    include ClassLogger
+
+    def initialize(uid, options={})
+      super(uid, options)
+      @settings = Settings.eft_proxy
+      @fake = (options[:fake] != nil) ? options[:fake] : @settings.fake
+      initialize_mocks if @fake
+    end
+
+    def default_message_on_exception
+      'Failed to connect with EFT system.'
+    end
+
+    def get_feed_internal
+      return {} unless Settings.features.cs_billing
+      get_parsed_response
+    end
+
+    def get_parsed_response
+      url = "#{@settings.base_url}/student/#{@uid}"
+      logger.info "get_parsed_response: Fake = #{@fake}; Making request to #{url} on behalf of user #{@uid}; cache expiration #{self.class.expires_in}"
+      response = get_response(
+        url,
+        basic_auth: {username: @settings.username, password: @settings.password},
+        on_error: {rescue_status: 404}
+      )
+      feed = {statusCode: response.code}
+      if response.code == 404
+        logger.debug "404 response from EFT Enrollment API for user #{@uid}"
+        feed.merge(body: 'No EFT data could be found for your account.')
+      else
+        logger.debug "EFT remote response: #{response.inspect}"
+        feed.merge response.parsed_response
+      end
+    end
+
+    def mock_json
+      read_file('fixtures', 'json', 'eft_enrollment.json')
+    end
+
+  end
+end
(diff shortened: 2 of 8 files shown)


app/controllers/my_eft_enrollment_controller.rb
```diff
@@ -0,0 +1,7 @@
+class MyEftEnrollmentController < ApplicationController
+
+  def get_feed
+    render json: Eft::MyEftEnrollment.from_session(session).get_feed_as_json
```
paulkerschen: @davidlimcheng Would `Eft::MyEftEnrollment` need to include either `Cache::JsonAddedCacher` or `Cache::JsonifiedFeed` in order to implement `get_feed_as_json`? (All things being equal, it's a bit more efficient to use the second option.)

Also, while I'm not a stickler for specs on what is basically scaffolding, we'll of course want them for both the controller and the model once actual behavior gets wired in.

app/models/eft/my_eft_enrollment.rb
```diff
@@ -0,0 +1,44 @@
...
+      url = "#{@settings.base_url}/student/#{@uid}"
+      logger.info "get_parsed_response: Fake = #{@fake}; Making request to #{url} on behalf of user #{@uid}; cache expiration #{self.class.expires_in}"
+      response = get_response(
+        url,
+        basic_auth: {username: @settings.username, password: @settings.password},
+      )
+      feed = {statusCode: response.code}
+      if response.code == 404
+        logger.debug "404 response from EFT Enrollment API for user #{@uid}"
+        feed.merge(body: 'No EFT data could be found for your account.')
```
paulkerschen: @davidlimcheng While this is a bit of a hidden convention, if you want custom 404 handling you might need to add `on_error: {rescue_status: 404}` to your `get_response` options: otherwise it will just raise an error. See Cal1Card::Photo for an example.

Verifying this kind of behavior is the kind of thing you'd write a spec for, of course. :)

spec/models/eft/my_eft_enrollment_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+    context 'fetching fake data feed' do
+      subject { fake_model.get_parsed_response }
+
+      it_behaves_like 'a polite HTTP client'
+
+      it 'has correctly parsed JSON' do
+        puts subject
+        expect(subject[:statusCode]).to eq 200
+        expect(subject["data"]["transaction-status"]).to eq "Confirmed"
+      end
```
davidlimcheng: @paulkerschen : For some reason, my JSON keys are getting parsed out as strings instead of symbols.  I know that symbols are preferred, but I absolutely could not figure out why it was getting parsed out this way.  Any ideas as to why this is happening?

davidlimcheng:
@paulkerschen Made the changes we talked about, but I left a question for you regarding the format of the parsed JSON.  Thanks!

spec/models/eft/my_eft_enrollment_spec.rb
raydavis: When we make hashes ourselves, we prefer symbols, but when we get a parsed JSON or XML response or get a SQL query result, the keys will always be strings. You'll see similar mishmashes of symbol-keys-by-us + string-keys-from-elsewhere all through the code. If the hash is going to get processed more by levels above the low-level proxy feed, we're more likely to get rid of the string keys.

raydavis: More concisely, "proxy" feeds tend to have string keys & "merged model" feeds tend to be symbolized.

spec/models/eft/my_eft_enrollment_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+
+  describe '#get_parsed_response' do
+
+    context 'fetching fake data feed' do
+      subject { fake_model.get_parsed_response }
+
+      it_behaves_like 'a polite HTTP client'
+
+      it 'has correctly parsed JSON' do
+        puts subject
```
raydavis: Gotta lose the puts, of course. :)

davidlimcheng: Rookie mistake =)

spec/models/eft/my_eft_enrollment_spec.rb
davidlimcheng: @raydavis I see - so if we do indeed parse out the strings to symbols, it would be something specific to the model or controller (usually), and would probably accompany some type of further parsing.  That makes sense - thanks for the explanation!

spec/models/eft/my_eft_enrollment_spec.rb
raydavis: Nah, we all end up submitting PRs with those devils slipping in. I think Paul K. & I both did it in the last week. :)

davidlimcheng:
Pesky `puts` has been amended, should be good to go pending travis.

raydavis:
Thanks for the hoop jumping, @davidlimcheng

PR #5336, opened 2016-05-04
SISRP-17895 Add simple end-of-BearFacts protection & lots of TODOs

raydavis:

The idea here is to stave off total disaster rather than to ensure a good UX in Fall 2016. I'll open follow-up tickets for remaining legacy-data mopping chores.


paulkerschen:
No one can say we weren't warned. Thanks @raydavis.

PR #5341, opened 2016-05-05
SISRP-18516 Feature flag and feed item for Class Info enrollment tab

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-18516

For purposes of the enrollment tab, the front will want to know if a given term is sourced from Campus Solutions; there are probably other purposes in the future.


paulkerschen:
Travis is _at least_ half pacified.

PR #5342, opened 2016-05-05
SISRP-18780, apply uniq filter on user-search result set

johncrossman:
https://jira.berkeley.edu/browse/SISRP-18780
https://jira.berkeley.edu/browse/SISRP-18774

* Remove dupes in post_processing of result set.
* Move `add_user_attributes` such that results of search-by-SID get name info

### DIFF
--- a/app/controllers/search_users_controller.rb
+++ b/app/controllers/search_users_controller.rb
@@ -11,14 +11,13 @@ class SearchUsersController < ApplicationController
 
   def search_users
     users = authorize_results by_id(id_param)
-    add_user_attributes users
-    render json: { users: decorate(users) }.to_json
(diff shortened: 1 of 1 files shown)


app/controllers/search_users_controller.rb
```diff
@@ -11,14 +11,13 @@ class SearchUsersController < ApplicationController
 
   def search_users
     users = authorize_results by_id(id_param)
-    add_user_attributes users
```
johncrossman: This call to `add_user_attributes` was moved in to `by_id` method such that `by_id_or_name` reaps its benefits, too.

paulkerschen:
Welcome back, @johncrossman!

PR #5345, opened 2016-05-05
SISRP-18516 Class Info enrollment tab front end

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-18516

Class Info pages for instructors get a new "Enrollment" tab between "Class Info" and "Roster," if it's a Campus Solutions term and the feature flag is turned on. Per requirements, each tab position gets its own URL, and tabs are changed from buttons to links.

Enrollment and waitlist summaries surface the fields we can currently get without database overhead (name, SID, email, units and grade option). More fields may come later. Tables are sortable and reverse-sortable by column, and like rosters can be filtered by section.


raydavis:
OK more feature flags = big win

PR #5346, opened 2016-05-05
SISRP-18764 - Update Textbook and Booklist to avoid instructor email overload

redconfetti:
https://jira.berkeley.edu/browse/SISRP-18764

### DIFF
--- a/app/models/textbooks/proxy.rb
+++ b/app/models/textbooks/proxy.rb
@@ -103,7 +103,7 @@ module Textbooks
     def get_as_json
       self.class.smart_fetch_from_cache(
         {id: "#{@slug}-#{@dept}-#{@course_catalog}-#{@section_numbers.join('-')}",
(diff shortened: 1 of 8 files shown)


paulkerschen:
Code looks good, @redconfetti, but four commits seems like a lot of history. What would you think about squashing them down to one?

redconfetti:
@paulkerschen So we prefer to keep updates singular, with a one-to-one ratio with assigned tasks, instead of separate so that the changes are easier to follow and understand?

paulkerschen:
Yes, @redconfetti, unless it's quite a large task, or has some very sharp dividing line. In this case I think the functional intent of all changes is closely aligned enough that it makes sense to group them in one commit.

redconfetti:
@paulkerschen Your zucchini is ready (green + squash)

paulkerschen:
Yum!

PR #5348, opened 2016-05-05
SISRP-18763, refactor out snake_case in search_users, uid_lookup and view_as templates

johncrossman:
https://jira.berkeley.edu/browse/SISRP-18763

The view-as templates have had a lot snake_case (e.g., ldap_uid). The work on Advisor student lookup has, at times, followed the same snake_case pattern because of shared code, causing bugs like SISRP-18763. It's time to refactor away the bad pattern.

### DIFF
--- a/src/assets/javascripts/angular/controllers/widgets/userSearchController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/userSearchController.js
@@ -41,13 +41,11 @@ angular.module('calcentral.controllers').controller('UserSearchController', func
   };
 
   var decorate = function(users) {
-    var firstName = 'first_name';
-    var lastName = 'last_name';
     var missingName = 'Name Not Provided';
 
     angular.forEach(users, function(user) {
       // Normalize user's person name for the UI.
-      user.name = user.name || user.defaultName || (user[firstName] || '').concat(' ', user[lastName] || '');
+      user.name = user.name || user.defaultName || (user.firstName || '').concat(' ', user.lastName || '');
       // Guard against whitespace-only name.
       if (/^\s+$/.test(user.name)) {
         user.name = missingName;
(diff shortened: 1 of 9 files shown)


src/assets/javascripts/angular/controllers/widgets/userSearchController.js
```diff
@@ -43,11 +43,13 @@ angular.module('calcentral.controllers').controller('UserSearchController', func
   var decorate = function(users) {
     var firstName = 'first_name';
     var lastName = 'last_name';
+    var studentId = 'student_id';
     var missingName = 'Name Not Provided';
 
     angular.forEach(users, function(user) {
       // Normalize user's person name for the UI.
       user.name = user.name || user.defaultName || (user[firstName] || '').concat(' ', user[lastName] || '');
+      user[studentId] = user[studentId] || user.studentId;
```
raydavis: Seems a little sad to stick client-side code with this, since I think all the results are coming back from a single API. Do you think it might be possible to normalize the contract on the server's side?

johncrossman: @raydavis I agree. I'll review.

johncrossman:
@raydavis  IMO, it is better – and less risky – to ditch these snake_case hacks than to continue working around them. Luckily, the scope of this refactor is limited to `search_users`, `uid_lookup` and `view_as` templates and the corresponding server-side controllers. I've done a good amount of verification locally.

I removed stubbed ids from `stored_users_controller_spec` because it was colliding with fixture data (e.g., two studentIds per user).

cc: @paulkerschen

raydavis:
Thanks, @johncrossman , I like it. I also appreciate how much the "prepare_to_render" method name reminds me of my own experience of higher education.

PR #5350, opened 2016-05-05
SISRP-18781 Enable SSL TLSv1.2 protocol for textbooks API, etc.

raydavis:
https://jira.berkeley.edu/browse/SISRP-18781


johncrossman:
Ah yes, our ol' friend SSL. Thanks for the added robustness, @raydavis

PR #5353, opened 2016-05-06
SISRP-18739, if search input is 'Doe' then search LDAP with *Doe*

johncrossman:
https://jira.berkeley.edu/browse/SISRP-18739

In previous logic, a search for 'Doe' would cause LDAP search with `Doe*`. This PR gives us the regex we want: `*Doe*`.


raydavis:
Sorry but I'm not sure about this, @johncrossman & @pfarestveit. When we've defaulted to substring matches in the past, the results have been unhappy. A population the size of UC Berkeley returns so many false positives that it's impossible for the searcher to guess how to winnow them down.

The original ticket seems to be asking for a search of last name as well as first name. Might there be some more focused way to deliver that? (Feel free to toss the question back at me if you don't have time to deal with it!)

paulkerschen:
As someone with no expertise whatsoever in this area, I'll throw in a performance question: http://stackoverflow.com/questions/9564120/using-wildcards-in-ldap-search-filters-queries

johncrossman:
@raydavis I'm going to close this PR and investigate a better way. And thank you @paulkerschen for the pointer to performance concern.

PR #5355, opened 2016-05-06
SISRP-18739, if search input is 'Doe' then search LDAP with *Doe*

johncrossman:
https://jira.berkeley.edu/browse/SISRP-18739

Based on feedback from @raydavis and @paulkerschen in PR #5353 

I'm now using `Net::LDAP::Filter.intersect` to construct chain all filters before running the LDAP search. If the input is a single token then the constructed filter is: `first_name=Doe* OR last_name=Doe*`.

### DIFF
--- a/app/models/calnet_ldap/client.rb
+++ b/app/models/calnet_ldap/client.rb
@@ -32,14 +32,26 @@ module CalnetLdap
     end
 
     def search_by_name(name, include_guest_users=false)
-      results = []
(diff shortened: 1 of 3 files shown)


johncrossman:
@paulkerschen Regarding performance: we have no leading wild-cards and the chaining of filters means one LDAP search instead of one per filter.

raydavis:
Lookin' sharp, @johncrossman -- thank you!

PR #5356, opened 2016-05-06
SISRP-18733 More workarounds for spotty EDO DB data

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-18733

Some gymnastics that I wish weren't necessary.

For complex reasons having to do with EDO DB implementation, checking for an 'ACTIVE' status code under ON rather than WHERE will screen out a small class of valid courses that exist in the cross-listing table as 'ACTIVE', but in the course table as 'DEPRECATED'.

However, for complex reasons having to do with EDO DB implementation, rewriting the query in this way will result in duplicate results for a small class of courses that link to both active, correctly formatted department names (e.g., 'PHYS ED') and deprecated, incorrectly formatted department names ('PHYSED') through the course/class translation table.

However, for complex reasons having to do with EDO DB implementation, we can screen out the duplicate results if we drop some of our more specific database-level column ordering. 

Further, for complex reasons having to do with EDO DB implementation, database-level ordering was inconsistent in any case and we're already resorting at a higher level in Ruby. Therefore dropping the database ordering has no impact on UX.

### DIFF
--- a/app/models/edo_oracle/queries.rb
+++ b/app/models/edo_oracle/queries.rb
@@ -3,7 +3,7 @@ module EdoOracle
     include ActiveRecordHelper
     include ClassLogger
 
-    CANONICAL_SECTION_ORDERING = 'dept_name, catalog_root, catalog_prefix nulls first, catalog_suffix nulls first, primary DESC, instruction_format, section_display_name, section_num'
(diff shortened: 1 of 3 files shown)


raydavis:
Well, gosh, if it's for _complex reasons_, we'd better get the change in there!

PR #5357, opened 2016-05-06
SISRP-16788 - Enrollment - Law: show different text for advisors depending on program

redconfetti:
https://jira.berkeley.edu/browse/SISRP-16788

* Adds static advisor contact list for display to Berkeley Law students
* Modifies text in appointment time section. Fixes logic for detecting if student is a Law student (see  [SISRP-17895](https://jira.berkeley.edu/browse/SISRP-17895))
* Fixes logic for displaying date in Enrollment Adjust section template, and fixing the exclusion of the 'Change Grading Option' and 'Withdraw' links
* Adds dummy feeds for holdsFactory

### DIFF
--- a/src/assets/templates/widgets/enrollment/consult_advisor.html
+++ b/src/assets/templates/widgets/enrollment/consult_advisor.html
@@ -1,34 +1,56 @@
-<div data-cc-spinner-directive="enrollment.holds.isLoading">
+<div class="cc-enrollment-card-advisors-consult" data-cc-spinner-directive="enrollment.holds.isLoading">
+
   <div data-ng-if="enrollment.holds.hasHolds" class="cc-flex">
     <div>
       <i class="fa fa-exclamation-circle cc-icon-red cc-enrollment-card-icon"></i>
     </div>
     <div>
-      You have a hold that may affect your ability to enroll in classes.
-      Consult with an advisor.
+      You have a hold that may affect your ability to enroll in classes. Consult with an advisor.
     </div>
   </div>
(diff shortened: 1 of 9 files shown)


src/assets/templates/widgets/enrollment/consult_advisor.html
```diff
@@ -1,34 +1,56 @@
...
+      </li>
+      <li>
+        JSP students should contact Margo Rodriguez at
+        <a href="mailto:mrodriguez@law.berkeley.edu">mrodriguez@law.berkeley.edu</a>.
+      </li>
+    </ul>
+  </div>
+
+  <div data-ng-if="!enrollmentTerm.isLawStudent">
+    <!-- This presentation not yet in use. See SISRP-15831 -->
```
johncrossman: @redconfetti Personally, I don't like having Jira numbers in place of a short explanation. Future devs might be too lazy (or unable) to read a long Jira thread.

johncrossman:
Looks good, @redconfetti   Waiting on Travis

src/assets/templates/widgets/enrollment/consult_advisor.html
redconfetti: @johncrossman okay, I'll just remove it and then they can look at the blame, and then hunt down the JIRA, if they care that much. I just figured this would help because Christian implemented this logic before the data feed providing the data is even available... it would help if people knew that it's not there, instead of pulling their hair out trying to find a UID for a user that has the data.

PR #5359, opened 2016-05-06
SISRP-18908 - Return Holds Section to Status, Holds, & Blocks

davidlimcheng:
More info at:
https://jira.berkeley.edu/browse/SISRP-18908

I left the whole `$watchGroup` logic in there because I'm not exactly sure how the behavior of `studentInfo.regStatus.code` works over time.

@johncrossman, let me know if I should PR this against QA Branch as well.

### DIFF
--- a/src/assets/javascripts/angular/controllers/widgets/academicsStatusHoldsBlocksController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/academicsStatusHoldsBlocksController.js
@@ -15,12 +15,14 @@ angular.module('calcentral.controllers').controller('AcademicsStatusHoldsBlocksC
   $scope.$watchGroup(['studentInfo.regStatus.code', 'api.user.profile.features.csHolds'], function(newValues) {
     var enabledSections = [];
 
-    if (newValues[0] !== null || newValues[1]) {
+    if (newValues[0] !== null) {
       enabledSections.push('Status');
     }
-    if (newValues[2]) {
+
+    if (newValues[1]) {
       enabledSections.push('Holds');
     }
+
     enabledSections.push('Blocks');
 
     $scope.statusHoldsBlocks.enabledSections = enabledSections;


src/assets/javascripts/angular/controllers/widgets/academicsStatusHoldsBlocksController.js
```diff
@@ -12,15 +12,14 @@ angular.module('calcentral.controllers').controller('AcademicsStatusHoldsBlocksC
   // for changes in order to display the corresponding UI elements.
   $scope.statusHoldsBlocks = {};
 
-  $scope.$watchGroup(['studentInfo.regStatus.code', 'api.user.profile.features.csHolds'], function(newValues) {
+  $scope.$watchGroup(['studentInfo.regStatus.code'], function(newValues) {
     var enabledSections = [];
 
     if (newValues[0] !== null || newValues[1]) {
```
johncrossman: @davidlimcheng `newValues[1]` is not needed, right? Our array will have only one element.

davidlimcheng: @johncrossman Yep, my bad.  Confusion stemmed from $watchGroup behavior - amended.

src/assets/javascripts/angular/controllers/widgets/academicsStatusHoldsBlocksController.js
```diff
@@ -12,15 +12,14 @@ angular.module('calcentral.controllers').controller('AcademicsStatusHoldsBlocksC
...
 
-    if (newValues[0] !== null || newValues[1]) {
+    if (newValues[0] !== null) {
       enabledSections.push('Status');
     }
-    if (newValues[2]) {
-      enabledSections.push('Holds');
-    }
+
+    enabledSections.push('Holds');
```
johncrossman: @davidlimcheng After closer inspection, I think we need `api.user.profile.features.csHolds` in the mix. We don't want  `enabledSections.push('Holds')` if the feature flag is false. Would the following code make sense?
```
if (newValues[0] !== null) {
  enabledSections.push('Status');
}
if (newValues[1]) {
  enabledSections.push('Holds');
}
```

davidlimcheng: @johncrossman Yeah, this makes sense.  I wasn't sure if we wanted feature flags to stick around after releasing, but it makes sense that we do want to keep them.  Will amend shortly.

davidlimcheng:
@johncrossman Voila, amended!

johncrossman:
Thanks @davidlimcheng  Waiting on T

PR #5360, opened 2016-05-08
SISRP-18736, if user search has error then clear previous results

johncrossman:
https://jira.berkeley.edu/browse/SISRP-18736

Reset `users` array (i.e., search results) on error.


paulkerschen:
Sensible!

PR #5365, opened 2016-05-09
SISRP-18941 - Remove Enrollment Verification and Transcript Links

redconfetti:
https://jira.berkeley.edu/browse/SISRP-18941

See QA PR #5367

### DIFF
--- a/src/assets/templates/academics_semesters.html
+++ b/src/assets/templates/academics_semesters.html
@@ -1,18 +1,9 @@
 <div class="cc-widget">
   <div class="cc-widget-title">
     <h2 class="cc-left">Semesters</h2>
-    <a data-ng-href="{{transcriptLink}}" class="cc-right cc-widget-title-link" data-ng-if="transcriptLink">
-      <strong>Request Transcripts</strong>
-    </a>
   </div>
 
   <div class="cc-widget-padding cc-flex-space-between-vertical-15">
-    <div data-ng-if="enrollmentVerificationLink">
-      <a data-ng-href="{{enrollmentVerificationLink}}">
-        Request your enrollment verification
-      </a>
-    </div>
-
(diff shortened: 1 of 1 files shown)


config/settings.yml
```diff
@@ -430,9 +430,11 @@ features:
   class_enrollment_summer_2016: false
   class_info_enrollment_tab: false
   course_manage_official_sections: false
+  cs_academic_enrollment_verification_link: true
```
raydavis: @redconfetti , to reduce the risk of unpleasant surprises, we prefer to configure {{settings.yml}} flags as we expect them to be set on production as of the next release. That's why there are so many "false" defaults.

It seems like these should follow the same rule. Does that sound right to you, too?

redconfetti: @raydavis I was assuming that we'd want them to be what they will be by default most of the time in the future, with an explicit override (as `false` in this case) in production.

Thanks for the clarification on our policy though. I'll change them to the settings we expect with the release. Is it a good idea to assume that we can omit the need to document this in the release notes, since it will have the correct configuration by default? Or should I include it just so that it's on the release teams radar?

raydavis: Thanks for being flexible, @redconfetti . If we know what Go-Live/release the enabling happens in and there's a slot for that release in Confluence, then we can put an instruction in. We have an anticipatory instruction for GoLive 7, for example, and that gives a place to note when the flag was enabled on the test environments.

redconfetti: @raydavis Okay. So I'll configure the code to default to false, and when these links become enabled again via a feature flag, we'll simply add a mention in the release notes for that explicit configuration of `true` for these values.

raydavis: Yep, that's been the protocol. Followed (eventually, we hope) by later deletion of the feature flag.

redconfetti:
I've simplified this update to remove the links. We'll simply add them back later.

PR #5370, opened 2016-05-10
SISRP-18807 Distinguish enrolled courses from zero-position waitlisted courses

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-18807

Various bits of CalCentral code have used "has waitlist position of zero?" as shorthand for "is not waitlisted in this section?" This will no longer fly in the EDO DB world, which conceives positionless waitlists for non-enrollment sections. Accordingly we need an explicit `waitlisted` section property.


raydavis:
Thank you for covering legacy data sources too, @paulkerschen

PR #5371, opened 2016-05-10
CLC-6287 / CLC-6289, bump up cache-expiration on Canvas::MergedUserSites and ServiceAlerts::Merged

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-6287
https://jira.ets.berkeley.edu/jira/browse/CLC-6289

Small steps towards a land of better performance, assuming such a place exists.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -256,7 +256,7 @@ cache:
     CampusOracle::CourseSections: NEXT_08_00
     CampusOracle::UserAttributes: NEXT_08_00
 
-    Canvas::MergedUserSites: <%= 15.minutes %>
+    Canvas::MergedUserSites: <%= 30.minutes %>
     Canvas::UserCourses: <%= 30.minutes %>
     Canvas::Groups: <%= 30.minutes %>
     Canvas::UserActivityStream: <%= 10.minutes %>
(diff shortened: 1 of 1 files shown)


config/settings.yml
```diff
@@ -275,7 +275,7 @@ cache:
 
     Textbooks::Proxy: <%= 24.hours %>
 
-    ServiceAlerts::Merged: <%= 2.minutes %>
+    ServiceAlerts::Merged: <%= 24.hours %>
```
paulkerschen: @johncrossman Bumping this up will require us to do a manual cache clear every time we (or others) add or remove a service alert. Granted that it's not the most terrible hardship, one Postgres call every two minutes (this feed isn't user-specific) still doesn't seem like much of a load. Do we have reason to think otherwise?

johncrossman: I had the same thought, @paulkerschen  

@raydavis might disagree but, IMO, clearing cache every time we add/remove a alert is reasonable. The length of cache should be determined by how often content is likely to change.

raydavis: As I wrote at the JIRA, I don't think it makes much difference one way or the other.

FWIW, though, cache lifespan is actually a negotiation between "how often the data changes", "how expensive the data is to get", and "how important fresh data is." That's why it's an art. :)

PR #5377, opened 2016-05-10
SISRP-19020 - Residency: Haas EWMBA student showing "California Residency" with blank status

dfkaye:
https://jira.berkeley.edu/browse/SISRP-19020

Suppresses residency block in template, if `residency.description` is not set by the controller's `parseCalResidency()` function - i.e., when `residency` is empty in the feed.

Also adding spinner for the residency block in loading state and when `residency.description` is not yet set.


paulkerschen:
Thanks @dfkaye.

PR #5392, opened 2016-05-11
[QA] NOJIRA Make Webmock gem available under production RAILS_ENV

paulkerschen:
For fake-proxy-under-load-test purposes.


paulkerschen:
After-the-fact JIRA https://jira.ets.berkeley.edu/jira/browse/CLC-6308

PR #5393, opened 2016-05-11
SISRP-16435, links for Advisors: Multi-Year Academic Planner and Schedule Planner

johncrossman:
https://jira.berkeley.edu/browse/SISRP-16435
https://jira.berkeley.edu/browse/SISRP-19058 (remove Academic Planner link on Advising Student Profile card)

* Generic Advising Resources card gets: multi_year_academic_planner, schedule_planner
* Student Profile card for Advisors gets: multi_year_academic_planner_student_specific, schedule_planner_student_specific

Eventually these links should be served by CS API but there is an urgent need. Thus we manage them via CalCentral settings.


paulkerschen:
So many ways to get a BCS URL... thanks @johncrossman.

PR #5397, opened 2016-05-12
SISRP-19184 - Profile Contact Info: click edit and data vanishes

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-19184

According to this [W3 page (section 4.3)](https://www.w3.org/TR/html-markup/syntax.html), `<select>` elements are not void elements, and must have a closing `</select>` tag - otherwise, the `<select>` will eat up all of the elements underneath it, which is what it was doing when the "edit" button was being pushed.

This bug also lives in other places, primarily in Profile, so I've created [SISRP-19203](https://jira.berkeley.edu/browse/SISRP-19203) to ensure the other ones get cleaned up as well.

### DIFF
--- a/src/assets/templates/widgets/profile/address.html
+++ b/src/assets/templates/widgets/profile/address.html
@@ -5,7 +5,8 @@
   </label>
 </div>
 <div class="cc-select">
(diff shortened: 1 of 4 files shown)


johncrossman:
Looks right to me, @davidlimcheng   Thank you.  We'll need a QA PR, too.

PR #5398, opened 2016-05-12
SISRP-17920 - Update bConnected information when a user doesn't have an account yet

dfkaye:
https://jira.berkeley.edu/browse/SISRP-17920

Note: This is 5.6 change


johncrossman:
Thanks @dfkaye  Looks right.

PR #5399, opened 2016-05-12
SISRP-19070 - SLR: Correcting Language of Status

dfkaye:
https://jira.berkeley.edu/browse/SISRP-19070

California Residency status text changes for 5.4

PR #5400, opened 2016-05-12
[QA] SISRP-19070 - SLR: Correcting Language of Status

dfkaye:
https://jira.berkeley.edu/browse/SISRP-19070

California Residency status text changes for 5.4

master PR https://github.com/ets-berkeley-edu/calcentral/pull/5399


src/assets/templates/california_residency_status.html
```diff
@@ -7,17 +10,13 @@
     data-cc-outbound-enabled="false">submit your Statement of Legal Residency</a>.
```
johncrossman: Link text should not include 'submit more'. See comment in PR #5399

PR #5402, opened 2016-05-12
[QA] CLC-6282 Roll back Signet and SSL gem updates

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6282

Based on load test results, revert #5259 (gem-update commit only) and #5350.


johncrossman:
Matches what we ran in load test. Thanks, @paulkerschen   Waiting on T

PR #5403, opened 2016-05-13
SISRP-16435, hide resource links on student overview resource when blank config

johncrossman:
https://jira.berkeley.edu/browse/SISRP-16435

Student overview should not surface blank URLs.


johncrossman:
*Note:* https://jira.berkeley.edu/browse/SISRP-19251 is specifically about ‘blank config -> no URLs for user’

PR #5408, opened 2016-05-13
SISRP-19171 - 5.4 - Delegated Access - In Linking Your Account page both Security key and Email address fields having a white space is graying out the Link Button

dfkaye:
https://jira.berkeley.edu/browse/SISRP-19171

Note: this is a **5.6** change.

Fixes the Delegated Access linking account input whitespace problem.  The updated `data-ng-pattern` directive now allows leading, trailing whitespace in the text inputs.  The `data-ng-trim="false"` directive is removed so that angular trims the field values before applying them to the model.

### DIFF
--- a/src/assets/templates/widgets/delegate/linking.html
+++ b/src/assets/templates/widgets/delegate/linking.html
@@ -19,7 +19,7 @@
         </label>
       </div>
       <div>
-        <input type="text" id="cc-delegate-linking-key" maxlength="20" data-ng-model="delegate.currentObject.data.securityKey" data-ng-pattern="/^[\S]+$/" data-ng-trim="false" required aria-required="true" />
+        <input type="text" id="cc-delegate-linking-key" maxlength="20" data-ng-model="delegate.currentObject.data.securityKey" data-ng-pattern="/^[\s]*[\S]+[\s]*$/" required aria-required="true" />
       </div>
       <div>
         <label for="cc-delegate-linking-email">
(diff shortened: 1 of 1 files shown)


src/assets/templates/widgets/delegate/linking.html
```diff
@@ -19,15 +19,15 @@ <h3 class="cc-delegate-linking-title">Link Your Account with Your Student</h3>
         </label>
       </div>
       <div>
-        <input type="text" id="cc-delegate-linking-key" maxlength="20" data-ng-model="delegate.currentObject.data.securityKey" data-ng-pattern="/^[\S]+$/" data-ng-trim="false" required aria-required="true" />
+        <input type="text" id="cc-delegate-linking-key" maxlength="20" data-ng-model="delegate.currentObject.data.securityKey" data-ng-pattern="/^[\s]*[\S]+[\s]*$/" required aria-required="true" />
```
johncrossman: @dfkaye Looks good. Can you verify that the value is being stripped (ie, trimmed) before we send it to campus solutions.

dfkaye: @johncrossman Affirmative.  Using console fu, we can see the text input values, `"   some   "`  and `"  value   "` are trimmed before mapping the request data to json as:

    proxyEmailAddress  "value"
    securityKey        "some"

johncrossman:
Thank you, @dfkaye

PR #5415, opened 2016-05-17
SISRP-18744 - Advisor student overview page needs a better UX while data elements are loading

dfkaye:
https://jira.berkeley.edu/browse/SISRP-18744

1) show spinner over residency while it loads
2) show spinner over Class Enrollments while it loads
3) display "No enrollments" message in Class Enrollments if there are none

### DIFF
--- a/src/assets/templates/academics_status_holds_blocks.html
+++ b/src/assets/templates/academics_status_holds_blocks.html
@@ -30,7 +30,14 @@
               <td colspan="2" data-ng-bind-html="studentInfo.regStatus.explanation"></td>
             </tr>
           </tbody>
-          <tbody data-ng-if="!residency.isLoading && residency.description">
+          <tbody data-ng-if="residency.isLoading">
+            <tr>
+              <td colspan="2">
+                <div data-cc-spinner-directive="residency.isLoading"></div>
+              </td>
+            </tr>
+          </tbody>
+          <tbody data-ng-if="!residency.isLoading">
             <tr>
               <th class="cc-table-subheader" scope="row">California Residency</th>
               <td>
--- a/src/assets/templates/widgets/enrollment_card.html
+++ b/src/assets/templates/widgets/enrollment_card.html
@@ -1,11 +1,13 @@
 <div class="cc-enrollment-card" data-ng-controller="EnrollmentCardController">
-  <div class="cc-widget" data-ng-if="enrollment.isLoading">
+  <div class="cc-widget" data-ng-if="enrollment.isLoading || enrollment.terms.length === 0">
     <div class="cc-widget-title">
       <h2>Class Enrollment</h2>
     </div>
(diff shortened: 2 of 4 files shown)


src/assets/templates/academics_status_holds_blocks.html
```diff
@@ -30,7 +30,14 @@ <h2 data-ng-bind="statusHoldsBlocks.enabledSections | andFilter"></h2>
               <td colspan="2" data-ng-bind-html="studentInfo.regStatus.explanation"></td>
             </tr>
           </tbody>
-          <tbody data-ng-if="!residency.isLoading && residency.description">
+          <tbody>
+            <tr>
+              <td colspan="2">
+                <div data-cc-spinner-directive="residency.isLoading && !residency.description"></div>
```
johncrossman: @dfkaye Couple questions:
* Why do we need  `!residency.description` in the expression above? The `tbody` block below only relies on `residency.isLoading`.
* Now that you've removed the `data-ng-if`, will the `tbody` block above cause unwanted padding in the UI?

src/assets/templates/widgets/enrollment_card.html
```diff
@@ -1,11 +1,13 @@
 <div class="cc-enrollment-card" data-ng-controller="EnrollmentCardController">
-  <div class="cc-widget" data-ng-if="enrollment.isLoading">
+  <div class="cc-widget" data-ng-if="enrollment.isLoading || !enrollment.terms.length">
     <div class="cc-widget-title">
       <h2>Class Enrollment</h2>
     </div>
-    <div data-cc-spinner-directive="enrollment.isLoading"></div>
+    <div class="cc-widget-padding" data-cc-spinner-directive="enrollment.isLoading">
+      <div class="cc-widget-text">No Enrollments</div>
+    </div>
```
johncrossman: @dfkaye Does this 'No Enrollments' div need `enrollment.terms.length === 0`?

src/assets/templates/academics_status_holds_blocks.html
dfkaye: @johncrossman `residency.description` is set by the controller once the residency feed comes in.  We suppress it in cases where there is no residency information (a `code` property) in the feed.

I tested the presence and absence of the empty `<tbody>` but have not noticed any extra padding either way.

src/assets/templates/widgets/enrollment_card.html
dfkaye: @johncrossman The check for `!enrollment.terms.length` is on the cc-widget div that contains the generic "Class Enrollment" heading and the spinner div for the "No enrollments" message.  Would you prefer `enrollment.terms.length === 0` instead of `!enrollment.terms.length`?

dfkaye:
@johncrossman updated per our conversation earlier - thanks for spotting those

johncrossman:
Thanks @dfkaye

PR #5416, opened 2016-05-17
CLC-6314 Fix LTI / CalCentral clashes

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6314

This may not fix all the problems we see in production but it should fix the first bunch.

Since this will be needed in a maintenance release, I'll prepare a QA PR after we check this one out.


johncrossman:
Makes sense to me. Thanks, @raydavis

PR #5419, opened 2016-05-18
CLC-6276, prepare CalCentral's course-capture controller for Fall 2016 / EDO db

johncrossman:
* Test coverage


johncrossman:
@paulkerschen or @redconfetti - Your experience w/ EDO db queries would be useful in reviewing this PR.

app/models/edo_oracle/queries.rb
```diff
@@ -312,6 +324,38 @@ def self.has_student_history?(ldap_uid, student_terms = nil)
...
+          sec."status-code" = 'A'
+          #{section_type_restriction}
+        ORDER BY
+          sec."term-id" desc,
+          crs."subjectArea",
+          crs."catalogNumber-number",
+          crs."catalogNumber-prefix" nulls first,
+          crs."catalogNumber-suffix" nulls first,
+          sec."component-code",
+          sec."sectionNumber"
```
paulkerschen: @johncrossman I've found that, unlike our legacy Oracle, the EDO DB doesn't have the separate parts of the catalog number reliably parsed out. In #5191 I introduced logic in Edo::UserCourses::Base to do post-query sorting in Ruby, and have restricted CANONICAL_SECTION_ORDERING to a minimal set of guaranteed-to-be-there fields.

app/models/edo_oracle/queries.rb
```diff
@@ -312,6 +324,38 @@ def self.has_student_history?(ldap_uid, student_terms = nil)
...
+      safe_query <<-SQL
+        SELECT
+          #{SECTION_COLUMNS}
+        FROM
+          SISEDO.CLASSSECTIONV00_VW sec
+        #{JOIN_SECTION_TO_COURSE}
+        WHERE
+          sec."term-id" = '#{term_id}' AND
+          crs."subjectArea" = '#{department}' AND
+          crs."catalogNumber-formatted" = '#{catalog_id}' AND
```
paulkerschen: @johncrossman Just as a note, this would fail to match in those cases where the left join couldn't find a matching row in the crs view. However, I think that after a lot of work by the integration team, all the section-without-course orphans have been happily reunited with their parents, so we should now be OK here.

app/models/edo_oracle/queries.rb
paulkerschen: @johncrossman Update: after looking once more at the query, is it safe to just remove everything from ORDER BY before sec."component-code"? Seems like those values would be the same across all results.

paulkerschen:
Looks good, @johncrossman. Left one question about ordering query results.

app/models/edo_oracle/queries.rb
johncrossman: @paulkerschen I'm going to close this PR in favor of #5425. The new PR is my interpretation of your feedback. If I'm off the mark I'll return to this branch and revise again.

PR #5421, opened 2016-05-18
[QA] CLC-6314 Fix LTI / CalCentral clashes

raydavis:
QA version of https://github.com/ets-berkeley-edu/calcentral/pull/5416


johncrossman:
Once again, looks good.

PR #5425, opened 2016-05-18
SISRP-17510, Force canonical course-code sorting on EdoOracle::UserCourses (revised)

johncrossman:

This PR is a revised version of #5419 

* post-query sorting of course_sections resultset in EdoOracle::UserCourses::SelectedSections
* Test coverage


app/models/edo_oracle/user_courses/selected_sections.rb
```diff
@@ -23,6 +23,26 @@ def get_selected_sections(term_yr, term_cd, course_ids)
...
+        course_sections term_yr, term_cd, department, catalog_id, false
+      end
+
+      private
+
+      def course_sections(term_yr, term_cd, department, catalog_id, only_secondary_sections)
+        term_id = Berkeley::TermCodes.to_edo_id term_yr, term_cd
+        sections = EdoOracle::Queries.get_course_sections term_id, department, catalog_id, only_secondary_sections
+        sections.map! { |section| HashConverter.symbolize section }
+        sort_by_course_code sections
```
paulkerschen: @johncrossman, my apologies: I think I left misleading feedback on #5419 because of delayed understanding. What I meant to say is that this call to sort the sections by course code looks like a no-op, because all sections will come back with the same course code. Is that right?

So I'd think that a version of #5419 without the course-code ORDER BY columns would do just fine - unless, of course, you have other reasons for wanting to take advantage of the EdoOracle::UserCourses::SelectedSections abstraction layer.

johncrossman:
Thanks @paulkerschen  I'll revise the former PR's branch.

PR #5426, opened 2016-05-18
CLC-6276, prepare CalCentral's course-capture controller for Fall 2016 / EDO db

johncrossman:
* Test coverage

I revised PR #5419 by removing the no-op bits of `ORDER BY`


paulkerschen:
Thanks @johncrossman!

PR #5433, opened 2016-05-19
SISRP-19030 - Dev: Pull Registration Status from Hub

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-19030

I'm pretty sure this will need more work before it's ready to be pulled.  I'd also love some guidance on writing the spec for `MyRegistrations::MyRegistrations`, when possible.  

Appreciate the feedback!


app/controllers/my_registrations_controller.rb
```diff
@@ -0,0 +1,9 @@
+class MyRegistrationsController < ApplicationController
```
raydavis: There's no use of this feed by front-end code in this PR, so I'm fuzzy on the need for the API.

Do you plan to have the browser integrate the feed directly into the profile-icon's Status dropdown and the My Academics widgets? If so, then we'll want an example feed in public/dummy/json to make front-end development easier.

On the other hand, if you plan to merge this data structure into higher-level merged models for MyBadges and MyAcademics, there's no need for a new route and controller.

app/models/hub_edos/registrations.rb
```diff
@@ -5,11 +5,10 @@ class Registrations < Proxy
 
     def initialize(options = {})
       super(options)
-      @term_id = options[:term_id]
     end
 
     def url
-      "#{@settings.base_url}/#{@campus_solutions_id}/registrations?term-id=#{@term_id}"
+      "#{@settings.base_url}/#{@campus_solutions_id}/registrations"
```
raydavis: Interesting. I copied the original request format from Russ Connacher's description. So leaving the term-id parameter off returns everything CS knows about?

app/models/my_registrations/my_registrations.rb
```diff
@@ -0,0 +1,53 @@
+module MyRegistrations
+  class MyRegistrations < UserSpecificModel
+
+    include Cache::LiveUpdatesEnabled
```
raydavis: Could be wrong, but I don't think we need LiveUpdatesEnabled on this. Also, if this is cached, we don't need to cache HubEdos::Registrations any more: MyRegistrations is its only caller, so nothing is gained by the double-copy in memory. If we drop LiveUpdatesEnabled, FreshenOnWarm, and the HubEdos::Registrations cache, MyRegistrations will still need UserCacheExpiry.

I'm not going to pretend that any of this is intuitively obvious, so I'd be glad to talk about it!

app/models/my_registrations/my_registrations.rb
```diff
@@ -0,0 +1,53 @@
...
+      registrations = get_registrations
+      terms = get_terms
+      {
+        :affiliations => registrations["affiliations"],
+        :registrations => match_terms(registrations["registrations"], terms)
+      }
+    end
+
+    def get_registrations
+      registrations = HubEdos::Registrations.new(user_id: @uid).get
```
raydavis: So this new class is forward-facing only, no coverage for Summer 2016 registration status? (As with my controller question, this just shows my lack of context for the new code's use.)

app/models/my_registrations/my_registrations.rb
```diff
@@ -0,0 +1,53 @@
...
+        :future => berkeleyTerms.future ? berkeleyTerms.future.campus_solutions_id : nil,
+        :previous => berkeleyTerms.previous ? berkeleyTerms.previous.campus_solutions_id : nil,
+        :grading_in_progress => berkeleyTerms.grading_in_progress ? berkeleyTerms.grading_in_progress.campus_solutions_id : nil
+      }
+    end
+
+    private
+
+    # Match registration terms with Berkeley::Terms-defined terms.
+    def match_terms(registrations, terms)
```
raydavis: This logic, at least, seems like a good candidate for an RSpec regression test. :) You'd basically mock out various "Berkeley::Terms.fetch" results and "HubEdos::Registrations" results, and make sure you got the expected output from get_feed_internal.

raydavis:
Looking good, @davidlimcheng -- thanks! As expected, we'll probably want to chat about next steps.

app/controllers/my_registrations_controller.rb
davidlimcheng: Yes, integrating the feed directly into the dropdown/widgets is the plan!  [SISRP-17751](https://jira.berkeley.edu/browse/SISRP-17751) is a good use case of this feed in the future - although the design, which shows Summer 2016 and Fall 2016 registration status, will have to currently pull from both legacy sources and the hub, eventually we'll want to use this feed alone to handle these cases.

Will amend with a dummy JSON.

app/models/hub_edos/registrations.rb
davidlimcheng: Yep, it returns all past registered terms for that SID, plus any future terms they are [term](https://en.wikipedia.org/wiki/Semantic_satiation)-activated for.

app/models/my_registrations/my_registrations.rb
davidlimcheng: I shall take a look, and definitely shoot you a ping if any of it becomes unclear.

app/models/my_registrations/my_registrations.rb
davidlimcheng: @raydavis Yep, there will be no coverage for Summer 2016 registration status until GoLive 7, when all of the pre-Fall 2016 registration data will be converted to CS (just spoke with Danny Grieb about this).  

Would it make sense to merge legacy feeds with this one?  Honestly, I didn't consider that case.

app/models/my_registrations/my_registrations.rb
davidlimcheng: Sounds good!  I'll write up some cases in my amendment.

app/models/my_registrations/my_registrations.rb
davidlimcheng: @raydavis So everything here makes sense but I have a few questions.  I removed the mentioned classes - but had to `extend Cache::Cacheable` to `MyRegistrations` to expose it to the `expire` method.  I also had to include `Cache::CachedFeed` since `JsonAddedCacher` doesn't include a `get_feed` method - I assume I no longer need `JsonAddedCacher` because our feed is already in JSON?  

Also, should I be adding this feed into any of the `CampusSolutions::DataExpiry` definitions?  Specifically, `CampusSolutions::EnrollmentTermExpiry` seems like it would make sense.

davidlimcheng:
@raydavis Pushed amendments, and left a comment for you on the now outdated diff.

raydavis:
Thank you for the dummy feed & RSpecs, @davidlimcheng .

Sorry about leaving you to wander alone in the twisty maze of cache modules! I should at least have pointed you to [some examples](https://github.com/ets-berkeley-edu/calcentral/pull/5183/files).
I've also thrown some existing notes into a [Confluence page](https://confluence.ets.berkeley.edu/confluence/display/MYB/How+to+Cache) for slightly easier reference.

Regarding your specific questions:

1. Including ``Cache::CachedFeed`` was the right move, but you don't need to ``extend Cache::Cacheable``: the include does that for you.

2. Since the feed is called directly from the browser but will also be referred to by other server-side side, JsonAddedCacher seems like the way to go. If the feed is _only_ going to be called from the browser, you'd use JsonifiedFeed (and probably have to add some JSON parsing to your RSpec tests).

3. I agree, the new EnrollmentTermExpiry seems to match up.

raydavis:
Looks great, @davidlimcheng , just could use that last bit of cache-fiddling.

davidlimcheng:
Thanks, @raydavis !  The confluence cache doc is going to be extremely useful.  Made those last changes, and removed extra "it"'s from the rspec.  It should be good to go now.

raydavis:
Thank you for your patience, @davidlimcheng , and for finally prodding me to make that damned Confluence page. :)

PR #5435, opened 2016-05-19
SISRP-19390 - 5.6 - Advising Resources - Add WebNow Documents link

redconfetti:
https://jira.berkeley.edu/browse/SISRP-19390

NOTE: This is no longer slated for GL 5.4b. Rich Meyer has confirmed that this is being pushed to another GL (5.6?)

### DIFF
--- a/src/assets/templates/widgets/advisor/advising_links.html
+++ b/src/assets/templates/widgets/advisor/advising_links.html
@@ -50,7 +50,8 @@
   <li data-ng-if="ucAdvisingResources.ucAdvisingLinks.multiYearAcademicPlannerGeneric.url">
     <a
       data-ng-href="{{ucAdvisingResources.ucAdvisingLinks.multiYearAcademicPlannerGeneric.url}}"
-    >Multi-Year Planner</a>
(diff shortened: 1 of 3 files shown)


src/assets/templates/widgets/advisor/advising_links.html
```diff
@@ -52,4 +52,10 @@
       data-ng-href="{{ucAdvisingResources.ucAdvisingLinks.multiYearAcademicPlanner.url}}"
     >Multi-Year Academic Planner</a>
```
johncrossman: @redconfetti On the line above, I notice we don't use:
```
data-ng-bind="ucAdvisingResources.ucAdvisingLinks.multiYearAcademicPlanner.name"
```
Not a problem, unless we change the desired name in `AdvisingResources` class.

johncrossman: @redconfetti  Since this enhancement is "no longer slated for GL 5.4b" and not blocking our testing of release candidate, you should amend the commit (per above) if you agree.

raydavis:
Thanks for addressing @johncrossman 's comments, @redconfetti

PR #5444, opened 2016-05-20
[QA] SISRP-19471 Increase ReservedCodeCacheSize for Torquebox

raydavis:
https://jira.berkeley.edu/browse/SISRP-19471

Based on my reading, this is very unlikely to be the cause of our recent systemic failures, but it may decrease their frequency while we investigate the real problem.

QA first, given the schedule.

### DIFF
--- a/script/start-torquebox.sh
+++ b/script/start-torquebox.sh
@@ -25,7 +25,7 @@ echo "------------------------------------------" | $LOGIT
 echo "`date`: Starting CalCentral..." | $LOGIT
 OPTS=${CALCENTRAL_JRUBY_OPTS:="-Xcompile.invokedynamic=false -Xcext.enabled=true -J-Djruby.thread.pool.enabled=true -J-Djava.io.tmpdir=$PWD/tmp"}
 export JRUBY_OPTS=$OPTS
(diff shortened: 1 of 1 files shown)


johncrossman:
@raydavis @dkawase  http://xmlandmore.blogspot.com/2014/08/jdk-8-revisiting-reservedcodecachesize.html  says:
```
In JDK 8, it chooses the following default values for linux-x64 platforms:
    bool TieredCompilation        = true       {pd product}     
    intx CompileThreshold         = 10000      {pd product}
    uintx ReservedCodeCacheSize   = 251658240  {pd product}

When tiered compilation is enabled, two things happen:
  1. CompileThreshold is ignored
  2. A bigger code cache is needed.  Internally, HotSpot will set it to be 240 MB (i.e., 48 MB * 5)

That's why we say that people don't need to set the following options anymore in JDK8:
  -XX:ReservedCodeCacheSize=256m -XX:+TieredCompilation 
  or
  -XX:CompileThreshold=8000
```
We are running:
* java-1.8.0-oracle.x86_64 1:1.8.0.71-1jpp.1.el6_7
* Linux ets-calcentral-prod-02.ist.berkeley.edu 2.6.32-573.18.1.el6.x86_64 #1 SMP Wed Jan 6 11:20:49 EST 2016 x86_64 x86_64 x86_64 GNU/Linux

Opinion?

johncrossman:
Talked to @raydavis and I agree that having explicit config is a good thing. That said, he'll update the Jira w/ this article URL.

PR #5447, opened 2016-05-20
SISRP-19992, translate_location spec targeting locations withouth room number

johncrossman:
https://jira.berkeley.edu/browse/SISRP-19992

The bug is not fixed. This PR only demonstrates what is broken. 

See `pending do` block.


raydavis:
Thanks for explaining the reason behind the pending-from-birth code, @johncrossman (TDD with a long pause between the two Ds).

PR #5449, opened 2016-05-20
SISRP-19664 - In EdoOracle::Queries spec, :textext should be :testext

redconfetti:
https://jira.berkeley.edu/browse/SISRP-19664


johncrossman:
Thanks @redconfetti   Let's see if those pass when they hit Bamboo  :)

redconfetti:
@johncrossman When I would run the entire file under TESTEXT environment, it told me it ran 0 tests. I had to explicitly specify a line number. I wonder if we should try to pass Bamboo a failing testext test just to make sure it's doing it's job.

PR #5450, opened 2016-05-24
SISRP-19804 - Add EFT Proxy Information, Update Back-End for Actual Response

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-19804

* Removed some unnecessary class inclusions, as the incoming feed is already JSON'd (I hope this is correct)
* Removed custom error handling, since the error messages coming from EFT should be more helpful in determining any problems
* Good ol' Oski info
* I do have a security token to enable this proxy (currently just for its QA environment) - what would be the best way to communicate this to all developers?

### DIFF
--- a/app/models/eft/my_eft_enrollment.rb
+++ b/app/models/eft/my_eft_enrollment.rb
@@ -4,6 +4,7 @@ module Eft
     include Cache::JsonifiedFeed
     include Cache::FeedExceptionsHandled
     include Cache::UserCacheExpiry
+    include CampusSolutions::CampusSolutionsIdRequired
     include Proxies::HttpClient
     include Proxies::Mockable
     include ClassLogger
@@ -16,7 +17,13 @@ module Eft
     end
 
     def default_message_on_exception
-      'Failed to connect with EFT system.'
+      "Failed to connect with EFT system"
+    end
+
+    def lookup_student_id
+      if @uid
+        @campus_solutions_id = CalnetCrosswalk::ByUid.new(user_id: @uid).lookup_campus_solutions_id
+      end
     end
 
     def get_feed_internal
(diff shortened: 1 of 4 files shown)


redconfetti:
LGTM I'd like another developer to also review

raydavis:
@davidlimcheng , sorry, I guess I'm easy to confuse today...

"Removed some unnecessary class inclusions, as the incoming feed is already JSON'd (I hope this is correct)"
Does that match the results of MyEftEnrollment get_feed when you test it against the API? I would normally expect "parsed_response" to return a hash.

"Removed custom error handling, since the error messages coming from EFT should be more helpful in determining any problems"
Would it be possible to add an RSpec mock (or extra fixture file) to illustrate what front-end code should expect from an error?

redconfetti:
@davidlimcheng I'd like to go over this together.

app/models/eft/my_eft_enrollment.rb
```diff
@@ -1,22 +1,24 @@
 module Eft
   class MyEftEnrollment < UserSpecificModel
     include Cache::CachedFeed
-    include Cache::JsonifiedFeed
```
redconfetti: @davidlimcheng I think we need to assess if we need to bring this back because it cached the response AFTER converting it to JSON, which saves on the performance of the server in not having to convert it to JSON each time.

app/models/eft/my_eft_enrollment.rb
```diff
@@ -25,21 +27,18 @@ def get_feed_internal
...
 
     def get_parsed_response
-      url = "#{@settings.base_url}/student/#{@uid}"
+      lookup_student_id
+      url = "#{@settings.base_url}eft_status?studentId=#{@campus_solutions_id}"
       logger.info "get_parsed_response: Fake = #{@fake}; Making request to #{url} on behalf of user #{@uid}; cache expiration #{self.class.expires_in}"
       response = get_response(
         url,
-        basic_auth: {username: @settings.username, password: @settings.password},
-        on_error: {rescue_status: 404}
```
redconfetti: We should probably bring this back so that the HTTP client library that we use doesn't raise an exception and log it as an error, because we do expect this EFT API to return a 404 normally, not as an exceptional circumstance.

davidlimcheng:
@raydavis @redconfetti I've amended my commit with pushes that include proper caching, and an rspec test that stubs a 404 response from the API.  Huge thanks for the help, Jason thanks a bunch for the detailed explanations regarding exception handling.

raydavis:
Looks great, @davidlimcheng , let's start exercising it IRL!

PR #5451, opened 2016-05-24
SISRP-19921 - Add FPP Proxy Information, Update Back-End for Actual Response

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-19921

* Fixture is exactly the same!


johncrossman:
Thanks @davidlimcheng  Looks good.  I'll add URL to verify-API script.

PR #5453, opened 2016-05-25
SISRP-19912 - Dev: emergency contact edit - backend prep work

dfkaye:
https://jira.berkeley.edu/browse/SISRP-19912

This is behind the feature flag, 'cs_profile_emergency_contacts'

Part of the longer SISRP-10778 story.  This PR is for work required to get a student's emergency contact data from the *correct* GET endpoint, instead of the hub's edos/student feed.

The fixture and the model spec are lengthy in order to verify certain transformations are done in the ruby so we don't have to do it in the JavaScript.

The JavaScript and template are modified to use the new data structure, but the UI is left in read-only state.

### DIFF
--- a/app/controllers/campus_solutions/emergency_contacts_controller.rb
+++ b/app/controllers/campus_solutions/emergency_contacts_controller.rb
@@ -1,0 +1,9 @@
+module CampusSolutions
+  class EmergencyContactsController < CampusSolutionsController
+
+    def get
+      json_passthrough CampusSolutions::EmergencyContacts, user_id: session['user_id']
+    end
+
+  end
+end
--- a/app/models/campus_solutions/emergency_contacts.rb
+++ b/app/models/campus_solutions/emergency_contacts.rb
@@ -1,0 +1,57 @@
+module CampusSolutions
+  class EmergencyContacts < CachedProxy
+
+    include CampusSolutionsIdRequired
+    include EmergencyContactsFeatureFlagged
+
+    def initialize(options = {})
+      super options
+      initialize_mocks if @fake
+    end
+
+    def xml_filename
+      'emergency_contacts.xml'
+    end
+
+    def build_feed(response)
+      return {} unless is_feature_enabled or response.parsed_response.blank?
+      normalize_response(response).parsed_response
+    end
+
+    def normalize_response(response)
+      # Perform several checks and transforms on the response because the campus_solutions API
+      # returns inconsistently structured data.
+
+      container = response.parsed_response['STUDENTS']['STUDENT']['EMERGENCY_CONTACTS']
+      return response if container.blank?
+
+      contacts = container['EMERGENCY_CONTACT']
+
+      contacts.sort! do |left, right|
+        # Put the primary_contact with 'Y' at the head position, values being either 'Y' or 'N'
+        right['PRIMARY_CONTACT'] <=> left['PRIMARY_CONTACT']
+      end
+
+      contacts.each do |item|
+        # The campus_solutions API returns the following structure:
+        # When a contact has one phone, emergency_phone points to that phone;
+        # When a contact has more than one phone, then emergency_phone points to an array of phones;
+        # We fix that by making emergency_phone into an array, always.
+        phones = item['EMERGENCY_PHONES']
+
+        if phones
+          phones['EMERGENCY_PHONE'] = Array.wrap phones['EMERGENCY_PHONE']
+        else
+          item['EMERGENCY_PHONES'] = {"EMERGENCY_PHONE" => []}
+        end
+      end
+
+      response
+    end
+
+    def url
+      "#{@settings.base_url}/UcApiEmergencyContactGet.v1/?EMPLID=#{@campus_solutions_id}"
+    end
+
+  end
+end
--- a/spec/models/campus_solutions/emergency_contacts_spec.rb
+++ b/spec/models/campus_solutions/emergency_contacts_spec.rb
@@ -1,0 +1,49 @@
+describe CampusSolutions::EmergencyContacts do
+
+  let(:user_id) {'1024600'}
+
+  shared_examples 'a proxy that gets data' do
+    let(:proxy) { CampusSolutions::EmergencyContacts.new(fake: true, user_id: user_id) }
+    subject { proxy.get }
+    it_should_behave_like 'a simple proxy that returns errors'
+    it_behaves_like 'a proxy that properly observes the emergency contact feature flag'
+    it_behaves_like 'a proxy that got data successfully'
+    it 'returns data with the expected structure' do
+
+      # Guard against the empty feed when testing against real proxy with testext context
+      students_feed = subject[:feed][:students]
+      if students_feed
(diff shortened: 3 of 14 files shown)


app/models/campus_solutions/emergency_contacts.rb
```diff
@@ -0,0 +1,58 @@
...
+      initialize_mocks if @fake
+    end
+
+    def xml_filename
+      'emergency_contacts.xml'
+    end
+
+    def build_feed(response)
+      return {} if response.parsed_response.blank?
+      sanitize_response(response).parsed_response
```
davidlimcheng: @dfkaye Should we insert a conditional feature-flag here as well?  i.e., return {} unless Settings.features.cs_profile_emergency_contacts

dfkaye: Thanks @davidlimcheng ~ I'm still amending this PR and can make that change, too.
Are you suggesting the flag check goes in `build_feed` at the start?

dfkaye: @davidlimcheng ~ all done amending this PR ~ have at it :)

app/models/campus_solutions/emergency_contacts.rb
```diff
@@ -0,0 +1,59 @@
...
+        # When a contact has one phone, emergency_phone points to that phone;
+        # When a contact has more than one phone, then emergency_phone points to an array of phones;
+        # We fix that by making emergency_phone into an array, always.
+        phones = item['EMERGENCY_PHONES']
+
+        if phones
+          entries = phones['EMERGENCY_PHONE']
+          phones['EMERGENCY_PHONE'] = [entries] unless entries.is_a?(Array)
+        else
+          item['EMERGENCY_PHONES'] = Hash["EMERGENCY_PHONE" => []]
```
raydavis: Why not just the usual ``{'EMERGENCY_PHONE' => []}``?

app/models/campus_solutions/emergency_contacts.rb
```diff
@@ -0,0 +1,59 @@
...
+        # Put the primary_contact with 'Y' at the head position, values being either 'Y' or 'N'
+        right['PRIMARY_CONTACT'] <=> left['PRIMARY_CONTACT']
+      end
+
+      contacts.each do |item|
+        # The campus_solutions API returns the following structure:
+        # When a contact has one phone, emergency_phone points to that phone;
+        # When a contact has more than one phone, then emergency_phone points to an array of phones;
+        # We fix that by making emergency_phone into an array, always.
+        phones = item['EMERGENCY_PHONES']
```
raydavis: I think you can replace this entire block with [one line](http://api.rubyonrails.org/v4.1.8/classes/Array.html#method-c-wrap):

```
item['EMERGENCY_PHONES'] = Array.wrap item['EMERGENCY_PHONES']
```

spec/models/campus_solutions/emergency_contacts_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+  context 'mock proxy' do
+    let(:fake_proxy) { true }
+    it_should_behave_like 'a proxy that gets data'
+  end
+
+  context 'real proxy', testext: true do
+    let(:fake_proxy) { false }
+    it_should_behave_like 'a proxy that gets data'
```
raydavis: Is it really true that we can rely on the BCS test API to always return "PhoneGuy" as the fourth emergency contact for UID 1024600? Generally, I'd prefer to just check on basic formatting of the payload.

app/controllers/campus_solutions/emergency_contacts_controller.rb
```diff
@@ -0,0 +1,11 @@
+module CampusSolutions
+  class EmergencyContactsController < CampusSolutionsController
+
+    before_filter :exclude_acting_as_users
+
+    def get
+      json_passthrough CampusSolutions::EmergencyContacts, user_id: session['user_id']
```
johncrossman: @dfkaye What is the caching strategy for this feed? `UserSpecificModel` on underlying model?

app/models/campus_solutions/emergency_contacts.rb
```diff
@@ -0,0 +1,59 @@
+module CampusSolutions
+  class EmergencyContacts < Proxy
+
+    include CampusSolutionsIdRequired
+    include ProfileFeatureFlagged
```
johncrossman: What about `cs_profile_emergency_contacts` feature flag?

spec/models/campus_solutions/emergency_contacts_spec.rb
dfkaye: Thanks for all those suggestions, @raydavis.

Regarding the real proxy case, I could reformat the shared payload tests to be more "conditional" - i.e., to check whether I have any contacts in the payload first, then assert on those only if they exist - e.g, if there are contacts, then the first is the primary, and if there's a second, it should not be marked primary... then check for any phones, and so on...

Would that be acceptable?

raydavis: If you think it's worthwhile, @dfkaye , sure. We have some similar testext "put a foot forward, feel the ground; put a foot forward, feel the ground..." checks floating around the codebase, I think.

app/models/campus_solutions/emergency_contacts.rb
dfkaye: @johncrossman I'm not sure what you mean here

app/controllers/campus_solutions/emergency_contacts_controller.rb
dfkaye: @johncrossman What would you recommend for that?

app/models/campus_solutions/emergency_contacts.rb
johncrossman: @dfkaye  I assume that if feature flag `cs_profile_emergency_contacts` is false then `/api/campus_solutions/emergency_contacts` should always be an empty feed. Instead of `ProfileFeatureFlagged`:
```
module CampusSolutions
  module EmergencyContactsFeatureFlagged
    def is_feature_enabled
      Settings.features.cs_profile_emergency_contacts
    end
end
```

app/controllers/campus_solutions/emergency_contacts_controller.rb
johncrossman: @dfkaye I would recommend a change on `EmergencyContacts` model:
```
module CampusSolutions
  class EmergencyContacts < CachedProxy
    ...
```
cc: @raydavis

raydavis: Good eye, @johncrossman . Yes, anything directly called from a controller needs to be cached (or at least made up of completely cached pieces).

dfkaye:
@raydavis @davidlimcheng @redconfetti @johncrossman 
Finally!  Fixes and changes based on yesterday's feedback are finally done - thanks for your patience

raydavis:
Thank you, @dfkaye

PR #5455, opened 2016-05-25
SISRP-18674 - Update Current List of Campus Links (Financial)

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-18674

Adds a couple of new links, and updates a couple of older ones.

### DIFF
--- a/public/json/campuslinks.json
+++ b/public/json/campuslinks.json
@@ -549,9 +549,9 @@
       ]
     },
     {
-      "name": "Billing Services",
-      "description": "Billing and payment options for students and parents",
-      "url": "http://studentbilling.berkeley.edu/",
+      "name": "Billing FAQ",
+      "description": "Frequently asked questions about billing, payments, financial aid, registration, and enrollment",
+      "url": "http://studentcentral.berkeley.edu/faqs#/faq3",
       "roles": {
         "student": true,
         "applicant": true,
@@ -788,7 +788,7 @@
     },
     {
       "name": "Cal Student Central",
-      "description": "A resourceful website with answers to the most frequently asked questions by students",
+      "description": "Providing support to your questions concerning financial aid, fees and billing, payments, disbursements, registration, and enrollment",
       "url": "http://studentcentral.berkeley.edu/",
       "roles": {
         "student": true,
@@ -1567,24 +1567,6 @@
       ]
     },
     {
-      "name": "e-bills",
-      "description": "Pay your CARS bill online with either Electronic Billing (e-Bill) or Electronic Payment (e-Check)",
-      "url": "https://bearfacts.berkeley.edu/bearfacts/student/CARS/ebill.do?bfaction=accessEBill ",
-      "roles": {
-        "student": true,
-        "applicant": true,
-        "staff": false,
-        "faculty": false,
-        "exStudent": true
-      },
-      "categories": [
-        {
-          "topcategory": "Finances",
-          "subcategory": "Billing \u0026 Payments"
-        }
-      ]
-    },
-    {
       "name": "Educational Opportunity Program",
       "description": "Guidance and resources for first generation and low-income college students.",
       "url": "http://eop.berkeley.edu",
@@ -2018,10 +2000,6 @@
         "exStudent": true
       },
       "categories": [
-        {
-          "topcategory": "Finances",
-          "subcategory": "Billing \u0026 Payments"
-        },
         {
           "topcategory": "Wellness",
           "subcategory": "Campus Health Center"
@@ -2506,7 +2484,7 @@
     },
     {
       "name": "Payment Options",
-      "description": "Learn more about the options for making payment either electronically or by check to your CARS account",
+      "description": "Learn more about the options for making payments either electronically or by check to your CARS account",
       "url": "http://studentbilling.berkeley.edu/carsPaymentOptions.htm",
       "roles": {
         "student": true,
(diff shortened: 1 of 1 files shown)


davidlimcheng:
Sorry, please do not merge yet - I forgot to remove some obsolete links.

davidlimcheng:
OK, should be good to go now.

public/json/campuslinks.json
```diff
@@ -1513,6 +1513,24 @@
       ]
     },
     {
+      "name": "Delegate Access",
+      "description": "Authorize others to access your billing information",
+      "url": "https://calcentral.berkeley.edu/profile/delegate",
```
johncrossman: @davidlimcheng Can we use a relative URL here?

johncrossman:
@davidlimcheng This PR has changes that go beyond the Jira description. These wording changes are all approved, yes? Also, does this work include link to "Tax 1098-T form"?

davidlimcheng:
@johncrossman Yep, the wording changes are actually coming from the [API doc](https://docs.google.com/a/berkeley.edu/document/d/1_utuo4ADFTup9C4aF2ENAqi7yxZ0K1QKhCOd-27KnN4/edit?usp=sharing), which Flint helped to flesh out.  

As for the "view form" part of 1098-T, that requires some additional CSS that enables separate inline links  on the same list item.  This will be added in a future PR, and I've created [SISRP-20452](https://jira.berkeley.edu/browse/SISRP-20452) to make sure it doesn't fall through the cracks.  Thanks for the attention to detail!

johncrossman:
Thanks, @davidlimcheng  All makes sense. Once the make-it-relative URL change is in, I will merge.

public/json/campuslinks.json
davidlimcheng: @johncrossman After doing some digging, it looks like we won't be able to use a relative URL due to this validator in `Links:Link`: 
```validates :url, :format => URI::regexp(%w(http https))```

Unless I am missing something - is there some way we can bypass this for our relative link?

Funny enough, whoever put this line here was already iffy about its usage, lol:
```    # This URL validator is quite primitive but probably sufficient for our needs```

Comment (johncrossman) [inline public/json/campuslinks.json]: @davidlimcheng I don't know this code well – you'll need to investigate the "bypass" option. IMO, a link to prod is both annoying and dangerous. E.g., it would derail QA automation.  cc: @raydavis

Comment (davidlimcheng) [inline public/json/campuslinks.json]: @johncrossman Makes sense, let me do a little more digging.

Comment (pfarestveit) [inline public/json/campuslinks.json]: @johncrossman, @davidlimcheng - I have no problem with a static production link appearing across all environments if the link is part of a list of equally static external links.  In an automated test, I'd just verify the link text and href instead of clicking the link and verifying the resulting page title (as I do with other external links).

Comment (johncrossman) [inline public/json/campuslinks.json]: @davidlimcheng If it's okay w/ QA then I'm less concerned.

@raydavis: Do you object to an absolute CalCentral URL in campusLinks?

Comment (raydavis) [inline public/json/campuslinks.json]: The campuslinks / My Campus design isn't really set up for anything _other_ than absolute URLs, so that's just gonna be how it is, I'm afraid.

Comment (davidlimcheng) [inline public/json/campuslinks.json]: @pfarestveit @johncrossman @raydavis FYI, this will appear in the Financial Resources card, which is not a list of equally static external links.  There are dynamic links that will be inserted in to the front-end, see comments at [SISRP-19215](https://jira.berkeley.edu/browse/SISRP-19215).  

What I can do is add this "Delegate Access" link in a similar fashion, since it will only be used on the Financial Resources card.  This will most likely be the case with the "Tax 1098-T" form - since it will require two different static links on the same list item - something that cannot be done within the current `campuslinks.json` schema.

Comment (raydavis) [inline public/json/campuslinks.json]: Ah, thank you, @davidlimcheng -- that makes sense.

Comment (davidlimcheng) [inline public/json/campuslinks.json]: Cool - so I'll go ahead and remove the "Delegate Access" and "Tax 1098" objects from this PR, since they will be included in the front-end, and amend the PR.

**Comment by davidlimcheng** [comment]
@johncrossman PR has been amended to remove "Delegate Access" and "1098-T" links.  I've updated [SISRP-20452](https://jira.berkeley.edu/browse/SISRP-20452) to include this work.  I thiiiink this should be good now.

File: public/json/campuslinks.json [hunk 2]
```diff
@@ -549,9 +549,9 @@
       ]
     },
     {
-      "name": "Billing Services",
-      "description": "Billing and payment options for students and parents",
-      "url": "http://studentbilling.berkeley.edu/",
+      "name": "Billing FAQ",
+      "description": "Frequently asked questions about billing, payments, financial aid, registration, and enrollment",
+      "url": "http://studentcentral.berkeley.edu/faqs#/faq3",
```
johncrossman: @davidlimcheng Do we want `#/faq3` in this URL?

johncrossman:
Posted one last question, @davidlimcheng   Looks good

public/json/campuslinks.json
davidlimcheng: @johncrossman Yep, that was the requirement.  Although there was a bug on the studentcentral site that wouldn't allow this drawer to open - see [SISRP-14826](https://jira.berkeley.edu/browse/SISRP-14826).  The URL will still take you to the correct page, so I think it's OK to leave in.

johncrossman:
Thanks for all the work, @davidlimcheng

PR #5462, opened 2016-05-29
SISRP-18781 Re-enable Textbooks API; limit Metaspace memory consumption

raydavis:
https://jira.berkeley.edu/browse/SISRP-18781


raydavis:
We cannot upgrade jruby-openssl versions past 0.9.7 because runaway memory consumption quickly brings CalCentral to a standstill. A fix is being discussed at: https://github.com/jruby/jruby-openssl/issues/86

johncrossman:
Thanks, @raydavis   We'll avoid the mo' meta blues.

PR #5463, opened 2016-05-30
SISRP-18680 FinAid Award Comparison Enrollment details unclear

mDibyo:
https://jira.berkeley.edu/browse/SISRP-18680

### DIFF
--- a/src/assets/javascripts/angular/controllers/widgets/finaid/finaidAwardsCompareController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/finaid/finaidAwardsCompareController.js
@@ -133,12 +133,35 @@ angular.module('calcentral.controllers').controller('FinaidAwardsCompareControll
     return _.map(titles, function(title) {
       var item = itemsByTitle[title];
       if (item) {
-        return {
-          title: item.title,
-          values: item.values || [{
-            subvalue: [item.value]
-          }]
-        };
+        switch (title) {
+          case 'Enrollment': {
+            _.forEach(item.values, function(value) {
+              value.subvalue = [value.subvalue[0] + ': ' + value.subvalue[1] + ' units'];
+            });
+            return {
+              title: item.title,
+              values: item.values
+            };
+          }
+          case 'Expected Family Contribution (EFC)':
+          case 'Berkeley Parent Contribution': {
+            return {
+              title: item.title,
+              values: [{
+                isAmount: !_.isNaN(_.toNumber(item.value)),
+                subvalue: [item.value]
+              }]
+            };
+          }
+          default: {
+            return {
+              title: item.title,
+              values: item.values || [{
+                subvalue: [item.value]
+              }]
+            };
+          }
+        }
       }
       return {
         title: title,
--- a/src/assets/templates/widgets/finaid_awards_compare_summary.html
+++ b/src/assets/templates/widgets/finaid_awards_compare_summary.html
@@ -7,7 +7,13 @@
     <div data-ng-if="item.values">
       <div data-ng-repeat="value in item.values track by $index" class="cc-text-right">
         <div>
-          <span data-ng-bind="'&nbsp;' + value.subvalue.join(' &bull; ')"></span>
+          <span data-ng-if="!value.isAmount">
+            <span data-ng-repeat="subvalue in value.subvalue">
+              <span data-ng-bind="subvalue"></span>
+              <span data-ng-if="!$last"> &bull; </span>
+            </span>
+          </span>
+          <span data-ng-if="!!value.isAmount" data-cc-amount-directive="value.subvalue[0]" data-cc-amount-directive-color="false"></span>
           <span data-ng-if="$first" data-cc-amount-change-symbol-directive="item.change"></span>
           <span data-ng-if="!$first" data-cc-amount-change-symbol-directive="changeTags.blank"></span>
         </div>


src/assets/javascripts/angular/controllers/widgets/finaid/finaidAwardsCompareController.js
```diff
@@ -133,12 +133,35 @@ angular.module('calcentral.controllers').controller('FinaidAwardsCompareControll
...
+          case 'Enrollment': {
+            _.forEach(item.values, function(value) {
+              value.subvalue = [value.subvalue[0] + ': ' + value.subvalue[1] + ' units'];
+            });
+            return {
+              title: item.title,
+              values: item.values
+            };
+          }
+          case 'Expected Family Contribution (EFC)':
```
davidlimcheng: @mDibyo Is there supposed to be a return in here for EFC?  If not, is there a reason why we don't let it fall under the default case?

mDibyo: @davidlimcheng The EFC case falls through into the next case (BPC). So lines 148-154 is executed. Essentially both EFC and BPC have the same procedure.

davidlimcheng: @mDibyo Ahhh I see, that makes sense.  Thanks, that is very good to know!

src/assets/templates/widgets/finaid_awards_compare_summary.html
```diff
@@ -7,7 +7,8 @@ <h4 class="show-for-small-only cc-text-light" data-ng-bind="type.title"></h4>
     <div data-ng-if="item.values">
       <div data-ng-repeat="value in item.values track by $index" class="cc-text-right">
         <div>
-          <span data-ng-bind="'&nbsp;' + value.subvalue.join(' &bull; ')"></span>
+          <span data-ng-if="!value.isAmount" data-ng-bind="'&nbsp;' + value.subvalue.join(' &bull; ')"></span>
```
redconfetti: @mDibyo - Did we test this functionality? I was under the impression that expressions aren't evaluated directly via ngBind, and that instead [ngBindTemplate](https://docs.angularjs.org/api/ng/directive/ngBindTemplate) was necessary to pull off something like this.

Perhaps it's needed only for bringing in more than one scope variable, and my thinking is wrong.

mDibyo: @redconfetti I have tested it. ngBind seems to do the job Check the JIRA for a screenshot.

I think the following illustrates the difference well:
<span ng-bind="firstName + ' ' + lastName"></span>
<span ng-bind-template="{{ firstName }} {{ lastName }}"></span>

redconfetti: Okay. I guess if it were more complicated and thus less readable, I'd advise using ngBindTemplate.

redconfetti: @mDibyo Actually, I checked our code and I see that we're not using ng-bind anywhere to provide more than one value from the scope. I think we should use ng-bind-template here for code consistency.

davidlimcheng: @mDibyo @redconfetti Although both work, I'd agree with Jason on the use of `ngBindTemplate`, just to keep consistent with the rest of our codebase.  I don't think there are many instances in our  code of `ngBind` being used to show more than one variable or character, whereas `ngBindTemplate` is used throughout in cases like this.

davidlimcheng: @mDibyo Will it be OK to amend this PR with the proposed change?

mDibyo: @redconfetti @davidlimcheng I have addressed this. I am sorry for the delay.

What you guys said makes sense. We want to be consistent. Directly making the change gave the following piece of code:
```html
<span data-ng-if="!value.isAmount" data-ng-bind-template=" {{ value.subvalue.join(' &bull; ') }}"></span>
```
Here, the number of values being bound is still technically one. Since there can be a varying number of elements in the array `value.subvalue`, I couldn't think of a nice template to bind. I think this was the reason the previous version of the code had been written with the inconsistent ng-bind.

So I instead switched to what you see currently:
```html
<span data-ng-if="!value.isAmount">
  <span data-ng-repeat="subvalue in value.subvalue">
    <span data-ng-bind="subvalue"></span>
    <span data-ng-if="!$last"> &bull; </span>
  </span>
</span>
```
I used ng-bind instead of ng-bind-template since now there's only one value. This seems to be a reasonably good solution. Let me know what you guys think.

mDibyo:
@davidlimcheng @redconfetti I have replied to our comment thread on the finaid_awards_comparesummary.html file, to provide context for the new changes.

src/assets/templates/widgets/finaid_awards_compare_summary.html
davidlimcheng: @mDibyo I think this is a great solution.  @redconfetti thoughts?

davidlimcheng:
This looks good to me.  Thanks for the work @mDibyo!  One failing Travis looks like false positive.

src/assets/templates/widgets/finaid_awards_compare_summary.html
redconfetti: It looks good to me. In fact, using `data-ng-bind` alone with the join he did would have been a good solution also. This solution is elegant as well.

PR #5464, opened 2016-05-30
SISRP-20522 Add Key to compare symbols to Compare Packages card

mDibyo:
https://jira.berkeley.edu/browse/SISRP-20522

### DIFF
--- a/src/assets/templates/widgets/finaid_awards_compare_current_prior.html
+++ b/src/assets/templates/widgets/finaid_awards_compare_current_prior.html
@@ -9,6 +9,22 @@
   </div>
 </div>
 
+<div class="cc-widget-text">
+  <h3>Key</h3>
+  <span class="cc-finaid-compare-awards-key-item">
+    <div data-cc-amount-change-symbol-directive="changeTags.deleted"></div>
+    Only in Prior
+  </span>
+  <span class="cc-finaid-compare-awards-key-item">
+    <div data-cc-amount-change-symbol-directive="changeTags.added"></div>
+    Only in Current
+  </span>
+  <span class="cc-finaid-compare-awards-key-item">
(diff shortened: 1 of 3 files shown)


src/assets/templates/widgets/finaid_awards_compare_current_prior.html
```diff
@@ -9,6 +9,18 @@
   </div>
 </div>
 
+<div class="cc-widget-text">
+  <div>
```
redconfetti: @mDibyo - Is this wrapping DIV necessary?

src/assets/templates/widgets/finaid_awards_compare_current_prior.html
```diff
@@ -9,6 +9,18 @@
...
+    <h3>Key</h3>
+  </div>
+  <div data-cc-amount-change-symbol-directive="changeTags.deleted"></div>
+  &nbsp;Only in Prior&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
```
redconfetti: I would wrap each of these in an inline element like `span`, apply `display:inline-block;` so that you can use a `margin` style to apply the spacing between these items in the Key. The use of `&nbsp;` doesn't seem right.

mDibyo: Yeah, I was thinking the same thing actually. Your solution makes sense. Done!

src/assets/templates/widgets/finaid_awards_compare_current_prior.html
mDibyo: @redconfetti No, it isn't actually. You are right.
I have made the change here, and also in finaid_awards_compare_section.html .

PR #5479, opened 2016-06-02
SISRP-20587 - 5.6 - My Academics Course Page (Instructor) - Apply 12 column card width

redconfetti:
https://jira.berkeley.edu/browse/SISRP-20664


johncrossman:
I see no prob here. Thanks, @redconfetti

PR #5485, opened 2016-06-06
SISRP-20508 Do not show empty Final Exam Schedule card

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-20508

Either students or instructors might need to see the final exam schedule, but only if the schedule contains something.


johncrossman:
Exam schedule for all. Thanks @paulkerschen

PR #5487, opened 2016-06-07
SISRP-19252 Filter instructing classes from My Classes under advisor view-as

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-19252


johncrossman:
+1

PR #5491, opened 2016-06-07
SISRP-19189 - Reorder Links as shown in Design for Financial Resources

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-19189

* Had to get rid of the "Registration Fees" campus links JSON object, since it was actually a duplicate of "Tuition and Fees" (just changed the name)
* Uses `data-ng-repeat-start` and `data-ng-repeat-end` to insert custom list items into an `ngRepeat` iterating through an array at a certain index
* Found out that the HTML for the EFT status items are not matching up with the design, created [SISRP-21011](https://jira.berkeley.edu/browse/SISRP-21011) to fix that
* Changed `campuslinks` to `campusLinks` just to keep casing consistent with the rest of the variables

### DIFF
--- a/src/assets/templates/widgets/finances_links.html
+++ b/src/assets/templates/widgets/finances_links.html
@@ -3,13 +3,11 @@
     <h2>Financial Resources</h2>
   </div>
   <div data-cc-spinner-directive="isLoading">
-    <div class="cc-list-link-container" data-ng-repeat="subcategory in campuslinks.data.subcategories">
+    <div class="cc-list-link-container" data-ng-repeat="subcategory in campusLinks.data.subcategories">
       <h3>{{subcategory}}</h3>
       <ul class="cc-list-links">
-        <li data-ng-repeat="link in campuslinks.data.links | orderBy:'name' | campusLinksSubcategoryFilter:subcategory">
-          <a data-ng-href="{{link.url}}"
-            data-ng-click="api.analytics.trackExternalLink('Campus links', subcategory, link.url)"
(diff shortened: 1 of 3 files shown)


davidlimcheng:
Just realized I forgot to make a last minute change - will amend now.

davidlimcheng:
Voila, good to go.

src/assets/templates/widgets/finances_links.html
```diff
@@ -30,18 +28,20 @@
             <strong>Manage Account</strong>
           </a>
         </li>
-        <li data-ng-if="subcategory === 'Billing & Payments'">
+        <li data-ng-repeat-start="link in campusLinks.data.links | campusLinksSubcategoryFilter:subcategory">
```
redconfetti: So the [ngRepeat start and end](https://docs.angularjs.org/api/ng/directive/ngRepeat#special-repeat-start-and-end-points) are needed because you can't wrap `LI` elements in a DIV, right?

davidlimcheng: @redconfetti Not necessarily - it's needed to include both the FPP and 1098-T Tax Form `<li>` elements underneath it within the scope of the `ngRepeat`.  By including these elements within the scope of the ngRepeat directive, I can specify something like ```<li data-ng-if="$index === 1">``` to insert an element into the the `campuslinks.data.links` array that `ngRepeat` is iterating over, after `campuslinks.data.links[1]`.

PR #5492, opened 2016-06-08
SISRP-21011 - Fix EFT Status HTML and use data-ng-bind

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-21011

* Fixed nested text (that aren't links), and used `data-ng-bind` for scope variables.

### DIFF
--- a/src/assets/stylesheets/_lists.scss
+++ b/src/assets/stylesheets/_lists.scss
@@ -30,6 +30,14 @@
       content: '\2013';
     }
   }
+  .cc-list-links-text {
+    margin-left: 16px;
+    &:before {
+      float: left;
+      margin-top: -2px;
+      padding-right: 10px;
+    }
+  }
 }
 .cc-list-link-container {
   padding: 0 15px;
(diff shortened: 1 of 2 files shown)


src/assets/stylesheets/_lists.scss
```diff
@@ -30,6 +30,14 @@
       content: '\2013';
     }
   }
+  .cc-list-links-text {
+    margin-left: 16px;
+    &:before {
+      float: left;
+      margin-top: -2px;
+      padding-right: 10px;
```
davidlimcheng: `padding-right: 10px` is due to the inclusion of `margin-left: -16px` under .cc-list-links > a > &:before. This essentially adds 2px to `padding-right`. I'm not sure why this was done this way, but since this CSS is used elsewhere on CalCentral, I figured I wouldn't mess with the existing code unless the changes can be verified everywhere else.

paulkerschen: Thanks for the explanation @davidlimcheng. We should probably create a follow-up ticket to see what's motivating those negative margins (I'm surprised they got past Christian). Do you have access to CLC JIRAs?

davidlimcheng: Thanks @paulkerschen - created!  [CLC-6349](https://jira.ets.berkeley.edu/jira/browse/CLC-6349)

PR #5494, opened 2016-06-08
SISRP-21079 Move legacy Calendar::Queries to CampusOracle::Calendar

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-21097

New queries will live in a parallel EdoOracle::Calendar class.

### DIFF
--- a/app/models/calendar/preprocessor.rb
+++ b/app/models/calendar/preprocessor.rb
@@ -12,7 +12,7 @@ module Calendar
 
     def get_entries
       entries = []
-      courses = Calendar::Queries.get_all_courses
+      courses = CampusOracle::Calendar.get_all_courses
       courses.each do |course|
         this_term_slug = Berkeley::TermCodes.to_slug(course['term_yr'], course['term_cd'])
         term = Berkeley::Terms.fetch.campus[this_term_slug]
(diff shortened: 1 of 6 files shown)


redconfetti:
@paulkerschen I really like this move for the Queries class under CampusOracle. There's no place like home... there's no place......

I wonder if we should have done this with [Oec::Queries](https://github.com/ets-berkeley-edu/calcentral/blob/7f31a3d4/app/models/oec/queries.rb) as well. The OEC and Calendar query classes were those two outliers that hit databases that weren't grouped under CampusOracle.

@johncrossman What do you think?

paulkerschen:
@redconfetti The OEC retrofit is scoped for next month, and I think that would be a great time to do a corresponding move on that queries class.

johncrossman:
Looks good, @paulkerschen  I agree with comments above in terms of consistency. However, OEC is a standalone app and `Calendar` is a CalCentral feature. That difference might argue in favor of keeping `Oec::Queries` separate.  cc: @redconfetti

johncrossman:
On second thought, I think it far better to move the OEC queries in the same way. The alternative might involve two directories under `/oec`: `edo_oracle` and `campus_oracle`. Shudder.

redconfetti:
@johncrossman Yeah. I think EdoOracle::Oec or even EdoOracle::Oec::Queries would be a good way of making them associated with the DB source, but also separate for the application they serve (as they aren't shared).

PR #5496, opened 2016-06-09
SISRP-20940 - Enroll FPP CS page looses calcentral header

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-20940

* Thanks @redconfetti for pointing me to the directive to use for these CS links.
* This will most probably cause merge conflicts with #5492, if that PR is merged first.  I'll rebase if so.


paulkerschen:
Thanks @davidlimcheng!

PR #5499, opened 2016-06-10
SISRP-19992 Correctly parse EDO DB locations without room numbers

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-19992

And un-pend the spec.

### DIFF
--- a/app/models/edo_oracle/course_sections.rb
+++ b/app/models/edo_oracle/course_sections.rb
@@ -48,7 +48,12 @@ module EdoOracle
       if meeting['location'] == 'Requested General Assignment'
         building_name = 'Room not yet assigned'
       else
-        building_name, room_number = meeting['location'].rpartition(/\s+/).reject &:blank?
+        # See spec for behavior of this regex.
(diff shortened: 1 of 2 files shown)


redconfetti:
LGTM, @johncrossman ?

app/models/edo_oracle/course_sections.rb
```diff
@@ -48,7 +48,12 @@ def translate_location(meeting)
       if meeting['location'] == 'Requested General Assignment'
         building_name = 'Room not yet assigned'
       else
-        building_name, room_number = meeting['location'].rpartition(/\s+/).reject &:blank?
+        # See spec for behavior of this regex.
+        if (partitioned = meeting['location'].match /\A(.*)\s+(\d[^\s]*)\Z/)
```
johncrossman: @paulkerschen Not sure where we draw the line on edge cases but this pattern does not match locations such as `Hearst Field Annex C57`

paulkerschen: @johncrossman I will amend and add that location to the spec. :)

johncrossman:
Thanks, @paulkerschen  Waiting on T

johncrossman:
All we need is one green Travis

PR #5500, opened 2016-06-10
SISRP-20587 - Course Page Enrollments section - Add 'Email to List' option

redconfetti:
https://jira.berkeley.edu/browse/SISRP-20670

### DIFF
--- a/app/models/edo_oracle/queries.rb
+++ b/app/models/edo_oracle/queries.rb
@@ -291,11 +291,15 @@ module EdoOracle
           enroll."WAITLISTPOSITION" AS waitlist_position,
           enroll."UNITS_TAKEN" AS units,
           TRIM(enroll."GRADING_BASIS_CODE") AS grading_basis,
-          plan."ACADPLAN_DESCR" AS major
+          plan."ACADPLAN_DESCR" AS major,
+          stdgroup."STDNT_GROUP" as ugrad_terms_in_attendance_group
         FROM SISEDO.ENROLLMENTV00_VW enroll
         LEFT OUTER JOIN
           SISEDO.STUDENT_PLAN_CC_V00_VW plan ON enroll."STUDENT_ID" = plan."STUDENT_ID" AND
           plan."ACADPLAN_TYPE_CODE" = 'MAJ'
+        LEFT OUTER JOIN
+          SISEDO.STUDENT_GROUPV00_VW stdgroup ON enroll."STUDENT_ID" = stdgroup."STUDENT_ID" AND
+          stdgroup."STDNT_GROUP" IN ('R1TA', 'R2TA', 'R3TA', 'R4TA', 'R5TA', 'R6TA', 'R7TA', 'R8TA')
         WHERE
           enroll."CLASS_SECTION_ID" = '#{section_id}'
           AND enroll."TERM_ID" = '#{term_id}'
--- a/app/models/edo_oracle/queries.rb
+++ b/app/models/edo_oracle/queries.rb
@@ -281,6 +281,28 @@ module EdoOracle
       SQL
     end
 
+    # Extended version of #get_enrolled_students used for rosters
+    def self.get_rosters(section_id, term_id)
+      safe_query <<-SQL
+        SELECT DISTINCT
+          enroll."CAMPUS_UID" AS ldap_uid,
+          enroll."STUDENT_ID" AS student_id,
+          enroll."STDNT_ENRL_STATUS_CODE" AS enroll_status,
+          enroll."WAITLISTPOSITION" AS waitlist_position,
+          enroll."UNITS_TAKEN" AS units,
+          TRIM(enroll."GRADING_BASIS_CODE") AS grading_basis,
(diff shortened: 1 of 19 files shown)


redconfetti:
Oooh, made PR too soon. I should disable the button when no users are present in the list.

src/assets/javascripts/angular/directives/academicsClassInfoEnrollmentDirective.js
```diff
@@ -7,8 +7,56 @@ angular.module('calcentral.directives').directive('ccAcademicsClassInfoEnrollmen
...
+        }
+
+        var students = [];
+        angular.forEach(scope.students, function(student) {
+          if (isStudentInSection(student)) {
+            this.push(student);
+          }
+        }, students);
+
+        return students;
```
paulkerschen: @redconfetti Now that we've invited Lodash to the party, would `_.filter` meet your needs here?
https://lodash.com/docs#filter

src/assets/javascripts/angular/directives/academicsClassInfoEnrollmentDirective.js
```diff
@@ -7,8 +7,56 @@ angular.module('calcentral.directives').directive('ccAcademicsClassInfoEnrollmen
...
+       * Returns list of student email addresses for the currently selected section
+       * @returns {string}
+       */
+      scope.studentsInSectionEmailList = function() {
+        var students = studentsInSelectedSection();
+        if (students) {
+          var studentEmails = [];
+          angular.forEach(students, function(student) {
+            this.push(student.email);
+          }, studentEmails);
```
paulkerschen: And `_.map` here?
https://lodash.com/docs#map

src/assets/templates/oec.html
```diff
@@ -34,13 +30,12 @@ <h1 class="cc-heading-page-title">OEC Control Panel</h1>
...
-                      data-ng-options="department.code as department.name for department in oecDepartments">
-                <option value="">All participating departments</option>
-              </select>
-            </div>
+            <select id="cc-page-oec-department"
+                  class="cc-inline-select"
+                  data-ng-model="taskParameters.options.departmentCode"
+                  data-ng-options="department.code as department.name for department in oecDepartments">
+              <option value="">All participating departments</option>
+            </select>
```
paulkerschen: Thanks for extra cleanup on the selects in here, @redconfetti

app/models/edo_oracle/queries.rb
```diff
@@ -279,6 +279,40 @@ def self.get_enrolled_students(section_id, term_id)
...
+        if unique_rosters[uid].blank?
+          current_roster['majors'] = [current_roster.delete("major")]
+          unique_rosters[uid] = current_roster
+        else
+          unique_rosters[uid]['majors'] << current_roster['major']
+          unique_rosters[uid]['majors'].sort!
+        end
+      end
+      unique_rosters.values
+    end
```
paulkerschen: @redconfetti This logic looks good. Our pattern in the past, though, has been to use the `EdoOracle::Queries` and `CampusOracle::Queries` classes for raw SQL and not much else. We tend to keep higher-order merging and translation logic in separate classes.

In this case, since `Rosters::Common` is the only consumer of this query, I think it might be cleaner to include this logic  as a private method in that class. What do you think?

src/assets/javascripts/angular/directives/academicsClassInfoEnrollmentDirective.js
redconfetti: @paulkerschen Good call. That's actually refactoring quite beautifully.

app/models/edo_oracle/queries.rb
redconfetti: @paulkerschen I've moved the logic over to Rosters::Common, and actually found that we're already performing a loop anyway to merge the acquired data with basic attributes, so I am simply just using that loop.

src/assets/templates/directives/academics_class_info_enrollment.html
```diff
@@ -1,9 +1,19 @@
 <div class="cc-widget">
   <div class="cc-widget-title">
     <h2 class="cc-left" data-ng-bind="title"></h2>
-    <div class="cc-select cc-right cc-academics-class-enrollment-select">
+    <div class="cc-academics-class-enrollment-options cc-right">
+      <a class="cc-button"
+        data-ng-href="http://mail.google.com/mail/u/0/?view=cm&fs=1&tf=1&source=mailto&bcc={{studentsInSectionEmailList()}}"
+        data-ng-disabled="studentsInSectionEmailList().length == 0">
```
paulkerschen: @redconfetti Should this be `===`, or did you want a type-converting comparison?

paulkerschen:
Looks good to me. One minor front-end question.

src/assets/templates/directives/academics_class_info_enrollment.html
redconfetti: @paulkerschen No, that's a good catch.

PR #5501, opened 2016-06-10
SISRP-20266 - Remove currentTerm displaying on Billing Summary

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-20266

* I wanted to actually get your opinions on this PR before merging
* This goes in-hand with the "Prior to Fall 2016" subheading shown on the CARS Summary card
* The requirement is to show a "Fall 2016" subheading during the transition period.  Once CARS is retired, we'll want to remove the subheading - the problem is, that date is currently a moving target.
* This PR makes the subheading conditional on the "Current Term" that is being returned by the API.  Since CS "Current Term" is kind of funky, I've included "Summer 2016" as part of the conditional
* I think it'd be a good idea to create a new CalShare task (as opposed to a JIRA ticket), which acts as the source of record in SIS, to remove this code for GL7.  This ensures it won't fall through the cracks.
* Thoughts?

### DIFF
--- a/src/assets/javascripts/angular/controllers/pages/billingDetailsController.js
+++ b/src/assets/javascripts/angular/controllers/pages/billingDetailsController.js
@@ -5,12 +5,16 @@ var angular = require('angular');
 angular.module('calcentral.controllers').controller('BillingDetailsController', function(apiService, financesFactory, $scope) {
   apiService.util.setTitle('My Finances');
   $scope.billingTerm = {
-    currentTerm: '',
+    // We want to show this text only during the transition period, so we'll remove this for GL7.
+    fallText: null,
     isLoading: true
   };
 
   var getCurrentTerm = function(data) {
-    $scope.billingTerm.currentTerm = data.data.feed.summary.currentTerm;
+    // Toggles the "Fall 2016" text needed for transition period, remove this for GL7.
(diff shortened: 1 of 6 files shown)


redconfetti:
I don't see any problem with this, especially given that it's temporary.

@davidlimcheng Do we create CalShare tasks in addition to JIRA's sometimes for important things like this that require followup?

paulkerschen:
@davidlimcheng, looks like a workable solution to a fussy requirement.

One question: is it necessary for `fallSubheader` to be an object containing both a string and a boolean? Or would it be simpler just to set `fallSubheader` to a string value when needed, and have the template show it only when present?

davidlimcheng:
@redconfetti Not usually - normally I'd just create a JIRA ticket.  But since this is so far ahead in the future, I want to make sure it doesn't slip through the sands of time.  I'll have to get the OK from Vannessa and Ben first, though.

@paulkerschen That's a great suggestion!  Will apply & amend.

davidlimcheng:
@redconfetti @paulkerschen Went ahead and made the changes to all of the areas that used the "currentTerm" from the API, with the suggested "truthy string value".  Hope this looks good, thanks for the input!

paulkerschen:
Thanks @davidlimcheng!

PR #5503, opened 2016-06-12
SISRP-21140 - Multiple phones for Emergency Contact should limit to one of each available phone type

dfkaye:
https://jira.berkeley.edu/browse/SISRP-21140

### DIFF
--- a/src/assets/templates/widgets/profile/emergency_phone_fields.html
+++ b/src/assets/templates/widgets/profile/emergency_phone_fields.html
@@ -4,13 +4,24 @@
       <strong>Phone Type</strong>
     </label>
   </div>
-  <div class="cc-select">
-    <select id="cc_page_widget_profile_emergency_phone_type" data-ng-model="phone.phoneType" data-ng-options="type.fieldvalue as type.xlatlongname for type in emergencyPhone.phoneTypes" data-ng-if="phone || !emergencyPhone.isAdding">
-      <option value="" disabled selected>Choose&hellip;</option>
-    </select>
-    <select id="cc_page_widget_profile_emergency_phone_type" data-ng-model="emergencyPhone.currentObject.data.phoneType" data-ng-options="type.fieldvalue as type.xlatlongname for type in emergencyPhone.phoneTypes" data-ng-if="!phone && emergencyPhone.isAdding">
-      <option value="" disabled selected>Choose&hellip;</option>
-    </select>
+
+  <div class="cc-page-widget-profile-section-emergency-phone-type-select">
+    <div class="cc-page-widget-profile-section-emergency-phone-type-editing" data-ng-if="phone || !emergencyPhone.isAdding">
+      <span data-ng-bind="phone.phoneTypeDescr" data-ng-if="phone.availablePhoneTypes.length === 1"></span>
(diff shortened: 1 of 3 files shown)


src/assets/templates/widgets/profile/emergency_phone_fields.html
```diff
@@ -4,13 +4,24 @@
...
+      <span data-ng-bind="phone.phoneTypeDescr" data-ng-if="phone.availablePhoneTypes.length === 1"></span>
+      <div class="cc-select" data-ng-if="phone.availablePhoneTypes.length > 1">
+        <select id="cc_page_widget_profile_emergency_phone_type" data-ng-model="phone.phoneType" data-ng-options="type.fieldvalue as type.xlatlongname for type in phone.availablePhoneTypes">
+          <option value="" disabled selected>Choose&hellip;</option>
+        </select>
+      </div>
+    </div>
+    <div class="cc-page-widget-profile-section-emergency-phone-type-adding" data-ng-if="!phone && emergencyPhone.isAdding">
+      <span data-ng-bind="emergencyPhone.currentObject.data.phoneTypeDescr" data-ng-if="emergencyPhone.currentObject.data.availablePhoneTypes.length === 1"></span>
+      <div class="cc-select" data-ng-if="emergencyPhone.currentObject.data.availablePhoneTypes.length > 1">
```
paulkerschen: @redconfetti Does your recent work with the `cc-select` class establish a new pattern for us to follow going forward?

paulkerschen:
Thanks for sorting out some messy requirements, @dfkaye. Code looks good to me; I've left one question on @redconfetti's recent changes to select styling.

src/assets/templates/widgets/profile/emergency_phone_fields.html
redconfetti: Yes. We do want to start using `cc-inline-select` instead of `cc-select`. I tried to apply it to the forms in the profile section, but I found that the height applied to the drop-down menu didn't match the other form fields, and I wasn't happy with the vertical spacing. 

This inconsistency is going to require some more investigation. I hate to half-implement things like this, but I just don't have the time to replace all select menus throughout all of CalCentral.

If David could use `cc-inline-select`, and make the necessary changes to adopt it, that would be very nice. I'm not sure why other inline text inputs aren't the same height as the ones used in this section of the site.

dfkaye: `cc-inline-select` comes through as unstyled - I don't see that anywhere in the .scss files

redconfetti: @paulkerschen - The 'cc-inline-select' style isn't merged in from #5500 yet, and even once it is I think this would provide complications for this task. I think we should create a Jira to complete the adoption of cc-inline-select throughout the application scheduled for another sprint.

paulkerschen: OK, so no merge blocker for this PR. Thanks, @redconfetti, feel free to create that JIRA.

redconfetti: Done with [SISRP-21290](https://jira.berkeley.edu/browse/SISRP-21290)

PR #5507, opened 2016-06-14
SISRP-21045 - Billing Activity : If Due Date is missing then it shows abrupt line ending

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-21045

* Only show due date if the API is bringing over `itemDueFlag !== null`


johncrossman:
Thx @davidlimcheng

PR #5509, opened 2016-06-14
SISRP-19922 - Remove 'GTE Forms' link from Advising Resources card

redconfetti:
https://jira.berkeley.edu/browse/SISRP-19922

### DIFF
--- a/src/assets/javascripts/angular/factories/advisingFactory.js
+++ b/src/assets/javascripts/angular/factories/advisingFactory.js
@@ -6,14 +6,14 @@ var angular = require('angular');
  * Advising Factory
  */
 angular.module('calcentral.factories').factory('advisingFactory', function(apiService) {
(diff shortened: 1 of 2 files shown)


src/assets/javascripts/angular/factories/advisingFactory.js
```diff
@@ -6,14 +6,14 @@ var angular = require('angular');
  * Advising Factory
  */
 angular.module('calcentral.factories').factory('advisingFactory', function(apiService) {
```
redconfetti: I'm simply updating the variable names in the commented out dummy paths, and placing them beneath each actual definition, so that they can be easily uncommented or commented during development.

johncrossman: Thank you for this cleanup, @redconfetti

PR #5510, opened 2016-06-14
SISRP-20698, all users get links to eForms on My Campus

johncrossman:
https://jira.berkeley.edu/browse/SISRP-20698

Rich confirmed that we these links should be available to all users.


paulkerschen:
Formidable, @johncrossman.

PR #5511, opened 2016-06-15
SISRP-19374 - 5.6 - Academics - Status, Holds, Blocks - Residency Message Catalog

dfkaye:
https://jira.berkeley.edu/browse/SISRP-19374

PR for the Ruby support.  
UI will be in separate PR.

### DIFF
--- a/app/models/berkeley/residency_message_code.rb
+++ b/app/models/berkeley/residency_message_code.rb
@@ -1,0 +1,70 @@
+module Berkeley
+  module ResidencyMessageCode
+    extend self
+
+    def residency_message_code(slr_status, residency_status, tuition_exception)
+      slr_status.strip! if slr_status
+      residency_status.strip! if residency_status
+      tuition_exception.strip! if tuition_exception
+
+      case slr_status
+        when 'N' # Not submitted
+          case residency_status
+            when '', 'PEND' # none or Pending
+              '2000'
+            else
+              ''
+          end
+        when 'A' # Awaiting
+          case residency_status
+            when '', 'PEND'
+              '2001'
+            else
+              ''
+          end
+        when 'R' # Received
+          case residency_status
+            when '', 'PEND'
+              '2002'
+            else
+              ''
+          end
+        when 'D', 'Y' # Completed
+          case residency_status
+            when 'NON' # Non-resident
+              case tuition_exception
+                when ''
+                  '2004'
+                when 'RA', 'RV' # AB540, Veteran/Dependent of Veteran
+                  '2005'
+                when 'RP' # Dependent Deceased Law Enf/FF
+                  '2007'
+                when 'RD', 'RDRA', 'RE', 'RF', 'RL' # Attorney Waiver, DACA Students Not CA or AB540, UC Employee Outside of CA, Faculty Spouse or Dependent, Dep of UC Emp Outside CA
+                  '2009'
+                when 'RM' # Military Member/Dep/Spouse
+                  '2010'
+                else
+                  ''
+              end
+            when 'RES' # Resident
+              case tuition_exception
+                when ''
+                  '2003'
+                when 'R8', 'RB', 'RP' # Credentialed Public School Emp, Grad Bureau Indian Affairs Sch, Dependent Deceased Law Enf/FF
+                  '2006'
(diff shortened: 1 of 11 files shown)


app/models/berkeley/residency_message_code.rb
```diff
@@ -0,0 +1,70 @@
+module Berkeley
+  module ResidencyMessageCode
+    extend self
+
+    def residency_message_code(slr_status , residency_status, tuition_exception)
```
johncrossman: @dfkaye  stray space before the comma

app/models/campus_solutions/residency_message.rb
```diff
@@ -0,0 +1,29 @@
+module CampusSolutions
+  class ResidencyMessage < GlobalCachedProxy
+
+    def initialize(options = {})
+      super options
+
+      puts "-----------------------------"
+      pp options
+      puts "-----------------------------"
```
johncrossman: @dfkaye These lines (from debugging) should be removed.

app/models/campus_solutions/residency_message.rb
```diff
@@ -0,0 +1,29 @@
+module CampusSolutions
+  class ResidencyMessage < GlobalCachedProxy
```
johncrossman: @dfkaye should we cache per `@message_nbr`?  I.e., add:
```
    def instance_key
      @message_nbr
    end
```

app/models/hub_edos/demographics.rb
```diff
@@ -23,5 +23,29 @@ def instance_key
       @instance_key
     end
 
+    def build_feed(response)
+      response = super(response)
+      return response unless !response['student']['residency'].blank?
```
johncrossman: @dfkaye re: the trailing conditional. Is the following more readable?
```
return response if response['student']['residency'].blank?
```
Also, are we guaranteed to get `response['student']`? If not, then the statement above will need to do nil check.

app/models/hub_edos/demographics.rb
```diff
@@ -23,5 +23,29 @@ def instance_key
...
+
+      # Adds residency.message.code to the response
+      slr_status = ''
+      residency_status = ''
+      tuition_exception = ''
+      slr = response['student']['residency']['statementOfLegalResidenceStatus'] || ''
+      if slr
+        slr_status = slr['code']
+      end
```
johncrossman: @dfkaye if the var `slr` gets empty string (ie, `''`) then the if expression will be true and then `slr['code']` will crash. You might want to extract a method here. I.e.,
```
slr_status = find_residence_status(response)
```
```

Comment (johncrossman) [inline app/models/hub_edos/demographics.rb]: Correction: `slr['code']` will not crash as I describe. :)  I'll let you decide on whether to leave as is. Extracting a reusable method might make sense since the lines below repeat the same pattern.

File: config/routes.rb [hunk 6]
```diff
@@ -192,6 +192,7 @@
   get '/api/campus_solutions/slr_deeplink' => 'campus_solutions/slr_deeplink#get', :via => :get, :defaults => { :format => 'json' }
   get '/api/campus_solutions/fpp_enrollment' => 'campus_solutions/fpp_enrollment#get', :via => :get, :defaults => { :format => 'json' }
   get '/api/campus_solutions/emergency_contacts' => 'campus_solutions/emergency_contacts#get', :via => :get, :defaults => { :format => 'json' }
+  get '/api/campus_solutions/residency_message' => 'campus_solutions/residency_message#get', :via => :get, :defaults => { :format => 'json' }
```
johncrossman: @dfkaye As I was told recently, the `:via => :get, ` is unnecessary (ie, obsolete) given the line's leading `get`.

johncrossman:
Thank you, @dfkaye  I think we can merge once comments above are addressed.

dfkaye:
@johncrossman Travis happy ~ Have another look...

app/models/hub_edos/demographics.rb
```diff
@@ -23,5 +23,24 @@ def instance_key
       @instance_key
     end
 
+    def build_feed(response)
+      response = super(response)
+      return response if response['student'].blank?
+      return response if response['student']['residency'].blank?
```
johncrossman: @dfkaye On a future commit, I recommend simplifying the above with:
```
return response if response['student'].blank? || response['student']['residency'].blank?
```

johncrossman:
Looks good, @dfkaye  Thx

PR #5513, opened 2016-06-15
SISRP-19374 - 5.6 - Academics - Status, Holds, Blocks - Residency Message Catalog - UI

dfkaye:
https://jira.berkeley.edu/browse/SISRP-19374

PR for the UI portion.

paired with Ruby support PR https://github.com/ets-berkeley-edu/calcentral/pull/5511

### DIFF
--- a/src/assets/javascripts/angular/controllers/widgets/academicsStatusHoldsBlocksController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/academicsStatusHoldsBlocksController.js
@@ -6,7 +6,7 @@ var _ = require('lodash');
 /**
  * Academics status, holds & blocks controller
  */
-angular.module('calcentral.controllers').controller('AcademicsStatusHoldsBlocksController', function(apiService, profileFactory, slrDeeplinkFactory, $q, $scope) {
+angular.module('calcentral.controllers').controller('AcademicsStatusHoldsBlocksController', function(apiService, profileFactory, slrDeeplinkFactory, residencyMessageFactory, $scope) {
   // Data for studentInfo and csHolds are pulled by the AcademicsController that
   // governs the academics template. The statusHoldsBlocks segment watches those
   // for changes in order to display the corresponding UI elements.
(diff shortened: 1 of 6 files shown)


src/assets/javascripts/angular/controllers/widgets/academicsStatusHoldsBlocksController.js
```diff
@@ -59,28 +59,41 @@ angular.module('calcentral.controllers').controller('AcademicsStatusHoldsBlocksC
...
 
   var parseCalResidency = function(residency) {
-    residency.fromTerm = _.get(residency, 'fromTerm.name');
+    // Turn 2014 Summer into Summer 2014.
+    residency.fromTerm.relabeled = residency.fromTerm.name.split(' ').reverse().join(' ');
 
-    residency.description = calResidencyCode[residency.code];
     angular.merge($scope.residency, residency);
+
+    var urlResidencyMessage = '/api/campus_solutions/residency_message';
```
redconfetti: @dfkaye - Shouldn't we be placing this inside of a factory? I'm pretty sure that's the convention we've setup so that all the API interfaces in use are contained in the collection of factories.

src/assets/javascripts/angular/controllers/widgets/academicsStatusHoldsBlocksController.js
```diff
@@ -59,28 +59,41 @@ angular.module('calcentral.controllers').controller('AcademicsStatusHoldsBlocksC
...
-    'NON': 'Non-Resident',
-    'RES': 'Resident'
-  };
-
   var getPerson = profileFactory.getPerson;
 
   var parseCalResidency = function(residency) {
-    residency.fromTerm = _.get(residency, 'fromTerm.name');
+    // Turn 2014 Summer into Summer 2014.
+    residency.fromTerm.relabeled = residency.fromTerm.name.split(' ').reverse().join(' ');
```
paulkerschen: @dfkaye It would be good to do this on the back end if we can. See usages of `Berkeley::TermCodes.normalized_english`:
https://github.com/ets-berkeley-edu/calcentral/blob/master/app/models/berkeley/term_codes.rb#L38

paulkerschen:
Looks good. Thanks @dfkaye

PR #5514, opened 2016-06-15
SISRP-21186 - Cannot add new emergency contact for the first time

dfkaye:
https://jira.berkeley.edu/browse/SISRP-21186

Always set samePhone and sameAddress to false to allow any address edits.


paulkerschen:
Thanks @dfkaye

PR #5515, opened 2016-06-16
SISRP-21299 Update campus directory links

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-21299


redconfetti:
@paulkerschen i was wondering when we're going to do this.

PR #5517, opened 2016-06-16
SISRP-21097 Queries and specs for EdoOracle::Calendar

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-21097

Some of our existing Oracle fanciness is moved into superclasses for reuse.

Also, CampusOracle::Calendar is explicitly told not to query CS terms.

*Note:* these new queries are not yet called in production. Nothing to QA for v78.


redconfetti:
@paulkerschen I really like this. Beautiful

PR #5521, opened 2016-06-16
SISRP-18965 - 5.6 - My Dashboard (Student) - Status, Blocks and Holds - Add Pre-Fall 2016 Terms to Registrations Feed, Refactor Reg

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-18965

* For Fall 2016 regstatus, I focused on keeping the exact same design as we do now
* Instead of detangling `academicsController.js`, I decided to insert the new feed into `academicsStatusHoldsBlocks.js`.  This will also reduce the number of extra calls needed when placing this card onto the Advisor - Student Overview page (which is the next step).
* Will need to apply the same refactoring to `academicsController.js` as well though, but that will be post GL5.6.
* I also had to simplify the Rspec tests, since `CampusOracle::Queries` does not have any information on any terms prior to the "current" term.  What it does have, though, does suffice to test the functionality of the model.
* I know this is asking a lot, but it would be great if I could get reviewed/merged sometime today if possible.  Trying to place this card onto Student Overview before the branch gets cut tomorrow.  It's a hefty one though, so I understand the need to carefully take the time to review.
* Thanks a bunch @paulkerschen once again for the direction.

### DIFF
--- a/app/models/my_registrations/my_registrations.rb
+++ b/app/models/my_registrations/my_registrations.rb
@@ -1,6 +1,7 @@
 module MyRegistrations
   class MyRegistrations < UserSpecificModel
 
+    include Berkeley::UserRoles
     include Cache::CachedFeed
     include Cache::JsonAddedCacher
     include Cache::UserCacheExpiry
@@ -9,8 +10,9 @@ module MyRegistrations
       registrations = get_registrations
       terms = get_terms
       {
-        :affiliations => registrations["affiliations"],
-        :registrations => match_terms(registrations["registrations"], terms)
+        affiliations: registrations['affiliations'],
+        terms: terms,
+        registrations: match_terms(registrations['registrations'], terms)
       }
     end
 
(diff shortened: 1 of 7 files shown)


src/assets/javascripts/angular/controllers/widgets/academicsStatusHoldsBlocksController.js
```diff
@@ -6,13 +6,19 @@ var _ = require('lodash');
...
   $scope.statusHoldsBlocks = {};
+  $scope.regStatus = {
+    summary: null,
+    explanation: null,
+    needsAction: null,
+    isLoading: true
+  };
 
-  $scope.$watchGroup(['studentInfo.regStatus.code', 'api.user.profile.features.csHolds'], function(newValues) {
+  $scope.$watchGroup(['regStatus.summary', 'api.user.profile.features.csHolds'], function(newValues) {
```
davidlimcheng: Since regStatus.summary can come from either CampusOracle or the Hub, we'll stop basing visibility on the MyBadges feed, and look to see if `$scope.regStatus.summary` has been populated instead.

app/models/my_registrations/my_registrations.rb
```diff
@@ -20,33 +22,85 @@ def get_registrations
...
+        result[:reg_status][:transitionTerm] = true if term_transition?
+      end
+
+      HashConverter.camelize response
+    end
+
+    def slug_to_edo_id(slug)
+      code = Berkeley::TermCodes.from_slug(slug)
+      Berkeley::TermCodes.to_edo_id(code[:term_yr], code[:term_cd])
+    end
```
paulkerschen: @davidlimcheng How about adding this method to Berkeley::TermCodes for reuse? I think I've felt the lack of it before...

app/models/my_registrations/my_registrations.rb
```diff
@@ -20,33 +22,85 @@ def get_registrations
...
+      matched_terms
+    end
+
+    def parse_legacy_term(term_name, term_id, term_english)
+      term_code = Berkeley::TermCodes.from_edo_id(term_id)
+      result = CampusOracle::Queries.get_person_attributes(@uid, term_code[:term_yr], term_code[:term_cd])
+      return if result.nil?
+      if result
+        result[:reg_status] = Notifications::RegStatusTranslator.new().translate_for_feed result['reg_status_cd']
+        result[:reg_status][:education_level] = Notifications::EducLevelTranslator.new().translate result['educ_level']
```
paulkerschen: @davidlimcheng No parens needed after `new` on these two lines

app/models/my_registrations/my_registrations.rb
```diff
@@ -1,16 +1,18 @@
...
     include Cache::JsonAddedCacher
     include Cache::UserCacheExpiry
 
     def get_feed_internal
       registrations = get_registrations
       terms = get_terms
-      {
-        :affiliations => registrations["affiliations"],
-        :registrations => match_terms(registrations["registrations"], terms)
+      response = {
```
paulkerschen: @davidlimcheng Presumably no need to assign a local variable `response` if this is the return value of the method?

spec/models/my_registrations/my_registrations_spec.rb
```diff
@@ -1,48 +1,36 @@
 describe MyRegistrations::MyRegistrations do
   before(:each) do
     @model = MyRegistrations::MyRegistrations.new(user_id: '61889')
+    @model.instance_variable_set(:@uid, 61889)
```
paulkerschen: @davidlimcheng If you check the superclass `UserSpecificModel`, you'll see the convention for initializing subclasses is to pass the UID alone as first argument (i.e. no `user_id` key). That should properly set the UID and save you having to call `instance_variable_set` on the newly initialized model.

spec/models/my_registrations/my_registrations_spec.rb
```diff
@@ -52,96 +40,54 @@
...
     it 'returns all matched terms' do
       matched_terms = @model.get_feed_internal
-      matched_terms[:registrations][:current].should have(0).items
-      matched_terms[:registrations][:running].should have(0).items
-      matched_terms[:registrations][:sis_current_term].should have(0).items
-      expect(matched_terms[:registrations][:next][0]["term"]["id"]).to eq 2168
-      expect(matched_terms[:registrations][:future][0]["term"]["id"]).to eq 2171
-      matched_terms[:registrations][:previous].should have(0).items
-      expect(matched_terms[:registrations][:grading_in_progress][0]["term"]["id"]).to eq 2159
+      matched_terms[:registrations][2165].should be_nil
```
paulkerschen: @davidlimcheng We prefer `expect` to `should` syntax these days, though it's a slow road rooting out all the old ones.

spec/models/my_registrations/my_registrations_spec.rb
```diff
@@ -52,96 +40,54 @@
...
-      matched_terms[:registrations][:current].should have(0).items
-      matched_terms[:registrations][:running].should have(0).items
-      matched_terms[:registrations][:sis_current_term].should have(0).items
-      matched_terms[:registrations][:next].should have(0).items
-      matched_terms[:registrations][:future].should have(0).items
-      matched_terms[:registrations][:previous].should have(0).items
-      matched_terms[:registrations][:grading_in_progress].should have(0).items
+      matched_terms[:registrations][2165].should be_nil
+      matched_terms[:registrations][2168].should be_nil
+      matched_terms[:registrations][2171].should be_nil
```
paulkerschen: @davidlimcheng `expect` here also

spec/models/my_registrations/my_registrations_spec.rb
paulkerschen: @davidlimcheng Also, instead of creating instance variables in a `before` block, we tend to use lazy initialized variables in a `let` block: (i.e. `let(:model} { MyRegistrations::MyRegistrations.new(user_id: '61889') }`, and then refer to `model` without the `@` sign. This is a good writeup on the rationale, if you're curious:

http://stackoverflow.com/questions/5359558/when-to-use-rspec-let

app/models/my_registrations/my_registrations.rb
davidlimcheng: @paulkerschen Ah yes, good catch.  I previously used HashConverter.camelize here, but decided to put it in a lower level method instead.

spec/models/my_registrations/my_registrations_spec.rb
davidlimcheng: Ahh, that makes total sense.  Will amend.

davidlimcheng:
Thanks @paulkerschen!  Amended with proposed changes.  Rspec gets more likable with every PR.

davidlimcheng:
Very weird, trying to find details results in a "The repository at ets-berkeley-edu/calcentral was not found." message on Travis-CI.  Has this happened before?

paulkerschen:
Travis is pacified. Thanks @davidlimcheng!

PR #5522, opened 2016-06-16
SISRP-20669 - Add 'Terms' column to enrollment lists

redconfetti:
https://jira.berkeley.edu/browse/SISRP-20669

### DIFF
--- a/app/models/rosters/common.rb
+++ b/app/models/rosters/common.rb
@@ -78,7 +78,7 @@ module Rosters
           attrs[:email] = attrs.delete :email_address
           attrs[:majors] = enrollments_by_uid[attrs[:ldap_uid]].collect { |e| e['major'] }
           if (enrollment_row = enrollments_by_uid[attrs[:ldap_uid]].first)
-            attrs[:terms_in_attendance] = enrollment_row['terms_in_attendance_group'].try(:[], 1)
+            attrs[:terms_in_attendance] = terms_in_attendance_code(enrollment_row['terms_in_attendance_group'], enrollment_row['academic_program_code'])
             attrs[:enroll_status] = enrollment_row['enroll_status']
             attrs[:grade_option] = Berkeley::GradeOptions.grade_option_from_basis enrollment_row['grading_basis']
             attrs[:units] = enrollment_row['units'].to_s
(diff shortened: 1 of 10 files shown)


app/models/rosters/common.rb
```diff
@@ -1,3 +1,4 @@
+# encoding: UTF-8
```
paulkerschen: @redconfetti Why is the encoding comment needed?

public/dummy/json/campus_rosters.json
```diff
@@ -92,8 +94,9 @@
       "last_name": "LBBERRR",
       "login_id": "333333",
       "majors": [
-        "Landscape Architecture BA"
+        "Education PhD"
       ],
+      "terms_in_attendance": 'G',
```
paulkerschen: @redconfetti ANGRY RED single quotes in JSON

spec/models/rosters/common_spec.rb
```diff
@@ -1,3 +1,4 @@
+# encoding: UTF-8
```
paulkerschen: @redconfetti Same encoding question here. If it's the em dash — that requires it, might be preferable to refer to it by codepoint: `"\u2014"`.

redconfetti: @paulkerschen It is the em dash. Thanks for the tip. I didn't know that one of those codes would be interpreted via JSON properly.

src/assets/javascripts/angular/controllers/pages/rosterController.js
```diff
@@ -19,6 +20,10 @@ angular.module('calcentral.controllers').controller('RosterController', function
     return (student.section_ccns.indexOf($scope.searchSection.ccn) !== -1);
   };
 
+  $scope.allStudentsCount = function() {
+    return _.get($scope, 'students.length', 0);
+  };
+
```
paulkerschen: @redconfetti Did this function sneak in from a different branch? I don't see it in use anywhere.

spec/models/rosters/campus_spec.rb
```diff
@@ -71,6 +51,7 @@
       expect(feed[:students].length).to eq 2
 
       student = feed[:students][0]
+      puts "student: #{student.inspect}"
```
paulkerschen: Stray puts, @redconfetti

PR #5523, opened 2016-06-17
CLC-6352 Lock down gulp-header to 1.8.2

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6352


paulkerschen:
marcel-hendrykx-yan-tregger-be-quite-hypnose-reelle-disques-jade

PR #5529, opened 2016-06-17
SISRP-18000 - 5.6 - Student Overview - Status, Blocks and Holds - Add Status, Blocks, and Holds Card to Student Overview

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-18000

* I knew that old code I threw away would come in handy...it totally did
* Separated the `status_holds_blocks.html` template from the `academicStatusHoldsBlocks.js` controller, so that the `userOverview.js` controller could own the data on the Student Overview screen.
* Regblock stuff should be removed in Fall 2016
* Uses the same type of "legacy or CS" parsing as the regular Status/Holds/Blocks card


davidlimcheng:
Just realized I should have flipped my `isLoading` flags inside of `.finally` blocks - one sec, amending.

davidlimcheng:
Kk, good to go.

src/assets/javascripts/angular/controllers/pages/userOverviewController.js
```diff
@@ -70,6 +105,92 @@ angular.module('calcentral.controllers').controller('UserOverviewController', fu
...
+  };
+
+  /**
+   * Parses any terms on or before the legacy cutoff.  Mirrors current functionality, this should be able to be removed in Fall 2016.
+   */
+  var parseLegacyTerm = function(term) {
+    $scope.regStatus.summary = term.regStatus.summary;
+    $scope.regStatus.explanation = term.regStatus.explanation;
+    $scope.regStatus.needsAction = term.regStatus.needsAction;
+  };
```
paulkerschen: @davidlimcheng This code looks good as before, but since there's so much overlap with academicsStatusBlocksHoldsController.js, the common methods should really be extracted into a shared service (perhaps called registrationService?) I know we're getting close to branch cut, so I'll let the room weigh in...

paulkerschen:
Merging! Thanks for the work, @davidlimcheng. Could we create a follow-up task to factor out the reused code?

davidlimcheng:
Sure thing, went ahead and opened [CLC-6353](https://jira.ets.berkeley.edu/jira/browse/CLC-6353).

src/assets/javascripts/angular/controllers/pages/userOverviewController.js
davidlimcheng: That's a great point - since this will be changing within a month or so for GL6 (redesign), I'll make a note of creating a service for any shared code in the next PR.

app/controllers/advising_student_controller.rb
```diff
@@ -39,6 +40,15 @@ def holds
     render json: CampusSolutions::MyHolds.new(student_uid_param).get_feed_as_json
   end
 
+  #TODO: Remove by Fall 2016
+  def blocks
+    render json: Bearfacts::Regblocks.new({user_id: student_uid_param}).get
+  end
+
+  def registrations
+    render json: MyRegistrations::MyRegistrations.new(student_uid_param).get_feed_internal
```
redconfetti: @davidlimcheng Is there a reason why we didn't use #get_feed (provided to MyRegistrations::MyRegistrations by [Cache::CachedFeed](https://github.com/ets-berkeley-edu/calcentral/blob/49a647ba8/app/models/my_registrations/my_registrations.rb#L5)

It appears that we're not using a cacheing mechanism here that could be used. This could be on purpose, which is why I ask.

PR #5532, opened 2016-06-17
SISRP-21396 Hide 'Delegate Access' financial link from delegates

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-21396


johncrossman:
image
LGTM.  But Rebecca De Mornay is not pleased.

PR #5536, opened 2016-06-20
SISRP-20587 - Adds spinner message for Enrollments/Rosters. Adds email link to names in Rosters.

redconfetti:
https://jira.berkeley.edu/browse/SISRP-20587

See QA PR #5537

### DIFF
--- a/src/assets/templates/widgets/roster.html
+++ b/src/assets/templates/widgets/roster.html
@@ -1,6 +1,6 @@
 <div data-ng-controller="RosterController" data-ng-class="{'cc-page-roster': !isLoading}">
 
-  <div data-cc-spinner-directive>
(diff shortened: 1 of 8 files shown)


src/assets/templates/widgets/roster.html
```diff
@@ -62,8 +62,24 @@ <h2 class="cc-visuallyhidden">
           <div data-ng-include="'widgets/roster_photo.html'"></div>
         </a>
       </div>
```
redconfetti: The changes being made below are made with the intention of NOT creating a bug in bCourses where the names are linkable, even though the students email is not present.

raydavis:
If you can't feed it, spin it!

PR #5539, opened 2016-06-20
SISRP-18649 Query rosters for all sections of a course in one go

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-18649

### DIFF
--- a/app/models/campus_oracle/queries.rb
+++ b/app/models/campus_oracle/queries.rb
@@ -170,7 +170,7 @@ module CampusOracle
       result = []
       use_pooled_connection {
         sql = <<-SQL
-      select roster.student_ldap_uid ldap_uid, roster.enroll_status, trim(roster.pnp_flag) as pnp_flag,
+      select roster.student_ldap_uid as ldap_uid, roster.enroll_status, trim(roster.pnp_flag) as pnp_flag,
         trim(person.first_name) as first_name, trim(person.last_name) as last_name, person.student_email_address, person.student_id, person.affiliations
       from calcentral_class_roster_vw roster, calcentral_student_info_vw person
       where roster.term_yr = #{term_yr.to_i}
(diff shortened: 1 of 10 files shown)


app/models/campus_oracle/queries.rb
```diff
@@ -184,6 +184,26 @@ def self.get_enrolled_students(ccn, term_yr, term_cd)
       stringify_ints! result
     end
 
+    # Version of get_enrolled_students for multiple CCNs.
+    def self.get_enrolled_students_for_ccns(ccns, term_yr, term_cd)
+      result = []
+      use_pooled_connection {
+        sql = <<-SQL
+      select roster.course_cntl_num, roster.student_ldap_uid ldap_uid, roster.enroll_status, trim(roster.pnp_flag) as pnp_flag,
```
johncrossman: @paulkerschen should this be `roster.student_ldap_uid AS ldap_uid`?

paulkerschen: @johncrossman I copied from the query above, which also does without the `AS`. I guess it's implicit in this Oracle dialect?

app/models/edo_oracle/queries.rb
```diff
@@ -303,7 +304,7 @@ def self.get_rosters(section_id, term_id)
           SISEDO.STUDENT_GROUPV00_VW stdgroup ON enroll."STUDENT_ID" = stdgroup."STUDENT_ID" AND
           stdgroup."STDNT_GROUP" IN ('R1TA', 'R2TA', 'R3TA', 'R4TA', 'R5TA', 'R6TA', 'R7TA', 'R8TA')
         WHERE
-          enroll."CLASS_SECTION_ID" = '#{section_id}'
+          enroll."CLASS_SECTION_ID" IN (#{ccns.join ','})
```
johncrossman: @paulkerschen the `ccns` won't be in quotes but does it matter? I believe I've seen EDO db play nice in either case.

app/models/rosters/campus.rb
```diff
@@ -35,14 +37,17 @@ def get_feed_internal
...
+
       crosslisted_courses.each do |course|
         course[:sections].each do |section|
           feed[:sections] << {
             ccn: section[:ccn],
             name: "#{course[:dept]} #{course[:catid]} #{section[:section_label]}"
           }
-          section_enrollments = get_enrollments(section[:ccn], course[:term_yr], course[:term_cd])
-          section_enrollments.each do |enr|
+          next unless enrollments[section[:ccn]]
```
johncrossman: @paulkerschen IMO, a `next` is best when it's the first line in an loop block. IMO, it's risky to have it in the middle of the block cuz it might go unnoticed when debugging.

spec/models/rosters/campus_spec.rb
```diff
@@ -120,15 +138,34 @@
...
+            },
+            {
+              ldap_uid: waitlisted_student_login_id,
+              student_id: enrolled_student_student_id,
+              first_name: 'First Name',
+              last_name: 'Last Name',
+              email_address: "#{waitlisted_student_login_id}@example.com"
+            }
+          ]
+        end
```
johncrossman: Thanks for giving the specs so much attention, @paulkerschen

johncrossman:
Nice work, @paulkerschen Ready to merge once I get your opinion on a comment.

app/models/edo_oracle/queries.rb
paulkerschen: @johncrossman Don't think it matters now, but yes, I'll add in quotes in case we get ambiguity with leading zeros or such down the road/

app/models/rosters/campus.rb
paulkerschen: @johncrossman That's fair! I'll rewrite using `try`.

app/models/campus_oracle/queries.rb
paulkerschen: I'll add `AS` here and to the old query just to keep consistent.

raydavis:
Comment has been duly opinionated.

PR #5542, opened 2016-06-21
SISRP-21595, do not strip my_higher_one_url when nil

johncrossman:
https://jira.berkeley.edu/browse/SISRP-21595

Nothin' from nothin' leaves nothin'.

### DIFF
--- a/app/models/campus_solutions/my_higher_one_url.rb
+++ b/app/models/campus_solutions/my_higher_one_url.rb
@@ -14,7 +14,8 @@ module CampusSolutions
 
     def get_higher_one_url
       return {} unless is_feature_enabled
-      proxy.build_url.strip
+      url = proxy.build_url
+      url && url.strip
     end
 
     private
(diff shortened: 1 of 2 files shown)


app/models/campus_solutions/my_higher_one_url.rb
```diff
@@ -14,7 +14,8 @@ def get_feed_internal
 
     def get_higher_one_url
       return {} unless is_feature_enabled
-      proxy.build_url.strip
+      url = proxy.build_url
+      url && url.strip
```
davidlimcheng: Is this essentially:
```
url.strip if url.present?
url
```
??

davidlimcheng:
Thanks @johncrossman!

app/models/campus_solutions/my_higher_one_url.rb
paulkerschen: @davidlimcheng Basically, except that the `url.strip` has to come last. The reason is that `.strip` returns a new string instead of modifying the old string, so the two lines of code in your comment, executed in order, would return the original, unstripped URL.

paulkerschen: @davidlimcheng Also, it's checking `url` for truthiness rather than presence, which is a bit more permissive (an empty string will pass the former but not the latter).

davidlimcheng: @paulkerschen Ahh I see, good to know.  Very elegant.

PR #5543, opened 2016-06-21
SISRP-21453 & SISRP-20157 - Profile Work Experience: Add and Edit buttons should be disabled when viewing-as, start and end dates not shown

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-21453
https://jira.berkeley.edu/browse/SISRP-20157

* Start/End dates stopped showing because they were reformatted in the returned EDO.  Fixed them everywhere they appeared.
* Had to use `data-ng-if` instead of `data-ng-disabled` on Add/Edit buttons.


davidlimcheng:
One sec, amending.

davidlimcheng:
OK sorry.  We only want Add/Edit if `preventActingAsUsersFromPosting: false`.  Should be good now.

PR #5549, opened 2016-06-22
SISRP-21401 - Make Payment for Fall 2016 button should be greyed out in view as mode

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-21401

* I actually had to talk myself thru this one:
* We want the "Make Payment" button to DISABLE only for view-as users (`&&` statement)
* We want the "View Official Monthly Payment" item to ENABLE for directlyAuthenticated or delegates (`||` statement)
* Hence, one `&&` and one `||` for the seemingly same link to HigherOne.


paulkerschen:
Booleans! My mnemonic is to remember that `!(x || y)` is the same as `!x && !y`. Looks right and thanks @davidlimcheng.

PR #5364, opened 2016-05-09
SISRP-18744 - Advisor student overview page needs a better UX while data elements are loading

dfkaye:
https://jira.berkeley.edu/browse/SISRP-18744

PR #5550, opened 2016-06-22
SISRP-21347 Do not show enrollment cards with no enrollments

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-21347

Since this is a partial revert of #5364, it will reintroduce the mystery-spinner issue first noted in https://jira.berkeley.edu/browse/SISRP-18744.

### DIFF
--- a/src/assets/templates/widgets/enrollment_card.html
+++ b/src/assets/templates/widgets/enrollment_card.html
@@ -1,13 +1,5 @@
-<div class="cc-enrollment-card" data-ng-controller="EnrollmentCardController">
-  <div class="cc-widget" data-ng-if="enrollment.isLoading || enrollment.terms.length === 0">
-    <div class="cc-widget-title">
-      <h2>Class Enrollment</h2>
-    </div>
-    <div class="cc-widget-padding" data-cc-spinner-directive="enrollment.isLoading">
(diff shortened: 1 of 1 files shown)


src/assets/templates/widgets/enrollment_card.html
```diff
@@ -1,13 +1,5 @@
...
-  <div class="cc-widget" data-ng-if="enrollment.isLoading || enrollment.terms.length === 0">
-    <div class="cc-widget-title">
-      <h2>Class Enrollment</h2>
-    </div>
-    <div class="cc-widget-padding" data-cc-spinner-directive="enrollment.isLoading">
-      <div class="cc-widget-text">No Enrollments</div>
-    </div>
-  </div>
-  <div data-ng-if="!enrollment.isLoading && enrollment.terms.length > 0">
+<div class="cc-enrollment-card" data-ng-controller="EnrollmentCardController" data-cc-spinner-directive="enrollment.isLoading">
```
davidlimcheng: @paulkerschen I didn't know you can access scope variables on the same attribute you declare the controller.  Very good to know.

raydavis:
Thanks, @paulkerschen -- I agree with the partial regression and with scheduling a fuller fix for next sprint.

PR #5553, opened 2016-06-22
SISRP-21615 Use EDO DB, not LDAP, as student id source in rosters

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-21615

### DIFF
--- a/app/models/rosters/common.rb
+++ b/app/models/rosters/common.rb
@@ -85,6 +85,7 @@ module Rosters
             attrs[:email] = attrs.delete :email_address
             attrs[:majors] = section_enrollments_by_uid[attrs[:ldap_uid]].collect { |e| e['major'] }
             if (enrollment_row = section_enrollments_by_uid[attrs[:ldap_uid]].first)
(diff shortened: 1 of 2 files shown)


app/models/rosters/common.rb
```diff
@@ -85,6 +85,7 @@ def get_enrollments(ccns, term_yr, term_cd)
             attrs[:email] = attrs.delete :email_address
             attrs[:majors] = section_enrollments_by_uid[attrs[:ldap_uid]].collect { |e| e['major'] }
             if (enrollment_row = section_enrollments_by_uid[attrs[:ldap_uid]].first)
+              attrs[:student_id] = enrollment_row['student_id']
```
redconfetti: @paulkerschen. I assume that your next commit will change some logic to ensure that the student ID isn't overwritten when not present?

paulkerschen: @redconfetti In fact we're safe there, since student_id is what EDO DB uses to join everything else; so we're guaranteed to have a student_id present in the enrollment row which correlates properly to all the other data we're pulling in.

PR #5555, opened 2016-06-22
SISRP-21513 - Updates terms in attendance value calculation

redconfetti:
Updates terms in attendance value calculation for Rosters feed to depend on enrollment specific academic career code

https://jira.berkeley.edu/browse/SISRP-21513

See #5559 for QA PR


raydavis:
Thank you for the fix, @redconfetti ! I'll merge QA first, given the timing.

PR #5568, opened 2016-06-22
SISRP-21625 - UAT 5.6 - Cal Central - My Finances - Billing Summary-Details - Fall 2016 Sorting incorrectly

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-21625

* Confirmed with functional owner that we want to keep case-insensitivity
* From AngularJS documentation:  "Note: If you notice numbers not being sorted as expected, make sure they are actually being saved as numbers and not strings."
* Due dates weren't sorting correctly because they were sorting on flags instead of the actual date

### DIFF
--- a/src/assets/javascripts/angular/controllers/widgets/billingController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/billingController.js
@@ -103,7 +103,7 @@ angular.module('calcentral.controllers').controller('BillingController', functio
   };
 
   var parseAmounts = function(value) {
-    return _.isNumber(value) ? value.toFixed(2) : value;
+    return _.isNumber(value) ? parseFloat(value.toFixed(2)) : value;
   };
 
   var parseBillingInfo = function(data) {
--- a/src/assets/templates/widgets/billing_activity.html
+++ b/src/assets/templates/widgets/billing_activity.html
@@ -96,8 +96,8 @@
               <th data-ng-click="changeSorting('itemType')" width="10%" scope="col">
                 Type <i data-ng-class="getSortClass('itemType')"></i>
               </th>
-              <th data-ng-click="changeSorting('itemDueFlag')" class="cc-table-center" width="20%" scope="col">
-                Due <i data-ng-class="getSortClass('itemDueFlag')"></i>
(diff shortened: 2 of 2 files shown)


src/assets/javascripts/angular/controllers/widgets/billingController.js
```diff
@@ -103,7 +103,7 @@ angular.module('calcentral.controllers').controller('BillingController', functio
   };
 
   var parseAmounts = function(value) {
-    return _.isNumber(value) ? value.toFixed(2) : value;
+    return _.isNumber(value) ? parseFloat(value.toFixed(2)) : value;
```
paulkerschen: @davidlimcheng If I'm reading this right, this first turns a number into a string with `toFixed`, and then back into a number with `parseFloat`. Is there any net change? What happens if you skip the calls to `parseAmounts` entirely?

davidlimcheng: @paulkerschen That's a great catch!  Interestingly, I remember having to use `.toFixed(2)` with the data initially, I believe because there were too many 0's after the decimal point.  It seems like it's returning good data now - but I'd love to check-in with the API developer before removing this.  I'll check in with him first thing tomorrow morning and update this PR.

davidlimcheng:
Please hold off on merging this PR until I get a chance to verify decimal points w/ the API dev.

paulkerschen:
@davidlimcheng You could also try an Angular number filter in the template
https://docs.angularjs.org/api/ng/filter/number
to force the display to two decimal points without changing the data type in the controller.

davidlimcheng:
@paulkerschen I had considered that, but since we are using a custom directive to present the amount (`data-cc-amount-directive="item.itemLineAmount"`), the only way to use a filter is at the directive level.  I just spoke with the API developer, and unfortunately he couldn't confirm whether the data will only send values at two decimal points.  

I think this is probably the safest route, although it does seem a bit redundant.

paulkerschen:
Following @davidlimcheng's diligent research on rounding solutions, this PR looks good to merge, but we'll do a QA version first.

PR #5579, opened 2016-06-23
[QA] SISRP-21616 Fix Residency message codes

raydavis:
QA FIRST fix for https://jira.berkeley.edu/browse/SISRP-21616

The original specs were incomplete and the implementation needed a complete overhaul.

PR #5583, opened 2016-06-23
SISRP-21616 Fix Residency message codes

raydavis:
Master-branch version of https://github.com/ets-berkeley-edu/calcentral/pull/5579


davidlimcheng:
That is some serious business logic, but code looks good to me @raydavis.

PR #5589, opened 2016-06-28
SISRP-21703 SISRP-21705 SISRP-21706 Show transcript data to advisors

raydavis:
https://jira.berkeley.edu/browse/SISRP-21703 and friends

Rather than move "shared controller" code into an AngularJS service or taking on the bigger job of giving the Semesters template its own controller, I decided to just give the front-end code what it wanted.


paulkerschen:
Thanks @raydavis, seems like a sensible place for the job.

PR #5592, opened 2016-06-28
SISRP-21150 - QA tests for new Finances UI

pfarestveit:
https://jira.berkeley.edu/browse/SISRP-21150

### DIFF
--- a/spec/ui_selenium/util/web_driver_utils.rb
+++ b/spec/ui_selenium/util/web_driver_utils.rb
@@ -110,6 +110,8 @@ class WebDriverUtils
   end
 
   def self.amt_to_s(amount)
+    # The new finances UI currently displays a 1.0 float as "1". But a direct conversion from float to string would produce "1.0",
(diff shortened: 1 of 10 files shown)


spec/ui_selenium/util/web_driver_utils.rb
```diff
@@ -109,6 +109,11 @@ def self.currency_to_f(dollar_amount)
     dollar_amount.delete('$, ').to_f
   end
 
+  def self.amt_to_s(amount)
+    trimmed_amount = (amount == amount.to_i ? amount.to_i : amount)
```
paulkerschen: @pfarestveit Seems like this line doesn't do anything? It returns `amount` unless `amount.to_i` is equal, but in that case, since they're equal, returning `amount` would do the same.

pfarestveit: @paulkerschen - it looks totally absurd I agree.  I'm adding a comment to explain the wonky hack.

paulkerschen:
One green Travis. Thanks @pfarestveit.

PR #5594, opened 2016-06-29
SISRP-21785 - Finances Billing - Balance filter showing paid transactions with zero balance

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-21785

Also fixes:
https://jira.berkeley.edu/browse/SISRP-21704

* Verified with functional that CS will never send amounts with more than two decimal places
* Since JS is terrible at keeping trailing zeroes on `floats`, decided to just leave the `float` as is, and append a `string` version of that float to the object for presentation, while any controller/filter/sort logic is still based on the `float`.
* Still need to fix the "all transactions" sort, which is breaking because it still sorts on `itemBalance` but is showing `itemLineAmount`


raydavis:
Thank you for the clear fix & explanation!

PR #5596, opened 2016-06-29
SISRP-21883 First fixes for View-As self issues

raydavis:
https://jira.berkeley.edu/browse/SISRP-21883

This first PR fixes the Delegate log-out issue and does basic server-side blocking of View-As-self requests. Front-end code needs to be changed to show error responses from ``/act_as``.


paulkerschen:
"Seems," madam? Nay, it is; I know not "seems."

PR #5602, opened 2016-06-30
SISRP-12998 Fix FinAid Estimated Cost of Attendance labels

raydavis:
https://jira.berkeley.edu/browse/SISRP-12998


paulkerschen:
We must adapt!

PR #5603, opened 2016-06-30
CLC-6370 Temporarily hardcode Fall 2016 term dates

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6370

The only thing to be said for this dumb temporary code is that it obviates some other dumb temporary code.


raydavis:
Well, _I_ think it's _very nice_.

PR #5604, opened 2016-06-30
SISRP-21798 - Front/Back End Dev for 5.6d - Status and Holds

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-21798
Design JIRA (see MVP for GL5.6d attached PDF): https://jira.berkeley.edu/browse/SISRP-13817

* First step of redesign for the "Status and Holds" card, with step two coming in GL6.0
* Removes blocks from status/holds card + popover + academics feed with `legacyRegblocks` feature flag
* Separates CSS from `_academics.scss` to its own `_status_holds.scss`
* Had to make the same changes in the Advisor User Overview view of this card as well or else it would have been broken - sorry for the big PR!
* There's a bit of table CSS I left that is not being used, just in case we need to go back to the table layout of this card for accessibility reasons.
* All messaging is taken directly from the Design, see spreadsheet attached to JIRA.


davidlimcheng:
Ahh, accidentally set the regblocks feature flag to false.  Will amend.

paulkerschen:
Looks good, @davidlimcheng, thanks.

PR #5608, opened 2016-07-01
CLC-6373 OEC: better section sorting, primaries first

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6373


app/models/oec/sis_import_sheet.rb
```diff
@@ -74,5 +72,16 @@ def sorted_rows
       non_cross_listed_rows
     end
 
+    def self.sortable(row)
+      [                                                   # Canonical sort order:
+        row['DEPT_NAME'].to_s,                            # - Department name;
+        row['CATALOG_ID'].try(:match, /\d+/).to_s.to_i,   # - Numeric portion of catalog ID;
```
johncrossman: @paulkerschen My not be a big deal...  We have catalog_ids like:
```
277.1
277.2
277.3
277.4
```
The logic above will give us '277' for all the above and thus unpredictable sort order?

paulkerschen: @johncrossman Having the string version of `row['CATALOG_ID']` as the next value is meant to account for cases like that.

johncrossman:
Thanks for clarification, @paulkerschen  +1

PR #5610, opened 2016-07-01
CLC-6377 OEC: Exclude online courses from room-share lookup

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6377


johncrossman:
Exclusion looks good.

PR #5611, opened 2016-07-01
CLC-6375 OEC: Surface student ID only for active students

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6375


johncrossman:
Unless this is Java => looks good.

PR #5619, opened 2016-07-06
SISRP-21252 - Financials - how will amount due alerts in CalCentral appear for CS financials?

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-21252

* Follows the same logic as before, just added the CS Financials data sources
* QA PR to come

### DIFF
--- a/src/assets/javascripts/angular/controllers/widgets/statusController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/statusController.js
@@ -7,6 +7,11 @@ var angular = require('angular');
  * Status controller
  */
 angular.module('calcentral.controllers').controller('StatusController', function(activityFactory, apiService, badgesFactory, financesFactory, holdsFactory, $http, $scope, $q) {
+  $scope.finances = {
+    carsFinances: {},
+    csFinances: {}
+  };
+
   // Keep track on whether the status has been loaded or not
   var hasLoaded = false;
 
@@ -33,20 +38,57 @@ angular.module('calcentral.controllers').controller('StatusController', function
     }
   };
 
-  var loadFinances = function(data) {
-    if (!data.summary) {
+  var loadCarsFinances = function(data) {
+    if (data.summary) {
+      $scope.finances.carsFinances = data.summary;
+    }
+  };
+
+  var loadCsFinances = function(data) {
+    if (_.get(data, 'feed.summary')) {
+      $scope.finances.csFinances = data.feed.summary;
+    }
+  };
+
+  var parseFinances = function() {
+    $scope.totalPastDueAmount = 0;
+    $scope.minimumAmountDue = 0;
+    var cars = {
+      pastDue: 0,
+      minDue: 0
+    };
+    var cs = {
+      pastDue: 0,
+      minDue: 0
+    };
+
+    if (!$scope.finances.carsFinances && !$scope.finances.csFinances) {
       return;
     }
-
-    if (data.summary.totalPastDueAmount > 0) {
+    if ($scope.finances.carsFinances) {
+      cars = {
+        pastDue: $scope.finances.carsFinances.totalPastDueAmount,
+        minDue: $scope.finances.carsFinances.minimumAmountDue
+      };
+      $scope.totalPastDueAmount += cars.pastDue;
+      $scope.minimumAmountDue += cars.minDue;
+    }
+    if ($scope.finances.csFinances) {
+      cs = {
+        pastDue: $scope.finances.csFinances.pastDueAmount,
+        minDue: $scope.finances.csFinances.amountDueNow
+      };
+      $scope.totalPastDueAmount += cs.pastDue;
+      $scope.minimumAmountDue += cs.minDue;
+    }
+    if (cars.pastDue > 0 || cs.pastDue > 0) {
       $scope.count++;
       $scope.hasAlerts = true;
-    } else if (data.summary.minimumAmountDue > 0) {
+    } else if (cars.minDue > 0 || cs.minDue > 0) {
       $scope.count++;
       $scope.hasWarnings = true;
     }
-    $scope.totalPastDueAmount = data.summary.totalPastDueAmount;
-    $scope.minimumAmountDue = data.summary.minimumAmountDue;
+
     $scope.hasBillingData = ($scope.minimumAmountDue !== null);
   };
 
@@ -117,13 +159,18 @@ angular.module('calcentral.controllers').controller('StatusController', function
       // Only fetch financial data for delegates who have been given explicit permssion.
       var includeFinancial = (!apiService.user.profile.delegateActingAsUid || apiService.user.profile.delegateViewAsPrivileges.financial);
       if (includeFinancial) {
-        var getFinances = financesFactory.getFinances().success(loadFinances);
+        var getCarsFinances = financesFactory.getFinances().success(loadCarsFinances);
+        var getCsFinances = financesFactory.getCsFinances().success(loadCsFinances);
         var getFinaidActivityOld = activityFactory.getFinaidActivityOld().then(loadActivity);
-        statusGets.push(getFinances, getFinaidActivityOld);
+        statusGets.push(getCarsFinances, getCsFinances, getFinaidActivityOld);
       }
 
       // Make sure to hide the spinner when everything is loaded
-      $q.all(statusGets).then(finishLoading);
+      $q.all(statusGets).then(function() {
+        if (includeFinancial) {
+          parseFinances();
+        }
+      }).then(finishLoading);
     }
   });
 });


src/assets/javascripts/angular/controllers/widgets/statusController.js
```diff
@@ -33,20 +38,34 @@ angular.module('calcentral.controllers').controller('StatusController', function
...
+  var loadCsFinances = function(data) {
+    if (!data.feed || !data.feed.summary) {
+      return;
+    }
+    $scope.finances.csFinances = data.feed.summary;
+  };
 
-    if (data.summary.totalPastDueAmount > 0) {
+  var parseFinances = function() {
+    if (!$scope.finances.carsFinances || !$scope.finances.csFinances) {
```
johncrossman: @davidlimcheng do we want `&&` here?

johncrossman: For example, what if `$scope.finances.carsFinances` is null and `$scope.finances.csFinances.pastDueAmount > 0`?  We want `$scope.hasAlerts = true`, right?

src/assets/javascripts/angular/controllers/widgets/statusController.js
```diff
@@ -117,13 +136,14 @@ angular.module('calcentral.controllers').controller('StatusController', function
...
+        var getCarsFinances = financesFactory.getFinances().success(loadCarsFinances);
+        var getCsFinances = financesFactory.getCsFinances().success(loadCsFinances);
         var getFinaidActivityOld = activityFactory.getFinaidActivityOld().then(loadActivity);
-        statusGets.push(getFinances, getFinaidActivityOld);
+        statusGets.push(getCarsFinances, getCsFinances, getFinaidActivityOld);
       }
 
       // Make sure to hide the spinner when everything is loaded
-      $q.all(statusGets).then(finishLoading);
+      $q.all(statusGets).then(parseFinances).then(finishLoading);
```
johncrossman: @davidlimcheng if `includeFinancial` is false then I don't think we should `parseFinances`.

src/assets/javascripts/angular/controllers/widgets/statusController.js
davidlimcheng: @johncrossman Hmm, that's true - but that would also mean that if for some reason `$scope.finances.carsFinances = null` because of some temporary server issue, and we still surface `$scope.minimumAmountDue`, there's a chance that `$scope.minimumAmountDue` can increase once the student logs on again, without a clear explanation.  I thought that it might've been better to not show this at all if that was the case - what do you think?

src/assets/javascripts/angular/controllers/widgets/statusController.js
davidlimcheng: @johncrossman Great catch, will amend once our other conversation finishes up.

src/assets/javascripts/angular/controllers/widgets/statusController.js
johncrossman: @davidlimcheng Hmm. I don't think we should suppress `csFinances.amountDueNow` just because our `carsFinances` data-source is unavailable. Ideally, we message the user when a backend service is unavailable.

johncrossman: @raydavis Let us know if you disagree w/ my opinion above.

raydavis: I agree, @johncrossman. When possible, we strongly prefer letting the user know the information is currently unavailable, rather than showing them a wrong result.

That's not always possible with complexly merged models. In this case, would the CARS error be surfaced any other way? If so, then that would count as some sort of explanation for the change in total amount due.

davidlimcheng: @raydavis @johncrossman It would, it would be surfaced by navigating to "My Finances" and seeing the "Problem reaching campus data" message we've got in there - and, same with a CS error.

davidlimcheng:
@johncrossman Made proposed changes, adding logic that should hold up in case either one of the data sources break.

src/assets/javascripts/angular/controllers/widgets/statusController.js
```diff
@@ -33,20 +35,55 @@ angular.module('calcentral.controllers').controller('StatusController', function
...
       return;
     }
+    $scope.finances.carsFinances = data.summary;
+  };
+
+  var loadCsFinances = function(data) {
+    if (!data.feed || !data.feed.summary) {
+      return;
+    }
+    $scope.finances.csFinances = data.feed.summary;
```
johncrossman: @davidlimcheng This is a good spot for lodash:
```
if (_.get(data, 'feed.summary') {
  $scope.finances.csFinances = data.feed.summary;
}
```

src/assets/javascripts/angular/controllers/widgets/statusController.js
```diff
@@ -7,6 +7,8 @@ var angular = require('angular');
  * Status controller
  */
 angular.module('calcentral.controllers').controller('StatusController', function(activityFactory, apiService, badgesFactory, financesFactory, holdsFactory, $http, $scope, $q) {
+  $scope.finances = {};
```
johncrossman: @davidlimcheng For the sake of clarity, I'd like the declaration above to show the expected elements. I.e., 
```
$scope.finances = {
  carsFinances: {},
  csFinances: {}
}
```
If that doesn't play nice with template logic then ignore this comment. :)

src/assets/javascripts/angular/controllers/widgets/statusController.js
```diff
@@ -33,20 +35,55 @@ angular.module('calcentral.controllers').controller('StatusController', function
     }
   };
 
-  var loadFinances = function(data) {
+  var loadCarsFinances = function(data) {
     if (!data.summary) {
       return;
     }
+    $scope.finances.carsFinances = data.summary;
```
johncrossman: @davidlimcheng Another matter of personal style: an affirmative conditional is easier to read. Instead of `if not data.summary` I like:
```
if (data.summary) {
  $scope.finances.carsFinances = data.summary;
}
``` 
And getting rid of the `return` is nice.

src/assets/javascripts/angular/controllers/widgets/statusController.js
```diff
@@ -33,20 +35,55 @@ angular.module('calcentral.controllers').controller('StatusController', function
...
+  var loadCsFinances = function(data) {
+    if (!data.feed || !data.feed.summary) {
+      return;
+    }
+    $scope.finances.csFinances = data.feed.summary;
+  };
+
+  var parseFinances = function() {
+    var totalPastDueAmount = 0;
+    var minimumAmountDue = 0;
```
johncrossman: @davidlimcheng Can we do away with the vars above by using:
```
$scope.totalPastDueAmount = 0;
$scope.minimumAmountDue = 0;
```

src/assets/javascripts/angular/controllers/widgets/statusController.js
```diff
@@ -33,20 +35,55 @@ angular.module('calcentral.controllers').controller('StatusController', function
...
 
-    if (data.summary.totalPastDueAmount > 0) {
+    if (!$scope.finances.carsFinances && !$scope.finances.csFinances) {
+      return;
+    }
+    if ($scope.finances.carsFinances) {
+      cars = {
+        pastDue: $scope.finances.carsFinances.totalPastDueAmount,
+        minDue: $scope.finances.carsFinances.minimumAmountDue
+      };
```
johncrossman: @davidlimcheng  I would declare `cars` before this conditional with:
```
cars = {
  pastDue: 0,
  minDue: 0
}
```
Then update the values in this if block, as needed. Then, line 77 does not need to worry that the var was never set.

src/assets/javascripts/angular/controllers/widgets/statusController.js
```diff
@@ -33,20 +35,55 @@ angular.module('calcentral.controllers').controller('StatusController', function
...
+      totalPastDueAmount += cars.pastDue;
+      minimumAmountDue += cars.minDue;
+    }
+    if ($scope.finances.csFinances) {
+      cs = {
+        pastDue: $scope.finances.csFinances.pastDueAmount,
+        minDue: $scope.finances.csFinances.amountDueNow
+      };
```
johncrossman: @davidlimcheng Same as above.

johncrossman:
@davidlimcheng I've put a handful of comments. Once I get your replies I'll merge.  I'm running out of the office. I'll be back online at 5PM. Ping @raydavis if you need feedback sooner. Thanks for the good work!

src/assets/javascripts/angular/controllers/widgets/statusController.js
davidlimcheng: Much, much cleaner, good call.

davidlimcheng:
@johncrossman Went ahead and pushed the proposed changes.  Thanks a bunch for the in-depth review, lots of good stuff in this PR!

johncrossman:
Thanks, @davidlimcheng  Nice work!

PR #5620, opened 2016-07-06
SISRP-20755 	SISRP-15391 Greetings, Terms API; farewell, Tele-BEARS

raydavis:
https://jira.berkeley.edu/browse/SISRP-20755
https://jira.berkeley.edu/browse/SISRP-15391

There's still plenty of work to do on SISRP-20755. But given the centrality of Terms data in our codebase and the size of the changes I've made so far, it seems best to merge the work in stages. Let me know if you disagree!


app/models/terms/proxy.rb
```diff
@@ -0,0 +1,77 @@
+module Terms
+  class Proxy  < BaseProxy
+    include ClassLogger
+    include Proxies::AuthenticatedApi
+    include Proxies::Mockable
+
+    def initialize(options = {})
+      super(Settings.terms_proxy, options)
+      @temporal_position = options[:temporal_position]
+      @as_of_date = options[:as_of_date]
```
johncrossman: @raydavis can we have a line of documentation to explain the required-ness of these args. They're treated as optional in `def request_options` and yet they compose the `instance_key` .

app/models/terms/proxy.rb
```diff
@@ -0,0 +1,77 @@
...
+        get_internal
+      end
+    end
+
+    def get_internal
+      response = get_response(url, request_options)
+      # When the campus is between terms, a request for the 'Current' term will return 404.
+      if response.code == 404 &&
+        (!response['apiResponse'] || !response['apiResponse']['message'] ||
+        response['apiResponse']['message']['description'] != 'No term found for the given date')
```
johncrossman: @raydavis In this comparison of `description` I would expect `==` instead of `!=`. Am I wrong?

raydavis: The goal is to log "unexpected" 404s (caused by bad API addresses, for example) at a higher level than "expected" 404s (caused in the normal course of business when we are between two terms). We don't want to pollute the production log files with repeated bogus ERROR-level messages. I'll try to clarify with a comment.

johncrossman:
Thank you for clarification, @raydavis   Waiting on T

johncrossman:
Another nail in TeleBears' coffin.

PR #5621, opened 2016-07-06
NOJIRA Tests which hit real server endpoints should be confined to textext

raydavis:
Currently these are breaking Travis CI due to CLC-6386. That SSLError issue needs to be fixed, but even so a RAILS_ENV=test rspec run should not rely on reaching live servers.


johncrossman:
TLS = Time Lost Sleuthing   
Thanks, @raydavis

PR #5623, opened 2016-07-06
CLC-6386 Use TLS1.2 SSL for RSS feeds

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6386

(The new ``request_options`` mix-ins can also be applied to some existing proxy classes -- I just haven't gotten around to it yet...)


johncrossman:
Thanks @raydavis  Looks good

PR #5632, opened 2016-07-11
SISRP-21312 - Hide CARS card for new students without CARS data

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-21312


raydavis:
Thank you for the very helpful JIRA comments!

PR #5637, opened 2016-07-13
CLC-6395, only show alert when we have 1+ videos in list

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-6395

See screenshot in the Jira.


paulkerschen:
Revenge of the radio star. Thanks @johncrossman

PR #5640, opened 2016-07-14
SISRP-22490 - Add Both Current and Next RegStatus to Status & Holds

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-22490

* If either `current` or `next` are `null`, then hide that regstatus
* This does **not** change any logic on `academicsController.js` that conditionally shows/hide the "Status and Holds" card - that will come in a future refactoring PR
* Includes "My Academics" and "User Overview" versions of this card
* Moves shared code into a service
* Keeps legacy messaging, but that will likely turn into message catalog stuff in a future PR


raydavis:
Looks much cleaner, @davidlimcheng

PR #5642, opened 2016-07-15
SISRP-22634 Resolve disagreement over "current term" between legacy & CS

raydavis:
https://jira.berkeley.edu/browse/SISRP-22634

Blocker bug resulting from merger of legacy and CS terms data.

### DIFF
--- a/app/models/berkeley/terms.rb
+++ b/app/models/berkeley/terms.rb
@@ -98,6 +98,9 @@ module Berkeley
 
       @current = @running || future_terms.pop
       if (@next = future_terms.pop)
+        # Temporary workaround for Summer 2016, during which legacy SIS says the "current term" is Fall 2016 but CS SIS
+        # says the "current term" is Summer 2016.
+        @sis_current_term ||= @next if @current.legacy? && @current.is_summer && !@next.legacy?
         @future = future_terms.pop
         unless future_terms.empty?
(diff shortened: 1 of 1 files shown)


app/models/berkeley/terms.rb
```diff
@@ -98,6 +98,9 @@ def init
 
       @current = @running || future_terms.pop
       if (@next = future_terms.pop)
+        # Temporary workaround for Summer 2016, during which legacy SIS says the "current term" is Fall 2016 but CS SIS
+        # says the "current term" is Summer 2016.
+        @sis_current_term ||= @next if @current.legacy? and @current.is_summer and !@next.legacy?
```
johncrossman: @raydavis I suspect the use of `and` works but I'd expect `&&`.  Am I wrong that `and` and `or` are intended for control-flow? E.g., `obj.do or obj.die`

paulkerschen: @raydavis @johncrossman The Ruby Style Guide, for one, thinks that `and` and `or` have such unexpected precedence that it's just not worth it: https://github.com/bbatsov/ruby-style-guide#no-and-or-or

raydavis: Total gaffe on my part, guys. I'll fix ASAP.

raydavis:
Blunder fixed.

PR #5645, opened 2016-07-15
SISRP-22431 - 6.0 - Enrollment Verification Options Pt.1 - Feature Flag, Front-End

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-22431

* Feature-flag driven
* No links yet - links to come following Links API implementation + SSO to National Student ClearingHouse API
* Included both a fixture and a dummy JSON - a little overboard maybe?
* We're always going to want to grab the three messages once this page hits, so I just grab the whole message set instead of making three separate API calls


raydavis:
Overboard seems like a good idea in this case!

PR #5649, opened 2016-07-18
SISRP-20755 Use fake Hub Term API in default developer.yml

raydavis:
A little late following our convention on this one...


johncrossman:
Just like "krab", fake: true.

PR #5653, opened 2016-07-18
SISRP-22733 Fix Law-specific Class Enrollment sections

raydavis:
https://jira.berkeley.edu/browse/SISRP-22733

@redconfetti , this is another one you'll likely be interested by.


redconfetti:
@raydavis After the back-end refactor all this logic is going to be thrown out, but I surely want to make sure we don't re-introduce this issue. Hm...

raydavis:
@redconfetti , if you don't think it will do harm, please merge -- it may have to be part of an emergency release if it tests out.

PR #5657, opened 2016-07-20
SISRP-22857 - 6.0 - Class Enrollment - Refactor Back-end and Front-end

redconfetti:
https://jira.berkeley.edu/browse/SISRP-22857


src/assets/templates/widgets/enrollment/plan.html
```diff
@@ -1,67 +1,65 @@
...
 
-      <div data-ng-if="enrollment.academicPlan.updateLink && enrollmentTerm.academicPlan.classes.length">
-        <div class="cc-flex cc-flex-space-between">
-          <div>
-            <strong>Current Academic Plan for <span data-ng-bind="enrollmentTerm.termDescr"></span></strong>
-          </div>
+  <div data-ng-if="api.user.profile.actAsOptions.canSeeCSLinks">
+    <div data-ng-if="enrollment.academicPlan.updateLink && !enrollmentInstruction.academicPlan.classes.length">
+      Build your multi-year
+      <a data-cc-campus-solutions-link-directive="enrollment.academicPlan.updateLink.url"
```
redconfetti: I can see that I missed updating these references to the academic plan data.

app/models/my_academics/class_enrollments.rb
```diff
@@ -0,0 +1,148 @@
...
+        type_key = [type_code, career_code]
+        types[type_key] = { instruction_type_code: type_code, career_code: career_code, academic_plans: [] } if types[type_key].blank?
+        types[type_key][:academic_plans] << plan
+      end
+      types
+    end
+
+    def get_enrollment_instruction_type_code(plan)
+      type_codes = []
+      ENROLLMENT_INSTRUCTION_TYPES.keys.each do |cpp_category|
```
redconfetti: Ray says : If we use `each`, we should get key/value pairs, which can simplify this code.

raydavis:
As said in the walkthrough, @redconfetti might reconsider some naming and some minor code simplification, but basically this looks great. Thank you!

app/models/my_academics/class_enrollments.rb
redconfetti: @raydavis - I fixed this in #5657

PR #5659, opened 2016-07-20
SISRP-21766 Show Section ID as "Class Number", course code as "Course"

raydavis:
https://jira.berkeley.edu/browse/SISRP-21766

### DIFF
--- a/src/assets/templates/academics_classinfo.html
+++ b/src/assets/templates/academics_classinfo.html
@@ -53,7 +53,7 @@
               <thead>
                 <tr>
                   <th width="25%">Section</th>
-                  <th>CCN</th>
+                  <th>Class Number</th>
                 </tr>
               </thead>
               <tbody>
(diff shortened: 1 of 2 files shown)


src/assets/templates/academics_classinfo.html
```diff
@@ -117,6 +117,24 @@ <h4 data-ng-if="!isInstructorOrGsi && classScheduleCount.total">Section Schedule
             </div>
           </div>
 
+          <h3 data-ng-if="selectedCourse.sections.length && isInstructorOrGsi">Class Number</h3>
+          <div class="cc-table" data-ng-if="selectedCourse.sections.length && isInstructorOrGsi">
```
davidlimcheng: @raydavis , would it make sense to wrap both the `h3` and `div` into a parent `div` to avoid having two `data-ng-if`'s?

raydavis: I was curious about that myself, @davidlimcheng , but decided to just follow the pattern already established in the template. I'll give it a quick bogus-y test now & amend the PR if it looks OK.

PR #5667, opened 2016-07-22
SISRP-23011 - My Academics Feed - College and Level - Remove FPF Role Logic

redconfetti:
The Roles added to the MyAcademics::CollegeAndLevel feed are no longer necessary after the refactor done in [SISRP-22857](https://jira.berkeley.edu/browse/SISRP-22857). This is a follow up to remove this data which is no longer needed.

https://jira.berkeley.edu/browse/SISRP-23011


johncrossman:
Looks right to me, @redconfetti   Waiting on T

PR #5671, opened 2016-07-26
SISRP-22752 - Add CNP to Status and Holds, Update Popover

davidlimcheng:
Also includes:
* SISRP-23069 - Update Status and Holds on User Overview (Advisors)
* SISRP-23070 - Refactor My Academics for Status and Holds Card
* NOJIRA - Misc Status & Hold Unit Test Fixes

Stuff:
* Apologies for the huge PR.  
* I realize that we're using undergraduate Add/Drop date to drive CNP action for Grad/Law students.  This logic is due to the lack of a "Grad CNP" date from the Terms API, but it was a functional decision to have Grad/Law CNP action taken 5 weeks into the term, which fits perfectly with undergraduate add/drop date.
* The logic for what we show and when can get pretty complex.  The [Design JIRA](https://jira.berkeley.edu/browse/SISRP-13817) lays out things pretty well, which also links to a [Google Spreadsheet](https://docs.google.com/spreadsheets/d/1Pl_eS5uxKT-hjcaVFWknslqhrDlm8Tlj7iH22_DbW4k/edit#gid=0) that really maps out the logic really well.
* Once this gets in I'll open up a QA PR as well.

### DIFF
--- a/app/controllers/hub_edo_controller.rb
+++ b/app/controllers/hub_edo_controller.rb
@@ -19,7 +19,8 @@ class HubEdoController < ApplicationController
   end
 
   def student_attributes
-    json_passthrough HubEdos::MyStudentAttributes
+    model = HubEdos::MyStudentAttributes.from_session session
+    render json: model.get_feed_internal
   end
 
   def work_experience
--- a/app/controllers/hub_edo_controller.rb
+++ b/app/controllers/hub_edo_controller.rb
@@ -18,6 +18,10 @@ class HubEdoController < ApplicationController
     json_passthrough HubEdos::MyStudent, options
   end
 
+  def student_attributes
+    json_passthrough HubEdos::MyStudentAttributes
+  end
+
   def work_experience
     # Delegates get an empty feed.
     return {
--- a/app/models/berkeley/term.rb
+++ b/app/models/berkeley/term.rb
@@ -19,6 +19,8 @@ module Berkeley
     attr_reader :classes_end
     # The end of instruction is one week after classes end. The week between is RRR week. The week after is for final exams.
     attr_reader :instruction_end
+    # The end of the drop/add period, used for Cancellation for Non-Payment for Grad and Law students.
+    attr_reader :end_drop_add
     # BearFacts and related systems set "CT"/"Current Term" to Fall (and "FT"/"Future Term" to
     # the next year's Spring) as soon as the Spring term is over. Summer terms are special-cased as
     # "CS" or "FS", and lack some SIS support. This quirk becomes important when configuring
@@ -60,12 +62,18 @@ module Berkeley
         @classes_start = @start
         @classes_end = @end
         @instruction_end = @end
+        @end_drop_add = false
       else
         @is_summer = false
         session = term_feed['sessions'].first
         @classes_start = session['beginDate'].to_date.in_time_zone.to_datetime
         @instruction_end = session['endDate'].to_date.in_time_zone.to_datetime.end_of_day
         @classes_end = @instruction_end.advance(days: -7)
+        session['timePeriods'].each do |timePeriod|
+          if timePeriod['period']['code'] == '140'
+            @end_drop_add = timePeriod['endDate']
+          end
+        end
       end
       self
     end
--- a/app/models/berkeley/term.rb
+++ b/app/models/berkeley/term.rb
@@ -15,11 +15,13 @@ module Berkeley
     attr_reader :start
     # The Fall/Spring semesters end two weeks after the end of formal classes.
     attr_reader :end
+    # The first day of instruction.  This date is used for Cancellation for Non-Payment logic to trigger the change from a
+    # CNP notification to a CNP warning, as well as the drop date for undergraduates.
     attr_reader :classes_start
     attr_reader :classes_end
     # The end of instruction is one week after classes end. The week between is RRR week. The week after is for final exams.
     attr_reader :instruction_end
-    # The end of the drop/add period, used for Cancellation for Non-Payment for Grad and Law students.
+    # The end of the drop/add period, used for Cancellation for Non-Payment as the drop date for Grad and Law students.
     attr_reader :end_drop_add
     # BearFacts and related systems set "CT"/"Current Term" to Fall (and "FT"/"Future Term" to
     # the next year's Spring) as soon as the Spring term is over. Summer terms are special-cased as
@@ -41,6 +43,8 @@ module Berkeley
     end
 
     def from_cs_api(api_feed)
+      # The code used to identify end date of drop/add period in the Terms API.
+      end_drop_add_code = '140'
       @raw_source = api_feed
       # The HubTerm API returns an array of matching terms, one for each applicable academic career. For general
       # campus-wide reference, we pick the Undergraduate entry.
@@ -70,7 +74,7 @@ module Berkeley
         @instruction_end = session['endDate'].to_date.in_time_zone.to_datetime.end_of_day
         @classes_end = @instruction_end.advance(days: -7)
         session['timePeriods'].each do |timePeriod|
-          if timePeriod['period']['code'] == '140'
+          if timePeriod['period']['code'] == end_drop_add_code
             @end_drop_add = timePeriod['endDate']
           end
         end
--- a/app/models/my_registrations/my_registrations.rb
+++ b/app/models/my_registrations/my_registrations.rb
@@ -22,6 +22,7 @@ module MyRegistrations
     end
 
     def get_terms
+      berkeley_term = Berkeley::Term.new
       berkeley_terms = Berkeley::Terms.fetch
       terms = {}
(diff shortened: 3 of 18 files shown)


app/models/berkeley/term.rb
```diff
@@ -19,6 +19,8 @@ class Term
     attr_reader :classes_end
     # The end of instruction is one week after classes end. The week between is RRR week. The week after is for final exams.
     attr_reader :instruction_end
+    # The end of the drop/add period, used for Cancellation for Non-Payment for Grad and Law students.
+    attr_reader :end_drop_add
```
redconfetti: @davidlimcheng, I don't like the idea of this name indicating that the date provided applies to any term. Also your comment states that it's for Grad and Law Students. This is true because as you say, we're using it for Grad and Law students... but isn't this technically serving the undergrad add/drop deadline date?

Shouldn't we make this :end_drop_add_undergrad ?

I'm wanting to avoid the following:
1. Other developers assume that add/drop deadline dates apply to all terms for any career
1. Other developers assume that this deadline date applies to Grad and Law students, when really it's sourced from the undergrad deadline date.

redconfetti:
@davidlimcheng, wow - this is huge.

app/models/berkeley/term.rb
davidlimcheng: @redconfetti You're right, this is definitely serving the undergrad add/drop deadline date.  Taking a look at `Berkeley::Term` though, @raydavis left this very helpful message for developers (lines 43-44):

```The HubTerm API returns an array of matching terms, one for each applicable academic career. For general campus-wide reference, we pick the Undergraduate entry.```

Thus, it's assumed that all of the methods in this class serve undergraduate dates.  I wanted to avoid appending "undergrad" to this method so as to keep it consistent with all the other names in the class - i.e. instead of `classes_end_undergrad` we just show `classes_end`.  Its usage for grad/law students is also closer to the front-end than this point of the flow, so I'd hate to disrupt the current naming pattern at this point.

That being said, definitely open to changing it if it makes more sense to!

Comment (raydavis) [inline app/models/berkeley/term.rb]: Nicely summarized, @davidlimcheng . One addition: recognizing the possible (but unproven) usefulness of the other careers in the array, I stowed the full list in the "raw_source" method of Berkeley::Term, so grab it if it's useful!

File: app/models/berkeley/term.rb [hunk 2]
```diff
@@ -60,12 +62,18 @@ def from_cs_api(api_feed)
...
         @instruction_end = @end
+        @end_drop_add = false
       else
         @is_summer = false
         session = term_feed['sessions'].first
         @classes_start = session['beginDate'].to_date.in_time_zone.to_datetime
         @instruction_end = session['endDate'].to_date.in_time_zone.to_datetime.end_of_day
         @classes_end = @instruction_end.advance(days: -7)
+        session['timePeriods'].each do |timePeriod|
+          if timePeriod['period']['code'] == '140'
```
raydavis: Could you make this "magic number" a constant and document its source? I didn't find it in JIRA tickets right off.

app/models/berkeley/term.rb
raydavis: The comment is useful, @davidlimcheng . For symmetry, could you also add a comment about what's "used for Cancellation for Non-Payment for Undergraduate students"?

app/models/my_registrations/my_registrations.rb
```diff
@@ -97,6 +125,18 @@ def parse_legacy_term(term_name, term_id, term_english)
       HashConverter.camelize response
     end
 
+    def term_start(cs_feed)
+      Berkeley::Term.new.from_cs_api(cs_feed).classes_start
```
raydavis: This series of methods feels oddly roundabout. Couldn't the calling code just get the Berkeley::Term object _once_ and refer to its methods directly rather than instantiating it over and over?

app/controllers/hub_edo_controller.rb
```diff
@@ -18,6 +18,10 @@ def student
     json_passthrough HubEdos::MyStudent, options
   end
 
+  def student_attributes
+    json_passthrough HubEdos::MyStudentAttributes
```
redconfetti: Why are we using HubEdos::MyStudentAttributes when HubEdos::StudentAttributes already performs cacheing?

src/assets/javascripts/angular/controllers/pages/userOverviewController.js
```diff
@@ -157,6 +153,22 @@ angular.module('calcentral.controllers').controller('UserOverviewController', fu
           }
         }
       });
+    });
+  };
+
+  var loadStudentAttributes = function() {
+    studentAttributesFactory.getStudentAttributes({
```
raydavis: Would it make sense to DRY this shared logic into statusHoldsService along with the shared matchTermIndicators? (Not a merge blocker, of course!)

app/controllers/hub_edo_controller.rb
raydavis: True, two layers of cache for a single unaltered feed is one too many.

raydavis:
Phew, I think that does it for me! Thanks for this massive effort, @davidlimcheng .

FWIW, this is a much bigger and extended PR then we would normally like to see land on the QA branch so late, so we'll need to double check with everyone after the master-branch version gets finalized.

src/assets/javascripts/angular/controllers/pages/userOverviewController.js
```diff
@@ -157,6 +153,22 @@ angular.module('calcentral.controllers').controller('UserOverviewController', fu
...
+    }).success(function(data) {
+      var studentAttributes = _.get(data, 'feed.student.studentAttributes.studentAttributes');
+      // Strip all positive student indicators from student attributes feed.
+      _.forEach(studentAttributes, function(attribute) {
+        if (_.startsWith(attribute.type.code, '+')) {
+          $scope.regStatus.positiveIndicators.push(attribute);
+        }
+      });
+
+      statusHoldsService.matchTermIndicators($scope.regStatus.positiveIndicators, $scope.regStatus.registrations);
```
redconfetti: We should consider putting this in a service if the code is very similar between the two controllers, and then passing in the `regStatus`.

src/assets/javascripts/angular/controllers/widgets/academicsStatusHoldsBlocksController.js
```diff
@@ -6,14 +6,16 @@ var _ = require('lodash');
 /**
  * Academics status, holds & blocks controller
  */
-angular.module('calcentral.controllers').controller('AcademicsStatusHoldsBlocksController', function(apiService, profileFactory, slrDeeplinkFactory, residencyMessageFactory, registrationsFactory, statusHoldsService, $scope) {
+angular.module('calcentral.controllers').controller('AcademicsStatusHoldsBlocksController', function(apiService, enrollmentVerificationFactory, profileFactory, slrDeeplinkFactory, studentAttributesFactory, residencyMessageFactory, registrationsFactory, statusHoldsService, $scope) {
```
redconfetti: Possibly remove 'Blocks' from this controller name?

app/controllers/hub_edo_controller.rb
redconfetti: @davidlimcheng We should just get rid of HubEdos::MyStudentAttributes. Both Ray and I agree we should let the caching happen at the proxy layer. We only cache in the wrapping model when it's a merged feed, using un-cached proxies beneath it.

app/models/berkeley/term.rb
redconfetti: @davidlimcheng Yeah. I'm just wanting to make sure that people don't assume that this value is sourced from data for the 'GRAD' or 'LAW' careers for the term, when really it's sourced from the 'UGRD' career. I'd hate for this to be something that other developers miss because of your comment not being clear enough.

app/controllers/hub_edo_controller.rb
davidlimcheng: @redconfetti @raydavis Thanks guys.  I've kept `HubEdos::MyStudentAttributes` for its usefulness in grabbing the session uid, but removed the caching from that layer.

app/models/berkeley/term.rb
davidlimcheng: Added more details comments in `term.rb`.

app/models/my_registrations/my_registrations.rb
davidlimcheng: This here is a great lesson learned!

src/assets/javascripts/angular/controllers/pages/userOverviewController.js
davidlimcheng: Yes definitely!  Especially after our discussion of passing `$scope` to a service.  Will fix this in a future PR.

src/assets/javascripts/angular/controllers/pages/userOverviewController.js
davidlimcheng: Agreed!

PR #5676, opened 2016-07-26
SISRP-22703 - Add Link to CS Enrollment Verification

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-22703

* Design JIRA:  https://jira.berkeley.edu/browse/SISRP-15681
* The feature flag for `enrollment_verification_deeplink` will be flipped once the CS mod is done.  This has been approved for the no-fly zone, if necessary.
* Deeplink is hardcoded since we are transitioning to the Links API, but to my understanding the API won't be ready to be used until post GL6.
* In the meantime, we'll be showing the link to the Google doc.
* For testing, we are able to use the same Google Doc, with the request that we prepend "test-" in front of the test ID used.

### DIFF
--- a/app/controllers/campus_solutions/enrollment_verification_deeplink_controller.rb
+++ b/app/controllers/campus_solutions/enrollment_verification_deeplink_controller.rb
@@ -1,0 +1,9 @@
+module CampusSolutions
+  class EnrollmentVerificationDeeplinkController < CampusSolutionsController
+
+    def get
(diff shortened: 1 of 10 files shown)


redconfetti:
@davidlimcheng This is a GoLive 6 task as well?

davidlimcheng:
@redconfetti Yessir

PR #5680, opened 2016-07-26
[QA]	SISRP-22752 - Add CNP to Status and Holds, Update Popover

davidlimcheng:
QA PR of #5671

### DIFF
--- a/app/models/my_registrations/my_registrations.rb
+++ b/app/models/my_registrations/my_registrations.rb
@@ -22,6 +22,7 @@ module MyRegistrations
     end
 
     def get_terms
+      berkeley_term = Berkeley::Term.new
       berkeley_terms = Berkeley::Terms.fetch
       terms = {}
       # ':previous' and ':grading_in_progress' methods are not included here because Campus::Oracle does not keep information prior to the current term.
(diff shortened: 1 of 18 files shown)


app/models/my_registrations/my_registrations.rb
```diff
@@ -30,6 +31,19 @@ def get_terms
         term = berkeley_terms.send term_method
         if term
           terms[term_method] = {id: term.campus_solutions_id, name: term.to_english}
+          # For 'current' and 'next' terms, we need the date of start and end of instruction to determine CNP status
+          if (term_method == :current || term_method == :next)
+            temporal_position = term_method == :current ? HubTerm::Proxy::CURRENT_TERM : HubTerm::Proxy::NEXT_TERM
+            cs_feed = HubTerm::Proxy.new(temporal_position: temporal_position).get_term
```
raydavis: Sorry, @davidlimcheng , in the mad rush yesterday, I think this still turned out unnecessarily noisy. How about amending it as:
```
berkeley_term = Berkeley::Term.new.from_cs_api(cs_feed)
...
  classesStart: berkeley_term.classes_start,
  end: berkeley_term.end,
  endDropAdd: berkeley_term.end_drop_add
```

That's how the class was meant to be used.

davidlimcheng: Sure thing, @raydavis.  Apologies for the twice-incorrect usage.

raydavis: NP, the intended usage might have been clearer if I'd implemented a class-level factory method rather than an instance-level initializer method. I confess, I took the latter approach just to make test writing a bit easier.

PR #5681, opened 2016-07-26
SISRP-23107 - Ensure parsed feeds do not include raw EDO Hub response

redconfetti:
https://jira.berkeley.edu/browse/SISRP-23107

See QA PR #5683


raydavis:
Thank you for the find and the fix, @redconfetti !

I'm afraid this probably also calls for a QA PR...

PR #5688, opened 2016-07-27
[QA] SISRP-23165 - Enrollment Card Holds Fix - QA Only

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-23165

* This is a QA only PR, since our Master and QA branches deviate on how to handle `holds` for the Enrollment Card.
* This change is needed since coercing an empty array into a boolean results in a `true` value - as described [in this helpful article](http://james.padolsey.com/javascript/truthy-falsey/)
* The JIRA also contains another bug regarding registration status, which I will open in a separate PR that matches a fix for Master (more details in JIRA comments)

### DIFF
--- a/src/assets/javascripts/angular/controllers/widgets/enrollment/enrollmentCardController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/enrollment/enrollmentCardController.js
@@ -183,7 +183,7 @@ angular.module('calcentral.controllers').controller('EnrollmentCardController', 
   var loadHolds = function() {
     return academicStatusFactory.getAcademicStatus().then(function(data) {
       $scope.enrollment.holds.isLoading = false;
-      $scope.enrollment.holds.hasHolds = !!_.get(data, 'data.feed.student.holds');
+      $scope.enrollment.holds.hasHolds = !!(_.get(data, 'data.feed.student.holds')).length;
     });
   };
 


src/assets/javascripts/angular/controllers/widgets/enrollment/enrollmentCardController.js
```diff
@@ -183,7 +183,7 @@ angular.module('calcentral.controllers').controller('EnrollmentCardController',
   var loadHolds = function() {
     return academicStatusFactory.getAcademicStatus().then(function(data) {
       $scope.enrollment.holds.isLoading = false;
-      $scope.enrollment.holds.hasHolds = !!_.get(data, 'data.feed.student.holds');
+      $scope.enrollment.holds.hasHolds = !!(_.get(data, 'data.feed.student.holds')).length;
```
johncrossman: @davidlimcheng might `_.get(data, 'data.feed.student.holds')` be null?

raydavis:
Sorry about missing that, @davidlimcheng & @pfarestveit -- my "testing" was a little sketchy!

src/assets/javascripts/angular/controllers/widgets/enrollment/enrollmentCardController.js
davidlimcheng: @johncrossman The way the data is set up, the Hub will always return an array, regardless of whether or not the student has any holds:

```
{
   statusCode:200,
   feed:{
      student:{
         academicStatuses:[...],
         holds:[],
         awardHonors:[],
         degrees:[]
      }
   },
   studentNotFound:null
}
```
So, I believe `_.get(data, 'data.feed.student.holds')` will always = `[]` or `[...]`

johncrossman: @davidlimcheng Understood. But beware of assuming "...the Hub will always..."  The Hub might wake up grumpy and decide to exclude `holds` when empty. :)

PR #5693, opened 2016-07-28
SISRP-23167 - CNP notification not expanding for Grad Student

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-23167

Master PR, QA PR to follow - please merge QA first.

PR #5694, opened 2016-07-28
[QA FIRST] SISRP-23167 - CNP notification not expanding for Grad Student

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-23167


johncrossman:
Master PR is #5693  Both look right.

PR #5696, opened 2016-07-28
[QA] SISRP-23217 Enable GRADUATE, UNDERGRAD CS Affiliations

raydavis:
https://jira.berkeley.edu/browse/SISRP-23217

QA-FIRST PR!

This bug was caused by a disconnect between two levels of CalCentral code. One class added the CS Affiliations as User Roles, then a class farther up carefully filtered them out again.

The motivation behind this split-brain behavior is far in the past, and we can instead now use the UserRoles class as a single trusted source.

### DIFF
--- a/spec/models/user/api_spec.rb
+++ b/spec/models/user/api_spec.rb
@@ -550,10 +550,7 @@ describe User::Api do
     context 'role lookup errors' do
       let(:edo_attributes) do
         {
-          roles: {
-            body: 'An unknown server error occurred',
-            statusCode: 503
-          }
+          roles: {}
         }
       end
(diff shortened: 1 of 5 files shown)


spec/models/user/api_spec.rb
```diff
@@ -550,10 +550,7 @@
     context 'role lookup errors' do
       let(:edo_attributes) do
         {
-          roles: {
```
raydavis: This was based on obsolete code. The current implementation returns an empty hash instead.

johncrossman:
Looks good, @raydavis  Waiting on green Travis

PR #5697, opened 2016-07-28
[QA] SISRP-23183 - Graduate students - Exception from CNP -Incomplete text in expanded view

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-23183

* Start to pull from correct hub property, despite the hub pulling the wrong field from CS.
* So for now, this message will be hardcoded in the front-end.
* Also fixes a typo for `<div data-ng-if="!registration.positveIndicators.R99">` =(


raydavis:
Just between you and me, I kind of like being able to find the message text in source code...

PR #5698, opened 2016-07-28
SISRP-23175 - CNP notification should have Yellow Color icon

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-23175

* Apparently there's no such class as `cc-icon-yellow` and it should be `cc-icon-gold`!
* Found a couple of other places referencing `cc-icon-yellow` so I changed those as well.
* Should I open a QA PR for this?


johncrossman:
+1 @davidlimcheng

PR #5702, opened 2016-07-28
[QA] SISRP-23173 - Header popover shows CNP and Registration alert but no Status Block Card exists

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-23173

* Apologies for this, prior sloppy code on my part.  This is why coding shouldn't happen past 10pm =\

PR #5705, opened 2016-07-28
SISRP-23173 - Header popover shows CNP and Registration alert but no Status Block Card exists

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-23173

Master PR of #5702.


johncrossman:
Thx, @davidlimcheng Waiting on a green T.

PR #5708, opened 2016-07-28
SISRP-23117 Guard against overlarge Oracle "SELECT IN ()" queries

raydavis:
https://jira.berkeley.edu/browse/SISRP-23117

### DIFF
--- a/app/models/user/basic_attributes.rb
+++ b/app/models/user/basic_attributes.rb
@@ -5,9 +5,13 @@ module User
     def attributes_for_uids(uids)
       return [] if uids.blank?
       uid_set = uids.to_set
-      attrs = CampusOracle::Queries.get_basic_people_attributes(uids).map do |result|
-        uid_set.delete result['ldap_uid']
-        transform_campus_row result
+      attrs = []
+      # Oracle dislikes ' IN ()' queries with more than 1000 items.
+      uids.each_slice(1000) do |next_batch|
+        CampusOracle::Queries.get_basic_people_attributes(next_batch).each do |result|
+          uid_set.delete result['ldap_uid']
+          attrs << transform_campus_row(result)
+        end
       end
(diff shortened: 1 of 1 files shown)


app/models/user/basic_attributes.rb
```diff
@@ -5,9 +5,12 @@ module BasicAttributes
     def attributes_for_uids(uids)
       return [] if uids.blank?
       uid_set = uids.to_set
-      attrs = CampusOracle::Queries.get_basic_people_attributes(uids).map do |result|
-        uid_set.delete result['ldap_uid']
-        transform_campus_row result
+      # Oracle dislikes ' IN ()' queries with more than 1000 items.
+      while (next_batch = uids.slice!(0..999)).present?
```
paulkerschen: Oh, for heaven's sake, I thought we'd caught all of those. Thanks for bagging this one, @raydavis. What do you think about the idiomatic writing `uids.each_slice(1000) do |next_batch|`?

app/models/user/basic_attributes.rb
```diff
@@ -5,9 +5,12 @@ module BasicAttributes
...
+        attrs = CampusOracle::Queries.get_basic_people_attributes(next_batch).map do |result|
```
paulkerschen: @raydavis Also, perhaps I'm reading wrong, but doesn't the assignment of `attrs` within the loop mean that you end up with only the results of the last iteration? (Submit 1003 UIDs, get 3 results.)

app/models/user/basic_attributes.rb
raydavis: I'd think I missed it. Thanks, @paulkerschen , I'll amend.

app/models/user/basic_attributes.rb
raydavis: Ouch. I'll _really_ amend.

raydavis:
Peace on earth, good will toward men, amended.

PR #5711, opened 2016-07-29
SISRP-23228 - 6.0d - Biz logic in CampusSolutions::Link modifies the cached CS response (localhost only)

dfkaye:
https://jira.berkeley.edu/browse/SISRP-23228

### DIFF
--- a/app/models/campus_solutions/link.rb
+++ b/app/models/campus_solutions/link.rb
@@ -37,22 +37,32 @@ module CampusSolutions
       if url_id
         # Need this due to side-effect of the lift process in normalize_response.
         link_id = url_id.downcase.camelize(:lower).to_sym
-        link = link_resources[:feed][:ucLinkResources][:links][link_id]
+        entry = link_resources[:feed][:ucLinkResources][:links][link_id]
 
-        if link
+        if entry
+          link = entry.slice(:url, :urlId, :description)
+
           if placeholders.present?
             placeholders.each do |k, v|
               link[:url] = link[:url].gsub("{#{k}}", v)
             end
           end
 
-          # promote properties we're interested in, delete the rest
-          link[:properties].each do |property|
+          # promote properties we're interested in
+          entry[:properties].each do |property|
             if property[:name] == 'NEW_WINDOW' && property[:value] == 'Y'
               link[:showNewWindow] = true
             end
+            if property[:name] == 'UCFROM'
+              link[:ucFrom] = property[:value]
+            end
+            if property[:name] == 'UCFROMLINK'
+              link[:ucFromLink] = property[:value]
+            end
+            if property[:name] == 'UCFROMTEXT'
+              link[:ucFromText] = property[:value]
+            end
           end
-          link.delete(:properties)
         end
       end
 
--- a/spec/models/campus_solutions/link_spec.rb
+++ b/spec/models/campus_solutions/link_spec.rb
@@ -1,7 +1,8 @@
 describe CampusSolutions::Link do
+  let(:placeholder_text) { "REPLACED_PLACEHOLDER" }
   let(:proxy) { CampusSolutions::Link.new(fake: fake_proxy) }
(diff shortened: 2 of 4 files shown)


app/models/campus_solutions/link.rb
```diff
@@ -37,22 +37,23 @@ def get_url(url_id, placeholders = {})
       if url_id
         # Need this due to side-effect of the lift process in normalize_response.
         link_id = url_id.downcase.camelize(:lower).to_sym
-        link = link_resources[:feed][:ucLinkResources][:links][link_id]
+        entry = link_resources[:feed][:ucLinkResources][:links][link_id]
+
+        if entry
+          link = { :url => entry[:url], :urlId => entry[:urlId], :description => entry[:description] }
```
paulkerschen: @dfkaye FYI, the idiomatic Rails way to write this would be: `link = entry.slice(:url, :urlId, :description)`

That will create `link` as a new hash without modifying `entry`.

app/models/campus_solutions/link.rb
```diff
@@ -37,22 +37,23 @@ def get_url(url_id, placeholders = {})
...
           end
 
-          # promote properties we're interested in, delete the rest
-          link[:properties].each do |property|
+          # promote properties we're interested in
+          entry[:properties].each do |property|
             if property[:name] == 'NEW_WINDOW' && property[:value] == 'Y'
               link[:showNewWindow] = true
             end
           end
```
paulkerschen: @dfkaye In this case it would be a bit more efficient to use `find` rather than `each`, since `find` will stop iteration if it finds a match. You could write:

```
if entry[:properties].find { |property| property[:name] == 'NEW_WINDOW' && property[:value] == 'Y' }
  link[:showNewWindow] = true
end
```

app/models/campus_solutions/link.rb
dfkaye: @paulkerschen I'll make that change - thanks!

app/models/campus_solutions/link.rb
dfkaye: @paulkerschen OK - Q: what if we want to pull up another property down the road? go back to `each`?

paulkerschen: @dfkaye Yes, if we start checking for multiple properties, `each` would make sense.

dfkaye: @paulkerschen replaced with `find` - *we can always revisit...* ;)

app/models/campus_solutions/link.rb
```diff
@@ -37,22 +37,32 @@ def get_url(url_id, placeholders = {})
...
           if placeholders.present?
             placeholders.each do |k, v|
               link[:url] = link[:url].gsub("{#{k}}", v)
             end
           end
 
-          # promote properties we're interested in, delete the rest
-          link[:properties].each do |property|
+          # promote properties we're interested in
+          entry[:properties].each do |property|
```
redconfetti: @dfkaye, yeah if you're going to iterate over the entire list of properties at least once in any situation, even if you're converting this to a hash, it makes sense to do it this way.

spec/models/campus_solutions/link_spec.rb
```diff
@@ -1,7 +1,8 @@
 describe CampusSolutions::Link do
+  let(:placeholder_text) { "REPLACED_PLACEHOLDER" }
   let(:proxy) { CampusSolutions::Link.new(fake: fake_proxy) }
   let(:url_id) { "UC_CX_TEST_LINK" }
-  let(:placeholder_text) { "REPLACED_PLACEHOLDER" }
+  let(:url_id_for_properties) { "UC_CX_GT_FAEMRLAON_ADD" }
```
redconfetti: If I have a child, I think I might name it Faemrlaon.

> Make sure you finish eating your brocoli Faemrlaon, or else you'll get no dessert.

PR #5714, opened 2016-08-01
SISRP-23306 After many a summer dies the Tele-BEARS

raydavis:
https://jira.berkeley.edu/browse/SISRP-23306


davidlimcheng:
Summer 2016, we hardly knew ye.

PR #5715, opened 2016-08-02
SISRP-23303 Rename misleading "lookup_student_id" Crosswalk method

raydavis:
https://jira.berkeley.edu/browse/SISRP-23303

A bit of clarification before our code learns about CS-provided identifiers. This method actually only looks for a _legacy_ (pre-CS) Student ID.


paulkerschen:
Thanks @raydavis.

PR #5716, opened 2016-08-02
SISRP-23473 Class calendar integration: query LDAP for campus email if not otherwise available

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-23473

This code will not be exercised often, but for Oski and a few others it makes all the difference in the world.


raydavis:
Merge all the things.

PR #5735, opened 2016-08-04
SISRP-22142 - Student Advising factory and card boilerplate, mechanism to disable L & S Advising card

redconfetti:
* Introduces myAdvisorFactory
* Introduces boilerplate for Advising Card showing up on My Academics page
* 'advising' feature flag will now override the L&S Advising card (which is enabled via `legacy_advising`) feature flag.

https://jira.berkeley.edu/browse/SISRP-22142

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -474,6 +474,7 @@ features:
   cs_transfer_credit: false
   hub_term_api: true
   enrollment_verification: false
+  legacy_advising: false
   legacy_regblocks: true
   manage_site_mailing_lists: false
   my_fin_aid: false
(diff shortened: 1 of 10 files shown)


config/settings.yml
```diff
@@ -474,6 +474,7 @@ features:
   cs_transfer_credit: false
   hub_term_api: true
   enrollment_verification: false
+  legacy_advising: false
```
paulkerschen: @redconfetti Should we change the feature flag check in MyAdvising::LegacyAdvising to match?

redconfetti: @paulkerschen Just fixed this in a follow up commit

paulkerschen:
Thanks much @redconfetti.

PR #5736, opened 2016-08-04
SISRP-22717 Merged model and controller for advising card

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-22717

We've been advised that the sis:staff:advisor ENF topic, which currently expires advising resources, will be used to expire advising appointments as well.


raydavis:
Another day, another integration point... Thanks, @paulkerschen !

PR #5738, opened 2016-08-04
SISRP-22619 - 6.0 - Add student-specific links to Personal Summary card

dfkaye:
https://jira.berkeley.edu/browse/SISRP-22619

### DIFF
--- a/app/controllers/advising_student_controller.rb
+++ b/app/controllers/advising_student_controller.rb
@@ -44,7 +44,8 @@ class AdvisingStudentController < ApplicationController
     links = json[:feed] && json[:feed][:ucAdvisingResources] && json[:feed][:ucAdvisingResources][:ucAdvisingLinks]
     if links
       # Advisors get only a subset of links
-      keys = [:ucServiceIndicator, :ucStudentAdvisor, :multiYearAcademicPlannerStudentSpecific, :schedulePlannerStudentSpecific]
+      keys = [:ucServiceIndicator, :ucStudentAdvisor, :multiYearAcademicPlannerStudentSpecific, :schedulePlannerStudentSpecific,
+              :studentAppointments, :studentAdvisorNotes, :studentWebnowDocuments]
       advising_links = links.select { |key| keys.include? key }
       json[:feed][:ucAdvisingResources][:ucAdvisingLinks] = advising_links
     end
--- a/app/models/campus_solutions/advising_resources.rb
+++ b/app/models/campus_solutions/advising_resources.rb
@@ -42,25 +42,55 @@ module CampusSolutions
         add_cs_link links, :multi_year_academic_planner_generic, 'MULTI_YEAR_ACADEMIC_PLANNER_GENERIC', 'Multi-Year Planner'
         add_cs_link links, :multi_year_academic_planner, 'MULTI_YEAR_ACADEMIC_PLANNER_STUDENT_SPECIFIC', 'Multi-Year Planner', "?UCemplid=#{lookup_student_id}"
         add_cs_link links, :schedule_planner, 'SCHEDULE_PLANNER_STUDENT_SPECIFIC', 'Schedule Planner', "?EMPLID=#{lookup_student_id}"
-
+        
         # LINK-API CALLS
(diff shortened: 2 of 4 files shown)


dfkaye:
Ready for review - Link URL table definitions have been migrated to QAT.

app/controllers/advising_student_controller.rb
```diff
@@ -46,6 +46,32 @@ def resources
       # Advisors get only a subset of links
       keys = [:ucServiceIndicator, :ucStudentAdvisor, :multiYearAcademicPlannerStudentSpecific, :schedulePlannerStudentSpecific]
       advising_links = links.select { |key| keys.include? key }
+
+      # LINK-API CALLS
+      appointment_system = CampusSolutions::Link.new.get_url('UC_CX_APPOINTMENT_STD_MY_APPTS', {
+        :EMPLID => "#{student_uid_param}"
+      })
+      appointment_system_link = appointment_system[:feed][:link]
+      if appointment_system_link
```
davidlimcheng: If we expect `appointment_system_link` to be empty sometimes (using context clues from the line below), would it make sense to use the `.try` method here so that we don't error out?  E.g. `appointment_system_link = appointment_system.try(:[], :feed).try(:[], :link)

paulkerschen: I agree with @davidlimcheng that null check are a good idea here. `try` is one option; explicit checks on each element is another.

```
if appointment_system && appointment_system[:feed]
  advising_links[:ucAppointmentSystem] = appointment_system[:feed][:link]
end
```

If empty links are unexpected enough that they should be logged as errors, a shared method like this one might be useful:
https://github.com/ets-berkeley-edu/calcentral/blob/master/app/models/advising/my_advising.rb#L79

dfkaye: @davidlimcheng @paulkerschen - CS Link always returns
```{ 
    status: <some status>, 
    feed: { link: <which could be nil>
}}```

The logging suggestion makes sense - but there are only two such link calls in this file. <del>If</del> *When* we add more links, that would be the time to encapsulate these in a helper.

Comment (raydavis) [inline app/controllers/advising_student_controller.rb]: And to complete our tour of the wild world of nil-safe deep hash references, there's:

```
response.fetch('apiResponse', {}).fetch('response', {}).fetch('terms', nil)
```

Comment (paulkerschen) [inline app/controllers/advising_student_controller.rb]: That's a good 'un too, @raydavis, though `fetch` has burned me before by being safe on an absent key, but not on a present key that happens to have a nil value.

**Comment by dfkaye** [comment]
@davidlimcheng @paulkerschen Ready for review again...

File: app/models/campus_solutions/advising_resources.rb [hunk 2]
```diff
@@ -44,25 +44,53 @@ def build_feed(response)
...
         })
-        appointment_system_link = appointment_system[:feed][:link]
         if appointment_system_link
-          links['uc_appointment_system'] = appointment_system_link
+          links[:uc_appointment_system] = appointment_system_link
+        end
+
+        # STUDENT-SPECIFIC LINKS
+        student_appointments_link = fetch_link('UC_CX_APPOINTMENT_STD_MY_APPTS', {
+          :EMPLID => "#{lookup_student_id}"
```
paulkerschen: @dfkaye Here and below, the result of `lookup_student_id` (which was called at the top of `get_internal`) will be stored in an instance variable at `@student[:campus_solutions_id]`. So you can just use that instance variable, instead of calling back out to the lookup method.

raydavis: Having spent some time last week looking at that innocuous sounding method, I have to +1 @paulkerschen 's comment: it fetches an object over and over from the cache every time it's called. Cache fetches are cheap, but they aren't _that_ cheap.

dfkaye: @raydavis @paulkerschen - making that change now - thanks

dfkaye: @raydavis @paulkerschen this one's ready (and passing travis) ~ I'll make a QA PR once this is merged to master ~ thanks

paulkerschen:
Looks good, @dfkaye, thanks.

PR #5742, opened 2016-08-05
SISRP-22720 Advising card front end

paulkerschen:
https://jira.berkeley.edu/browse/SISRP-22720

This PR assumes that the advising card will show to undergraduates only. If this assumption doesn't pass validation, a last fix may follow.


redconfetti:
@paulkerschen, The original L&S Advising card was restricted to only L&S students. I asked Chara is this one is meant for more than just L&S students, and she said yes. You might want to check in with her on the restrictions of this card.

Is it possible to be an undergrad but not in an L&S program?

PR #5756, opened 2016-08-08
SISRP-23757 - Ensure Summer 2016 Regstatus Expires after End of Instruction

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-23757

* Since the Terms API will start bumping Summer 2016 out of "Current" and replace it with Fall 2016, this problem of not expiring visibility for Summer regstatus should actually take care of itself - this is just a precaution.
* Let me know if I should make a QA PR for this as well, since Summer 2016 term ends on 8/12 according to the Terms API.  But as I had stated, this problem should take care of itself.


davidlimcheng:
Sounds good @raydavis , will open one shortly.

PR #5760, opened 2016-08-09
SISRP-23714 Handle Hub EDOs feeds which lack identifiers

raydavis:
https://jira.berkeley.edu/browse/SISRP-23714

This is a very focused fix for the regression in QA. It leaves the underlying issue (loss of the noStudentId flag from component Hub feeds) in place.

### DIFF
--- a/app/models/hub_edos/user_attributes.rb
+++ b/app/models/hub_edos/user_attributes.rb
@@ -69,12 +69,13 @@ module HubEdos
       # CS Identifiers simply treat 'student-id' as a synonym for the Campus Solutions ID / EmplID, regardless
       # of whether the user has ever been a student. (In contrast, CalNet LDAP's 'berkeleyedustuid' attribute
       # only appears for current or former students.)
-      if (cs_id_hash = edo[:identifiers].select {|id| id[:type] == 'student-id'}.first)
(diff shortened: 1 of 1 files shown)


app/models/hub_edos/user_attributes.rb
```diff
@@ -69,12 +69,13 @@ def extract_ids(edo, result)
       # CS Identifiers simply treat 'student-id' as a synonym for the Campus Solutions ID / EmplID, regardless
       # of whether the user has ever been a student. (In contrast, CalNet LDAP's 'berkeleyedustuid' attribute
       # only appears for current or former students.)
-      if (cs_id_hash = edo[:identifiers].select {|id| id[:type] == 'student-id'}.first)
+      identifiers = edo[:identifiers]
+      if identifiers.present? && (cs_id_hash = identifiers.select {|id| id[:type] == 'student-id'}.first)
```
paulkerschen: @raydavis Is the blank-identifiers case worth logging an error, or shall we wait for the less focused fix?

raydavis: @kerschen: Blank identifiers are logged by the component feeds, so that would be double-dipping

paulkerschen:
Thanks @raydavis.

PR #5775, opened 2016-08-10
SISRP-23764 - Academics > Advising > Appointment Self Service - Return to CalCentral link is not working

dfkaye:
https://jira.berkeley.edu/browse/SISRP-23764

### DIFF
--- a/app/models/campus_solutions/link.rb
+++ b/app/models/campus_solutions/link.rb
@@ -68,6 +68,7 @@ module CampusSolutions
           link[:name] = link.delete(:description)
           link[:title] = link.delete(:comments)
           if !link[:showNewWindow]
+            link[:isCsLink] = true
             link['IS_CS_LINK'] = true
           end
         end
--- a/app/models/advising/my_advising.rb
+++ b/app/models/advising/my_advising.rb
@@ -78,7 +78,7 @@ module Advising
 
     def fetch_link(link_key)
       if (link_feed = CampusSolutions::Link.new.get_url link_key)
-        link = link_feed.try(:[], :feed).try(:[], :link).try(:[], :url)
+        link = link_feed.try(:[], :feed).try(:[], :link)
       end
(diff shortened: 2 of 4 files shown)


app/models/campus_solutions/link.rb
```diff
@@ -68,6 +68,7 @@ def get_url(url_id, placeholders = {})
           link[:name] = link.delete(:description)
           link[:title] = link.delete(:comments)
           if !link[:showNewWindow]
+            link[:isCsLink] = true
             link['IS_CS_LINK'] = true
```
raydavis: Is the "IS_CS_LINK" property used by any higher-level code, @dfkaye ? I don't see it mentioned anywhere as a raw string.

dfkaye: @raydavis Yes - the link collection that gets built in app/models/campus_solutions/advising_resources.rb apparently camelCases it

raydavis: Sorry to bug you again, @dfkaye , but if it camelCased "IS_CS_LINK", wouldn't that have the same effect as if the "IS_CS_LINK" was replaced by "isCsLink"? I guess I'm just wondering, if you took "IS_CS_LINK" out, would you end up with the same payload going to the browser?

dfkaye: @raydavis I tried that - it turned `:isCsLink` into `iscslink`

raydavis: Thanks, @dfkaye . Pretty confusing situation -- we should probably try to clear this redundancy out of the pipes at some point, so as not to have developers copy-and-pasting "isCsLink" and "IS_CS_LINK" combos all over the codebase...

paulkerschen: +1 @raydavis - I for one would support a convention that symbolizes-and-camelizes string keys in these feeds, but leaves symbol keys alone.

PR #5780, opened 2016-08-10
CLC-6425 Add pagination to Canvas Course Sections API call

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6425

As usual when Instructure adds pagination, we found out by our payloads suddenly being restricted to 10 items.


paulkerschen:
How nice to see our code conventions doing their job! Thanks @raydavis.

PR #5781, opened 2016-08-10
SISRP-23981 - Advising Card - Restrict view from UCBX students

redconfetti:
https://jira.berkeley.edu/browse/SISRP-23981

Still awaiting confirmation that this is a blocker for GL 6.0d in [SISRP-23981](https://jira.berkeley.edu/browse/SISRP-23981)


raydavis:
Thanks, @redconfetti . We already provide a concurrentEnrollmentStudent user role in api/my/status (along with the ClassEnrollments version), so I just want to do a quick check on what the relation (if any) is -- BRB...

raydavis:
CalCentral's basic user attributes feed at api/my/status bases the user's "concurrentEnrollmentStudent" role on the CalNet 'AFFILIATE-TYPE-CONCURR ENROLL' affiliation. That user role is already used on My Academics in the Profile card and the Status & Holds card.

This PR adds a new "isConcurrentStudent" property to the JavaScript scope shared by most of the My Academics cards. This property comes from the CS Academic Status API, and comes to the browser via the CollegeAndLevel section of My Academics, which is _also_ used by the Profile card.

What happens if the two sources conflict?

PR #5793, opened 2016-08-13
SISRP-24173 Fix term-transition loss of registration status feed

raydavis:
https://jira.berkeley.edu/browse/SISRP-24173


raydavis:
EXPLANATION: All available terms are already in the Berkeley::Terms object. That logic has the smarts to know how to handle the between-terms situation so that CalCentral always has a "current" term even when the SIS does not (because of being between the official end of one term and the official beginning of the other).

When MyRegistrations unnecessarily tried to recreate the "current" term in its own code, it missed that nuance.

PR #5801, opened 2016-08-15
SISRP-24217 Switch to LDAP as preferred source for bulk attributes

raydavis:
https://jira.berkeley.edu/browse/SISRP-24217

We already prefer LDAP for the more detailed single-person attributes.


app/models/user/basic_attributes.rb
```diff
@@ -6,14 +6,19 @@ def attributes_for_uids(uids)
       return [] if uids.blank?
       uid_set = uids.to_set
       attrs = []
-      # Oracle dislikes ' IN ()' queries with more than 1000 items.
-      uids.each_slice(1000) do |next_batch|
-        CampusOracle::Queries.get_basic_people_attributes(next_batch).each do |result|
-          uid_set.delete result['ldap_uid']
-          attrs << transform_campus_row(result)
+      # Primary source for bulk attributes is CalNet LDAP.
+      attrs.concat CalnetLdap::UserAttributes.get_bulk_attributes(uid_set)
```
paulkerschen: @raydavis Minor, but at this point might we as well get rid of the empty-array initialization and just write `attrs = CalnetLdap::UserAttributes.get_bulk_attributes(uid_set)`?

paulkerschen: @raydavis Also looks like `CalnetLdap::UserAttributes.get_bulk_attributes` has a chance of returning nil, so we should probably fall back to an empty array either here or in that method.

paulkerschen:
Looks good; a couple of minor comments. Also, with bulk LDAP now on the front lines, might be time to pursue this TODO through queries and/or experimental benchmarking of our own:
https://github.com/ets-berkeley-edu/calcentral/blob/master/app/models/calnet_ldap/client.rb#L12

raydavis:
Thank you for the very helpful review, @paulkerschen -- amendment & CalNet query coming.

raydavis:
Have to close this while we sort out performance issues.

PR #5803, opened 2016-08-15
SISRP-13602 - UX DESIGN: Lack of Visual Arrow Indicator on Activity Card in Billing Details Page

dfkaye:
https://jira.berkeley.edu/browse/SISRP-13602

**Not ready for merge** - renaming-refactor coming soon...

Adds cross-browser arrow/triangle toggle solution to data table `<tbody>` tags in the following:
- billing/details
- academics (status holds)
- finaid table directive
- my finances account (cars) activity

### DIFF
--- a/src/assets/stylesheets/_widgets.scss
+++ b/src/assets/stylesheets/_widgets.scss
@@ -11,11 +11,11 @@
 
   // See https://git.io/vITBH
   // For an explanation of this magic
-  &:before, &:after {
+  &::before, &::after {
     content: " ";
     display: table;
   }
-  &:after {
+  &::after {
     clear: both;
   }
 }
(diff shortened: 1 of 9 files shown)


davidlimcheng:
@dfkaye This looks great to me.  I do have a few changes I'd like to make to the Status & Holds card (to keep consistent with the work I showed Flint yesterday), but I can open that PR.  Also, I'm not sure if we should be implementing this to the finaid table directive, just because it that directive touches so many cards where real estate can already be pretty full.  For example:  
screen shot 2016-08-16 at 3 18 14 pm

I'd recommend removing it from the directive for now, but I'd also like to see what @redconfetti and @raydavis think about it too.

raydavis:
I'd defer to Flint or another designer on that question, @davidlimcheng & @dfkaye . For the code piece, your and/or @redconfetti 's reviews are good enough for me!

davidlimcheng:
@raydavis Makes sense!  @dfkaye In my opinion, I think it'd be a good idea to go over these tables with Flint before merging.

dfkaye:
@davidlimcheng - I've visited this with Flint and flattened out as much as I can - have another look.

@redconfetti - What do you think of this solution for getting expand/collapse arrows on `<tbody>` elements? The CSS is used in multiple places so there are... interesting workarounds...

src/assets/stylesheets/_widgets.scss
```diff
@@ -11,11 +11,11 @@
 
   // See https://git.io/vITBH
   // For an explanation of this magic
-  &:before, &:after {
+  &::before, &::after {
```
redconfetti: [Bye bye IE8](https://css-tricks.com/almanac/selectors/a/after-and-before/#article-header-id-0)

src/assets/stylesheets/_widgets.scss
```diff
@@ -193,6 +193,58 @@
...
+    .cc-toggle-triangle {
+      &::after {
+        border-bottom: 4px solid transparent;
+        border-right: 4px solid $body-font-color;
+        border-top: 4px solid transparent;
+      }
+    }
+  }
+}
+.cc-widget-table-hover-opened {
```
redconfetti: @dfkaye This looks a lot like .cc-widget-table-hover. Could we dry this up using [SCSS extend](http://sass-lang.com/guide#topic-7)?

redconfetti:
@dfkaye I'd like to review this with someone that understand's the approach you're taking here, within the context of the existing triangle classes that we use with list elements.

src/assets/stylesheets/_widgets.scss
dfkaye: @redconfetti - I added the second colon because the atom editor plugin for scss files keeps shouting that both colons are necessary. Should I remove the extra colons?

redconfetti: @dfkaye I would check to see if IE8 isn't a supported browser any longer. If not, I have no problem with this.

PR #5806, opened 2016-08-17
[QA FIRST] SISRP-24387 - UX DEV: Status and Holds and Popover (CNP Emergency Update)

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-24387

* Lots of logic additions to cater to the CNP extension
* Also found a Promise bug within the Advisors User Overview page that would run a certain function before finishing all API calls/parsing

### DIFF
--- a/src/assets/javascripts/angular/controllers/pages/userOverviewController.js
+++ b/src/assets/javascripts/angular/controllers/pages/userOverviewController.js
@@ -136,8 +136,8 @@ angular.module('calcentral.controllers').controller('UserOverviewController', fu
   var loadRegistrations = function() {
     advisingFactory.getStudentRegistrations({
       uid: $routeParams.uid
-    }).success(function(data) {
-      _.forOwn(data.terms, function(value, key) {
+    }).then(function(data) {
+      _.forOwn(data.data.terms, function(value, key) {
         if (key === 'current' || key === 'next') {
           if (value) {
             $scope.regStatus.terms.push(value);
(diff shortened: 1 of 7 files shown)


src/assets/javascripts/angular/controllers/pages/userOverviewController.js
```diff
@@ -136,16 +136,16 @@ angular.module('calcentral.controllers').controller('UserOverviewController', fu
   var loadRegistrations = function() {
     advisingFactory.getStudentRegistrations({
       uid: $routeParams.uid
-    }).success(function(data) {
-      _.forOwn(data.terms, function(value, key) {
+    }).then(function(data) {
+      _.forOwn(data.data.terms, function(value, key) {
```
raydavis: Sorry, @davidlimcheng , I'm feeling dim at this point. The data.data is to fix a bug unrelated to the CNP extension?

redconfetti: Okay @davidlimcheng explained to @raydavis and I that the need to use `data.data` is because of success() method removing some sort of wrapper.

davidlimcheng: @raydavis:  I had to change `.success()` to `.then()` so that `loadStudentAttributes` would have data available in `$scope.regStatus.registrations`.  And as you said, `.success()` unwraps the data, whereas `.then()` does not - so I did have to add append an extra `.data` here.

raydavis: Thanks, @davidlimcheng !

PR #5809, opened 2016-08-18
SISRP-24414 - Hide My Academics - Advising widget for Law students

redconfetti:
https://jira.berkeley.edu/browse/SISRP-24414

See QA PR #5810


raydavis:
Yep. More pedantically, a "law" user role results if Hub Affiliations returns a ['students']['affiliations'] entry with status "ACT" and type "LAW".

raydavis:
There are other ways to decide whether someone is a law student (see StudentInfo's isLawStudent), but Status & Holds is already using the user profile version, so this card might as well too.

redconfetti:
@raydavis I agree. This is just an attempt to quickly get something in to reduce support tickets. The ultimate solution we're looking to implement is an actual extraction of the type of logic we have applied to the Class Enrollment card to determine which type of student they are in each active term... and apply it to the MyStatus API data feed, so that we can use it as needed in other cards.

PR #5813, opened 2016-08-22
SISRP-23500 - 7.0 - My Academics - Semesters (Law & Unofficial Transcript)

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-23500

* Since the design calls for the link to be disabled for delegates (not shown in the JIRA, but discussed w/ Chara), I moved the delegate logic over to the HTML.

### DIFF
--- a/src/assets/templates/academics_semesters.html
+++ b/src/assets/templates/academics_semesters.html
@@ -1,6 +1,17 @@
 <div class="cc-widget">
   <div class="cc-widget-title">
     <h2 class="cc-left">Semesters</h2>
+    <span class="cc-right cc-widget-title-link">
+      <a data-ng-if="!api.user.profile.delegateActingAsUid && !isAdvisingStudentLookup"
+        data-ng-href="{{requestTranscript.link}}"
+        data-ng-attr-title="{{requestTranscript.title}}">
+        <strong>Request Transcript</strong>
+      </a>
+      <a class="cc-academics-link-disabled"
+        data-ng-if="api.user.profile.delegateActingAsUid">
+        <strong>Request Transcript</strong>
+      </a>
+    </span>
   </div>
 
   <div class="cc-academics-enrollment-verification-link" data-ng-if="api.user.profile.features.enrollmentVerification && (api.user.profile.isDirectlyAuthenticated || api.user.profile.actingAsUid) && !isAdvisingStudentLookup">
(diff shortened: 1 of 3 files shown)


src/assets/templates/academics_semesters.html
```diff
@@ -1,6 +1,17 @@
...
   <div class="cc-widget-title">
     <h2 class="cc-left">Semesters</h2>
+    <span class="cc-right cc-widget-title-link">
+      <a data-ng-if="(api.user.profile.isDirectlyAuthenticated || api.user.profile.actingAsUid) && !isAdvisingStudentLookup"
+        data-ng-href="{{requestTranscript.link}}"
+        data-ng-attr-title="{{requestTranscript.title}}">
+        <strong>Request Transcript</strong>
+      </a>
+      <a class="cc-academics-link-disabled"
+        data-ng-if="api.user.profile.delegateActingAsUid">
```
davidlimcheng: I was contemplating having this logic here instead:
```
data-ng-if="!(api.user.profile.isDirectlyAuthenticated || api.user.profile.actingAsUid) && api.user.profile.delegateActingAsUid"
```
But I figured this was overkill and it was just cluttering the HTML.  Thoughts?  Do you think having that extra bit of logic in there is more secure, or is it just a redundant check?  I ask because I've run into this a couple times, and would like to know the preference of our team for future features.

redconfetti: @davidlimcheng - When `delegateActingAsUid` is true, is `!isAdvisingStudentLookup` also true? Is the delegate experience supposed to mirror the advising experience?

raydavis: Your sample alternative looks like redundant noise to me, @davidlimcheng . In fact, I wonder if a more readable if-clause for the active link might be
```
!api.user.profile.delegateActingAsUid && !isAdvisingStudentLookup
```

raydavis: @redconfetti , the intent is that these flags indicate incompatible states.

davidlimcheng: @raydavis That is much more readable, and I think it covers all of our bases.  Will amend now.

davidlimcheng:
Amended!

src/assets/templates/academics_semesters.html
redconfetti: @davidlimcheng - This is good now, but I'm curious if it should be the following?
```
<a class="cc-academics-link-disabled" data-ng-if="api.user.profile.delegateActingAsUid || isAdvisingStudentLookup">
```

Do we want the unlinked text to also show for advisors?

davidlimcheng: @redconfetti Good question - according to Chara, the link shouldn't show up at all for advisors on the User Overview page.

PR #5817, opened 2016-08-22
SISRP-22618 - 7.0 - Student Overview - Personal Summary (Graduation Terms for Undergraduates)

redconfetti:
https://jira.berkeley.edu/browse/SISRP-22618


redconfetti:
Essentially I had to move the logic needed to determine the expected graduation term logic to the `academicsService`, so that it can be shared between the AcademicProfileController and the UserOverviewController

raydavis:
Makes sense. Thanks, @redconfetti

PR #5824, opened 2016-08-24
SISRP-19059 HubEdos cleanup before refactoring

raydavis:
https://jira.berkeley.edu/browse/SISRP-19059

I'm submitting fairly fine-grained PRs in case we need to cherry-pick any commits into an emergency release. See the commit messages for details.

### DIFF
--- a/app/models/hub_edos/registrations.rb
+++ b/app/models/hub_edos/registrations.rb
@@ -1,4 +1,6 @@
 module HubEdos
+  # For reasons unknown, Registrations seems to be the one HubEdos payload which doesn't include a "students" wrapper,
+  # which is why this class doesn't inherit from HubEdos::Student.
   class Registrations < Proxy
 
     def initialize(options = {})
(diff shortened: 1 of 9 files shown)


raydavis:
Travis caught an RSpec failure I missed the last time I ran locally. Amending...

raydavis:
Should be ready for review...

app/models/hub_edos/registrations.rb
```diff
@@ -13,26 +15,5 @@ def json_filename
...
-
-    def build_feed(response)
-      resp = parse_response response
-      get_students(resp)
-    end
-
-    def wrapper_keys
-      %w(apiResponse response any)
-    end
-
```
davidlimcheng: @raydavis Just trying to clarify my understanding of this - are we removing this code since the superclass `Proxy` already provides these methods?

raydavis: Yes, most of this code is taken care of by ``HubEdos::Proxy``. The ``smart_fetch_from_cache`` should have been taken out back when the MyRegistrations class was added, but got missed.

raydavis:
As an editorial aside, I personally find the ResponseHandler + Proxy + Student combo maybe a bit _too_ fine-grained to make for easy reading. But since that's a matter of taste, I left it in place and just tried to make the code match the class/module names better.

app/models/hub_edos/registrations.rb
davidlimcheng: Ah, gotcha.

davidlimcheng:
LGTM, @raydavis.

PR #5826, opened 2016-08-24
SISRP-19059 Add caching wrappers for Affiliations, Contacts, Demographics

raydavis:
Continuing the https://jira.berkeley.edu/browse/SISRP-19059 saga.

(The next step will be changing User::Api and its friends to use only the Hub endpoints that are needed.)

This PR also gets rid of a never-used (but distracting) piece of front-end wishful thinking.


davidlimcheng:
LGTM, @raydavis.  I wonder why `academic_standing.html` was never implemented.

raydavis:
Thanks, @davidlimcheng . The unused Academic Standing UX sure looks like it would be redundant with the My Academics page, so I can see wanting to leave it out.

PR #5831, opened 2016-08-26
SISRP-24774 - Concurrent Enrollment Student (Sep 24th) - shouldn't display any REG status.

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-24774

Notes for this PR:
* Since we are showing up to two terms (current and next), we need to enable the "Status" section to show if either of these terms are applicable to show registration status.  The changes in `$scope.$watchGroup` reflects this need.
* Uses term-specific career code "UCBX" to determine whether or not we should show that term's regstatus, which is more reliable due to possible conflicting `roles`.


raydavis:
Looks OK, but the front-end logic is getting pretty tricksy... :)

PR #5833, opened 2016-08-26
SISRP-159357 - Final Exam Schedule (Conflicts)

tiffanyqi:
JIRA: https://jira.berkeley.edu/browse/SISRP-15935
- What students' final exam schedule would look like during the enrollment period (unofficial), and
- Whether or not students' proposed schedule has a time conflict

All work done on this card and its assumptions is documented here: https://docs.google.com/document/d/11Au0Vg_sTWE_NDqZj6JJtnAfJ0WW5fAhDWTjJ6RXRr0/edit#
Also shows on the advising card.

Note:
I deleted some lines in my_academics_controller_spec.rb and filtered_for_delegate.rb. This is because those two specs are using outdated Bearfacts data / fixtures, which no longer works for this card. JIRA: https://jira.berkeley.edu/browse/SISRP-25001


raydavis:
Hey, @tiffanyqi , congratulations... but you still have a couple of hoops to jump through before this can get a good code review.

1. A bunch of regression tests are failing, as shown at the bottom of this link: https://s3.amazonaws.com/archive.travis-ci.org/jobs/155464396/log.txt
We don't merge without successful test runs, and these failures may be pointing out problems that will require some real changes in your code. By the way, you can run a full regression test before submitting a PR with this command line:

``RAILS_ENV=test LOGGER_LEVEL=ERROR rspec``

2. Very fine-grained commits are great when we're on our own machine making changes to our own working branch on our own drive. But in a long-lived shared codebase, they make it harder to track history, create emergency releases, etc. Would you mind doing a "squash" to reduce the number of commits in the PR?

tiffanyqi:
@raydavis Yup! I plan on fixing those problems religiously when I get back on Monday. Thanks, and I look forward to the code review :)

app/models/berkeley/term.rb
```diff
@@ -19,6 +19,8 @@ class Term
     # CNP notification to a CNP warning, as well as the drop date for undergraduates.
     attr_reader :classes_start
     attr_reader :classes_end
+    # Final exams start
+    attr_reader :final_exam_week_start
```
redconfetti: I don't see an update to the Rspec for spec/models/berkeley/term_spec.rb to test support for this new property

tiffanyqi: @redconfetti oh, good point. Let me go add that now. Thanks!

src/assets/stylesheets/_exam_schedule.scss
```diff
@@ -3,11 +3,29 @@
...
+  .cc-widget-exam-schedule-time-conflicts-left {
+    margin-left: -5px;
+  }
+  .cc-widget-exam-schedule-text {
+    padding: 5px;
+    padding-left: 20px;
+  }
+  .cc-widget-exam-schedule-conflicts {
+    border-left: 5px solid $cc-color-widget-list-blue-dark;
+    // border-left: 5px solid #f1a91e;
```
redconfetti: Should remove commented out styles

src/assets/templates/academics_semester.html
```diff
@@ -19,8 +19,7 @@ <h1 class="cc-heading-page-title cc-left">
   <div class="medium-8 columns cc-academics-row-margin cc-clearfix" data-ng-hide="selectedStudentSemester && !selectedCourses.length && selectedSemester.timeBucket === 'future'">
     <div data-ng-include="'academics_semester_classes.html'"></div>
 
-    <div data-ng-include="'widgets/final_exam_schedule.html'" data-ng-if="examSchedule.length && !studentInfo.isLawStudent && selectedSemester.timeBucket === 'current'"></div>
-
+<!--   <div data-ng-include="'widgets/final_exam_schedule.html'" data-ng-if="!studentInfo.isLawStudent && selectedSemester.timeBucket === 'current'"></div> -->
```
redconfetti: Why is this commented out? Shouldn't this just be removed?

Will this widget still work? If so we should perhaps feature flag it. That way we can reconfigure CalCentral to re-enable the old exam schedule if it is decided that this new one doesn't pass QA.

redconfetti:
@tiffanyqi - I just passed over this really quick. Will come back and finish reviewing later.

app/models/berkeley/final_exam_schedule.rb
```diff
@@ -0,0 +1,50 @@
+module Berkeley
```
redconfetti: It would be good to have an Rspec for this as well, even if it just simply implements a simple call to Berkeley::FinalExamSchedule#fetch and confirms some of the results, especially results that are transformed by the actual code from the CSV data.

redconfetti: I know you've probably tested this from the Rails console to make sure it works properly several times, but if you actually have the test coded out formally in an Rspec file, then we can run it a million times even after you've left to make sure we haven't broken it, and that it's still doing it's job like it should.

src/assets/templates/academics_semester.html
tiffanyqi: @redconfetti fixed. There is a feature flag here, which has completely replaced the old exam schedule. This is ok because BearFacts is no longer relevant, and the widget should only display schedules for only current or future semesters (as discussed earlier!).

app/models/berkeley/final_exam_schedule.rb
tiffanyqi: @redconfetti added, thanks!

app/models/my_academics/exams.rb
```diff
@@ -1,59 +1,193 @@
...
+        cs_data_available = false
 
-      feed['studentFinalExamSchedules']['studentFinalExamSchedule'].as_collection.each do |exam|
-        exam_data = exam['studentFinalExamScheduleKey']
+        semester[:classes].each do |course|
+          course[:sections].select{|x| course[:role] == 'Student' && x[:is_primary_section]}.each do |section|
+            course = {
+              name: course[:course_code],
+              number: course[:course_code].gsub(/[^0-9]/, '').to_i,
+              time: (section[:schedules][:recurring] != [] ? section[:schedules][:recurring][0][:schedule] : nil), # hopefully first
```
redconfetti: I think this could be better written as:
```
time: section[:schedules][:recurring].first,
```

If you are totally sure that it will always be an Array, either empty or with at least one item, then this code will work.

```
>> schedules = ['A','B','C']
=> ["A", "B", "C"]
>> schedules.first
=> "A"
>> schedules = []
=> []
>> schedules.first
=> nil
```

Are you sure that the My Academics - Schedules feed will always provide an array? If not then you should perhaps use `section[:schedules][:recurring).to_a`.

app/models/berkeley/final_exam_schedule.rb
```diff
@@ -0,0 +1,50 @@
...
+    #   Exam Group,Day,Date,Time,Class Days,Class Times,Course Exceptions
+    #   1,Monday,5/8/17,8-11A,MWF,8:00A 8:30A,
+    #   9,Wednesday,5/10/17,8-11A,,,"CHEM 1A, CHEM 1B, CHEM 3A, CHEM 3B"
+    # Output: {
+    #   B-M-8:00A=>{:exam_day=>"Monday",:exam_time=>"8-11A",:exam_slot=>1},
+    #   B-CHEM 1A=>{:exam_day=>"Wednesday",:exam_time=>"8-11A",:exam_slot=>9},
+    #   ...
+    # } Each key is the term_code - one_day - start_time, which maps to a exam slot, time, and day
+    #   If the course is an exception, the key is term_code - course_code
+    def self.fetch
```
redconfetti: You should wrap all of this in `self.class.fetch_from_cache { // code goes here }` to ensure that it actually caches it.

app/models/my_academics/exams.rb
```diff
@@ -1,59 +1,193 @@
+require 'date'
+
 module MyAcademics
   class Exams
     include AcademicsModule
     include ClassLogger
     include DatedFeed
+    include Cache::CachedFeed
```
redconfetti: I don't think these are needed because MyAcademics::Merged is already cacheing the data after the merger.

public/csv/sp_final_exam_schedule.csv
```diff
@@ -0,0 +1,23 @@
+Exam Group,Day,Date,Time,Class Days,Class Times,Course Exceptions
+1,Monday,5/8/17,8-11A,MWF,8:00A 8:30A,
+2,Monday,5/8/17,11:30-2:30P,TuTh,2:00P 2:30P,
+3,Monday,5/8/17,3-6P,,,"ECON 1, ECON 100B"
+4,Monday,5/8/17,7-10P,MWF,9:00A 9:30A,
+5,Tuesday,5/9/17,8-11A,MWF,1:00P 1:30P,
+6,Tuesday,5/9/17,11:30-2:30P,MWF,2:00P 2:30P,
+7,Tuesday,5/9/17,3-6P,MWF,10:00A 10:30A,
+8,Tuesday,5/9/17,7-10P,MWF,11:00A 11:30A,
+9,Wednesday,5/10/17,8-11A,,,"CHEM 1A, CHEM 1B, CHEM 3A, CHEM 3B"
```
redconfetti: I think this was [updated](http://schedule.berkeley.edu/examsp.html)

app/models/my_academics/exams.rb
```diff
@@ -1,59 +1,193 @@
...
+              time: (section[:schedules][:recurring] != [] ? section[:schedules][:recurring][0][:schedule] : nil), # hopefully first
+              waitlisted: section[:waitlisted]
+            }
 
-        exam_datetime = exam_data['examDate'].to_date
-        if exam_datetime.blank?
-          logger.warn "Error parsing date in final exams feed for user #{@uid}, exam data: #{exam_data.unwrap}"
-          next
+            # when CS data has populated
+            if !section[:schedules][:oneTime].empty?
```
redconfetti: Might be a good idea to make this `!section[:schedules][:oneTime].to_a.empty?`, so that if :oneTime is not an array or is not present, it doesn't fail altogether

spec/models/my_academics/exams_spec.rb
```diff
@@ -1,81 +1,165 @@
...
-  it 'should properly handle a student with an exam in an unparseable room' do
-    proxy = Bearfacts::Exams.new({:user_id => '865826', :fake => true})
-    Bearfacts::Exams.stub(:new).and_return(proxy)
-
-    feed = {}
-    MyAcademics::Exams.new('865826').merge(feed)
-
-    feed[:examSchedule].should_not be_nil
-    feed[:examSchedule][0][:locations][0][:raw].should == 'F295 HAAS'
+  let(:feed) do
```
redconfetti: This might be easier to follow and maintain if you put your cases/scenarios in their own containers.

For instance, I defined all the variations of studentPlans in [MyAcademics::ClassEnrollments spec](https://github.com/ets-berkeley-edu/calcentral/blob/master/spec/models/my_academics/class_enrollments_spec.rb#L57), so that I could flexibly insert them into the tests where I want them. I even made `student_plans` be an empty array by default.

```
let(:student_plans) { [] }
```

Further down in my Rspec file, plans in a students feed could show up where I override `student_plans` with a populated array, like in [this context](https://github.com/ets-berkeley-edu/calcentral/blob/master/spec/models/my_academics/class_enrollments_spec.rb#L331) and [that context](https://github.com/ets-berkeley-edu/calcentral/blob/master/spec/models/my_academics/class_enrollments_spec.rb#L349).

src/assets/templates/academics.html
```diff
@@ -9,7 +9,7 @@ <h1 class="cc-heading-page-title">My Academics</h1>
       class="medium-8 columns cc-academics-row-margin"
       data-ng-if="selectedTeachingSemester && !api.user.profile.hasStudentHistory && !api.user.profile.roles.student">
       <div data-ng-include="'academics_semester_classes.html'"></div>
-      <div data-ng-include="'widgets/final_exam_schedule.html'" data-ng-if="examSchedule.length && (selectedTeachingSemester.timeBucket === 'current')"></div>
+      <div data-ng-include="'widgets/final_exam_schedule.html'" data-ng-if="api.user.profile.sid && examSchedule.length && (selectedTeachingSemester.timeBucket === 'current')"></div>
```
redconfetti: I know that this 2 column widget container is present for instructors / faculty. Shouldn't it's logic also use the feature flag: `api.user.profile.features.finalExamSchedule`?

src/assets/templates/academics.html
```diff
@@ -29,7 +29,7 @@ <h1 class="cc-heading-page-title">My Academics</h1>
     <div class="medium-6 large-4 columns cc-column-3">
       <div data-ng-include="'widgets/enrollment_card.html'" data-ng-if="api.user.profile.features.csEnrollmentCard"></div>
 
-      <div data-ng-include="'widgets/final_exam_schedule.html'" data-ng-if="examSchedule.length && api.user.profile.roles.student && !studentInfo.isLawStudent"></div>
+      <div data-ng-include="'widgets/final_exam_schedule.html'" data-ng-if="api.user.profile.roles.student && !studentInfo.isLawStudent && api.user.profile.features.finalExamSchedule"></div>
```
redconfetti: Why does the ngIf statement not use `examSchedule.length` in this one?

src/assets/templates/widgets/final_exam_schedule.html
```diff
@@ -2,15 +2,35 @@
...
-  <div class="cc-widget-padding" data-ng-if="examSchedule.length">
-    <div data-ng-repeat="exam in examSchedule" class="cc-widget-exam-schedule-item cc-clearfix-container">
-      <div class="cc-widget-exam-schedule-time cc-left">
-        <span data-ng-bind="exam.date.epoch * 1000 | date:'EEE MMM d'"></span><br>
-        <span data-ng-bind="exam.time" class="cc-text-small cc-text-light"></span>
-      </div>
-      <div class="cc-widget-exam-schedule-content">
-        <strong data-ng-bind="exam.course_code"></strong><br>
-        <div class="cc-text-small" data-cc-map-link-builder-directive data-locations="exam.locations"></div>
+  <div data-ng-hide="!examSchedule.length">
```
redconfetti: So we want to hide the content inside of this card when there are no class exams found? Why don't we simply hide the entire card? Also, why are you using `data-ng-hide` instead of `data-ng-if`?

src/assets/templates/widgets/final_exam_schedule.html
```diff
@@ -2,15 +2,35 @@
...
+            <div data-ng-repeat="course in exam" class="cc-widget-exam-schedule-item cc-clearfix-container">
+              <div class="cc-widget-exam-schedule-time cc-left">
+                <span data-ng-bind="course.exam_date"></span><br>
+                <span data-ng-bind="course.exam_time" class="cc-text-small cc-text-light"></span>
+              </div>
+              <div class="cc-widget-exam-schedule-content">
+                <strong data-ng-bind="course.name"></strong>
+                <span data-ng-if="course.waitlisted">(waitlisted)</span>
+                <div class="cc-text-small" data-ng-bind="course.exam_location"></div>
+                <!-- <div class="cc-text-small" data-cc-map-link-builder-directive data-locations="course.exam_location"></div> -->
```
redconfetti: Please remove commented code.

redconfetti:
@tiffanyqi - I'm curious if it is fine to eternally use the manually populated CSV, or if there will ever be a functional requirement where the data in CampusSolutions that becomes available in the middle of the term has to be used instead of the CSV data?

src/assets/templates/academics.html
tiffanyqi: I do not believe this exam card is slated for instructor view. Currently, instructors can see when their exams are on the Class page (under their sections), although they will not see their final exams until halfway through the semester when CS data is loaded.

src/assets/templates/widgets/final_exam_schedule.html
tiffanyqi: I used data-ng-hide because I saw someone else using it, and thought it did a similar thing. Should I change it back?

Students should know whether or not they have any exams. If the card is gone altogether, there may be more confusion around this fact.

Also note, that using `examSchedule.length` actually won't work, because this data structure includes course information if they are enrolled. I haven't yet determined how to make it so that when students are enrolled in classes, but have no exams, it shows "No exam schedules listed".

src/assets/templates/widgets/final_exam_schedule.html
tiffanyqi: Do you know how to use this directive so that it links to the map of the location displayed? I commented it out because this was something I wanted to hopefully work in the future, but didn't know how at the moment.

tiffanyqi:
@redconfetti to clarify, once Campus Solutions data becomes available, the exam data assigned by the CSV will be removed for user view.

src/assets/templates/academics.html
redconfetti: If you look at line 10 it has an ngIf with `selectedTeachingSemester && !api.user.profile.hasStudentHistory && !api.user.profile.roles.student`. That's basically checking to see if the user is a teacher, because only a teacher has a `selectedTeachingSemester` in the $scope, doesn't have a student history, and isn't a student.

You'll notice that if you log in as a faculty member and view the My Academics page that teachers get a 2 column view, with the left column using up 8 spaces out of the 12 (2 columns).

You should probably remove this altogether if this card does not support data for instructors / faculty. I assumed that it would still be able to determine the final exam dates for instructors and display the same data to them.

src/assets/templates/widgets/final_exam_schedule.html
redconfetti: `data-ng-if` makes it so that the DIV doesn't even show up in the DOM for the page, where-as the `data-ng-hide` causes the `display:none` style to be applied to the DIV. It is preferred that you use ngIf instead of ngHide in the front-end code when possible.

Because ngIf causes the DIV to have an isolated scope, you often have to reference the parent scope outside of the HTML element that is using the `data-ng-if` via `$parent`. In cases where you have many nested elements using `data-ng-if`, you are then forced to use `$parent.$parent.myProperty` to reference a value. In such cases, we then consider the use of `data-ng-hide` as more appropriate.

If `examSchedule.length` won't work, because it's not an enumerable object (like an Array), then why are you using it in the code? Perhaps you should try `data-ng-if="examSchedule !== undefined"`. To display the "No exam schedules listed" message you would simply add a DIV with the inverse logic right after the DIV that shows schedules when present.
```
<div data-ng-if="examSchedule === undefined">
  No exam schedules listed
</div>
```

redconfetti:
@tiffanyqi , where does the official exam data provided by Campus Solutions come from? MyAcademics::Semesters feed? If so, where in the code is the logic that detects this?

app/models/berkeley/final_exam_schedule.rb
tiffanyqi: added!

tiffanyqi:
@redconfetti it is in the `MyAcademics::Semesters` feed, located in :semesters -> classes -> sections -> schedules -> oneTime

src/assets/templates/academics.html
tiffanyqi: got it, I'll remove it. thanks!

src/assets/templates/widgets/final_exam_schedule.html
tiffanyqi: My problem with this section is that even if a student is enrolled in classes, and those classes do **not** have a final exam, `examSchedule.length` or `examSchedule === undefined` would not solve that problem, because all enrolled classes will be contained in the `examSchedule`. Even if I use the field inside called `exams`, interim data would appear rather than CS data (because there wouldn't be any CS data if there are no official exams)

src/assets/templates/widgets/final_exam_schedule.html
raydavis: FWIW, when I want to work on something in the future, I usually save a local working branch in my Git repository and then go pick it up again.

src/assets/templates/academics.html
redconfetti: @tiffanyqi - We should implement the ngIf on this element with `!examSchedule.length` so that it doesn't show at all when a student is not active for any term.

src/assets/templates/widgets/final_exam_schedule.html
redconfetti: We discussed this and we decided that you should implement a solution to iterate through each term, and if any of the terms have an exam, that your examSchedule data should include a boolean called `hasExams` with true or false. This will be used by the front-end to determine if there are exams in ANY of the terms, and thus display a message if there are none. If there are at least 1, then the card will display that one in which ever term it applies to.

You might not need to transform the data feed on the back-end, but simply implement a controller for your card and include the following function registered on the $scope.
```
  $scope.hasExams = function() {
    var hasExams = false;
    angular.forEach($scope.examSchedule, function(term) {
      if (_.get(term, 'exams.length') > 0) {
        hasExams = true;
      }
    });
    return hasExams;
  };
```

The controller will have to require the lodash library like so. Reference other controllers to see how they do this.
```
var _ = require('lodash');
```

You'll have to discuss this further with Flint.

app/models/my_academics/exams.rb
```diff
@@ -1,59 +1,192 @@
...
+              time: section[:schedules][:recurring].to_a.first.try(:[], :schedule),
+              waitlisted: section[:waitlisted]
+            }
 
-        exam_datetime = exam_data['examDate'].to_date
-        if exam_datetime.blank?
-          logger.warn "Error parsing date in final exams feed for user #{@uid}, exam data: #{exam_data.unwrap}"
-          next
+            # when CS data has populated
+            if !section[:schedules][:oneTime].to_a.empty?
```
redconfetti: You don't have to negate the result of `empty?`, you can instead use [any?](http://devdocs.io/ruby~2.2/array#method-i-any-3F) method

app/models/my_academics/exams.rb
```diff
@@ -1,59 +1,192 @@
...
+              section[:schedules][:oneTime].each do |exam|
+                # hardcoded month parsing and final exam determination
+                date = exam[:date].split(' ')
+                month = date[1].split('/')[0]
+                # only include oneTime that are in May and Spring, or in December and Fall
+                if (semester[:termCode] == 'B' && month == '5') || (semester[:termCode] == 'D' && month == '12')
```
redconfetti: @tiffanyqi - We need to confirm this with Student Records that there will never be a oneTime meeting in these months. That may be true for upcoming terms, like Fall 2016 and Spring 2017, but what if in the far future of Berkeley there are other types of meetings added to the system.

You should create a Jira about this, and if we aren't 100% sure that this logic is solid, we should lobby to have some sort of full proof indicator added to these schedules such as `isMidterm` or `isFinal` booleans.

app/models/my_academics/exams.rb
```diff
@@ -1,59 +1,192 @@
...
+    #        B-CHEM 1A=>{:exam_day=>"Wednesday",:exam_time=>"8-11A",:exam_slot=>9},... }
+    # Output: outputed final exams
+    #  [{:cs_data_available=>false,:name=>"Fall 2016",:term=>"D",:term_year=>"2016",:curr=>true,:time_bucket=>"current"
+    #    :courses=>[{:name=>"BIO ENG 131",:number=>131,:time=>"MWF 2:00P-2:59P"},...],
+    #    :exams=>{5=>[{:name=>"BIO ENG 131",:number=>131,:time=>"MWF 2:00P-2:59P",:waitlisted=>nil,:exam_time="8-11A",..},...}]}
+    #    }
+    def assign_exams(academics_data, final_exam_conversion)
+      academics_data.each do |semester|
+        semester[:courses].each do |course|
+          if course[:exam_date_no_day] == nil # then no cs data available
```
redconfetti: We should update this to use `cs_data_available`

app/models/my_academics/exams.rb
```diff
@@ -1,59 +1,192 @@
...
+            # if it's a grad course, it most likely won't be in exams week, but check anyway
+            if course[:number] > 199
```
redconfetti: most grad classes do not have exams, but if they do, check. so if the exam_slot is true, then you should add the times, dates, and locations from the exam_slot

app/models/my_academics/exams.rb
```diff
@@ -1,59 +1,192 @@
...
+              course[:exam_slot] = final_exam_conversion[course[:name]]
 
-        ampm = exam_data['startTimeAmPmFlag'].to_text == 'A' ? 'AM' : 'PM'
-        time = "#{exam_data['startTime'].to_time} #{ampm}"
+            # otherwise it's a undergrad course
+            elsif course[:time]
+              puts "course[:time]: #{course[:time].inspect}"
```
redconfetti: REMOVE! 👎

app/models/my_academics/exams.rb
```diff
@@ -1,59 +1,192 @@
...
-      # TODO Replace or remove by Fall 2016.
-      return unless legacy_student? && current_term.legacy?
-      proxy = Bearfacts::Exams.new({:user_id => @uid})
-      feed = proxy.get[:feed]
-      return unless feed.present? && matches_current_year_term?(feed)
+      academics_data = parse_academic_data(data[:semesters])
+      data[:examSchedule] = assign_exams(academics_data, Berkeley::FinalExamSchedule.fetch)
+    end
+
+    # Parses the my academics semesters feed, and prepares data to be populated
```
redconfetti: A better way of documenting the input and output for a specific function is to include a test in exams_spec.rb that gives it an example input, and then validates the output matches what is expected.

The reason for having this in the form of an actual test is because if this changes in the future, the developer will be forced to update the actual code for the test... whereas they might neglect your comments, and then you have stale documentation.

src/assets/templates/widgets/final_exam_schedule.html
redconfetti: Yeah. You could copy this code into a Jira ticket task to add the location in a future release. Link the Jira to your current Jira.

app/models/berkeley/final_exam_schedule.rb
redconfetti: I think you should move that Berkeley::FinalExamSchedule tests from `exams_spec.rb` to `/spec/models/berkeley/final_exam_schedule_spec.rb`

app/models/my_academics/exams.rb
```diff
@@ -1,59 +1,187 @@
...
+              course[:exam_time] = final_exam_conversion[exam_key][:exam_time]
+              course[:exam_slot] = final_exam_conversion[exam_key][:exam_slot].to_i
+              course[:exam_date] = determine_exam_date(semester, final_exam_conversion[exam_key][:exam_day])
+            end
+          end
+        end
+        sort_exams(semester)
+      end
+      # sort by semester, current first
+      academics_data.sort_by{|s| s[:curr] ? 0 : 1 }
```
redconfetti: Can we replace the `s` with `semester`

spec/models/my_academics/exams_spec.rb
```diff
@@ -1,81 +1,455 @@
...
-    MyAcademics::Exams.new('865826').merge(feed)
+  # [{:cs_data_available=>false,:name=>"Fall 2016",:term=>"D",:term_year=>"2016",:curr=>true,:time_bucket=>"current"
+  #   :courses=>[{:name=>"BIO ENG 131",:number=>131,:time=>"MWF 2:00P-2:59P"},...],
+  #   :exams=>{5=>[{:name=>"BIO ENG 131",:number=>131,:time=>"MWF 2:00P-2:59P",:waitlisted=>nil,:exam_time="8-11A",..},...}]}
+  #   }
+  let(:result) do
+    MyAcademics::Exams.new(uid).merge(feed)
+  end
+
+  # Input:
```
redconfetti: We discussed how you can document this in the form of an Rspec test for your #assign_exam method

app/models/my_academics/exams.rb
tiffanyqi: Received confirmation from student records that it is ok for me to assume final exams in May and in December. Still will need to speak with Adam about differences between meeting and exam to get some sort of exam type checker

src/assets/templates/widgets/final_exam_schedule.html
tiffanyqi: Confirmed with Flint. When there is no exams but enrolled, should say "No exam data available" under a semester. When a student is not enrolled in one semester, the semester shouldn't slow. When a student isn't enrolled for both semesters, the card shouldn't show at all. will modify this, thanks!

redconfetti: @tiffanyqi It's interesting how during development we bump into the things that the designers didn't consider.

spec/models/berkeley/final_exam_schedule_spec.rb
```diff
@@ -0,0 +1,42 @@
+describe Berkeley::FinalExamSchedule do
+
+  # {
```
redconfetti: Are you going to convert this input / output  comment to some sort of test... Living documentation!

redconfetti: Or wait, you're just noting what the CSV data is structured like right? That seems okay actually.

spec/models/my_academics/exams_spec.rb
```diff
@@ -1,81 +1,535 @@
...
+
+  ## TEST CASES
+
+  context 'as a student' do
+    context 'with no semesters' do
+      let(:semesters) { [] }
+
+      it 'should parse academic data correctly' do
+        result = subject.parse_academic_data(feed[:semesters])
+        expect(result.length).to eq 0
```
redconfetti: Woot!

app/models/my_academics/exams.rb
redconfetti: @tiffanyqi - Should we hold off on merging until after you know when to switch from interim (csv) to official data? Or will you just do a follow PR for that?

tiffanyqi: @redconfetti It's probably ok to merge whenever devs think it's ready, since batch will probably run in October (and it's only September), we still have a lot of time, and changing the date should be super easy and painless.

redconfetti:
@raydavis - I've reviewed this and suggested all I can, ready for you to take a look over it. The only outstanding issue is that we'll need to schedule a follow up JIRA for sometime in October to implement the proper logic for the card to switch from the CSV based exam data to the official exam schedules that CS is providing for each student.

app/models/berkeley/final_exam_schedule.rb
```diff
@@ -0,0 +1,42 @@
...
+      fetch_from_cache do
+        semesters_csv =  {
+          B: 'public/csv/sp_final_exam_schedule.csv', # http://schedule.berkeley.edu/examsp.html
+          D: 'public/csv/fa_final_exam_schedule.csv' # http://schedule.berkeley.edu/examsf.html
+        }
+        course_to_exam = Hash.new
+
+        semesters_csv.each do |semester, semester_csv|
+          CSV.foreach(semester_csv,{:headers=>true}) do |row|
```
lyttam: A friendly suggestion: anytime I have nested loops or nested if..else blocks, I like to refactor them into separate methods if possible because it makes the code a little easier to read and maintain.  Your main method is called `fetch`, but the work being done inside this `semesters_csv.each` is more like processing data, so you could move lines 16-36 into its own method called `processSemester` or something like that.  Then your code reads more like "for each semester, processSemester".  Then for the next loop, "for each row, formatExam". And same for the days loop on line 27.  

That way, a n00b like me looks at the code and goes "ok, this fetch function grabs the data from the cache and then does some processing work on it" and I can look inside that process method if I need to know the implementation details.

tiffanyqi: Thanks @lyttam! I've pushed and separated those two out, as well as separated `parse_academic_data` into `parse_courses` and `parse_cs_exams` as @nkutub noted a couple weeks ago.

app/models/my_academics/exams.rb
```diff
@@ -1,59 +1,195 @@
...
-          date: format_date(exam_datetime, '%a %B %-d'),
-          time: time,
-          locations: locations,
-          course_code: course_code
-        }
+    ## Support Functions
+
+    # Selects the proper term depending what term we're in
+    def select_term(semester)
+      if semester[:time_bucket] == 'current' || semester[:timeBucket] == 'current'
```
raydavis: ECHO echo cho ho o o o

paulkerschen: Do upstream classes disagree about returning `:time_bucket` versus `:timeBucket`? If so, we might want to head upstream ourselves to settle that disagreement.

app/models/my_academics/exams.rb
```diff
@@ -1,59 +1,195 @@
...
 
-        exams << {
-          date: format_date(exam_datetime, '%a %B %-d'),
-          time: time,
-          locations: locations,
-          course_code: course_code
-        }
+    ## Support Functions
+
+    # Selects the proper term depending what term we're in
```
raydavis: It's difficult to follow but it looks like we start off by looping around multiple semesters but then switch to a choice of at-most-two in this method.

1. parse_academic_data ignores past and summer terms, but might loop around a list of current semester, next semester, next semester after that, next next semester, ...
2. But then select_term treats the next-semester, the next-next-semester, the next-next-semester all as one single semester.

Rather than keeping the mystery alive, could we explicitly specify which one or two semesters will be checked up at the top in parse_academic_data, and loop only around those?

app/models/berkeley/term.rb
```diff
@@ -73,6 +78,8 @@ def from_cs_api(api_feed)
         @classes_start = session['beginDate'].to_date.in_time_zone.to_datetime
         @instruction_end = session['endDate'].to_date.in_time_zone.to_datetime.end_of_day
         @classes_end = @instruction_end.advance(days: -7)
+        @final_exam_week_start = @instruction_end.advance(days: 3)
+        @final_exam_cs_data_available = @end.advance(days: -56)
```
raydavis: Unfortunately, as outlined at SISRP-23422, we still don't know for sure how CS will store or publish per-class exam schedules. It may be safer to go with a configurable CS-data-available date for Fall 2016.

app/models/my_academics/exams.rb
```diff
@@ -1,59 +1,195 @@
...
+          courses << course
         end
+      end
+    end
 
-        ampm = exam_data['startTimeAmPmFlag'].to_text == 'A' ? 'AM' : 'PM'
-        time = "#{exam_data['startTime'].to_time} #{ampm}"
+    # If CS data is available, it set exams for the courses
+    def parse_cs_exams(semester, section, course)
+      section[:schedules][:oneTime].each do |exam|
```
raydavis: Unfortunately, as outlined at SISRP-23422, we still don't know for sure how CS will store or publish per-class exam schedules. A subset of classes have repurposed one-off section meetings to show some midterm exam times, but that mechanism doesn't provide enough data for our purposes. I'll let you decide whether it's worth keeping this in as a placeholder, but if so it'll need a "TODO" comment to flag the open issue.

public/dummy/json/academics_cs_no_conflicts.json
```diff
@@ -0,0 +1 @@
...-12-13T08:00:99+00:00":[{"name":"INTEGBI 35AC","number":35,"time":"TuTh 2:00P-3:29P","waitlisted":null,"exam_location":"Dwinelle 155","exam_time":"8-11A","exam_slot":"2016-12-13T08:00:99+00:00","exam_date":"Tue 12/13"}],"2016-12-15T08:00:99+00:00":[{"name":"POLI SCI 112B","number":112,"time":"TuTh 5:00P-6:29P","waitlisted":true,"exam_location":"","exam_time":"11:30-2:30P","exam_slot":14,"exam_date":"Thu 12/15"}]}},{"cs_data_available":false,"name":"Spring 2017","term":"B","term_year":"2017","current":false,"time_bucket":"future","exams":{"5":[{"name":"INTEGBI 35AC","number":35,"time":"TuTh 2:00P-3:29P","waitlisted":null,"exam_location":"","exam_time":"8-11A","exam_slot":5,"exam_date":"Tue 12/13"}],"7":[{"name":"POLI SCI 112B","number":112,"time":"TuTh 5:00P-6:29P","waitlisted":true,"exam_location":"","exam_time":"11:30-2:30P","exam_slot":14,"exam_date":"Thu 12/15"}],"13":[{"name":"MATH 53","number":53,"time":"MWF 4:00P-4:59P","waitlisted":null,"exam_location":"","exam_time":"8-11A","exam_slot":13,"exam_date":"Thu 12/15"}],"14":[{"name":"MATH 54","number":54,"time":"MWF 4:00P-4:59P","waitlisted":null,"exam_location":"","exam_time":"8-11A","exam_slot":13,"exam_date":"Thu 12/15"}]}}]}
```
raydavis: Please format this and the other new dummy JSON feeds for easier readability.

app/models/my_academics/exams.rb
redconfetti:
```
# You could slice
>> { timeBucket: 'current', somethingElse: 'banana' }.slice(:timeBucket, :time_bucket)
=> {:timeBucket=>"current"}
>> { time_bucket: 'current', somethingElse: 'apple' }.slice(:timeBucket, :time_bucket)
=> {:time_bucket=>"current"}

# then run `values` on it to get only the values, then use `first` to get the first value from the resulting array
>> {:time_bucket=>"current"}.values
=> ["current"]
>> {:time_bucket=>"current"}.values.first
=> "current"
```

`if semester.slice(:timeBucket, :time_bucket).values.first == 'current'`

raydavis: I agree with @paulkerschen on this one: if we're getting ``:time_bucket`` in some ``semesters`` elements but getting ``:timeBucket`` in other ``semesters`` elements, that's a bug in MyAcademics::Semesters which should get fixed.

public/dummy/json/academics_cs_no_conflicts.json
tiffanyqi: I've done it this way so that we didn't have to see that I've added 1000 extra lines of code. Would you prefer for me to remove them completely, or format them all properly with this in mind? Thanks!

app/models/my_academics/exams.rb
tiffanyqi: `:time_bucket` was just another field I've added, and kept them both at that line. I can change this easily; thanks all!

public/dummy/json/academics_cs_no_conflicts.json
raydavis: If they seem really useful, then they can be added. If they were more just for your own trying-stuff-out, then probably better to leave them out, if only to cut down on the maintenance burden.

Any JSON that _is_ put in, though, should be formatted. Better 1000 readable lines than one unreadable one. :)

(Besides, I'm curious about what the change was to ``academics_teaching.json``!)

app/models/my_academics/exams.rb
tiffanyqi: @raydavis according to the follow ups to SISRP-23422, it appears SR and the API team are confident that once the final exam batch process runs, it will populate in the exams field of the Class Section API. As a next step, do you think we should include this exams field in the MyAcademics feed with the sample data that Adam has provided us, then use that in the examSchedule?

app/models/my_academics/exams.rb
tiffanyqi: @raydavis do you mean:
1. At the top of `parse_academic_data` to say something like `current_semester = get_the_current_semester_from_data` and `future_semester = get_future_sem_from_data`, combine them in an array, and loop over it?
2. `select_term` is only used the support helper functions, which is only used for the future and current semester.
 ``
`return Berkeley::Terms.fetch.next.code == 'C' ? Berkeley::Terms.fetch.future : Berkeley::Terms.fetch.next`
```
This line above says that if the "future" semester is summer, that's not what we want, so take the semester after that, which is fall. If it is not summer, then the future semester, spring, is what we want. Is this function a good way of specifying that, and if not, what is a better way to do this? Thanks!

Comment (raydavis) [inline app/models/my_academics/exams.rb]: Yes, @tiffanyqi , something like that. It's confusing to pass a "semester" argument to a method and then have the method work on a completely different semester. Could you use the ``select_term`` logic at the _top_ of your class's workflow rather than calling it from the inner loop? (As I admitted, though, I have a hard time following the current logic, so if that wouldn't work, please walk me through it. Thanks!)

File: app/models/my_academics/exams.rb (hunk 29 above)
Comment (raydavis) [inline app/models/my_academics/exams.rb]: There are still some open questions regarding how to interpret the upstream data, @tiffanyqi , so if we'd like to merge your PR before they're all settled, we still need some safety measures in the code. (Merging after the v82 branch is cut and then following up with final server-side logic sounds OK to me, and so does waiting -- I don't know if you or anyone else has an opinion...)

Comment (tiffanyqi) [inline app/models/my_academics/exams.rb]: @raydavis IMO I can remove the current exam calculation (the stuff in `parse_cs_exams`) and have it available for merging by today (if other devs are ok with it). I can create a follow up with accurate final exam calculation by next week or so with whatever data source we end up using. Also, what safety measures are you referring to?

File: app/models/berkeley/term.rb (hunk 28 above)
Comment (tiffanyqi) [inline app/models/berkeley/term.rb]: @raydavis would you prefer I remove all `cs_data_available` code and add it back in later when things are confirmed, or leave this field in `term.rb` temporarily?

Comment (raydavis) [inline app/models/berkeley/term.rb]: If you don't want to add another configuration flag, how about temporarily using something like:

```
    def determine_cs_data_available(semester)
      # TODO Follow-up with real CS data.
      false
      # term = select_term(semester)
      # term.final_exam_cs_data_available < Time.now
    end
```

File: app/models/my_academics/exams.rb (hunk 29 above)
Comment (raydavis) [inline app/models/my_academics/exams.rb]: "safety measures" -- just meant a "TODO" comment or something to keep the untested knowingly-incomplete CS side of the code from accidentally running on the servers.

Comment (tiffanyqi) [inline app/models/my_academics/exams.rb]: Great, I'll add this in.

File: app/models/berkeley/term.rb (hunk 28 above)
Comment (tiffanyqi) [inline app/models/berkeley/term.rb]: Perfect! This is what I'll do, thanks!

File: app/models/my_academics/exams.rb (hunk 27 above)
Comment (tiffanyqi) [inline app/models/my_academics/exams.rb]: @raydavis perhaps we can walk through this code together then! I'm having a bit of a hard time understanding what your confusion is as I believe `data.reject` is a clean way to select these semesters and loop through them. I'd love to understand what you're having trouble with and fix it.

File: app/models/my_academics/exams.rb [hunk 31]
```diff
@@ -1,59 +1,156 @@
...
+    ## Support Functions
 
-        exams << {
-          date: format_date(exam_datetime, '%a %B %-d'),
-          time: time,
-          locations: locations,
-          course_code: course_code
-        }
+    # Selects the proper term depending what term we're in
+    def select_term(semester)
```
raydavis: @tiffanyqi I see that you've added the right way to find the Berkeley::Term object by slug, so this code should be deleted, right?

tiffanyqi: @raydavis oops, removed. Thanks!

raydavis:
Thanks for all your work on this, @tiffanyqi -- let's see how it plays on the server!

PR #5836, opened 2016-08-30
SISRP-24832 - Remove withdraw link for law students (INC306025)

nkutub:
https://jira.berkeley.edu/browse/SISRP-24832

use badges factory to get law student status and hide link.

### DIFF
--- a/src/assets/javascripts/angular/controllers/widgets/studentResourcesController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/studentResourcesController.js
@@ -3,7 +3,7 @@
 var angular = require('angular');
 var _ = require('lodash');
 
-angular.module('calcentral.controllers').controller('StudentResourcesController', function(studentResourcesFactory, $scope) {
+angular.module('calcentral.controllers').controller('StudentResourcesController', function(studentResourcesFactory, userService, $scope) {
   $scope.studentResources = {};
   $scope.isLoading = true;
 
@@ -16,9 +16,15 @@ angular.module('calcentral.controllers').controller('StudentResourcesController'
     angular.extend($scope.studentResources, resources);
   };
 
+  // Identify Law students to suppress withdrawal link
+  var setStudentRole = function() {
+    $scope.isLawStudent = userService.profile.roles.law;
+  };
+
   var loadInformation = function() {
     loadStudentResources()
     .then(parseStudentResources)
+    .then(setStudentRole)
     .then(function() {
       $scope.isLoading = false;
     });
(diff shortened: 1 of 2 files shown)


src/assets/javascripts/angular/controllers/widgets/studentResourcesController.js
```diff
@@ -26,8 +26,8 @@ angular.module('calcentral.controllers').controller('StudentResourcesController'
 
   // Get the student Info from the badges factory to determin if is law student
   var getStudentInfo = badgesFactory.getBadges().success(loadStudentInfo);
-  var loadInformation = function() {
 
+  var loadInformation = function() {
```
raydavis: As an FYI, we'd normally prefer to include this in the more substantial commit rather than keeping it separate in the shared history.

.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+// Place your settings in this file to overwrite default and user settings.
```
raydavis: Was this an intended part of the commit, @nkutub ?

nkutub: no, this is part of my editor settings that I added to .gitignore but somehow I lost that in the rebase

src/assets/javascripts/angular/controllers/widgets/studentResourcesController.js
```diff
@@ -3,22 +3,34 @@
...
-angular.module('calcentral.controllers').controller('StudentResourcesController', function(studentResourcesFactory, $scope) {
+angular.module('calcentral.controllers').controller('StudentResourcesController', function(studentResourcesFactory, badgesFactory, $scope) {
   $scope.studentResources = {};
   $scope.isLoading = true;
 
   var loadStudentResources = function() {
     return studentResourcesFactory.getStudentResources();
   };
 
+  var loadStudentInfo = function(data) {
```
raydavis: As another FYI, our codebase currently has two different feeds that indicate "in the Law school": the StudentInfo one (which was effective for both legacy & CS data) and a newer CS-specific "law" role in "api/my/status" (which you can see used in "src/assets/templates/widgets/finances_links.html"). Someone will have to deal with the ambiguity at some point, and StudentInfo is likely to be the loser since the code is mostly legacy and totally a mess. But for right now it's dealer's choice.

nkutub: Yes, I did see that but for the sample I was using this value did not come across. I then found the studentInfo and it did return the value.

src/assets/javascripts/angular/controllers/widgets/studentResourcesController.js
```diff
@@ -3,22 +3,34 @@
...
+    }
+    $scope.isLawStudent = data.studentInfo.isLawStudent;
+  };
+
   var parseStudentResources = function(data) {
     var resources = _.get(data, 'data.feed.resources');
     angular.extend($scope.studentResources, resources);
   };
 
+  // Get the student Info from the badges factory to determin if is law student
```
paulkerschen: @nkutub Note typo.

nkutub:
removed the legacy studentInfo and replaced with myStatus roles implimentation through userService

raydavis:
Thanks, @nkutub !

PR #5837, opened 2016-08-30
SISRP-24762 - Undergraduate Students with no reg, 0 units. have s09(Before Aug 30th) - There is CNP warning

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-24762

* Removes CNP notification/warning for students who are "Not Enrolled"
* I chose to use `ng-hide` in this case (on `status_and_holds.html`) due to the simplicity it offered in this already complicated logic.  The design already requires a ton of logic due to the number of factors going into whether or not we show CNP (timing, reg status, term, etc), and using an exclusion-based directive like `ng-hide` for this very particular case was the most direct, simple solution in my opinion.


src/assets/templates/widgets/status_and_holds.html
```diff
@@ -44,6 +44,7 @@ <h4 data-ng-bind="registration.name"></h4>
                   ((registration.academicCareer.code !== 'UGRD') && !registration.pastAddDrop) ||
                   (registration.summary !== 'Officially Registered' && registration.positiveIndicators.R99)) &&
                   !registration.isLegacy && !registration.isSummer && registration.summary !== 'Officially Registered'"
+                  data-ng-hide="registration.summary === 'Not Enrolled' && (!registration.positiveIndicators.R99 && !registration.positiveIndicators.ROP)"
```
raydavis: Hey, @davidlimcheng , I see what you mean about the crazy ng-if clause, but it also seems pretty confusing to have both an ng-if and an ng-hide on the same element. Maybe it's time to move this decision out of the HTML template and into the controller?

Also interested in @redconfetti 's opinion.

redconfetti: @davidlimcheng Something that I like to do is take something cryptic like the 'R99' and 'ROP' properties and place them inside of a function that implies what these represent such as `isNotExempt` or `isExemptAndNotTemporarilyProtected`, just so that anyone reading the code understands the business logic better.

That said, I also think that it's confusing that you're using an ng-if and ng-hide together. It seems that you should put all the logic that determines that it shows into a function that perhaps relies on other functions in the controller.

davidlimcheng: @raydavis @redconfetti I totally agree - I think it's time to move this pie over to bake in the controller.  I've spoken with Chara and confirmed that this fix is good to push out to 7.1, so I'll open a PR once I get back from vacation, and close this PR for now.

Thanks for the input!  Hopefully we'll be able to tame this thing a bit over in the controller.

PR #5844, opened 2016-08-31
SISRP-24125 - My Academics>[term name] - GPA Calculator (ID and add CS total-units-at-risk-in-letter-grade-attempts equivalent)

nkutub:
bug fix where I cloned the response obj to be changed in the college_and_level rather than the response obj being changed itself which led to a bug downstream when response was being referenced by the hub_gpa_units module.

followed API spec for pulling the equivalent of "totalUnitsAttempted" field from CS similar to what was provided by legacy

### DIFF
--- a/app/models/my_academics/college_and_level.rb
+++ b/app/models/my_academics/college_and_level.rb
@@ -27,30 +27,34 @@ module MyAcademics
 
     def bearfacts_college_and_level
       response = Bearfacts::Profile.new(user_id: @uid).get
-      feed = response.delete :feed
+      # response is a pointer to an obj in memory and should not be modified, other functions may need to use it later
+      result = response.clone
+      feed = result.delete :feed
       # The Bear Facts API can return empty profiles if the user is no longer (or not yet) considered an active student.
       # Partial profiles can be returned for incoming students around the start of the term.
       if (feed.nil? || feed['studentProfile']['studentGeneralProfile'].blank? || feed['studentProfile']['ugGradFlag'].blank?)
-        response[:empty] = true
+        result[:empty] = true
       else
-        response.merge! parse_bearfacts_feed(feed)
+        result.merge! parse_bearfacts_feed(feed)
       end
-      response
+      result
     end
 
     def hub_college_and_level
       response = HubEdos::AcademicStatus.new(user_id: @uid).get
-      if (status = parse_hub_academic_status response)
-        response[:careers] = parse_hub_careers status
-        response[:level] = parse_hub_level status
-        response[:termName] = parse_hub_term_name status
-        response[:termsInAttendance] = status['termsInAttendance'].to_s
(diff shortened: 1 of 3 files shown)


app/models/my_academics/college_and_level.rb
```diff
@@ -27,30 +27,34 @@ def merge(data)
...
     def hub_college_and_level
       response = HubEdos::AcademicStatus.new(user_id: @uid).get
-      if (status = parse_hub_academic_status response)
-        response[:careers] = parse_hub_careers status
-        response[:level] = parse_hub_level status
-        response[:termName] = parse_hub_term_name status
-        response[:termsInAttendance] = status['termsInAttendance'].to_s
-        response.merge! parse_hub_plans status
+      # response is a pointer to an obj in memory and should not be modified, other functions may need to use it later
+      result = response.clone
```
raydavis: I get that this is safer than what we have, but I'm curious. Are any fields _other_ than the ones parsed from response[:feed] used? If not, then there's no need to clone and delete. We can just say

```
result = {}
if (status = parse_hub_academic_status response)
  result[:careers] = parse_hub_careers status
...
```

It's nicer to do that when we can, since it clarifies what to expect from the final result.

(I could ask the same question about the bearfacts chunk of code, but that's going away in a sprint or so anyway.)

app/models/my_academics/gpa_units.rb
```diff
@@ -18,15 +18,18 @@ def merge(data)
 
     def hub_gpa_units
       response = HubEdos::AcademicStatus.new(user_id: @uid).get
-      if (status = parse_hub_academic_status response)
+      # response is a pointer to an obj in memory and should not be modified, other functions may need to use it later
+      result = response.clone
```
raydavis: Because this data structure has always been so focused, I'm pretty sure that _no_ fields other than the ones parsed from response[:feed] are used. If not, then there's no need to clone and delete. We can just say

```
result = {}
if (status = parse_hub_academic_status response)
  result[:cumulativeGpa] = (cumulativeGpa = parse_hub_cumulative_gpa status) && cumulativeGpa.to_s
```

or even 

```
{
  cumulativeGpa: (cumulativeGpa = parse_hub_cumulative_gpa status) && cumulativeGpa.to_s
```

app/models/my_academics/college_and_level.rb
raydavis: This work-from-a-munged-raw-response convention entered the codebase as part of a bulk move away from explicit XML parsing in a range of different legacy APIs. I suspect the bulk copy-and-pastiness of the task led to more blanket use than was actually needed. At any rate, the CollegeAndLevel question is a distraction from @nkutub 's main task here, so we've agreed to leave it open for now.

raydavis:
Thanks for migrating the feature and for dealing with the unexpected extra work, @nkutub !

PR #5846, opened 2016-09-01
NOJIRA Additional logging to help production support & bug investigation

raydavis:
* @redconfetti suggested it was time to start noticing when supposedly active students don't have a Campus Solutions ID, & I agree with him.
* DEBUG-level logging of raw responses from most upstream APIs is invaluable when investigating issues or generating tests, but our CalNet LDAP client missed that step.

### DIFF
--- a/app/models/calnet_ldap/client.rb
+++ b/app/models/calnet_ldap/client.rb
@@ -88,9 +88,10 @@ module CalnetLdap
       ActiveSupport::Notifications.instrument('proxy', {class: self.class, search: args}) do
         response = @ldap.search args
         if response.nil?
-          logger.error "LDAP error returned: #{@ldap.get_operation_result}"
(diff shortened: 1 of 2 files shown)


app/models/calnet_ldap/client.rb
```diff
@@ -88,9 +88,10 @@ def search(args = {})
       ActiveSupport::Notifications.instrument('proxy', {class: self.class, search: args}) do
         response = @ldap.search args
         if response.nil?
-          logger.error "LDAP error returned: #{@ldap.get_operation_result}"
+          logger.error "LDAP search with args #{args} returned: #{@ldap.get_operation_result}"
```
redconfetti: @raydavis - Will `@ldap.get_operation_result` and `response` output in a useful way as a string via the implicit call to`to_s` ?

raydavis: Yes.

PR #5847, opened 2016-09-01
SISRP-24867 - 7.0 - Remove link to appointments from advising card

redconfetti:
https://jira.berkeley.edu/browse/SISRP-24867

### DIFF
--- a/src/assets/javascripts/angular/controllers/pages/academicsController.js
+++ b/src/assets/javascripts/angular/controllers/pages/academicsController.js
@@ -202,7 +202,7 @@ angular.module('calcentral.controllers').controller('AcademicsController', funct
                                  ($scope.hasRegStatus ||
                                  ($scope.regblocks && !$scope.regblocks.noStudentId) ||
                                  ($scope.numberOfHolds));
-    $scope.showLegacyAdvising = !$scope.filteredForDelegate && $scope.api.user.profile.features.legacyAdvising && !$scope.api.user.profile.features.advising && $scope.isLSStudent;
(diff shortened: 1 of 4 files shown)


src/assets/javascripts/angular/controllers/pages/academicsController.js
```diff
@@ -202,7 +202,7 @@ angular.module('calcentral.controllers').controller('AcademicsController', funct
                                  ($scope.hasRegStatus ||
                                  ($scope.regblocks && !$scope.regblocks.noStudentId) ||
                                  ($scope.numberOfHolds));
-    $scope.showLegacyAdvising = !$scope.filteredForDelegate && $scope.api.user.profile.features.legacyAdvising && !$scope.api.user.profile.features.advising && $scope.isLSStudent;
```
redconfetti: This was the original logic from when the new Advising card was implemented to replace the legacy L&S Advising card. If the new 'Advising' card is enabled, it causes the legacy 'L&S Advising' card to be disabled, even if the 'legacyAdvising' feature flag is TRUE.

With GL 7.0, both cards should be shown, because the new Advising card will now ONLY show advisor contacts to students.

src/assets/templates/academics.html
```diff
@@ -31,10 +31,10 @@ <h1 class="cc-heading-page-title">My Academics</h1>
 
       <div data-ng-include="'widgets/final_exam_schedule.html'" data-ng-if="examSchedule.length && api.user.profile.roles.student && !studentInfo.isLawStudent"></div>
 
-      <div data-ng-include="'widgets/my_advising.html'" data-ng-if="showAdvising"></div>
```
redconfetti: Because both cards can now display at the same time, it was decided that the new Advising card should display beneath the L&S Advising card (legacy / BHive).

PR #5848, opened 2016-09-01
SISRP-16948 Fix class calendar checkbox handling

raydavis:
https://jira.berkeley.edu/browse/SISRP-16948


raydavis:
I think the only circumstances under which the old code would actually have worked were probably the ones used to test it: enable the checkbox, stay on the same screen, and wait for the checkbox to actually fill in. However, a refresh or a re-entry would show the checkbox to be empty no matter what.

redconfetti:
@raydavis LGTM

PR #5852, opened 2016-09-02
SISRP-24681 - Adds SID to the Student Overview

lyttam:
https://jira.berkeley.edu/browse/SISRP-24681

- Adds SID so advisors can see on the Student Overview
- Cleaned up inline styles from college_and_level and replaced with the table style used elsewhere


raydavis:
Thanks for the style cleanup, @lyttam

PR #5854, opened 2016-09-07
NOJIRA, nightly Selenium on Bamboo: do not run my_profile_basic_info_spec

johncrossman:
Nightly UI tests (Selenium) fail often w.r.t MyProfile specs. Is it due to a sleepy Hub API? Nonetheless, the fix is to only run 'em locally.  cc: @pfarestveit


johncrossman:
@pfarestveit Amended. No change in `my_profile_delegate_access_spec`

PR #5855, opened 2016-09-07
SISRP-21644 Exclude inactive majors from class roster views

raydavis:
https://jira.berkeley.edu/browse/SISRP-21644

Depends on changes in the EDO DB.


redconfetti:
LGTM

PR #5670, opened 2016-07-22
SISRP-22922 - CalCentral Prod - My Academics - Fall 2016 enrollment no longer appearing for student

redconfetti:
https://jira.berkeley.edu/browse/SISRP-22922

PR #5857, opened 2016-09-08
CLC-6441 Include 'stop further enrollment' sections in OEC queries

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6441

OEC was not invited to the party in #5670.


johncrossman:
OEC is always the last to know. Thanks, @paulkerschen

PR #5868, opened 2016-09-13
SISRP-25535 / SISRP-25533 - Removes line separator and adds tentative message

tiffanyqi:
https://jira.berkeley.edu/browse/SISRP-25533
https://jira.berkeley.edu/browse/SISRP-25535

Two very small bugs that are now fixed that:
1) Show the message that final exams are tentative
2) Removes the line if there is only one semester present

### DIFF
--- a/public/dummy/json/academics_interim_two_sets_conflicts.json
+++ b/public/dummy/json/academics_interim_two_sets_conflicts.json
@@ -84,7 +84,7 @@
       "slug": "fall-2016",
       "termCode": "D",
       "termYear": "2016",
-      "timeBucket": "future",
+      "timeBucket": "current",
       "campusSolutionsTerm": true,
       "gradingInProgress": null,
       "classes": [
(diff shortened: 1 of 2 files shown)


raydavis:
Hey, @tiffanyqi , these changes look fine. But in the vast territory of your earlier PR, I missed the new "curr" property being slipped into the My Academics "semesters" feed. 

``"curr": true`` seems to mean exactly the same thing as ``"time_bucket": "current"``. For example, ``!semester.curr`` can be replaced by ``(semester[:timeBucket] !== 'current')``.

It would be better to work directly with ``time_bucket`` than to add more bulk to the feed and to make your exams JS code use different logic than the other JS code in Academics. Would you be OK with changing that bit in a follow-up PR?

tiffanyqi:
@raydavis sure, I'll change that real quick. Should I preface the commit with 'NOJIRA'?

raydavis:
If it was me, since it's all aimed at the same release, I'd just re-use the JIRA number used for the initial PR and note the follow-up in JIRA. (I have to do after-thought changes a lot!)

tiffanyqi:
@raydavis added! https://github.com/ets-berkeley-edu/calcentral/pull/5869

Thanks!

PR #5874, opened 2016-09-14
QA SISRP-25495 Fix CS ID logging

raydavis:
QA version of https://github.com/ets-berkeley-edu/calcentral/pull/5873

Given the sample logs from Tuesday's load test, I'm worried about overwhelming ops with noise if this doesn't go in. PR reviewers should feel free to push back, though, since the change shouldn't otherwise impact end users.


davidlimcheng:
This LGTM and I don't see a reason not to merge this, but as I've been very removed from GL7 QA week til now, I'll defer to @redconfetti on timing of this merge.

PR #5875, opened 2016-09-14
SISRP-25554 - Advising appointments, advisor info, etc. will not show if link is unavailable: UC_CX_APPOINTMENT_STD_ADD, UC_CX_APPOINTMENT_STD_MY_APPTS

dfkaye:
https://jira.berkeley.edu/browse/SISRP-25554

### DIFF
--- a/spec/models/advising/my_advising_spec.rb
+++ b/spec/models/advising/my_advising_spec.rb
@@ -137,8 +137,7 @@ describe Advising::MyAdvising do
       end
       it 'reports failure' do
         expect(Rails.logger).to receive(:error).with(/Could not retrieve CS link/).exactly(2).times
-        expect(subject[:statusCode]).to eq 500
-        expect(subject[:errored]).to eq true
+        expect(subject[:feed][:links]).not_to be
       end
     end
   end
--- a/app/models/advising/my_advising.rb
+++ b/app/models/advising/my_advising.rb
@@ -59,9 +59,6 @@ module Advising
           manageAppointments: manage_appointments_link,
           newAppointment: new_appointment_link
         }
-      else
-        advising_feed[:statusCode] = 500
-        advising_feed[:errored] = true
       end
     end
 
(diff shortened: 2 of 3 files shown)


johncrossman:
Travis needs attention, @dfkaye

dfkaye:
looking at it, thx @johncrossman

dfkaye:
@johncrossman  - wow - that details log is chock full of stack traces - I can't tell what got missed

johncrossman:
@dfkaye I typically download the raw log from Travis then search on "rspec".  In this case I see:
```
rspec ./spec/models/advising/my_advising_spec.rb:138 
  # Advising::MyAdvising fake proxies missing links reports failure
```

johncrossman:
@dfkaye Better yet, search on "Failures"

dfkaye:
travis passing now

spec/models/advising/my_advising_spec.rb
```diff
@@ -137,8 +137,7 @@
       end
       it 'reports failure' do
         expect(Rails.logger).to receive(:error).with(/Could not retrieve CS link/).exactly(2).times
-        expect(subject[:statusCode]).to eq 500
-        expect(subject[:errored]).to eq true
+        subject
```
raydavis: @dfkaye What _do_ we expect the response to look like in case of an error? That's important since it looks like ``my_advising.html`` checks it.

raydavis: E.g., maybe ``expect(subject[:links]).to be_nil``?

redconfetti: Good call. It would be nice if it asserted what the behavior should be when an error occurs.

dfkaye: @raydavis - Previously, if there were no links, then none of the other information would display, even if it was returned successfully, and an "unavailable" message displayed instead.  In the case that links are not returned, we still want to show the other information.  That was the motivating case for this PR.  Make sense?

dfkaye: @raydavis @redconfetti -- PR amended, travis building, should be ready in 5 to 10 minutes...

raydavis:
Thanks for the amendment, @dfkaye

PR #5879, opened 2016-09-16
SISRP-18299 - "Status" portion of status popover taking long time to load

dfkaye:
https://jira.berkeley.edu/browse/SISRP-18299

First part of UX fix - show a spinner while the status portion of popover is still loading by moving spinner out of the status template


lyttam:
Looks good!

PR #5880, opened 2016-09-17
SISRP-25536 Add feature flag for Class Calendar sign-up

raydavis:
https://jira.berkeley.edu/browse/SISRP-25536

We were a bit premature with that CS retrofit...


paulkerschen:
Eek.

PR #5884, opened 2016-09-20
SISRP-24500 - CalCentral prod is calling /api/eft_status?studentId= w…

nkutub:
https://jira.berkeley.edu/browse/SISRP-24500

Just added a check not to call the eft_enrolment for legacy students. I am assuming that there is no alternative for this data from non-CS sources and thus returning empty data for legacy students is sufficient.

### DIFF
--- a/app/models/bearfacts/proxy.rb
+++ b/app/models/bearfacts/proxy.rb
@@ -23,7 +23,7 @@ module Bearfacts
     end
 
     def get
-      return {} unless legacy_student?
+      return {} unless has_legacy_data?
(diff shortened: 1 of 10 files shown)


raydavis:
@nkutub , I don't think this will do what you want. ``legacy_student?`` is an ambiguous method name. What it means is "Had an account before Fall 2016." It does _not_ mean "Has no CS ID." A student who was registered in Spring 2016 and in Fall 2016 will have legacy records, but will also still be in CS.

The check that you want is for a nil Campus Solutions ID.

raydavis:
Looks great, @nkutub . Thank you for improving that method name!

PR #5885, opened 2016-09-21
SISRP-25121 Add support for Memcached analysis

raydavis:

Adds library and Rake task support for more in-depth analysis of Memcached usage.

### DIFF
--- a/lib/memcached.rb
+++ b/lib/memcached.rb
@@ -1,0 +1,167 @@
+module Memcached
+  require 'net/telnet'
+  extend self
+
+  def servers
+    result = []
+    hosts = ENV['hosts']
+    if hosts.blank?
+      config = CalcentralConfig.load_settings
+      if config && config.cache && config.cache.servers
+        hosts = config.cache.servers
+      end
+    else
+      hosts = hosts.split ','
+    end
+    hosts.each do |host|
+      host_tuple = host.split(':')
+      hostname = host_tuple.first
+      port = host_tuple[1] if host_tuple.size > 1
+      port ||= 11211
+      result << {
+        host: hostname,
+        port: port
+      }
+    end
+    result
+  end
+
+  def send(server, command, response_over='END')
+    hostname = server[:host]
+    port = server[:port]
+    parsed = nil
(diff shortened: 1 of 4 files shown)


lib/tasks/memcached.rake
```diff
@@ -7,113 +7,21 @@ namespace :memcached do
...
-          stats.merge!({
-            get_hit_percentage:"#{"%0.2f" % (raw_stats["get_hits"].to_i * 100/raw_stats["cmd_get"].to_i)}%",
-            get_miss_percentage:"#{"%0.2f" % (raw_stats["get_misses"].to_i * 100/raw_stats["cmd_get"].to_i)}%"
-          })
-        end
-        pp stats
-      end
-    end
-
+    pp Memcached.all_stats
```
lyttam: +1 for un-duplicating code!

lib/memcached.rb
```diff
@@ -0,0 +1,167 @@
...
+    end
+    result
+  end
+
+  def send(server, command, response_over='END')
+    hostname = server[:host]
+    port = server[:port]
+    parsed = nil
+    connected_host = Net::Telnet::new('Host' => hostname, 'Port' => port, 'Timeout' => 3)
+    connected_host.cmd('String' => command, 'Match' => /^#{response_over}/) do |c|
```
lyttam: Not sure how much it matters, but would it make sense to validate `command` before executing it in order to prevent malicious and/or stupid commands being run?  Or perhaps make `send` a private method in this module?

raydavis: Since this isn't exposed through any front-end, the only people who would be able to send malicious and/or stupid commands are those who are able to ssh into the cluster and run the code from the command line, or developers on their local machines trying stuff out. Especially for the trying-stuff-out case, I feel OK about passing through any command that a dev wants to try out. Does it still ring warning bells for you?

lyttam: No, that seems like a good balance of security and flexibility. Thanks for the explanation.

davidlimcheng:
Looks great to me, @raydavis , and thanks again for the rundown!

PR #5886, opened 2016-09-21
SISRP-25121 Reduce caching load of MyGroups

raydavis:

First of about a half-dozen cache structure changes to reduce load.

Removed LiveUpdatesEnabled because the "My Groups" feed is not volatile or important enough to require an update every 2 minutes.

Removed JsonAddedCacher because the feed's not long and complex enough to benefit from cached JSON, and it's referred to from another class on the server side.


paulkerschen:
Thanks, @raydavis, makes sense.

PR #5887, opened 2016-09-22
SISRP-25121 Reduce system load of Live Updates, My Badges; remove legacy Blocks

raydavis:

Reviewers should probably look at each commit separately. I tried to include reasonable explanations for the cache-related commits.

The long-painful "StudentInfo" portion of MyBadges would have made optimization more complex, & in production consists entirely of obsolete or duplicated data. Ripping it out involved ripping out legacy BearFacts-era registration block code as well.


app/models/my_academics/college_and_level.rb
```diff
@@ -37,6 +37,7 @@ def bearfacts_college_and_level
       else
         result.merge! parse_bearfacts_feed(feed)
       end
+      result[:isLawStudent] = result[:majors].present? && result[:majors].first[:college] == Berkeley::Departments.get('CLLAW')
```
nkutub: why are you only checking the first major in the feed? Is it not possible for there to be multiple colleges where the law college is not the first?

raydavis: Basically because the logic is copied from https://github.com/ets-berkeley-edu/calcentral/pull/5887/files#diff-461e0d41340bc6419c1c13a47e406b9aL52 and I don't want to spend time rethinking legacy BearFacts-era code. (This whole section will probably join Registration Blocks in SIS Heaven soon.)

nkutub: I'm cool with that :)

app/models/my_academics/college_and_level.rb
```diff
@@ -54,6 +55,7 @@ def hub_college_and_level
         result[:empty] = true
       end
       result.delete(:feed)
+      result[:isLawStudent] = result[:careers].present? && result[:careers].include?('Law')
```
davidlimcheng: @raydavis, I see that we're adding an `isLawStudent` property here, but all of the replaced HTML checks are looking at `api.user.profile.roles` instead of the merged Academics feed.  Is there another usage for these added properties that I'm not seeing?

app/models/my_badges/google_calendar.rb
```diff
@@ -13,9 +12,7 @@ def initialize(uid)
     def fetch_counts(params = {})
       @google_mail ||= User::Oauth2Data.get_google_email(@uid)
       @rewrite_url ||= !(Mail::Address.new(@google_mail).domain =~ /berkeley.edu/).nil?
-      self.class.fetch_from_cache(@uid) do
```
davidlimcheng: Were we double caching this because it was being cached here, and the `MyBadges::Merged` feed as well?

davidlimcheng: Oh, duh, just saw what you wrote on the commit.  Nevermind! =)

app/models/my_academics/college_and_level.rb
raydavis: No, @davidlimcheng , & so I thought about leaving it out entirely. But then I thought it might be worthwhile as a comparison point. Obviously I don't feel strongly, though!

Since we're generally pretty strict about "No wishful-thinking maybe-it'll-be-useful-someday," I think your question convinces me it should just be deleted. I'll amend the PR.

raydavis:
@davidlimcheng , I've updated the PR to remove the unused ``isLawStudent`` property.

davidlimcheng:
Thanks @raydavis, everything looks good to me.

PR #5888, opened 2016-09-22
SISRP-25897 - loadErrorDirective.js errors out when app gets 401 from photo api

dfkaye:
https://jira.berkeley.edu/browse/SISRP-25897

### DIFF
--- a/src/assets/javascripts/angular/directives/loadErrorDirective.js
+++ b/src/assets/javascripts/angular/directives/loadErrorDirective.js
@@ -10,7 +10,13 @@ angular.module('calcentral.directives').directive('ccLoadErrorDirective', functi
     link: function(scope, elm, attrs) {
       elm.bind('error', function() {
         scope.$apply(function() {
-          scope[attrs.ccLoadErrorDirective].loadError = true;
+          // Defensive programming, in case of failed calls to photo APIs which
+          // cause the photo field not to be initialized on the scope of its
+          // respective controller.
+          var context = scope[attrs.ccLoadErrorDirective];
+          if (context) {
+            context.loadError = true;
+          }
         });
       });
     }


raydavis:
@dfkaye , I don't have any objection to the fallback, but I think this leaves the incorrect call to the photo API in place. Is there any way to stop that?

nkutub:
I agree with @raydavis, the root of the error is the fact that the 401 status code is returned from the server and in HTML the standard is to rais it as a document level error. 

I think there are three ways to do this since the only way we know if there is a 401 error is after the fact that we did the call and it will raise on the root scope of angular. 

option 1: prevent 401 errors and refactor all the 401 return codes for images on the backend and return an unauthorized standard image. 

option 2: use an http interceptor at the root scope of angular to handle the error gracfully.

option 3: not use ng-src and roll our own custom directive to call the image and load it using $http.get (see example here https://github.com/dougmoscrop/angular-img-http-src/blob/master/index.js)

raydavis:
I was puzzled by @nkutub 's comment so I just chatted with him about it. Turns out he was talking about a different context & a slightly different problem. After sussing it out together, we agree that the front-end shouldn't be doing anything on the landing page that would trigger this, and so the real bug is the call out to api/my/photo.

dfkaye:
@raydavis @nkutub - this angular-only fix just quiets the console - if you'd prefer not to merge this, in favor of addressing the api instead, I'm OK with that

nkutub:
I addressed the api/my/photo call in https://github.com/ets-berkeley-edu/calcentral/pull/5891. So I think we are good with merging this one. @raydavis , do you have any concerns on merge?

raydavis:
Nah, @nkutub , like I said above I don't have any objection to the fallback - it just didn't address the bigger problem. I'll merge now.

PR #5890, opened 2016-09-23
SISRP-23614 - UX Design Bug Status and Holds Card modify text

nkutub:
simple wording change in HTML nothing more.


davidlimcheng:
10/10

PR #5891, opened 2016-09-23
SISRP-25716 CalCentral landing page throws TypeError: Cannot set prop…

nkutub:
…erty 'loadError' of undefined
Even though the header HTML is visually hidden on the login page, the image tags still try to request and load images. This was the case for the gear_popover.html widget. This created an error on the page. Simply added condition not to display if user is not authenticated yet, this is the same condition to display the entire widget.


raydavis:
I break it, @nkutub fixes it -- sweet!

PR #5892, opened 2016-09-23
SISRP-25546 - Student Overview - Personal Summary Card - Advising Res…

nkutub:
…ources - Multi-Year Planner page error

UPDATE: Adding link to https://jira.berkeley.edu/browse/SISRP-25546 for review convenience.

I do not have access to CS to test a link into it, but this change is so minor I figure I would push it anyway.


raydavis:
We usually try to avoid blind "theoretically this should work" fixes, but in this case it may be the fastest way forward.

PR #5894, opened 2016-09-24
SISRP-23422 Final Exam Schedule Use CS Data

tiffanyqi:
https://jira.berkeley.edu/browse/SISRP-23422

Note: We still check whether CS data is available based on whether we are 8 weeks before the end of the term. If functionals do not approve, we can change the date in a following pull request.

Update: alternate exams that are not in finals week (in which instructors deliberately change the final exam time) will NOT be in this data, but there is a field called "Alternate Method" somewhere in which we can specify that the exam may be different. Still getting details.

May be missing some test cases. Open to any ideas you may have; thanks!

### DIFF
--- a/app/models/edo_oracle/queries.rb
+++ b/app/models/edo_oracle/queries.rb
@@ -165,6 +165,33 @@ module EdoOracle
       SQL
     end
 
+    # No Campus Oracle equivalent.
+    def self.get_section_final_exam(term_id, section_id)
+      safe_query <<-SQL
+        SELECT DISTINCT
+          exam."term-id" AS term_id,
+          exam."session-id" AS session_id,
+          exam."date" AS exam_date,
+          exam."startTime" AS exam_start_time,
+          exam."endTime" AS exam_end_time,
+          exam."location-descr" AS location
+        FROM
+          SISEDO.EXAMV00_VW exam
+        JOIN SISEDO.CLASSSECTIONV00_VW sec ON (
+          exam."cs-course-id" = sec."cs-course-id" AND
+          exam."term-id" = sec."term-id" AND
+          exam."session-id" = sec."session-id" AND
+          exam."offeringNumber" = sec."offeringNumber" AND
+          exam."sectionNumber" = sec."sectionNumber"
+        )
+        WHERE
+          sec."term-id" = '#{term_id}' AND
+          sec."id" = '#{section_id}' AND
+          exam."type-code" = 'FIN'
+        ORDER BY exam_date
+      SQL
+    end
+
     # EDO equivalent of CampusOracle::Queries.get_sections_from_ccns
     # Changes:
     #   - 'course_cntl_num' is replaced with 'section_id'
--- a/app/models/my_academics/exams.rb
+++ b/app/models/my_academics/exams.rb
@@ -39,13 +39,19 @@ module MyAcademics
     def parse_courses(semester, courses, cs_data_available)
       semester[:classes].each do |course|
         course[:sections].select{|x| course[:role] == 'Student' && x[:is_primary_section]}.each do |section|
-          # TODO: properly uses CS data when it becomes available
           course = {
             name: course[:course_code],
             number: course[:course_code].gsub(/[^0-9]/, '').to_i,
             time: section[:schedules][:recurring].to_a.first.try(:[], :schedule),
             waitlisted: section[:waitlisted]
           }
+          if cs_data_available && section[:final_exams].any?
+            exam = section[:final_exams].first
+            course[:exam_location] = exam[:location] || 'Location TBD'
+            course[:exam_date] = parse_cs_exam_date(exam)
+            course[:exam_time] = parse_cs_exam_time(exam)
+            course[:exam_slot] = parse_cs_exam_slot(exam)
+          end
           courses << course
         end
       end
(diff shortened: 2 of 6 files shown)


spec/models/my_academics/exams_spec.rb
```diff
@@ -381,6 +407,74 @@
...
+        expect(fall_courses[1][:exam_location]).to_not be
+
+        expect(spring_courses.length).to eq 3
+        expect(spring_courses[0][:exam_location]).to eq 'Kroeber 221'
+        expect(spring_courses[0][:exam_date]).to eq 'Mon 12/12'
+        expect(spring_courses[1][:exam_location]).to_not be
+      end
+
+      it 'should transition data correctly' do
+        pp merged_result
```
raydavis: Oopps, looks like a stray work-in-progress line got through.

app/models/edo_oracle/queries.rb
```diff
@@ -165,6 +165,33 @@ def self.get_section_meetings(term_id, section_id)
       SQL
     end
 
+    # No Campus Oracle equivalent.
+    def self.get_section_final_exam(term_id, section_id)
```
davidlimcheng: I see this being done in other methods as well - but are the terms "course ID" and "section ID" interchangeable?  If so, would it make sense to consolidate the variable names down to one consistent name?  More a question stemming from curiosity than anything else.

Since this is a history question, might be better suited for @raydavis.

raydavis: Good question, @davidlimcheng . "Course ID" & "Section ID" do refer to different things, but descriptions sometime get fuzzy around the aspects of the course which only apply to "primary" sections: grades, exams, etc. Eventually a student's transcript (usually) shows a final grade for the Course, but the grade records themselves are entered for the Section.

app/models/my_academics/exams.rb
```diff
@@ -88,10 +94,37 @@ def sort_exams(semester)
 
     # Determine whether CS data is available depending on where we are in the semester
     def determine_cs_data_available(semester)
-      # TODO Follow-up with real CS data based on exams from Class Section API
-      # term = select_term(semester)
-      # return term.final_exam_cs_data_available < Time.now
-      return false
+      term = Berkeley::Terms.fetch.campus[semester[:slug]]
+      # Still calculated by whether or not we are 8 weeks before the end of the term or not. (Adjust later)
+      return term.final_exam_cs_data_available < Time.now
```
davidlimcheng: Can we define a "current_date" variable that will take `terms.fake_now` from `Settings.yml` if it was available, to facilitate testing for this feature?  Something like 

```
current_date = Settings.terms.fake_now || Time.now
```

Should work.  This way, we can "fake" the current day on CalCentral to something within 8 weeks before the end of a term, allowing the QA team to test this feature in a more realistic context.

app/models/edo_oracle/queries.rb
davidlimcheng: I see, thanks @raydavis.  Still seems a bit weird to me, but sounds like a necessity.

raydavis: The distinction is probably easiest to see for an instructor who teaches 2 different primary sections of the same "course". With some exceptions, a student can only be enrolled in one primary section, and the instructor's final grade is attached to the specific section enrollment, but the student sees the course name (e.g., "CHEM 1A") on their transcript.

davidlimcheng: @raydavis Totally, understood.  The weird part to me is that the call on this function uses `@course_id`, which implies a return of all final exam times/locations for every section in that course, whereas the inputs for this function uses `section_id`, which implies a return of a final exam time/location for one particular section.  Judging from the, code it seems like the former is the case.  But I look at the SQL and it indeed is matching on `sec."id"`!

In any case it doesn't seem like a big deal, and I'm sure I'll understand it more the closer I work with it =)

raydavis: Ah, light finally dawns! Thank you for persisting, @davidlimcheng . Yes, tracing through the call stack, there's more confusion than usual in the variable naming of EdoOracle::CourseSections, possibly due to not including the test case of multiple primary sections (more common for instructors than for students, but it does sometimes happen for students, too). The one caller of the class specifies "CCN" (equivalent to Section ID), so it's definitely _section_ specific:

```
section_data = EdoOracle::CourseSections.new(course[:term_id], section[:ccn]).get_section_data
```

The ``@course`` variable should really be renamed ``section`` to reduce confusion. @tiffanyqi , I'll leave it up to you whether you want to do that change yourself, or just add a comment about it.

davidlimcheng: Ahhh got it, so it indeed is section.  Thanks for digging deep, @raydavis!

tiffanyqi: Thank you confirming @raydavis @davidlimcheng!

spec/models/my_academics/exams_spec.rb
tiffanyqi: Removed, thanks!

app/models/my_academics/exams.rb
tiffanyqi: Added, can you check to see if this is what you wanted? Thanks!

davidlimcheng: Yes, perfect - thank you @tiffanyqi!

davidlimcheng:
Looks great, nice job @tiffanyqi!

PR #5896, opened 2016-09-26
SISRP-25121 Reduce caching load of MyAcademics

raydavis:

Front-end code does not check My Academics for Live Updates & so the feed should not be repeatedly pinged by server-side warming.
MyAcademics::Merged can be cached in JSON form thanks to the addition of FilteredForAdvisor.
This also fixes an inconsistency between Advisor View-As & User Overview.

### DIFF
--- a/app/controllers/advising_student_controller.rb
+++ b/app/controllers/advising_student_controller.rb
@@ -20,7 +20,7 @@ class AdvisingStudentController < ApplicationController
   end
 
   def academics
-    render json: filtered_academics.to_json
+    render json: MyAcademics::FilteredForAdvisor.from_session('user_id' => student_uid_param).get_feed_as_json
   end
 
   def academic_status
@@ -57,27 +57,6 @@ class AdvisingStudentController < ApplicationController
   end
 
   private
-
-  def filtered_academics
-    feed = MyAcademics::Merged.from_session('user_id' => student_uid_param).get_feed
-    if (semesters = feed[:semesters])
-      semesters.each do |s|
-        s.delete :slug
-        (classes = s[:classes]) && classes.each do |c|
-          c.delete :slug
-          c.delete :url
-          (sections = c[:sections]) && sections.each do |section|
-            section.delete :url
-          end
-        end
-      end
-    end
-    (memberships = feed[:otherSiteMemberships]) && memberships.each do |membership|
-      membership.delete :slug
-      membership[:sites].each { |site| site.delete :site_url } if membership[:sites]
-    end
-    feed
-  end
 
   def authorize_for_student
     raise NotAuthorizedError.new('The student lookup feature is disabled') unless is_feature_enabled
--- a/app/models/canvas_csv/provide_course_site.rb
+++ b/app/models/canvas_csv/provide_course_site.rb
@@ -253,11 +253,10 @@ module CanvasCsv
         # Get all sections for which this user is an instructor, sorted in a useful fashion.
         # Since this mostly matches what's shown by MyAcademics::Teaching for a given semester,
         # we can simply re-use the academics feed (so long as course site provisioning is restricted to
-        # semesters supported by My Academics). Ideally, MyAcademics::Teaching would be efficiently cached
-        # by user_id + term_yr + term_cd. But since we currently only cache at the level of the full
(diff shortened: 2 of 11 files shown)


app/models/canvas_csv/provide_course_site.rb
```diff
@@ -257,7 +257,8 @@ def candidate_courses_list
         # by user_id + term_yr + term_cd. But since we currently only cache at the level of the full
         # merged model, we're probably better off selecting the desired teaching-semester from that bigger feed.
         semesters = []
-        academics_feed = MyAcademics::Merged.new(@uid).get_feed
+        academics_feed = {}
+        MyAcademics::Teaching.new(@uid).merge academics_feed
         if (teaching_semesters = academics_feed[:teachingSemesters])
```
paulkerschen: @raydavis Is the calculation here that ProvideCourseSite is used seldom enough that it's not worth hitting and caching all endpoints in order to get a cached MyAcademics::Teaching? If so, might be worth updating the explanation in the comment above.

raydavis: Out-of-date comment. MyAcademics::Teaching draws solely from separately cached data sources (which is why the frequent warming of MyAcademics::Merged didn't totally kill us before). Thank you for noticing the discrepancy, @paulkerschen -- I'll amend.

raydavis: Comment revised, @paulkerschen .

app/controllers/advising_student_controller.rb
```diff
@@ -58,27 +58,6 @@ def student_attributes
...
-        s.delete :slug
-        (classes = s[:classes]) && classes.each do |c|
-          c.delete :slug
-          c.delete :url
-          (sections = c[:sections]) && sections.each do |section|
-            section.delete :url
-          end
-        end
-      end
-    end
```
paulkerschen: @raydavis: While @johncrossman knows more about this feature than I do, I think the intent of filtering out these slugs and URLs was to prevent links from appearing in the advisor dashboard, which shares `academics_semesters.html` with My Academics. Those links will break if followed from the advisor dashboard, since no view-as session is active.

davidlimcheng:
Aside from @paulkerschen's outstanding comment, LGTM!

app/controllers/advising_student_controller.rb
raydavis: Thanks again, @paulkerschen! I was thrown off by the fact that Advisor view-as censors URLs of class websites (along with the Teaching feed), and sloppily assumed that this was after the same result. But instead the existing Student Overview logic is missing _both_ of those deliberate suppressions.

Both need to be applied in the JSON response no matter how Advisors reach it. That bug will get fixed by the switch to FilteredForAdvisor Not showing the links isn't a question of permissions, though; it's just a matter of HTML rendering. I'll amend the PR again for that.

Does all that match your and/or @johncrossman 's understanding?

src/assets/templates/academics_semesters.html
```diff
@@ -21,10 +21,10 @@ <h2 class="cc-left">Semesters</h2>
   <div class="cc-widget-padding cc-flex-space-between-vertical-15">
     <div class="cc-table cc-academics-semester cc-academics-semester-{{semester.timeBucket}}" data-ng-repeat="semester in semesters | limitTo:pastSemestersLimit">
       <div>
-        <h3 data-ng-if="semester.hasEnrollmentData && semester.slug"><a class="cc-left" data-ng-href="/academics/semester/{{semester.slug}}" data-ng-bind="semester.name"></a></h3>
-        <h3 data-ng-if="!semester.hasEnrollmentData || !semester.slug" data-ng-bind="semester.name"></h3>
+        <h3 data-ng-if="semester.hasEnrollmentData && semester.slug && !isAdvisingStudentLookup"><a class="cc-left" data-ng-href="/academics/semester/{{semester.slug}}" data-ng-bind="semester.name"></a></h3>
```
raydavis: @paulkerschen , I've added link-suppression for Student Overview here. The same logic was already in effect elsewhere in this template.

paulkerschen:
Looks good, @raydavis, thanks.

PR #5897, opened 2016-09-26
CLC-6444 Remove CampusOracle::Oec and friends

paulkerschen:
Prior to addressing the performance issue in https://jira.ets.berkeley.edu/jira/browse/CLC-6444, simplify OEC code by clearing out the connection to legacy Oracle.

This does remove the ability to run the OEC tool against terms before 2016-D, but I can think of no scenario in which we'd ever need to do so.


johncrossman:
Looks good to me, @paulkerschen Waiting on Mr. T

PR #5900, opened 2016-09-26
SISRP-25802 Back-End Development for Student Overview (Advisor) - Student Success Card

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-25802

* Decision to make Student Success a merged, uncached feed is based on discussions with Flint Hahn about the future use of this card.  As seen in the [design](https://jira.berkeley.edu/browse/SISRP-19694), this card is designed to summarize important data about the student, and additional data is most likely to be added in the future.
* This could include data we are already pulling, which more likely than not is an already cached proxy.  That being said, we can always rearrange the caching scheme for this feed the if/when it gets larger.
* Some elements on the card is already being pulled by `MyAcademics::GpaUnits` (a part of `MyAcademics::Merged`, which the advisor student overview page is already pulling), and instead of rewriting the "legacy or not-legacy" logic of `MyAcademics::GpaUnits`, I opted to include necessary fields in the existing feed.
* Also includes two new CS endpoints in the verify_sis_endpoints.sh script.


app/models/my_academics/gpa_units.rb
```diff
@@ -24,8 +24,11 @@ def hub_gpa_units
       if (status = parse_hub_academic_status response)
         # GPA is passed as a string to force a decimal point for whole values.
         result[:cumulativeGpa] = (cumulativeGpa = parse_hub_cumulative_gpa status) && cumulativeGpa.to_s
-        result[:totalUnits] = (totalUnits = parse_hub_total_units status) && totalUnits.to_f
```
davidlimcheng: On second thought, the extra computational work here across all users with a `MyAcademics` feed probably isn't worth it, especially considering only advisors will be using the added properties.  Maybe it is worth it to write a `StudentSuccess:Units` model that would be very similar to this one, and cache the `StudentSuccess::Merged` feed until we find a reason to break up the cache by proxy.  

@raydavis, would love your opinion here, maybe we can chat about this tomorrow.

app/models/my_academics/gpa_units.rb
```diff
@@ -24,8 +24,11 @@ def hub_gpa_units
...
         result[:totalUnitsAttempted] = (totalUnitsAttempted = parse_hub_total_units_attempted status) && totalUnitsAttempted.to_f
+        result[:totalUnitsNotForGpa] = (totalUnitsNotForGpa = parse_hub_total_units_not_for_gpa status) && totalUnitsNotForGpa.to_f
```
raydavis: ``totalUnitsAttempted`` = Total units for GPA. Are there any units which are neither ``['type']['code'] == 'Not For GPA'`` nor ``['type']['code'] == 'For GPA'``?

raydavis:
@davidlimcheng Would you mind amending your commit message to prefix the JIRA number?

app/models/my_academics/gpa_units.rb
```diff
@@ -38,7 +41,10 @@ def parse_hub_cumulative_gpa(status)
 
     def parse_hub_total_units(status)
       if (units = status['cumulativeUnits']) && (total_units = units.find { |u| u['type'] && u['type']['code'] == 'Total'})
-        total_units['unitsPassed']
+        {
+          totalUnits: total_units['unitsPassed'].to_f,
+          transferUnitsAccepted: total_units['unitsTransferEarned'].to_f
```
davidlimcheng: Include testing units.

davidlimcheng: Do a sanity check if 'For GPA' and 'Not For GPA' add up to 'Total', if not, log a warning.

app/models/student_success/student_success_module.rb
```diff
@@ -0,0 +1,22 @@
+module StudentSuccess
+  module StudentSuccessModule
+    extend self
+
+    def parse_term_gpa(response)
+      term_gpa = response[:feed] && response[:feed][:ucAaTermData] && response[:feed][:ucAaTermData][:ucAaTermGpa]
```
lyttam: I think you could write this as 

`term_gpa = response.fetch(:feed, {}).fetch(:ucAaTermData, {}).fetch(:ucAaTermGpa, nil)`

Or perhaps there are better alternatives?  I've been using `fetch` because I find it easier to read, but would like to know if anyone prefers a different approach.

raydavis: When you're willing to let the default value be "nil" (like you have here), @paulkerschen convinced me that ``.try(:[],: ucAaTermData).try(:[], :ucAaTermGpa)`` is better since it won't throw an exception if the top object is nil.

davidlimcheng: Ah, will amend this to use `.try()`.  Thanks guys!

app/models/my_academics/gpa_units.rb
davidlimcheng: As we discussed this morning, there are also the following, which are included in the "Total" object:

- `unitsTest` - Units obtained from testing, including AP tests
- `unitsTransferEarned` - Units obtained from transfer credit
- `unitsOther` - Units that exceed limit from other categories, such as transfer credits that exceed the UCB limit.

davidlimcheng:
@raydavis, @lyttam:  Amended with fixes, ready for review.

app/models/my_academics/gpa_units.rb
```diff
@@ -53,9 +81,16 @@ def oracle_gpa_units
       {
         cumulativeGpa: student_info['cum_gpa'].nil? ? nil: student_info['cum_gpa'].to_s,
         totalUnits: student_info['tot_units'].nil? ? nil : student_info['tot_units'].to_f,
-        totalUnitsAttempted: student_info['lgr_tot_attempt_unit'].nil? ? nil : student_info['lgr_tot_attempt_unit'].to_f
+        totalUnitsAttempted: student_info['lgr_tot_attempt_unit'].nil? ? nil : student_info['lgr_tot_attempt_unit'].to_f,
+        isLegacy: true
```
davidlimcheng: @raydavis, I've added an `isLegacy` flag here, as it got pretty confusing on the front-end when you're expecting the extra hub data we're adding, and it didn't show up.  I figure we can use this to easily hide/show newer hub data when necessary.

app/models/student_success/outstanding_balance.rb
```diff
@@ -0,0 +1,19 @@
...
+      @student_uid_param = opts[:user_id]
+    end
+
+    def merge(data={})
+      data[:outstandingBalance] = outstanding_balance
+    end
+
+    def outstanding_balance
+      response = CampusSolutions::StudentOutstandingBalance.new(user_id: @student_uid_param).get
+      parse_outstanding_balance response
```
raydavis: Why is this broken out into a separate module instead of just included here?

app/models/student_success/term_gpa.rb
```diff
@@ -0,0 +1,20 @@
...
+      @student_uid_param = opts[:user_id]
+    end
+
+    def merge(data={})
+      data[:termGpa] = term_gpa
+    end
+
+    def term_gpa
+      response = CampusSolutions::StudentTermGpa.new(user_id: @student_uid_param).get
+      parse_term_gpa response
```
raydavis: Same question as with ``parse_outstanding_balance`` -- seems like we're making devs unnecessarily jump around from file to file to follow the logic, but maybe I'm missing something?

app/models/my_academics/gpa_units.rb
```diff
@@ -53,9 +81,16 @@ def oracle_gpa_units
...
       }
     end
 
+    def unit_sum_check(unit_total, unit_sum)
```
raydavis: It's unusual for us to move a simple log message into a separate method unless the logic is needed in multiple places. What was the thinking here?

app/models/my_academics/gpa_units.rb
raydavis: Makes sense!

app/models/my_academics/gpa_units.rb
davidlimcheng: Gahh, sloppiness and eagerness.  I originally had the arithmetic here but decided it was better above - will amend to remove this method.

app/models/student_success/outstanding_balance.rb
davidlimcheng: I originally thought the `StudentSuccess` module would be doing a lot more work, but `MyAcademics::GpaUnits` ended up carrying a lot of it - makes sense to get rid of the module since we've only got two small methods in there.  I'll amend removing the module.

app/models/student_success/term_gpa.rb
davidlimcheng: Nope!  See other comment.

davidlimcheng:
@raydavis Commit has been amended to remove the Student Success module, and extra logging method on `gpaUnits`.

raydavis:
Thanks, @davidlimcheng -- easier to follow now!

PR #5905, opened 2016-09-27
SISRP-23155 Empty semester card if user has only waitlisted courses

tiffanyqi:
https://jira.berkeley.edu/browse/SISRP-23155

### DIFF
--- a/app/models/my_academics/semesters.rb
+++ b/app/models/my_academics/semesters.rb
@@ -31,14 +31,24 @@ module MyAcademics
           semester[:hasEnrollmentData] = true
           semester[:summaryFromTranscript] = (semester[:timeBucket] == 'past')
           semester[:classes] = map_enrollments(enrollment_terms[term_key]).compact
+          semester[:has_enrolled_classes] = has_enrolled_classes?(enrollment_terms[term_key])
           merge_grades(semester, transcript_terms[term_key])
         else
           semester[:hasEnrollmentData] = false
           semester[:summaryFromTranscript] = true
+          semester[:has_enrolled_classes] = false
           semester[:classes] = map_transcripts transcript_terms[term_key][:courses]
           semester[:notation] = translate_notation transcript_terms[term_key][:notations]
         end
         semester unless semester[:classes].empty?
+      end
+    end
+
+    def has_enrolled_classes?(enrollment_term)
+      !!enrollment_term.find do |course|
+        if course[:role] == 'Student' && course[:sections]
+          course[:sections].find{|section| !section[:waitlisted]}
+        end
       end
     end
 
--- a/spec/models/my_academics/semesters_spec.rb
+++ b/spec/models/my_academics/semesters_spec.rb
@@ -64,6 +64,7 @@ describe MyAcademics::Semesters do
     format = opts[:format] || ['LEC', 'DIS', 'SEM'].sample
     section_number = opts[:section_number] || "00#{rand(9)}"
     is_primary_section = opts[:is_primary_section] || false
+    waitlisted = opts[:waitlisted]
     section = {
       associated_primary_id: opts[:associated_primary_id],
       ccn: opts[:ccn] || random_ccn,
@@ -81,6 +82,7 @@ describe MyAcademics::Semesters do
           schedule: 'MWF 11:00A-12:00P'
         }]
       },
+      waitlisted: waitlisted,
       instructors: [{name: random_name, uid: random_id}]
     }
     if opts[:edo_source]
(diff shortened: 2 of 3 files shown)


app/models/my_academics/semesters.rb
```diff
@@ -31,17 +31,31 @@ def semester_feed(enrollment_terms, transcript_terms)
...
+      no_enrolled_classes = true
+      enrollment_term.map do |course|
+        if course[:role] == 'Student'
+          course[:sections].each do |section|
+            no_enrolled_classes = false if !section[:waitlisted]
+          end
+        end
+      end
+      return no_enrolled_classes
+    end
```
nkutub: I am not sure I understand what the logic is trying to do here? map returns an array based on operation done for each element in enrolment_term, but I think you want to return only one value false if you see a waitlisted section., just use an each instead of map.

app/models/my_academics/semesters.rb
```diff
@@ -31,17 +31,31 @@ def semester_feed(enrollment_terms, transcript_terms)
           semester[:hasEnrollmentData] = true
           semester[:summaryFromTranscript] = (semester[:timeBucket] == 'past')
           semester[:classes] = map_enrollments(enrollment_terms[term_key]).compact
+          semester[:no_enrolled_classes] = check_enrolled_classes(enrollment_terms[term_key])
```
nkutub: is there a reason you are using a negative flag here instead of has_enrolled_classes. It is easier to read than for example no_enroled_classes = false which is a double negative and hard to read

spec/models/my_academics/semesters_spec.rb
```diff
@@ -64,6 +64,7 @@ def course_enrollment_section(opts={})
     format = opts[:format] || ['LEC', 'DIS', 'SEM'].sample
     section_number = opts[:section_number] || "00#{rand(9)}"
     is_primary_section = opts[:is_primary_section] || false
+    waitlisted = opts[:waitlisted] || nil
```
nkutub: Is this needed? will waitlisted not equal to nil if opts[:waitlisted] is nil and thus the || is superfluous?

app/models/my_academics/semesters.rb
nkutub: a way to short circut the search in a collection and not have to loop through all items if you find the that the first item has what your looking for. using .find { } might be faster.

app/models/my_academics/semesters.rb
tiffanyqi: `has_enrolled_classes` is way cleaner, will change this. Thanks!

app/models/my_academics/semesters.rb
tiffanyqi: Would it be clearer to use `enrollment_term.each` instead of `map` ? I'm not sure how to use `.find` in this context, since it is looping through the `enrollment_term`, `course`, and `section` before it can state whether or not `has_enrolled_classes` is true.

spec/models/my_academics/semesters_spec.rb
tiffanyqi: Oops, you're right this is not needed. Ready to be checked when you are!

app/models/my_academics/semesters.rb
tiffanyqi: Fixed via our chat

src/assets/templates/academics_semesters.html
```diff
@@ -24,8 +24,13 @@ <h2 class="cc-left">Semesters</h2>
...
         <h4 data-ng-if="semester.notation" data-ng-bind="semester.notation"></h4>
-        <a class="cc-right cc-academics-semester-title-link" data-ng-href="/academics/booklist/{{semester.slug}}" data-ng-if="semester.hasEnrollmentData && semester.slug && semester.timeBucket !== 'past'  && !isAdvisingStudentLookup">Book List</a>
-        <table>
+        <a class="cc-right cc-academics-semester-title-link" data-ng-href="/academics/booklist/{{semester.slug}}" data-ng-if="semester.hasEnrollmentData && semester.slug && semester.timeBucket !== 'past'">Book List</a>
+        <br>
+        <div class="cc-flex-space-between-vertical-15" data-ng-if="semester.no_enrolled_classes">
+          You are currently not enrolled in any classes for
+          <span data-ng-bind="semester.name"></span>.
+        </div>
+        <table data-ng-if="!semester.no_enrolled_classes">
```
nkutub: I think you missed this is should be semester.has_enrolled_classes.

src/assets/templates/academics_semesters.html
```diff
@@ -24,8 +24,13 @@ <h2 class="cc-left">Semesters</h2>
         <h3 data-ng-if="semester.hasEnrollmentData && semester.slug && !isAdvisingStudentLookup"><a class="cc-left" data-ng-href="/academics/semester/{{semester.slug}}" data-ng-bind="semester.name"></a></h3>
         <h3 data-ng-if="!semester.hasEnrollmentData || !semester.slug || isAdvisingStudentLookup" data-ng-bind="semester.name"></h3>
         <h4 data-ng-if="semester.notation" data-ng-bind="semester.notation"></h4>
-        <a class="cc-right cc-academics-semester-title-link" data-ng-href="/academics/booklist/{{semester.slug}}" data-ng-if="semester.hasEnrollmentData && semester.slug && semester.timeBucket !== 'past'  && !isAdvisingStudentLookup">Book List</a>
-        <table>
+        <a class="cc-right cc-academics-semester-title-link" data-ng-href="/academics/booklist/{{semester.slug}}" data-ng-if="semester.hasEnrollmentData && semester.slug && semester.timeBucket !== 'past'">Book List</a>
+        <br>
+        <div class="cc-flex-space-between-vertical-15" data-ng-if="semester.no_enrolled_classes">
```
nkutub: I think this should be !semester.has_enrolled_classes

tiffanyqi: Oops, just fixed. Thanks for the catch!

src/assets/templates/academics_semesters.html
tiffanyqi: Oops, just fixed. Thanks for the catch!

PR #5906, opened 2016-09-28
SISRP-25752 - 7.1 - Retrofit Concurrent Enrollment links to the link API

dfkaye:
https://jira.berkeley.edu/browse/SISRP-25752

### DIFF
--- a/app/models/my_academics/class_enrollments.rb
+++ b/app/models/my_academics/class_enrollments.rb
@@ -144,7 +144,26 @@ module MyAcademics
     end
 
     def get_links
-      Settings.campus_solutions_links.class_enrollment.as_json['table']
+      cs_links = {}
+
+      campus_solutions_link_settings = [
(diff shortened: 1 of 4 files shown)


app/models/my_academics/class_enrollments.rb
```diff
@@ -144,7 +144,26 @@ def get_active_career_terms
     end
 
     def get_links
-      Settings.campus_solutions_links.class_enrollment.as_json['table']
```
raydavis: @dfkaye , when we add new configuration settings to the code, we also document their existence in the central ``config/settings.yml`` file. Similarly, if you're removing settings, it's important to delete them from that file. Looks like that step might be missing in this PR?

dfkaye: @raydavis - ok - will amend in a bit - thanks

redconfetti: 💐 beautiful

PR #5907, opened 2016-09-28
SISRP-24520 - Add Term based CPP data to User Profile

redconfetti:
https://jira.berkeley.edu/browse/SISRP-24520

### DIFF
--- a/app/controllers/advising_student_controller.rb
+++ b/app/controllers/advising_student_controller.rb
@@ -32,7 +32,7 @@ class AdvisingStudentController < ApplicationController
   end
 
   def registrations
-    render json: MyRegistrations::MyRegistrations.new(student_uid_param).get_feed_as_json
+    render json: MyAcademics::Registrations.new(student_uid_param).get_feed_as_json
   end
 
   def advising
--- a/app/models/my_academics/academics_module.rb
+++ b/app/models/my_academics/academics_module.rb
@@ -28,21 +28,9 @@ module MyAcademics
       }
     end
 
-    # TODO: Refactor dependencies to rely on #parse_hub_academic_statuses
-    # Students with multiple careers have an academic status for each career
-    def parse_hub_academic_status(response)
-      status = response[:feed] && response[:feed]['student'] && (response[:feed]['student']['academicStatuses'].try :first)
-      status if status.present?
-    end
-
     def parse_hub_academic_statuses(response)
       status = response[:feed] && response[:feed]['student'] && response[:feed]['student']['academicStatuses']
       status if status.present?
-    end
-
-    def parse_hub_holds(response)
-      holds = response[:feed] && response[:feed]['student'] && (response[:feed]['student']['holds'])
-      holds if holds.present?
     end
 
     def course_info(campus_course)
--- a/app/models/my_academics/class_enrollments.rb
+++ b/app/models/my_academics/class_enrollments.rb
@@ -5,24 +5,10 @@ module MyAcademics
     include Cache::UserCacheExpiry
     include CampusSolutions::EnrollmentCardFeatureFlagged
 
-    # Used to identify different enrollment instruction presentations
-    ENROLLMENT_INSTRUCTION_TYPES = {
-      plan: [
-        {instruction_type_code: 'fpf', match: '25000FPFU'},
-        {instruction_type_code: 'haas_mba', match: '70141MBAG'},
-        {instruction_type_code: 'haas_ewmba', match: '701E1MBAG'},
-        {instruction_type_code: 'haas_execmba', match: '70364MBAG'},
-      ],
-      career: [
-        {instruction_type_code: 'law', match: 'LAW'},
-        {instruction_type_code: 'concurrent', match: 'UCBX'},
-      ]
-    }
-
     def get_feed_internal
       return {} unless is_feature_enabled && user_is_student?
       HashConverter.camelize({
-        enrollmentTermInstructionTypes: get_enrollment_term_instruction_types,
+        enrollmentTermInstructionTypes: get_career_term_roles,
         enrollmentTermInstructions: get_enrollment_term_instructions,
         enrollmentTermAcademicPlanner: get_enrollment_term_academic_planner,
         hasHolds: user_has_holds?,
@@ -30,19 +16,35 @@ module MyAcademics
       })
     end
 
-    def get_enrollment_term_instruction_types
+    # Groups student plans into groups based on roles (e.g. 'default', 'fpf', 'concurrent')
+    def grouped_student_plan_roles
+      grouped_roles = {}
+      active_plans.each do |plan|
+        role_code = plan[:role]
+        career_code = plan[:career][:code]
+        role_key = [role_code, career_code]
+        grouped_roles[role_key] = { role: role_code, career_code: career_code, academic_plans: [] } if grouped_roles[role_key].blank?
+        grouped_roles[role_key][:academic_plans] << plan
+      end
+      grouped_roles
+    end
+
+    # Returns unique couplings of current career-terms and current student plan roles
+    def get_career_term_roles
       career_terms = get_active_career_terms
-      instruction_types = get_enrollment_instruction_types
-      term_instruction_types = []
-      instruction_types.keys.each do |type_key|
-        instruction_type = instruction_types[type_key]
+
+      grouped_roles = grouped_student_plan_roles
+      career_term_plan_roles = []
+
+      grouped_roles.keys.each do |role_key|
+        student_plan_role = grouped_roles[role_key]
         career_terms.each do |career_term|
-          if (instruction_type[:career_code] == career_term[:acadCareer])
-            term_instruction_types << instruction_type.merge({term: career_term.slice(:termId, :termDescr)})
+          if (student_plan_role[:career_code] == career_term[:acadCareer])
+            career_term_plan_roles << student_plan_role.merge({term: career_term.slice(:termId, :termDescr)})
           end
         end
       end
-      term_instruction_types
+      career_term_plan_roles
     end
 
     def get_enrollment_term_academic_planner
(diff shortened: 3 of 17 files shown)


app/controllers/advising_student_controller.rb
```diff
@@ -32,7 +32,7 @@ def enrollment_instructions
   end
 
   def registrations
-    render json: MyRegistrations::MyRegistrations.new(student_uid_param).get_feed_as_json
+    render json: MyAcademics::Registrations.new(student_uid_param).get_feed_as_json
```
redconfetti: I simply renamed this class because it was redundant and all alone under the `MyRegistrations` module.

app/models/my_academics/academics_module.rb
```diff
@@ -28,23 +28,11 @@ def semester_info(term_key)
...
-
     def parse_hub_academic_statuses(response)
       status = response[:feed] && response[:feed]['student'] && response[:feed]['student']['academicStatuses']
       status if status.present?
     end
 
-    def parse_hub_holds(response)
-      holds = response[:feed] && response[:feed]['student'] && (response[:feed]['student']['holds'])
-      holds if holds.present?
-    end
```
redconfetti: The only dependency of `parse_hub_holds` was MyAcademics::ClassEnrollments and MyAcademics::CollegeAndLevel, so I moved it under MyAcademics::CollegeAndLevel and updated it return a parsed holds feed, which is currently only a boolean indicating that holds are present.

app/models/my_academics/class_enrollments.rb
```diff
@@ -5,44 +5,46 @@ class ClassEnrollments < UserSpecificModel
...
-        {instruction_type_code: 'fpf', match: '25000FPFU'},
-        {instruction_type_code: 'haas_mba', match: '70141MBAG'},
-        {instruction_type_code: 'haas_ewmba', match: '701E1MBAG'},
-        {instruction_type_code: 'haas_execmba', match: '70364MBAG'},
-      ],
-      career: [
-        {instruction_type_code: 'law', match: 'LAW'},
-        {instruction_type_code: 'concurrent', match: 'UCBX'},
-      ]
-    }
```
redconfetti: These constants were moved to MyAcademics::CollegeAndLevel as STUDENT_PLAN_ROLES. "Enrollment Instruction Types" are now known simply as "roles" or "student plan roles".

Within MyAcademics::ClassEnrollment, the grouped plans will still be referred to as "career term roles" (roles paired with career-terms), or 'enrollment instruction types' (representing each variant of the class enrollment card to be displayed by the front-end).

app/models/my_academics/class_enrollments.rb
```diff
@@ -5,44 +5,46 @@ class ClassEnrollments < UserSpecificModel
...
+      grouped_roles = {}
+      active_plans.each do |plan|
+        role_code = plan[:role]
+        career_code = plan[:career][:code]
+        role_key = [role_code, career_code]
+        grouped_roles[role_key] = { role: role_code, career_code: career_code, academic_plans: [] } if grouped_roles[role_key].blank?
+        grouped_roles[role_key][:academic_plans] << plan
+      end
+      grouped_roles
+    end
```
redconfetti: To match the terminology used for the 'student plan roles' provided by MyAcademics::CollegeAndLevel, I renamed this function after what it does, which is provide a grouped list of student plan roles that apply to the current student.

app/models/my_academics/class_enrollments.rb
```diff
@@ -64,69 +66,20 @@ def get_enrollment_term_instructions
...
-        })
+    def college_and_level
+      worker = Proc.new do
+        feed = {}
+        MyAcademics::CollegeAndLevel.new(@uid).merge(feed)
+        feed.try(:[], :collegeAndLevel)
       end
-      flat_plan
+      @college_and_level ||= worker.call
     end
```
redconfetti: The logic involving the extraction of a students academic plan data (CPP) is now moved over to the MyAcademics::CollegeAndLevel class. It is shared by both the My Academics Merged feed and now the Class Enrollments feed, each which perform their own cacheing. The ClassEnrollment class now only does it's own job which is to group the plans by role type (FPF, Concurrent, Law, Haas, Default), and then correlate the plans with the career-terms, thus making an "enrollment instruction type" object for each variant of the class enrollment card that is displayed by the front-end.

app/models/my_academics/college_and_level.rb
```diff
@@ -4,6 +4,20 @@ class CollegeAndLevel
     include ClassLogger
     include User::Student
 
+    # Student Plan Roles represent the type of student based on their student plans
+    STUDENT_PLAN_ROLES = {
+      plan: [
+        {student_plan_role_code: 'fpf', match: '25000FPFU'},
+        {student_plan_role_code: 'haas_mba', match: '70141MBAG'},
```
redconfetti: Now that I think of it, we should perhaps remove the Haas roles, because they're not in use yet with this release.

app/models/my_academics/college_and_level.rb
```diff
@@ -200,6 +225,79 @@ def parse_bearfacts_feed(feed)
       feed
     end
 
+    def get_academic_status
+      @academic_status ||= HubEdos::AcademicStatus.new({user_id: @uid}).get
+    end
```
redconfetti: `#get_academic_status` and the methods defined below it come from MyAcademics::ClassEnrollment. I prefer the way they work, because they are separate methods with their own jobs, instead of doing everything within `#parse_hub_plans`. 

Because MyAcademics::CollegeAndLevel uses a loop above in #parse_hub_plans in the last release, I decided to allow it to continue doing that to extract majors, minors, and the last expected graduation term. This is the most efficient manner of extracting the data, instead of looping through the plans again and again in separate methods just to get these.

I still preferred to expand #flatten_plan so that it identifies the "type" (major or minor), college, and expected graduation term, just like MyAcademics::CollegeAndLevel did before within `#parse_hub_plans `.

We're now extracting all relevant information with #flatten_plan, and then processing misc abstractions within #parse_hub_plans. I hope you prefer this approach @raydavis

app/models/my_academics/gpa_units.rb
```diff
@@ -21,7 +21,8 @@ def hub_gpa_units
       result = {}
       #copy needed feilds from response obj
       result[:errored] = response[:errored]
-      if (status = parse_hub_academic_status response)
+      # TODO: Consult with SR concerning GPA displayed when multiple academic careers present
+      if (status = parse_hub_academic_statuses(response).try :first)
```
redconfetti: I wanted to remove `#parse_hub_academic_status`, so I simply retrofitted MyAcademics::GpaUnits so that it does what it did before, which is extract GPA/Unit data from ONLY the first Academic Status.

We'll fix this once we consult with Student Records for a future release.

public/dummy/json/academics.json
```diff
@@ -40,7 +40,11 @@
     ],
     "termName": "Fall 2013",
     "termsInAttendance": "4",
-    "lastExpectedGraduationTerm": "2017 Spring"
+    "lastExpectedGraduationTerm": {
```
redconfetti: I decided that it would be better if we have the term code, the string we're provided, and the English formatted string (thanks `Berkeley::TermCodes.normalized_english`). I think a bug was reported regarding this anyway. @davidlimcheng?

spec/models/my_academics/class_enrollments_spec.rb
```diff
@@ -34,112 +34,103 @@
       }
     }
   end
-  let(:edo_hub_academic_status_feed) do
+  let(:college_and_level_feed) do
```
redconfetti: I'm now letting the spec stub out the college and level feed. It's easier to maintain this way.

spec/models/my_academics/college_and_level_spec.rb
```diff
@@ -3,18 +3,200 @@
   let(:uid) { '61889' }
   let(:campus_solutions_id) { '1234567890' }
   let(:legacy_campus_solutions_id) { '11667051' }
-  let(:hub_status_proxy) { HubEdos::AcademicStatus.new(user_id: uid, fake: true) }
   let(:bearfacts_proxy) { Bearfacts::Profile.new(user_id: uid, fake: true) }
   let(:fake_spring_term) { double(is_summer: false, :year => 2015, :code => 'B') }
   let(:feed) { {}.tap { |feed| subject.merge feed } }
+
+  # Hub Academic Status - Response / Feed
```
redconfetti: I really couldn't test the behavior elegantly by modifying the Fixture version of the Hub Academic Status, so I decided to completely build the feed, including only the consumed elements, using `let` statements with over-rides within the various contexts. I think it's beautiful, I hope you do too.

src/assets/javascripts/angular/controllers/widgets/enrollment/enrollmentCardController.js
```diff
@@ -171,8 +171,7 @@ angular.module('calcentral.controllers').controller('EnrollmentCardController',
    * @return {Boolean}            True when any type code provided matches
    */
   $scope.isInstructionType = function(instruction, typeCodes) {
-    return typeCodes.indexOf(instruction.instructionTypeCode) !== -1;
-    // return instruction.instructionTypeCode === typeCode;
+    return typeCodes.indexOf(instruction.role) !== -1;
```
redconfetti: No longer called `instructionTypeCode`, it's now `role`.

src/assets/javascripts/angular/services/academicsService.js
```diff
@@ -52,7 +52,7 @@ angular.module('calcentral.services').service('academicsService', function() {
   var expectedGradTerm = function(collegeAndLevel) {
     var careers = _.get(collegeAndLevel, 'careers');
     if (isNotGradOrLawStudent(careers) && collegeAndLevel.lastExpectedGraduationTerm) {
-      return collegeAndLevel.lastExpectedGraduationTerm;
+      return collegeAndLevel.lastExpectedGraduationTerm.name;
```
redconfetti: No longer a string but now a hash/JSON object that contains `name`.

public/dummy/json/academics.json
davidlimcheng: @redconfetti makes complete sense to me.

davidlimcheng: Although, I do question the usefulness of `description`.  If we have `name` and are expecting to only expose that in the front-end, why pass `description` to the front-end at all?

davidlimcheng:
Wow, what a doozy.  Looks good to me @redconfetti, great work dude!

app/models/my_academics/college_and_level.rb
nkutub: not sure what you mean by not part of this release, I have ticket SISRP-24519 that looks like it would depend on this and needs to be in tomorrow.

app/models/my_academics/college_and_level.rb
```diff
@@ -41,74 +55,85 @@ def bearfacts_college_and_level
...
-      end
+      end.uniq.compact
     end
 
-    def parse_hub_level(status)
-      status['currentRegistration'].try(:[], 'academicLevel').try(:[], 'level').try(:[], 'description')
+    def parse_hub_level(statuses)
+      level = statuses.collect do |status|
+        status['currentRegistration'].try(:[], 'academicLevel').try(:[], 'level').try(:[], 'description')
+      end.uniq.compact.to_sentence
```
redconfetti:
```
Push it to the limit
Walk along the razor's edge
but don't look down, just keep your head
and you'll be finished
```

app/models/my_academics/college_and_level.rb
```diff
@@ -200,6 +225,79 @@ def parse_bearfacts_feed(feed)
...
+        flat_plan[:plan].merge!({
+          code: plan.try(:[], 'code'),
+          description: plan.try(:[], 'description')
+        })
+
+        if (hub_plan['expectedGraduationTerm'])
+          expected_grad_term_name = hub_plan['expectedGraduationTerm'].try(:[], 'name')
+          flat_plan[:expectedGraduationTerm] = {
+            code: hub_plan['expectedGraduationTerm'].try(:[], 'id'),
+            description: expected_grad_term_name,
```
raydavis: As @davidlimcheng notes below, generally when we use ``normalized_english`` to fix Campus Solution's reverso season names, we only ship the readable version back to the browser, ignoring the CS version. Not a merge blocker, though!

raydavis:
This is impressive & imposing, @redconfetti ! I'm not about to pretend that I've run through the entire workflow in my head, but the resulting source files look good from here. I'm OK to merge if you don't want to make any last minute amendments.

public/dummy/json/academics.json
redconfetti: @davidlimcheng - description is the original value, but I can remove it. I'll do that before we merge this, since we don't use it.

redconfetti:
@raydavis Once this goes green, it's ready for you to merge. Or perhaps @davidlimcheng can merge it.

PR #5911, opened 2016-09-28
SISRP-25956 - Residency exception from unknown message code

dfkaye:

### DIFF
--- a/app/models/berkeley/residency_message_code.rb
+++ b/app/models/berkeley/residency_message_code.rb
@@ -46,7 +46,7 @@ module Berkeley
       end
       if message_code.nil?
         logger.warn "Cannot determine message code for residency '#{residency_status}', SLR '#{slr_status}', tuition exception '#{tuition_exception}'"
-        ''
+        {}
       else
         message_code
       end
--- a/spec/models/berkeley/residency_message_code_spec.rb
+++ b/spec/models/berkeley/residency_message_code_spec.rb
@@ -40,7 +40,7 @@ describe Berkeley::ResidencyMessageCode do
     describe 'resident' do
       let(:residency_status) { 'RES' }
       let(:slr_status) { 'ANYTHING' }
-      describe 'no tuition exception' do
+      context 'when no tuition exception' do
         let(:tuition_exception) { '' }
         it {should eq '2003'}
       end
@@ -73,7 +73,7 @@ describe Berkeley::ResidencyMessageCode do
     describe 'non-resident' do
       let(:residency_status) { 'NON' }
       let(:slr_status) { 'ANYTHING' }
-      describe 'no tuition exception' do
+      context 'when no tuition exception' do
         let(:tuition_exception) { '' }
         it {should eq '2004'}
       end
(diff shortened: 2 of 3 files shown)


raydavis:
Thanks, @dfkaye, so far so good. The JIRA lists two problems:

> 1. What's the unknown code?
> 2. We need to recover from unknown codes without throwing a TypeError exception.

Looks like this PR fixes (2). Any idea what's up with (1)? That's important, since the meaning of the code is probably something we need to convey to the student (rather than ignoring).

dfkaye:
@raydavis - By 'cannot determine message code from <params/>' we mean that the combination of params used in this case does not resolve to a residency message that can be fetched from the message catalog.  That's actually OK - the UI shouldn't say anything about a message that can't be found...

But I think now the misleading part is the use of `logger.warn` rather than assigning the logged statement as part of the return object when no message code can be matched. I suppose another refactoring like return `{ code: 2001 }` in a success case and `{ unmatched: "no code found for x,yz" }` in the no-match case.

Would that make things easier/clearer going forward?

raydavis:
> the combination of params used in this case does not resolve to a residency message that can be fetched from the message catalog. That's actually OK - the UI shouldn't say anything about a message that can't be found...

I agree, the UI shouldn't say anything about a message that can't be found.

But I don't think it's OK for CalCentral to not know what a message code means. That would usually indicate one of the following:

A) We were not told about one of the residency states, and so we need to be educated in a hurry.
B) Upstream data sources added a new state or changed one without telling us, and so we need to hustle to find out what happened.
C) There's a bug in our code.

Any of these requires action by the service team, which is why the condition is logged.

dfkaye:
Ok @raydavis - should I make the log statement less mysterious? e.g., "no residency message catalog number etc."?

raydavis:
No, @dfkaye , there's nothing wrong with the log message. What's wrong is our ignorance. Someone on the team -- you or someone else, if you want to leave the JIRA open -- needs to contact the API / data owners and ask them what's going on.

dfkaye:
@raydavis - ah - I see what's happening - I wasn't being clear - Only certain combinations of the params can be matched to a message code - the rest not considered valid and result in no message code to be sent.  The logging statement makes it appear that a non-matching combination is an error condition - whereas it's just a report that an non-matching combination was tried.

I'll work on this in the morning with Chara to see whether some combinations were missed and/or how to handle non-matching combinations more gracefully.

raydavis:
It does seem as though we've hit the wall on conversation through GitHub, @dfkaye. :)

My point is pretty basic: Upstream data sources aren't simply trying out random combinations and shrugging when one doesn't make sense. When certain combinations are _intended_ to repress the "Residency" display, then our code should explicitly treat them as successful. But when combinations are _unknown_ to us, then either our code needs to change or the upstream data sources need to be alerted; at any rate, some human communication is called for.

raydavis:
At the JIRA ticket, the data owner confirmed that this was a mistake in upstream systems & it's now been corrected, so no need for more development work here. Merge ahoy!

PR #5912, opened 2016-09-28
CLC-6444 Improve DB performance on OEC enrollments query

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6444

Bamboo can do: https://bamboo.ets.berkeley.edu/bamboo/browse/CPB-CPT384-7/

Used to be, when gathering student and enrollment data for OEC, we would query Oracle only for sections we cared about, and would run two separate queries on those sections: a `get_enrollments` query that returned section-ID/UID associations, and a `get_students` query that used `SELECT DISTINCT` to return unique rows of student data.

Now that we have more than 4000 evaluated sections in production, our giant list is hobbling the database. Ditch the list, run a single query to get all enrollment data for the term (accepting many duplicate student data rows), and post-process in code.

Our standard DB-facing code does not perform well on a result set of 200,000-plus rows. We save memory by handling rows and columns as separate arrays, merging them into the more convenient hash format only when validation requires.


paulkerschen:
Further testing against production data suggests the performance problems are not yet licked. Closing this for the moment.

PR #5917, opened 2016-09-29
SISRP-26013 - SIR option display broken in Calcentral

nkutub:
https://jira.berkeley.edu/browse/SISRP-26013

simple fix, looks like previous code did not account for multiple acceptance options and was not checking for each specific option to render UI elements.


raydavis:
I added a link to the JIRA ticket, @nkutub

PR #5918, opened 2016-09-29
SISRP-25837 - Class Enrollment card - Allow FPF only for a single term

redconfetti:
https://jira.berkeley.edu/browse/SISRP-25837


raydavis:
Ugh. +1 to your hopes for "a proper term-specific data set"!

PR #5919, opened 2016-09-29
CLC-6444 OEC: Reduce memory overhead of enrollments query and worksheet processing

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6444
Bamboo: https://bamboo.ets.berkeley.edu/bamboo/browse/CPB-CPT384-13/

This PR does two things, split into two commits.

Part one reworks the enrollments query. Used to be, when gathering student and enrollment data for OEC, we would query Oracle only for sections we cared about, and would run two separate queries on those sections: a get_enrollments query that returned section-ID/UID associations, and a get_students query that used SELECT DISTINCT to return unique rows of student data.

Now that we have more than 4000 evaluated sections in production, our giant list is hobbling the database. We ditch the list, run a single query to get all student and enrollment data for the term, and post-process in code.

Our standard DB-facing code does not perform well on a result set of 200,000-plus rows. We save memory by handling rows and columns as separate arrays, and merge them into the more convenient hash format only when validation requires.

Part two reduces memory footprint by changing the internal representation of worksheet rows from hashes to arrays. This is fairly easily done because every worksheet comes with a known sequence of headers and can map between string keys and integer indexes on the fly.

### DIFF
--- a/app/models/edo_oracle/connection.rb
+++ b/app/models/edo_oracle/connection.rb
@@ -5,13 +5,13 @@ module EdoOracle
       Settings.edodb
     end
 
-    def self.safe_query(sql)
+    def self.safe_query(sql, opts={})
       result = []
       return result if fake?
       use_pooled_connection do
         result = connection.select_all sql
       end
-      stringify_ints! result
+      opts[:do_not_stringify] ? result : stringify_ints!(result)
     rescue => e
       logger.error "Query failed: #{e.class}: #{e.message}\n #{e.backtrace.join("\n ")}"
       []
--- a/app/models/edo_oracle/oec.rb
+++ b/app/models/edo_oracle/oec.rb
@@ -80,15 +80,39 @@ module EdoOracle
       SQL
     end
 
-    def self.get_enrollments(term_id, select_clause, ccns_clause)
-      safe_query <<-SQL
-        #{select_clause}
-        WHERE
-          enroll."TERM_ID" = '#{term_id}'
-          AND #{ccns_clause}
-          AND (enroll."STDNT_ENRL_STATUS_CODE" = 'E')
-        ORDER BY ldap_uid
+    # See http://www.oracle.com/technetwork/issue-archive/2006/06-sep/o56asktom-086197.html for explanation of
+    # query batching with ROWNUM.
+    def self.get_batch_enrollments(term_id, batch_number, batch_size)
+      mininum_row_exclusive = (batch_number * batch_size)
+      maximum_row_inclusive = mininum_row_exclusive + batch_size
+      sql = <<-SQL
+        SELECT * FROM (
+          SELECT /*+ FIRST_ROWS(n) */ enrollments.*, ROWNUM rnum FROM (
+            SELECT DISTINCT
+              enroll."CLASS_SECTION_ID" as section_id,
+              enroll."CAMPUS_UID" AS ldap_uid,
+              enroll."STUDENT_ID" AS sis_id,
+              TRIM(name."NAME_GIVENNAME" || ' ' || name."NAME_MIDDLENAME") AS first_name,
+              name."NAME_FAMILYNAME" AS last_name,
+              email."EMAIL_EMAILADDRESS" AS email_address
+            FROM SISEDO.ENROLLMENTV00_VW enroll
+            LEFT OUTER JOIN SISEDO.PERSON_EMAILV00_VW email ON (
+              email."PERSON_KEY" = enroll."STUDENT_ID" AND
+              email."EMAIL_TYPE_CODE" = 'CAMP')
+            LEFT OUTER JOIN SISEDO.PERSON_NAMEV00_VW name ON (
+              name."PERSON_KEY" = enroll."STUDENT_ID" AND
+              name."NAME_TYPE_CODE" = 'PRI')
+            WHERE
+              enroll."TERM_ID" = '#{term_id}'
+              AND (enroll."STDNT_ENRL_STATUS_CODE" = 'E')
+            ORDER BY section_id, sis_id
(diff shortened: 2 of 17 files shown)


spec/models/oec/create_confirmation_sheets_task_spec.rb
```diff
@@ -46,7 +46,7 @@
     let(:template) { double(id: 'template_id', title: 'TEMPLATE', mime_type: 'application/vnd.google-apps.spreadsheet') }
     let(:spreadsheet) { double(worksheets: [courses_worksheet, report_viewers_worksheet]) }
     let(:courses_worksheet) { double(title: 'Courses', rows: [Oec::CourseConfirmation.new.headers]) }
-    let(:report_viewers_worksheet) { double(title: 'Report Viewers', rows: [Oec::Supervisors.new.headers]) }
+    let(:report_viewers_worksheet) { double(title: 'Report Viewers', rows: [Oec::SupervisorConfirmation.new.headers]) }
```
paulkerschen: This is an existing mistake in the test that I only noticed with new, stricter validation from Oec::WorksheetRow.

spec/models/oec/merge_confirmation_sheets_task_spec.rb
```diff
@@ -136,7 +136,7 @@ def mock_sheet(filename)
         confirmation.each do |confirmation_row|
           merged_confirmation_row = merged_supervisor_confirmation.find { |row| row['LDAP_UID'] == confirmation_row['LDAP_UID'] }
           supervisors_row = supervisors_worksheet.find { |row| row['LDAP_UID'] == confirmation_row['LDAP_UID'] }
-          merged_course_confirmation.headers.each do |header|
+          merged_supervisor_confirmation.headers.each do |header|
```
paulkerschen: As above.

app/models/edo_oracle/connection.rb
```diff
@@ -5,13 +5,13 @@ def self.settings
...
 
-    def self.safe_query(sql)
+    def self.safe_query(sql, opts={})
       result = []
       return result if fake?
       use_pooled_connection do
         result = connection.select_all sql
       end
-      stringify_ints! result
+      opts[:stringify] == false ? result : stringify_ints!(result)
```
johncrossman: @paulkerschen - So if `opts[:stringify]` is nil then we invoke `stringify_ints`?  If that's intentional then perhaps a more readable scheme would be:
```
opts[:do_not_stringify] ? result : stringify_ints!(result)
```

paulkerschen: Good call @johncrossman, I'll amend.

app/models/edo_oracle/oec.rb
```diff
@@ -80,15 +80,39 @@ def self.get_courses(term_id, filter_clause)
...
-      safe_query <<-SQL
-        #{select_clause}
-        WHERE
-          enroll."TERM_ID" = '#{term_id}'
-          AND #{ccns_clause}
-          AND (enroll."STDNT_ENRL_STATUS_CODE" = 'E')
-        ORDER BY ldap_uid
+    # See http://www.oracle.com/technetwork/issue-archive/2006/06-sep/o56asktom-086197.html for explanation of
+    # query batching with ROWNUM.
+    def self.get_batch_enrollments(term_id, batch_number, batch_size)
```
johncrossman: @paulkerschen would it be safer to have a signature of:
```
 def self.get_batch_enrollments(term_id, batch_number)
```
and encapsulate `Settings.oec.enrollments_batch_size`? This is a minor concern but future devs might not understand the importance of a conservative batch size.

johncrossman: Makes sense, @paulkerschen  Keep as is, I'd say.

app/models/oec/merged_sheet_validation.rb
```diff
@@ -124,23 +126,28 @@ def build_and_validate_export_sheets
         end
       end
 
-      Oec::Queries.students_for_cntl_nums(@term_code, ccns + suffixed_ccns.keys).each do |student_row|
-        validate_and_add(students, Oec::Worksheet.capitalize_keys(student_row), %w(LDAP_UID))
-      end
+      log :warn, "Retrieving enrollment data for the #{@term_code} term. This will take a few minutes...."
```
johncrossman: "The Chicago Manual of Style describes the use of the ellipsis at great length, referring to the “three dot, four dot, and rigorous” methods (11.51 ff)."  Please explain your four-dot style decision, @paulkerschen  ;)

app/models/oec/queries.rb
```diff
@@ -12,26 +12,47 @@ def courses_for_cntl_nums(term_code, ccns)
...
+        enrollments.rows.each do |enrollment_row|
+          string_section_id = enrollment_row[section_id_idx].to_i.to_s
+          if section_ids.include? string_section_id
+            # Since we skipped OracleBase.stringify_ints! in the upstream query, stringify section IDs now.
+            enrollment_row[section_id_idx] = string_section_id
+            rows << enrollment_row
+          end
+        end
+
+        break if enrollments.rows.count < Settings.oec.enrollments_batch_size
```
johncrossman: @paulkerschen I understand the logic here but an inline comment might be good. E.g., "We know we're done when actual count is less than requested count"

johncrossman:
I'm done with review. Looks good, @paulkerschen  I put one request for inline documentation. Once that's in I'll merge.

paulkerschen:
Thanks for review, @johncrossman, amendments are in.

johncrossman:
Once again, OEC breaks the Internet. Thanks, @paulkerschen

PR #5921, opened 2016-09-29
SISRP-24519 - 7.1 Advising Card Design Requests for Haas Graduate Programs

redconfetti:
https://jira.berkeley.edu/browse/SISRP-24519

### DIFF
--- a/src/assets/javascripts/angular/controllers/widgets/myAdvisingController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/myAdvisingController.js
@@ -6,18 +6,47 @@ var _ = require('lodash');
 /**
  * My Advising controller
  */
-angular.module('calcentral.controllers').controller('MyAdvisingController', function(myAdvisingFactory, $scope) {
+angular.module('calcentral.controllers').controller('MyAdvisingController', function(academicsFactory, myAdvisingFactory, $scope) {
   $scope.myAdvising = {
-    isLoading: true
+    feedsLoading: {
+      myAdvising: true,
+      academicsRoles: true
+    },
+    roles: {}
   };
 
-  var loadAdvisingInfo = function() {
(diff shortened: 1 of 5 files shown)


src/assets/javascripts/angular/controllers/widgets/myAdvisingController.js
```diff
@@ -6,18 +6,47 @@ var _ = require('lodash');
...
+          roles.haasEveningWeekendMba ||
+          roles.haasExecMba ||
+          roles.haasMastersFinEng ||
+          roles.haasMbaPublicHealth ||
+          roles.haasMbaJurisDoctor) {
+      return true;
+    }
+    return false;
+  };
+
```
nkutub: do you think that isHassStudent Logic would eventually need to be used elsewhere in the code? not only would this potentially need to be used in the frontend but also potentially this logic could be needed on backend to structure/parese/filter a feed. do you think is should be more centralizes or even moved to the backend in some fashion.

redconfetti: @nkutub I'm not sure where we'll need it. Right now I just am trying to make it present so that we can use it on the front-end. These are "academic roles", as opposed to the "roles" in the /api/my/status feed which are based on the users affiliations.

From what I hear so far, isHaasStudent wouldn't help, because Haas PhD students still get the default list of advising contacts. In [SISRP-20691](https://jira.berkeley.edu/browse/SISRP-20691) only 3 of the 6 Haas 'roles' apply (FTMBA, EWMBA, MFE) as well as an additional one (Undergrad in Business Administration - UGBA). It seems like we should just continue to create these camelCased role flags and let the front-end pick and choose them, group them together, as needed. We'll refactor if needed as we move forward.

raydavis:
Thanks for the review, @nkutub !

PR #5922, opened 2016-09-30
SISRP-25804 - Front-End Development for Student Overview (Advisor) - Student Success - GPA Section w/ HighCharts Graph

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-25804

* `advising_registrations.json` is unrelated, I just noticed we didn't have a dummy JSON for that feed and added one.
* The card currently does not match the "click to expand" design, due to the need for chart background to dynamically change color on mouse hover.  This is possible, but is not something I see to be a blocker for release at the moment.
* PRs for the next sections (Units, Oustanding Balance) to come.
* Lots of good stuff at the HighChartsJS [Documentation](http://api.highcharts.com/highcharts) 
* Make sure to run `npm install` after pulling this PR to your local machine to grab Highcharts from `package.json`!


src/assets/javascripts/angular/controllers/pages/userOverviewController.js
```diff
@@ -162,6 +179,45 @@ angular.module('calcentral.controllers').controller('UserOverviewController', fu
...
+      termGpa.push(term.termCumGpa);
+    });
+    if (termGpa.length < 2) {
+      return;
+    }
+    // The last element of the data series must also contain custom marker information to show the GPA.
+    termGpa[termGpa.length - 1] = {
+      y: termGpa[termGpa.length - 1],
+      dataLabels: {
+        color: termGpa[termGpa.length - 1] >= 2 ? '#2b6281' : '#cf1715',
```
raydavis: It's unusual for us to inline color codes. Is this difficult to implement with CSS?

davidlimcheng: @raydavis Unfortunately, the Highcharts options for markers do not have an option to apply CSS stylings:

```
marker: {
enabled: null
fillColor: undefined
height: null
lineColor: "#ffffff"
lineWidth: 0
radius: 4
states: {
hover: {…}
select: {…}
}
symbol: undefined
width: null
}
```

davidlimcheng: Ah but I see that the options for dataLabels do!  

```
dataLabels: {
align: "center"
allowOverlap: false
backgroundColor: undefined
borderColor: undefined
borderRadius: 0
borderWidth: 0
className: undefined
```

This means that I can hopefully change this to CSS, and I see that the line option has a className option as well.  Let me give it a shot, and amend if it works out.

src/assets/javascripts/angular/controllers/pages/userOverviewController.js
```diff
@@ -32,6 +32,23 @@ angular.module('calcentral.controllers').controller('UserOverviewController', fu
...
+  };
+  $scope.studentSuccess = {
+    gpaChart: {
+      xAxis: {
+        floor: 0,
+        visible: false
+      },
+      yAxis: {
+        min: 0,
+        max: 4.0,
```
lyttam: Is 4.0 an absolute max? Or is there a case where it might change?

src/assets/javascripts/angular/controllers/pages/userOverviewController.js
davidlimcheng: OK, after researching and trying out the options for a bit, I was able to apply a CSS class at the chart-container level, which applies to the line color - so I was able to get rid of the inline color in the directive.  For the data label and markers though, this must be done through [styled mode](http://www.highcharts.com/docs/chart-design-and-style/style-by-css), and unfortunately even then, the marker cannot access any CSS classes.

src/assets/javascripts/angular/controllers/pages/userOverviewController.js
davidlimcheng: Great question - judging from the [GPA Calculator](https://github.com/ets-berkeley-edu/calcentral/blob/master/src/assets/javascripts/angular/controllers/widgets/academicsGpaController.js) it looks like 4.0 is the absolute max, but I'll validate with the functional teams as well.

src/assets/javascripts/angular/factories/advisingFactory.js
```diff
@@ -15,6 +15,9 @@ angular.module('calcentral.factories').factory('advisingFactory', function(apiSe
   var urlAdvisingResources = '/api/advising/resources/';
   // var urlAdvisingResources = '/dummy/json/advising_resources.json';
   var urlAdvisingRegistrations = '/api/advising/registrations/';
+  // var urlAdvisingRegistrations = 'dummuy/json/advising_registrations.json';
```
lyttam: Just a lil typo: dummuy

src/assets/javascripts/angular/controllers/pages/userOverviewController.js
raydavis: Even "A+" grades are mapped to 4.0 at UC Berkeley, so I'm pretty sure that's a safe max for the graph.

src/assets/javascripts/angular/factories/advisingFactory.js
davidlimcheng: =O nice catch, amended.

raydavis:
Thanks, @davidlimcheng , I look forward to seeing this one in action. I'll merge as soon as Travis finishes ruminating.

src/assets/javascripts/angular/controllers/pages/userOverviewController.js
```diff
@@ -162,6 +179,45 @@ angular.module('calcentral.controllers').controller('UserOverviewController', fu
...
+        }
+      },
+      marker: {
+        enabled: true,
+        fillColor: termGpa[termGpa.length - 1] >= 2 ? '#2b6281' : '#cf1715',
+        radius: 3,
+        symbol: 'circle'
+      }
+    };
+    $scope.highCharts.dataSeries.push(termGpa);
```
lyttam: You might want to consider making this name more generic ($scope.highCharts) in case we decide to swap out highcharts for a different charting library in the future.

PR #5923, opened 2016-09-30
SISRP-25001 - Test Fixtures Still Rely on Bearfacts / Legacy Data

lyttam:
https://jira.berkeley.edu/browse/SISRP-25001

- Breaks unit test dependency on BearFacts; replaces detailed low-level unit tests with higher-level tests for delegate and advisor access
- Fixes MyAcademicsController.residency so it always returns JSON

### DIFF
--- a/spec/models/my_academics/filtered_for_delegate_spec.rb
+++ b/spec/models/my_academics/filtered_for_delegate_spec.rb
@@ -1,77 +1,111 @@
 describe MyAcademics::FilteredForDelegate do
-  let(:provider_classes) do
-    [
-      MyAcademics::CollegeAndLevel,
-      MyAcademics::TransitionTerm,
-      MyAcademics::GpaUnits,
-      MyAcademics::Semesters,
-      MyAcademics::Exams
-    ]
-  end
 
-  let(:uid) { '61889' }
-
-  before do
-    allow(Settings.features).to receive(:cs_delegated_access).and_return(is_feature_enabled)
-    fake_classes = Bearfacts::Proxy.subclasses + [ Regstatus::Proxy ]
-    fake_classes.each do |klass|
-      allow(klass).to receive(:new).and_return klass.new(user_id: uid, fake: true)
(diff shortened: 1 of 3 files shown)


lyttam:
I didn't find any BearFacts fixtures that looked like they should be removed, but if someone could check my work I'd appreciate it.

spec/models/my_academics/filtered_for_delegate_spec.rb
```diff
@@ -1,79 +1,61 @@
...
+        allow(subject).to receive(:is_feature_enabled).and_return(false)
+      end
+
+      it 'returns an empty feed' do
+        expect(subject.get_feed_internal).to eq({})
       end
     end
-  end
 
-  context 'when delegate permissions do not include grades', if: CampusOracle::Connection.test_data? do
```
raydavis: Unlike the deletions in the controller RSpec, this deletion would lose a regression check -- there's no place _other_ than FilteredForDelegate to ensure that the grade suppression is still effective. So I'm afraid this part of the PR will need to be dropped or amended -- sorry the JIRA was unclear!

raydavis:
Thanks for the controller fix & the much tidier controller RSpec, @lyttam -- both of those portions look ready to go.

spec/models/my_academics/filtered_for_delegate_spec.rb
lyttam: Thanks, I'll take some time to re-implement that test sans BearFacts.  FYI @johncrossman this PR may not make the cut, but I already discussed with @raydavis and it's ok.

johncrossman: Thanks letting me know, @lyttam

lyttam: I added tests for delegates with and without access to see grades.  PR updated, check it out!  @raydavis @johncrossman

johncrossman:
Thanks @lyttam  Test coverage looks good.

PR #5927, opened 2016-10-01
SISRP-25993 - Final Exam Schedule Incorporate Course Exceptions

tiffanyqi:
https://jira.berkeley.edu/browse/SISRP-25993
- Incorporate classes that have an alternate method of final exams or no exam at all.

Note: I did a hacky thing, in order to send a class that has no exam and an alternate method to the bottom, I set its Time field to something like the year 999999. If this isn't a good way to go, I'm open to other suggestions.

I apologize for this being late, not sure whether this should be cherry-picked into QA, but this should be the final piece to the final exam schedule feature.

### DIFF
--- a/app/models/edo_oracle/queries.rb
+++ b/app/models/edo_oracle/queries.rb
@@ -171,25 +171,28 @@ module EdoOracle
     def self.get_section_final_exam(term_id, section_id)
       safe_query <<-SQL
         SELECT DISTINCT
-          exam."term-id" AS term_id,
-          exam."session-id" AS session_id,
+          sec."term-id" AS term_id,
+          sec."session-id" AS session_id,
+          sec."id" AS section_id,
+          sec."finalExam" AS exam_type,
           exam."date" AS exam_date,
           exam."startTime" AS exam_start_time,
           exam."endTime" AS exam_end_time,
           exam."location-descr" AS location
         FROM
           SISEDO.EXAMV00_VW exam
-        JOIN SISEDO.CLASSSECTIONV00_VW sec ON (
+        RIGHT JOIN SISEDO.CLASSSECTIONV00_VW sec ON (
           exam."cs-course-id" = sec."cs-course-id" AND
           exam."term-id" = sec."term-id" AND
           exam."session-id" = sec."session-id" AND
           exam."offeringNumber" = sec."offeringNumber" AND
-          exam."sectionNumber" = sec."sectionNumber"
+          exam."sectionNumber" = sec."sectionNumber" AND
+          exam."type-code" = 'FIN'
         )
         WHERE
           sec."term-id" = '#{term_id}' AND
           sec."id" = '#{section_id}' AND
-          exam."type-code" = 'FIN'
+          sec."finalExam" IN ('A', 'Y')
         ORDER BY exam_date
       SQL
     end
--- a/app/models/my_academics/exams.rb
+++ b/app/models/my_academics/exams.rb
@@ -47,7 +47,7 @@ module MyAcademics
           }
           if cs_data_available && section[:final_exams].any?
             exam = section[:final_exams].first
-            course[:exam_location] = exam[:location] || 'Location TBD'
+            course[:exam_location] = choose_cs_exam_location(exam)
             course[:exam_date] = parse_cs_exam_date(exam)
             course[:exam_time] = parse_cs_exam_time(exam)
             course[:exam_slot] = parse_cs_exam_slot(exam)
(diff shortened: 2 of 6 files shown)


app/models/edo_oracle/queries.rb
```diff
@@ -171,25 +171,28 @@ def self.get_section_meetings(term_id, section_id)
...
-          exam."session-id" AS session_id,
+          sec."term-id" AS term_id,
+          sec."session-id" AS session_id,
+          sec."id" AS section_id,
+          sec."finalExam" AS exam_type
           exam."date" AS exam_date,
           exam."startTime" AS exam_start_time,
           exam."endTime" AS exam_end_time,
-          exam."location-descr" AS location
+          exam."location-descr" AS location,
```
johncrossman: @tiffanyqi I would expect this trailing comma to be a problem. I.e., bad SQL.

app/models/edo_oracle/queries.rb
```diff
@@ -171,25 +171,28 @@ def self.get_section_meetings(term_id, section_id)
...
           exam."offeringNumber" = sec."offeringNumber" AND
-          exam."sectionNumber" = sec."sectionNumber"
+          exam."sectionNumber" = sec."sectionNumber" AND
+          exam."type-code" = 'FIN'
         )
         WHERE
           sec."term-id" = '#{term_id}' AND
           sec."id" = '#{section_id}' AND
-          exam."type-code" = 'FIN'
+          (sec."finalExam" = 'A' OR  sec."finalExam" = 'Y')
```
johncrossman: @tiffanyqi please remove the extra space after the `OR`

johncrossman: @tiffanyqi also, this could be simplified to be `sec."finalExam" IN ('A', 'Y')`

app/models/edo_oracle/queries.rb
tiffanyqi: Amended. Thank you for the review despite the weekend!

app/models/edo_oracle/queries.rb
tiffanyqi: Updated!

app/models/edo_oracle/queries.rb
```diff
@@ -171,25 +171,28 @@ def self.get_section_meetings(term_id, section_id)
     def self.get_section_final_exam(term_id, section_id)
       safe_query <<-SQL
         SELECT DISTINCT
-          exam."term-id" AS term_id,
-          exam."session-id" AS session_id,
+          sec."term-id" AS term_id,
+          sec."session-id" AS session_id,
+          sec."id" AS section_id,
+          sec."finalExam" AS exam_type
```
johncrossman: @tiffanyqi ah, here is the missing comma  :)

tiffanyqi: Oh no! Amended again, thank you!

app/models/edo_oracle/queries.rb
```diff
@@ -171,25 +171,28 @@ def self.get_section_meetings(term_id, section_id)
...
           SISEDO.EXAMV00_VW exam
-        JOIN SISEDO.CLASSSECTIONV00_VW sec ON (
+        RIGHT JOIN SISEDO.CLASSSECTIONV00_VW sec ON (
           exam."cs-course-id" = sec."cs-course-id" AND
           exam."term-id" = sec."term-id" AND
           exam."session-id" = sec."session-id" AND
           exam."offeringNumber" = sec."offeringNumber" AND
-          exam."sectionNumber" = sec."sectionNumber"
+          exam."sectionNumber" = sec."sectionNumber" AND
+          exam."type-code" = 'FIN'
```
johncrossman: @tiffanyqi Just curious...  I would expect `exam."type-code" = 'FIN'` to live in the WHERE clause. JOIN clauses are ordinarily limited to tableA vs tableB comparisons. Does the RIGHT JOIN add special consideration?

app/models/my_academics/exams.rb
```diff
@@ -125,6 +125,20 @@ def parse_cs_exam_slot(exam)
         return Time.parse(date.strftime('%y-%m-%d') + ' ' + time.strftime('%H:%M'))
       elsif date
         return Time.parse(date.strftime('%y-%m-%d'))
+      elsif exam[:exam_type] == 'A' # alternate exams that aren't included
+        return Time.new(99999998)
+      else # no exams should be at the end
+        return Time.new(99999999)
```
johncrossman: @tiffanyqi Can you give me a little more info on how this causes _alternate exam_ and _no exam_ courses to "be at the end" on the UI? As far as I can tell, neither `final_exam_schedule.html` nor server-side logic has sorting logic per `exam_slot` value.

app/models/edo_oracle/queries.rb
tiffanyqi: Good question! We had that there at first, however, it is possible that there is no information from the SISEDO.EXAMV00_VW, thus breaking the whole query. We moved it in the RIGHT JOIN so that it filters there and still doesn't break. @nkutub can give you more info about this if I've missed anything.

app/models/my_academics/exams.rb
tiffanyqi: There's a `sort_exams` method in line 82. Essentially it takes all the exam_slots and sorts them in order by `Time` if CS data is available (and if it gets to this method, it implies it has), or by `Integer` if it's not available. Does this help?

app/models/edo_oracle/queries.rb
tiffanyqi: Good question! Initially we had that line in WHERE, but since it is possible that there is no information from the SISEDO.EXAM view, we moved it to the RIGHT JOIN so as to not break. @nkutub helped me with this query, and perhaps can give more info about this decision.

johncrossman:
Thanks for the answers, @tiffanyqi  Merging. Cherry-picking to QA depends on priority.

PR #5928, opened 2016-10-03
SISRP-25993-QA - Final Exam Schedule Incorporate Course Exceptions

tiffanyqi:
JIRA: https://jira.berkeley.edu/browse/SISRP-25993
Master PR: https://github.com/ets-berkeley-edu/calcentral/pull/5927

Incorporate classes that have an alternate method of final exams or no exam at all.

### DIFF
--- a/app/models/edo_oracle/course_sections.rb
+++ b/app/models/edo_oracle/course_sections.rb
@@ -38,9 +38,12 @@ module EdoOracle
     end
 
     def get_section_final_exam
+      # EdoOracle::Queries.get_section_final_exam accepts a section_id, which is a unique id
+      # that defines a section per course. Thus, @course_id, despite its naming convention,
+      # still works here.
       final_exams = EdoOracle::Queries.get_section_final_exam(@term_id, @course_id).map do |exam|
         {
-          type: exam['type'],
+          exam_type: exam['exam_type'],
           location: exam['location'],
           exam_date: exam['exam_date'],
           exam_start_time: exam['exam_start_time'],
--- a/app/models/my_academics/exams.rb
+++ b/app/models/my_academics/exams.rb
@@ -47,7 +47,7 @@ module MyAcademics
           }
           if cs_data_available && section[:final_exams].any?
             exam = section[:final_exams].first
-            course[:exam_location] = exam[:location] || 'Location TBD'
+            course[:exam_location] = choose_cs_exam_location(exam)
             course[:exam_date] = parse_cs_exam_date(exam)
             course[:exam_time] = parse_cs_exam_time(exam)
             course[:exam_slot] = parse_cs_exam_slot(exam)
@@ -125,6 +125,20 @@ module MyAcademics
         return Time.parse(date.strftime('%y-%m-%d') + ' ' + time.strftime('%H:%M'))
       elsif date
         return Time.parse(date.strftime('%y-%m-%d'))
+      elsif exam[:exam_type] == 'A' # alternate exams that aren't included
+        return Time.new(99999998)
+      else # no exams should be at the end
+        return Time.new(99999999)
+      end
+    end
+
+    def choose_cs_exam_location(exam)
+      if exam[:location]
+        return exam[:location]
+      elsif exam[:exam_type] == 'A'
+        return 'Final exam time not yet provided.'
+      else
+        return 'Location TBD'
       end
     end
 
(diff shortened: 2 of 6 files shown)


johncrossman:
@tiffanyqi a couple requests:
1. add a link to the master PR (merged earlier) in this PR
1. update release instructions w.r.t `final_exam_schedule` flag

tiffanyqi:
@johncrossman all done. Please let me know if you need anything else!

johncrossman:
Thanks @tiffanyqi  We also need `final_exam_schedule` set to true in all environments.  (Fyi, Darlene will make the change in prod)

tiffanyqi:
@johncrossman do you mean going into each environment and setting those to true? And once I've done this, to write that in the environments under specifying the `final_exam_schedule` is true?

Also note: I do not have SSH access, so I heard I will need some assistance with the dev.local.yml and those environments (I think?)

johncrossman:
@tiffanyqi I don't mind making the change for `final_exam_schedule`.  The unwritten rule says, keep all YAML files in sync with latest code changes. More importantly, the YAML in qa must match the expected configs of the next prod release. That's why I want to get feature flags in order.

tiffanyqi:
@johncrossman totally understand. Would this be a new QA&master PR with all the YAML changes?

johncrossman:
@tiffanyqi  No. I'm talking about YAML changes on the network, not what is in our Git repo.  On any given server, we want the feature flags in `~/.calcentral_config/production.local.yml` to override the hard-coded stuff in Git. As you said, a developer needs proper SSH access & sudo rights in order to modify  `~/.calcentral_config/production.local.yml`

tiffanyqi:
@johncrossman Thank you for clarifying. Via our conversation, thank you for getting the feature flags in order since I do not have credentials!

app/models/edo_oracle/course_sections.rb
```diff
@@ -40,7 +40,7 @@ def get_section_schedules
     def get_section_final_exam
       final_exams = EdoOracle::Queries.get_section_final_exam(@term_id, @course_id).map do |exam|
```
nkutub: There is still an inconsistency in the naming of course_id vs sections_id in the get_section_final_exam method. If there is an inconsistency in the variable names, might be good to explain that in comments someplace.

app/models/my_academics/exams.rb
```diff
@@ -125,6 +125,20 @@ def parse_cs_exam_slot(exam)
         return Time.parse(date.strftime('%y-%m-%d') + ' ' + time.strftime('%H:%M'))
       elsif date
         return Time.parse(date.strftime('%y-%m-%d'))
+      elsif exam[:exam_type] == 'A' # alternate exams that aren't included
+        return Time.new(99999998)
```
nkutub: this elsif is redundant since the final fallthrough else at the bottom returns the same thing. It could be removed and still yield the same result.

app/models/edo_oracle/course_sections.rb
nkutub: There is still inconsistency in the parameter being passed of @course_id vs the actual variable name in get_section_final_exam which is section_id. Maybe a comment would be good to explain the naming difference.

app/models/my_academics/exams.rb
tiffanyqi: I added this because I wanted to group alternates with each other, and no exams with each other. Side note, all `exam[:exam_type]` is returning null....so it definitely won't go through this code. Do you want to take a look at this together sometime?

johncrossman: @nkutub the trailing digit is the (very subtle) diff here: 9999999**8** v. 9999999**9**.

nkutub: missed that, thanks @johncrossman . My bad @tiffanyqi

johncrossman:
@tiffanyqi when you have a moment, please reply to @nkutub's comments.

app/models/edo_oracle/course_sections.rb
tiffanyqi: Added in this one (but not in `queries.rb`). How is the message?

nkutub: totally makes sense, good job keeping it simple.

tiffanyqi: Cool, thanks! Do you want a similar message in queries.rb, or do you think this will be sufficient?

johncrossman:
Thanks @tiffanyqi & @nkutub

PR #5929, opened 2016-10-03
SISRP-20510 - Hold icon on Enrollment card rendering poorly on iOS

nkutub:
https://jira.berkeley.edu/browse/SISRP-20510

removed inconsistency in displaying notification icons. using <span> tags instead of <div> tags for enrollment card holds icon is consistent with other card using similar icons. This also removed the required cc-flex style which created an inconsistent look/spacing on safari based browsers.

### DIFF
--- a/src/assets/templates/widgets/enrollment/consult_advisor.html
+++ b/src/assets/templates/widgets/enrollment/consult_advisor.html
@@ -1,12 +1,12 @@
 <div class="cc-enrollment-card-advisors-consult">
 
-  <div data-ng-if="enrollmentInstruction.hasHolds" class="cc-flex">
-    <div>
+  <div data-ng-if="enrollmentInstruction.hasHolds">
+    <span>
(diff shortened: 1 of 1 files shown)


src/assets/templates/widgets/enrollment/consult_advisor.html
```diff
@@ -1,12 +1,12 @@
 <div class="cc-enrollment-card-advisors-consult">
 
-  <div data-ng-if="enrollmentInstruction.hasHolds" class="cc-flex">
-    <div>
+  <div data-ng-if="enrollmentInstruction.hasHolds" >
```
johncrossman: @nkutub if you don't mind, remove stray space in `hasHolds" >`

nkutub: done!

johncrossman:
Thanks @nkutub

PR #5930, opened 2016-10-03
SISRP-26305 - MyAcademics::Exams for UID 1064465: NoMethodError: undefined method gsub for nil:NilClass

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-26305

* `course` variable name was overwriting the original `|course|` input parameter, this became a problem for students taking two of the same classes, as the 2nd `course` would try to parse the first, already-parsed `course`.


johncrossman:
Thanks, @davidlimcheng

PR #5934, opened 2016-10-04
SISRP-26282 - Class Page - Section Overview - No open position count when enrollments not present

redconfetti:
https://jira.berkeley.edu/browse/SISRP-26282

See QA PR #5935

### DIFF
--- a/spec/models/rosters/campus_spec.rb
+++ b/spec/models/rosters/campus_spec.rb
@@ -210,171 +210,189 @@ describe Rosters::Campus do
       let(:term_id) { Berkeley::TermCodes.to_edo_id(term_yr, term_cd) }
       let(:fake_oracle) { {} }
       let(:fake_edo) { fake_campus }
-      before do
-        expect(EdoOracle::Queries).to receive(:get_rosters).with([ccn1, ccn2], term_id).and_return enrollments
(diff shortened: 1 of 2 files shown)


spec/models/rosters/campus_spec.rb
```diff
@@ -210,172 +210,190 @@
       let(:term_id) { Berkeley::TermCodes.to_edo_id(term_yr, term_cd) }
       let(:fake_oracle) { {} }
       let(:fake_edo) { fake_campus }
-      before do
-        expect(EdoOracle::Queries).to receive(:get_rosters).with([ccn1, ccn2], term_id).and_return enrollments
-        # A method rather than a let so that modifications during a test don't persist.
-        def attributes
+      let(:feed) { feed = Rosters::Campus.new(user_id, course_id: campus_course_id).get_feed }
+
+      context 'when enrollments are present' do
```
redconfetti: I had to make this new sub-context with the `before` block that sets up the expectation for enrollments to be present, and thus for `User::BasicAttributes` to receive the call to `#attributes_for_uids`.

I then added a new context where this doesn't occur, because there are no enrollments present, on line 380

davidlimcheng:
LGTM

PR #5937, opened 2016-10-04
SISRP-26327, only show Teaching Card when user has teaching assignments

johncrossman:
https://jira.berkeley.edu/browse/SISRP-26327

If you are faculty with no teaching assignments then you get no Teaching Card.


johncrossman:
I'll submit a QA PR once this is merged.

paulkerschen:
A teacher not teaching is no teacher at all. Thanks @johncrossman.

PR #5938, opened 2016-10-04
CLC-6453, instructor role-code 'PROXY' assists instructor(s) with grading (e.g., Head GSI)

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-6453

cc: @paulkerschen 

The new 'PROXY' instructor role-code is coming soon. At the very least, we should not `raise ArgumentError` when the role-code lands. Future PRs will address specific business logic concerns. E.g., OEC.

Fyi, (personal) Bamboo build in-progress: https://bamboo.ets.berkeley.edu/bamboo/browse/CPB-CJT85-2

### DIFF
--- a/app/models/edo_oracle/adapters/common.rb
+++ b/app/models/edo_oracle/adapters/common.rb
@@ -8,18 +8,22 @@ module EdoOracle
 
       def adapt_dept_name_and_catalog_id(row, user_courses)
         dept_name, catalog_id = user_courses.parse_course_code row
-        row.merge!({'dept_name' => dept_name, 'catalog_id' => catalog_id})
+        row.merge!({
+          'dept_name' => dept_name,
+          'catalog_id' => catalog_id
+        })
       end
 
       def adapt_instructor_func(row)
         return unless (code = row.delete 'role_code')
         row['instructor_func'] = case code
-          when 'PI' then '1'    # Teaching and In Charge, equivalent of Teaching and Instructor of Record (1)
-          when 'TNIC' then '2'  # Teaching but Not in Charge, equivalent of Teaching but not Instructor of Record (2)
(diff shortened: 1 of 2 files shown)


app/models/edo_oracle/adapters/common.rb
```diff
@@ -8,18 +8,22 @@ def adapt_course_cntl_num(row)
 
       def adapt_dept_name_and_catalog_id(row, user_courses)
         dept_name, catalog_id = user_courses.parse_course_code row
-        row.merge!({'dept_name' => dept_name, 'catalog_id' => catalog_id})
+        row.merge!({
+                     'dept_name' => dept_name,
+                     'catalog_id' => catalog_id
+                   })
```
paulkerschen: @johncrossman Not a merge blocker, but I think our usual practice for multiline hashes is to indent only two more spaces, and have the `})` back at the same indent level as `row.merge!({`.

app/models/edo_oracle/adapters/common.rb
```diff
@@ -35,15 +39,18 @@ def adapt_primary_secondary_cd(row)
...
       end
 
       def term_id(term_year, term_code=nil)
         unless term_code
-          term_year, term_code = term_year.split('-')
+          term_year, term_code = term_year.split '-'
         end
-        Berkeley::TermCodes.to_edo_id(term_year, term_code)
+        Berkeley::TermCodes.to_edo_id term_year, term_code
```
paulkerschen: @johncrossman Also not a merge blocker, but unless an idiom suggests otherwise, we tend to include parentheses around multiple arguments.

johncrossman: Thanks, @paulkerschen  I'll amend both of the above.

PR #5942, opened 2016-10-04
CLC-6448 CLC-6449 OEC: special handling for Law and Spanish/Portuguese depts

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6448
https://jira.ets.berkeley.edu/jira/browse/CLC-6449

Note that because of this special handling, we no longer get to set evaluation type ahead of time in EdoOracle::Adapters::Oec.

### DIFF
--- a/app/models/oec/sis_import_task.rb
+++ b/app/models/oec/sis_import_task.rb
@@ -67,9 +67,8 @@ module Oec
           skip_course course, 'course with non-evaluated instruction format'
         else
           course['course_id_2'] = course['course_id']
-          set_dept_form course
-          # Course rows sourced from EdoOracle::Adapters::Oec may have evaluation type already set.
-          set_evaluation_type course unless course['evaluation_type']
+          set_dept_form(worksheet, course)
+          set_evaluation_type course
           worksheet[row_key] = Oec::Worksheet.capitalize_keys course
         end
       end
@@ -126,7 +125,7 @@ module Oec
       end
     end
 
-    def set_dept_form(course)
+    def set_dept_form(worksheet, course)
       return if course['cross_listed_flag'].present?
 
       # Sets 'dept_form' to either 'MCELLBI' or 'INTEGBI' for BIOLOGY courses; otherwise uses 'dept_name'.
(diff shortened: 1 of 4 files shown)


redconfetti:
Looks good to me

app/models/oec/sis_import_task.rb
```diff
@@ -67,9 +67,9 @@ def import_course(worksheet, course)
           skip_course course, 'course with non-evaluated instruction format'
         else
           course['course_id_2'] = course['course_id']
-          set_dept_form course
-          # Course rows sourced from EdoOracle::Adapters::Oec may have evaluation type already set.
-          set_evaluation_type course unless course['evaluation_type']
+          set_dept_form(worksheet, course)
+          # LAW and SPANISH are special cases that don't use F/G evaluation types.
+          set_evaluation_type course unless %w(LAW SPANISH).include? course['dept_form']
```
johncrossman: @paulkerschen would it make sense to move this LAW/SPANISH logic into `set_evaluation_type`? I.e.,
```
return if %w(LAW SPANISH).include? course['dept_form']
```
Or do you like the readability of having it here?

app/models/oec/sis_import_task.rb
```diff
@@ -136,6 +136,10 @@ def set_dept_form(course)
       if (mapping = Oec::CourseCode.catalog_id_specific_mapping(course['dept_name'], course['catalog_id']))
         @dept_forms[mapping] ||= Oec::CourseCode.where(dept_code: mapping.dept_code, catalog_id: '').pluck(:dept_name).first
         course['dept_form'] = @dept_forms[mapping]
+      # The Spanish and Portuguese department is another special case where all courses use the SPANISH department form,
+      # regardless of subject area.
+      elsif worksheet.dept_code == 'LPSPP'
+        course['dept_form'] = 'SPANISH'
```
johncrossman: @paulkerschen I assume this is good as is but...  Do we need `if worksheet.dept_code == 'LPSPP'` to be the first conditional in this `if ... elsif ... else` block?

johncrossman:
LGTM, too.  I posted two questions.

app/models/oec/sis_import_task.rb
paulkerschen: @johncrossman You'd be forgiven for thinking so, but the first conditional under `if` has to do with `BIOLOGY` courses, as the comment muddledly attempts to express.

app/models/oec/sis_import_task.rb
paulkerschen: @johncrossman I am convinced! Amending.

PR #5945, opened 2016-10-04
CLC-6451 OEC: Separately flag cross-listed and room-shared _GSI rows

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6451

Avoid bogus errors when faculty and GSIs turn out to be different people.


johncrossman:
GSIs get some love. Thanks, @paulkerschen

PR #5947, opened 2016-10-04
SISRP-26403 - Remove showing of exam conflicts if a student has more …

tiffanyqi:
…than one no exam
https://jira.berkeley.edu/browse/SISRP-26403
QA PR after this is merged.

- Adds "no exam" if a class' section doesn't have any final exams
- Removed the Time.new(9999999) vs. Time.new(999998) thing and just consolidated it to Time.new(9999)
- Show "location TBD" if a class has an exam, but not sure where the exam is
- Change messaging around whether a class is alternate
- Remove conflict messaging / visuals if there are multiple "no exams" or alternate final exams (this is hacky, if it's not good, any other suggestions?"

### DIFF
--- a/app/models/my_academics/exams.rb
+++ b/app/models/my_academics/exams.rb
@@ -45,12 +45,17 @@ module MyAcademics
             time: section[:schedules][:recurring].to_a.first.try(:[], :schedule),
             waitlisted: section[:waitlisted]
           }
-          if cs_data_available && section[:final_exams].any?
-            exam = section[:final_exams].first
-            parsed_course[:exam_location] = choose_cs_exam_location(exam)
-            parsed_course[:exam_date] = parse_cs_exam_date(exam)
-            parsed_course[:exam_time] = parse_cs_exam_time(exam)
-            parsed_course[:exam_slot] = parse_cs_exam_slot(exam)
+          if cs_data_available
+            if section[:final_exams].any?
+              exam = section[:final_exams].first
+              parsed_course[:exam_location] = choose_cs_exam_location(exam)
+              parsed_course[:exam_date] = parse_cs_exam_date(exam)
+              parsed_course[:exam_time] = parse_cs_exam_time(exam)
+              parsed_course[:exam_slot] = parse_cs_exam_slot(exam)
+            else
+              parsed_course[:exam_slot] = 'none'
+              parsed_course[:exam_location] = 'No exam.'
+            end
           end
           courses << parsed_course
         end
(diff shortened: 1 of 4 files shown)


src/assets/templates/widgets/final_exam_schedule.html
```diff
@@ -10,8 +10,8 @@ <h3 data-ng-bind="semester.name"></h3>
       <div class="cc-widget-exam-schedule-text" data-ng-if="!semester.cs_data_available && semester.exams">Dates and times are subject to change. Please check your syllabus for the official exam day and time for each class.</div>
       <div class="cc-widget-exam-schedule-text" data-ng-if="!semester.exams && isUndergraduate">No exam schedules listed.</div>
       <div data-ng-repeat="(exam_slot, exam) in semester.exams">
-        <div ng-class="{'cc-widget-exam-schedule-conflicts': exam.length > 1, 'cc-widget-exam-schedule-time-conflicts-left': exam.length > 1}">
-          <div class="cc-widget-exam-schedule-conflicts-text" data-ng-if="exam.length > 1">
+        <div ng-class="{'cc-widget-exam-schedule-conflicts': exam.length > 1 && exam_slot != '9999-01-01T00:00:00-08:00', 'cc-widget-exam-schedule-time-conflicts-left': exam.length > 1 && exam_slot != '9999-01-01T00:00:00-08:00'}">
+          <div class="cc-widget-exam-schedule-conflicts-text" data-ng-if="exam.length > 1 && exam_slot != '9999-01-01T00:00:00-08:00'">
```
davidlimcheng: @tiffanyqi I think I understand what we're going for here, but it's a bit hard to grasp as it is.  Can we translate the `exam_slot != '9999...'` logic in the controller, with a comment as to what's going on, and surface it to the HTML in a more readable manner?

app/models/my_academics/exams.rb
```diff
@@ -45,12 +45,17 @@ def parse_courses(semester, courses, cs_data_available)
...
-            parsed_course[:exam_slot] = parse_cs_exam_slot(exam)
+          if cs_data_available
+            if section[:final_exams].any?
+              exam = section[:final_exams].first
+              parsed_course[:exam_location] = choose_cs_exam_location(exam)
+              parsed_course[:exam_date] = parse_cs_exam_date(exam)
+              parsed_course[:exam_time] = parse_cs_exam_time(exam)
+              parsed_course[:exam_slot] = parse_cs_exam_slot(exam)
+            else
+              parsed_course[:exam_slot] = Time.new(9999)
```
redconfetti: Why this time instead of just using `nil`?

src/assets/templates/widgets/final_exam_schedule.html
davidlimcheng: Would it make sense to set `exam_slot` to `'none'` (or something like that), and iterate through `courses` with [Array#rotate](http://ruby-doc.org/core-2.0.0/Array.html#method-i-rotate) to get all of the `exam_slot == 'none'` ones to the bottom?

redconfetti: I'm pretty sure that JSON supports a null value. I just tested this at [jsonlint.com](http://www.jsonlint.com/):
```
{
	"something": null,
	"somethingElse": 1
}
```
It said that it's valid.

I see what you're saying about sorting classes that have no exam. Can you sort extract them from the list, sort the list, then append them to the bottom?

nkutub: I second @redconfetti's approach

paulkerschen: @redconfetti's approach is a good one. The alternative would be to not use `sort_by,` but use `sort` with a custom block that treats a missing `:exam_slot` as greater than anything else:
http://blog.crowdint.com/2011/01/05/custom-sorting-with-ruby.html

tiffanyqi: Okay I'll try that. Thanks for the comments everyone!
My other question, would you want all instances of `exam_slot != '9999-01-01T00:00:00-08:00'` to be changed to `exam_slot != 'none'` instead?

paulkerschen: @tiffanyqi Seems it would be cleanest not to include an `exam_slot` key in the hash at all. Do you need it for any reason other than sorting?

paulkerschen: @tiffanyqi I see! If you need the alert for multiple exams without an exam slot, then a string value of `none` with `exam_slot !== 'none'` in the front end makes sense to me.

tiffanyqi: @paulkerschen okay i'll try that :D, thanks!

nkutub: check it out https://repl.it/Do1U/0

paulkerschen: Thanks for the repl, @nkutub! See also https://repl.it/Do1U/6 for an implementation that avoids adding in bogus dates at all, which perhaps avoids some confusion.

tiffanyqi: @nkutub thank you! I amended it, and it definitely looks much better than before. Thank you for your input everyone! Please review to see if this looks better.

app/models/my_academics/exams.rb
tiffanyqi: @redconfetti Changed to `none`, please see if this works for you now. Thanks!

PR #5946, opened 2016-10-04
NOJIRA, CalNetLdap search returns berkeleyeduaffiliations as Enumerable

johncrossman:
This is a reaction to Bamboo build failure: https://bamboo.ets.berkeley.edu/bamboo/browse/MYB-CLCMVPMASTER-JOB1-4600/test/case/60932119

This PR only adjust expectation in the spec. CalCentral code is not touched – it references the `:berkeleyeduaffiliations` attribute in only `Berkeley::UserRoles`:
```
affiliations = ldap_record[:berkeleyeduaffiliations].to_a
```
*Bottom line:* It's not a problem if CalNetLdap returns an `Enumerable` instead of `Net::BER::BerIdentifiedArray`.  Our code will not know the difference.

PR #5948, opened 2016-10-04
NOJIRA, CalNetLdap search returns berkeleyeduaffiliations as Enumerable

johncrossman:
Master PR is #5946 

The [Bamboo breakage we saw on master](https://bamboo.ets.berkeley.edu/bamboo/browse/MYB-CLCMVPMASTER-4601) is likely to impact QA branch, too.

### DIFF
--- a/spec/models/calnet_ldap/client_spec.rb
+++ b/spec/models/calnet_ldap/client_spec.rb
@@ -2,7 +2,7 @@ describe CalnetLdap::Client do
 
   it 'should initialize with a configured Net::LDAP object' do
     ldap = subject.instance_variable_get :@ldap
-    expect(ldap).to be_an_instance_of Net::LDAP
+    expect(ldap).to be_a Net::LDAP
     expect(ldap.host).to eq 'nds-test.berkeley.edu'
     expect(ldap.port).to eq 636
     if ENV['RAILS_ENV'] == 'test'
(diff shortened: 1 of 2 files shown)


spec/models/calnet_ldap/client_spec.rb
```diff
@@ -2,7 +2,7 @@
 
   it 'should initialize with a configured Net::LDAP object' do
     ldap = subject.instance_variable_get :@ldap
-    expect(ldap).to be_an_instance_of Net::LDAP
+    expect(ldap).to be_a Net::LDAP
```
redconfetti: What is `be_a` used for? The class constant itself rather than an instance of the class?

redconfetti: Nevermind, I'm looking up. [type check matchers](https://www.relishapp.com/rspec/rspec-expectations/v/2-0/docs/matchers/type-check-matchers)

I see that the kind_of check used by `be_a` sees if the object includes functionality inherited or mixed in from Net::LDAP, whereas `be_an_instance_of` performs an `#instance_of?` call that only returns true if the object is of a specific class.

spec/models/calnet_ldap/user_attributes_spec.rb
```diff
@@ -8,12 +8,12 @@
     let(:ldap_result) do
       {
         dn: ['uid=61889,ou=people,dc=berkeley,dc=edu'],
-        objectclass: ['top', 'eduPerson', 'inetorgperson', 'berkeleyEduPerson', 'organizationalperson', 'person', 'ucEduPerson'],
+        objectclass: %w(top eduPerson inetorgperson berkeleyEduPerson organizationalperson person ucEduPerson),
```
redconfetti: oh so ruby

spec/models/calnet_ldap/client_spec.rb
johncrossman: thanks, @redconfetti  You're right – in this case, we don't need the restrictiveness of `be_an_instance_of`.

PR #5951, opened 2016-10-05
SISRP-26308 - Profile > Emergency Contact > While Adding Additional phone number not choosing a phone type will not save phone number or display any warning message that phone type is missing

dfkaye:
https://jira.berkeley.edu/browse/SISRP-26308

I'll wait til this is reviewed before making a QA PR...

### DIFF
--- a/src/assets/javascripts/angular/directives/telPhoneInputDirective.js
+++ b/src/assets/javascripts/angular/directives/telPhoneInputDirective.js
@@ -20,18 +20,18 @@ angular.module('calcentral.directives').directive('ccTelPhoneInputDirective', fu
    *  + `valid` boolean property set to true if a pattern was matched, false if
    *   no pattern is matched.
    */
-  var reIntlPhone = /^(\+?((?:\d ?){6,14}\d))?$/;
-  var reUsPhone = /^([\d][\W])?(\d{3})\W?[\W]?(\d{3})[\W]?(\d{4})?$/;
+  var reIntlPhone = /^(\+?((?:\d ?){6,14}\d))?\s?$/;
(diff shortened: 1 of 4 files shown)


src/assets/javascripts/angular/directives/telPhoneInputDirective.js
```diff
@@ -60,9 +61,10 @@ angular.module('calcentral.directives').directive('ccTelPhoneInputDirective', fu
         });
       });
 
-      elm.on('change', function() {
-        // The change handler updates the view only when an invalid field value
-        // is detected, in order to show error messages.
+      elm.on('blur', function() {
```
davidlimcheng: According to the [jQuery](http://api.jquery.com/on/) docs, `change`, when used on an input, triggers when the element loses focus.  A similar definition is given for `blur` [as well](https://api.jquery.com/blur/), so it does seem like these two can be interchangeable in this situation.

dfkaye: @nkutub , @davidlimcheng - I was seeing weird behavior earlier where a click on cancel in the edit case - so I used blur and that behavior stopped. I'm not seeing it now change (probably because I changed disabled attribute logic) - so I'll can change it back.

dfkaye: @nkutub , @davidlimcheng - changes are pushed... travis building now...

davidlimcheng:
LGTM, will merge upon @nkutub's approval.

nkutub:
Looks good!

src/assets/javascripts/angular/directives/telPhoneInputDirective.js
nkutub: I would change it back and see if it is still doing it

PR #5952, opened 2016-10-05
CLC-6454 OEC: Update affiliation types in SIS import

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6454

CALDAP affiliation strings are so 2014.


johncrossman:
Thanks @paulkerschen

PR #5963, opened 2016-10-06
SISRP-26595 - Advisor Resources link hover text isn't always included in the markup

dfkaye:
https://jira.berkeley.edu/browse/SISRP-26595

+ Problem caused by not using the cc-campus-solutions-link-item-directive for items from the Link API.
+ The change here will fix the 2 eForms links, and the Multi-Year Planner link.
+ The WebNow and Schedule of Classes links come from another API that does not provide hover text.
+ The Academic Progress Report should not be visible in 7.1

### DIFF
--- a/src/assets/templates/widgets/advisor/advising_links.html
+++ b/src/assets/templates/widgets/advisor/advising_links.html
@@ -24,28 +24,28 @@
     ></div>
   </li>
   <li data-ng-if="csLinks.ucEformsCenter.url">
-    <a
-      data-ng-href="{{csLinks.ucEformsCenter.url}}"
-      data-ng-bind="csLinks.ucEformsCenter.name"
-    ></a>
+    <div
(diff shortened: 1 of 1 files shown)


davidlimcheng:
@dfkaye Just so I know - why don't the WebNow and Schedule of Classes link come from the Link API as well?

dfkaye:
@davidlimcheng - those two are not supported in the Link URL table on the Campus Solutions side yet.

davidlimcheng:
@dfkaye Hmm in that case, should we add some hardcoded hover text in the controller, just to be consistent?  I'd check in with the designer just in case, but if other reviewers are cool with merging this as-is, I'm OK with that.

davidlimcheng:
@dfkaye @nkutub I'll go ahead and merge this PR since #5964 has been merged, and I don't think what I brought up is a blocker at all.  But, I still think it would be good to have consistency with our hover text!

PR #5965, opened 2016-10-06
SISRP-26276 - Delegates can no longer see final exams

tiffanyqi:
https://jira.berkeley.edu/browse/SISRP-26276

Simply removed the requirement that only a student can see final exams. This should work.

### DIFF
--- a/src/assets/templates/academics.html
+++ b/src/assets/templates/academics.html
@@ -28,7 +28,7 @@
     <div class="medium-6 large-4 columns cc-column-3">
       <div data-ng-include="'widgets/enrollment_card.html'" data-ng-if="api.user.profile.features.csEnrollmentCard"></div>
 
-      <div data-ng-include="'widgets/final_exam_schedule.html'" data-ng-if="api.user.profile.roles.student && !api.user.profile.roles.law && api.user.profile.features.finalExamSchedule"></div>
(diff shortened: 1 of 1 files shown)


tiffanyqi:
Actually, delegates should not see final exams (but advisors can). This will be repurposed for that.

src/assets/templates/academics.html
```diff
@@ -28,7 +28,7 @@ <h1 class="cc-heading-page-title">My Academics</h1>
     <div class="medium-6 large-4 columns cc-column-3">
       <div data-ng-include="'widgets/enrollment_card.html'" data-ng-if="api.user.profile.features.csEnrollmentCard"></div>
 
-      <div data-ng-include="'widgets/final_exam_schedule.html'" data-ng-if="api.user.profile.roles.student && !api.user.profile.roles.law && api.user.profile.features.finalExamSchedule"></div>
+      <div data-ng-include="'widgets/final_exam_schedule.html'" data-ng-if="!api.user.profile.roles.law && api.user.profile.features.finalExamSchedule"></div>
```
johncrossman: @tiffanyqi When a delegate is in view-as mode, the `api.user.profile.roles.student` would be true if the target user is a student. Did this work when verified locally?

tiffanyqi: Apparently when a delegate views the card, they do not see it. I suppose removing this wouldn't make a difference in this case....

PR #5968, opened 2016-10-06
SISRP-26567 - MyAcademics::Exams for UID 1507981: NoMethodError: undefined method final_exam_cs_data_available for nil:NilClass

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-26567

* Good bug - led to the decision not to surface Final Exams to delegates, more info in the JIRA.
* Also included a CSS fix
* Also included an `!isAdvisingStudentLookup` conditional that was accidentally removed in #5905.  This makes it so that advisors don't see the "Book List" link on the User Overview page.  As it stands, if an advisor clicks on that link, they will be taken to a 404 page.
* QA PR to come upon merge.

### DIFF
--- a/src/assets/templates/academics.html
+++ b/src/assets/templates/academics.html
@@ -28,7 +28,7 @@
     <div class="medium-6 large-4 columns cc-column-3">
       <div data-ng-include="'widgets/enrollment_card.html'" data-ng-if="api.user.profile.features.csEnrollmentCard"></div>
 
-      <div data-ng-include="'widgets/final_exam_schedule.html'" data-ng-if="api.user.profile.roles.student && !api.user.profile.roles.law && api.user.profile.features.finalExamSchedule"></div>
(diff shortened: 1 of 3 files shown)


davidlimcheng:
@johncrossman Caught that we also need to include a delegate check on the front-end - amended.  Good catch John!

src/assets/templates/academics.html
```diff
@@ -28,7 +28,7 @@ <h1 class="cc-heading-page-title">My Academics</h1>
     <div class="medium-6 large-4 columns cc-column-3">
       <div data-ng-include="'widgets/enrollment_card.html'" data-ng-if="api.user.profile.features.csEnrollmentCard"></div>
 
-      <div data-ng-include="'widgets/final_exam_schedule.html'" data-ng-if="api.user.profile.roles.student && !api.user.profile.roles.law && api.user.profile.features.finalExamSchedule"></div>
+      <div data-ng-include="'widgets/final_exam_schedule.html'" data-ng-if="api.user.profile.roles.student && !api.user.profile.roles.law && !api.user.profile.delegateActingAsUid && api.user.profile.features.finalExamSchedule"></div>
```
johncrossman: @davidlimcheng Thanks for tossing in this extra bit of protection.

johncrossman:
Looks good to me @davidlimcheng   Waiting on Travis.

PR #5978, opened 2016-10-07
SISRP-22665 - Displays profile image instead of default image on advisor Student Lookup

lyttam:
https://jira.berkeley.edu/browse/SISRP-22665

- Adds a new endpoint to return a photo for a specified UID (only available to advisors and superusers)
- Adds a directive that calls that endpoint and displays the photo

I was hoping to move the authorization logic from the angular templates into the PhotoController for better security.  That way a single endpoint could return either the requested person's photo, or the generic bear photo if the requestor is not authorized (e.g., a delegate).  

HOWEVER, the generic bear photo is an SVG.  I tried having the endpoint send the SVG but it wouldn't render.  It works when that file is requested by delegate_student_img.html.  Didn't want to spend too much time figuring it out, but if anyone knows a quick solution I'd love to hear it!  That would allow me to replace lines 85-95 in academics_student_profile.html with the new directive and let the controller handle the authorization on the server side.

### DIFF
--- a/app/controllers/photo_controller.rb
+++ b/app/controllers/photo_controller.rb
@@ -1,10 +1,25 @@
 class PhotoController < ApplicationController
-
   before_filter :api_authenticate_401
 
   def my_photo
-    if (photo_feed = User::Photo.fetch session['user_id'], session)
-      data = photo_feed[:photo]
+    send_photo get_photo(session['user_id'])
+  end
+
+  def photo
+    if current_user.policy.can_view_other_user_photo?
+      send_photo get_photo(uid_param)
+    else
+      render :nothing => true, :status => 403
+    end
+  end
+
+  def get_photo(uid)
+    photo_feed = User::Photo.fetch uid, session
+    photo_feed.try(:[], :photo)
+  end
+
+  def send_photo(data)
+    if data
       send_data(
         data,
         type: 'image/jpeg',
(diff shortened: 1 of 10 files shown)


johncrossman:
@lyttam if student does not have a photo (e.g., Oski w/ UID: 61889) what will Advisor see on student overview?

app/controllers/photo_controller.rb
```diff
@@ -1,10 +1,28 @@
...
-    if (photo_feed = User::Photo.fetch session['user_id'], session)
-      data = photo_feed[:photo]
+    send_photo get_photo(session['user_id'])
+  end
+
+  def photo
+    user = User::Api.from_session(session).get_feed
+    has_advisor_role = user.try(:[], :roles).try(:[], :advisor)
+    is_super_user = user.try(:[], :isSuperuser)
+    if (has_advisor_role || is_super_user)
```
lyttam: @johncrossman I moved that logic into `AuthenticationStatePolicy` per your suggestion.  `PhotoController` looks much cleaner now, thanks!

lyttam:
@johncrossman It will show a white box with the alt text "Oski Ursa-Major  Bear's photo".  Is that ok or should I have it show what Oski sees when directly authenticated? 

image

johncrossman:
Thanks, @lyttam   Looks good.

PR #5982, opened 2016-10-10
SISRP-26666 - Advising Resources - should not include 'Schedule of Classes - Scheduler View' link

dfkaye:
https://jira.berkeley.edu/browse/SISRP-26666

+ removes the Schedule of Classes link from Dashboard Advisor Resources
+ adds hover text to WebNow Documents link in Dashboard Advisor Resources
+ adds hover text to student-specific Schedule Planner link on Student Overview (Advisor Resources)

Release Notes for 7.1 updated - remove `cs_advising_scheduler_view` flag from all environments:
https://confluence.ets.berkeley.edu/confluence/pages/viewpage.action?spaceKey=MYB&title=Release-Specific+Instructions

PR #5984, opened 2016-10-10
[QA] SISRP-26666 - Advising Resources - should not include Schedule Planner - Scheduler View link

dfkaye:
__Not ready for merge__

Master PR https://github.com/ets-berkeley-edu/calcentral/pull/5982
Jira https://jira.berkeley.edu/browse/SISRP-26666


src/assets/templates/widgets/advisor/advising_links.html
```diff
@@ -53,16 +53,6 @@
       data-link="csLinks.ucAppointmentSystem"
     ></div>
   </li>
-
-  <!-- This URL is mislabelled in the API, it's in fact the scheduler view link -->
-  <li data-ng-if="api.user.profile.features.csAdvisingSchedulerView && links.classSearch.url">
```
johncrossman: @davidlimcheng this seems to be the only use of feature flag `cs_advising_scheduler_view`. If that's true, let's (1) get rid of it in bundled YAML files and (2) update release instructions to have it removed from network. Thanks

davidlimcheng: @dfkaye ***

PR #5985, opened 2016-10-10
SISRP-26648 - Delegates get infinite spinner in popover when viewing as student

dfkaye:
https://jira.berkeley.edu/browse/SISRP-26648

### DIFF
--- a/src/assets/templates/widgets/gear_popover.html
+++ b/src/assets/templates/widgets/gear_popover.html
@@ -27,7 +27,8 @@
     <div>
       <div class="cc-launcher-status-loading"
         data-cc-loading-directive="statusLoading"
-        data-cc-loading-process="Loading...">
+        data-cc-loading-process="Loading..."
+        data-ng-if="statusLoading">
       </div>
       <div data-ng-if="!statusLoading && !api.user.profile.delegateActingAsUid && !api.user.profile.advisorActingAsUid && (count !== 0 || hasBillingData || hasRegistrationData)">
         <div class="cc-popover-title">
--- a/src/assets/javascripts/angular/controllers/widgets/statusController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/statusController.js
@@ -195,8 +195,10 @@ angular.module('calcentral.controllers').controller('StatusController', function
       $scope.hasAlerts = false;
       $scope.hasWarnings = false;
 
-      // We use this to show the spinner
-      $scope.statusLoading = 'Process';
+      // We use this to show the spinner.
+      if (!apiService.user.profile.delegateActingAsUid && !apiService.user.profile.advisorActingAsUid) {
+        $scope.statusLoading = 'Process';
+      }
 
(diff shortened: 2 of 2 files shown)


src/assets/templates/widgets/gear_popover.html
```diff
@@ -29,7 +29,7 @@
         data-cc-loading-directive="statusLoading"
         data-cc-loading-process="Loading...">
       </div>
-      <div data-ng-if="!statusLoading && !api.user.profile.delegateActingAsUid && !api.user.profile.advisorActingAsUid && (count !== 0 || hasBillingData || hasRegistrationData)">
+      <div data-ng-if="!statusLoading && api.user.profile.isDirectlyAuthenticated && (count !== 0 || hasBillingData || hasRegistrationData)">
```
dfkaye: @davidlimcheng - do you mean wrap the two <div>s - one for loading, one for showing, inside one with  `data-ng-if="api.user.profile.isDirectlyAuthenticated"`... something like that?

davidlimcheng: @dfkaye - yeah, that makes sense - so basically move the check over to the <div> on line 27.  I would keep it separated as `!api.user.profile.delegateActingAsUid && !api.user.profile.advisorActingAsUid` though, because if we use `isDirectlyAuthenticated` instead, that might break our regular "view-as" mode as superusers.  cc @johncrossman

src/assets/templates/widgets/gear_popover.html
```diff
@@ -24,12 +24,12 @@
     </div>
   </button>
   <div class="cc-popover cc-dropdown cc-launcher-dropdown-menu" data-ng-show="api.popover.status('cc-popover-menu')">
-    <div>
+    <div data-ng-if="!api.user.profile.delegateActingAsUid && !api.user.profile.advisorActingAsUid">
```
johncrossman: @dfkaye does this conditional belong on the outer div? I.e.,
```
<div class="cc-popover ...  data-ng-if="!api.user.profile.delegateActingAsUid && !api.user.profile.advisorActingAsUid">
  <div>
```
Might there be problems with an empty `cc-popover` div?

johncrossman: @dfkaye Now that you mention it, I see that `profile_popover.html` is inside your modified `<div>`. That means that "Log out" will be hidden from delegates and advisors. You'll need to test locally.

dfkaye: @johncrossman - updated - should be ok to go now

johncrossman:
@dfkaye this will visibly change the gear-popover: the top grey border is gone.
image
How about a change in `statusController.js` such that `$scope.statusLoading` is never true if `api.user.profile.delegateActingAsUid || api.user.profile.advisorActingAsUid`.

dfkaye:
@johncrossman - This one has been updated again - have another look...

johncrossman:
+1

PR #5986, opened 2016-10-10
[QA] SISRP-26648 - Delegates get infinite spinner in popover when viewing as student

dfkaye:
__Not ready for merge__

Master PR https://github.com/ets-berkeley-edu/calcentral/pull/5985

Jira https://jira.berkeley.edu/browse/SISRP-26648


src/assets/templates/widgets/gear_popover.html
```diff
@@ -29,7 +29,7 @@
         data-cc-loading-directive="statusLoading"
         data-cc-loading-process="Loading...">
       </div>
-      <div data-ng-if="!statusLoading && !api.user.profile.delegateActingAsUid && !api.user.profile.advisorActingAsUid && (count !== 0 || hasBillingData || hasRegistrationData)">
+      <div data-ng-if="!statusLoading && api.user.profile.isDirectlyAuthenticated && (count !== 0 || hasBillingData || hasRegistrationData)">
```
johncrossman: @dfkaye I'm unclear on how this will fix the problem. Can you explain. (Did PR #5879 introduce the bug?)

dfkaye: @johncrossman #5879 addressed a long-loading problem that affected selenium tests - the goal was to show the new spinner and allow the rest of the popover with links to appear as soon as that data was available. The new problem is that delegates `viewing-as` a student are seeing the spinner which never disappears as there is no alert or warning data available to be loaded. So the goal here to suppress the spinner and alert section if the user is not the student.

johncrossman: Thanks @dfkaye  But how would the change above change the behavior of the spinner? The `<div>` containing the spinner is not impacted by your change. Also, prior to this PR, the condition of the `ng-if` evaluated to `false` when delegate was in view-as mode. That's what we want, right?

dfkaye: @johncrossman - Not sure I understand what you're getting at - however, see my lastest reply to @davidlimcheng in the master PR - https://github.com/ets-berkeley-edu/calcentral/pull/5985 - with respect to wrapping both loading and alert divs...

johncrossman:
@dfkaye @davidlimcheng I'd like this PR to be closed. Once we have the corresponding 'master' merged we can turn to the QA branch. Thanks

PR #5987, opened 2016-10-11
SISRP-25708 - Adds My Academics > Degree Progress card for GRAD

lyttam:
https://jira.berkeley.edu/browse/SISRP-25708

Some of this will get refactored when the UGRD version of the card is implemented.  Just wanted to get this merged so I can start on the [Progress Report card](https://jira.berkeley.edu/browse/SISRP-22302), which will share the same code.

 Degree Progress has been pulled from the 7.2 release so I added a feature flag.

### DIFF
--- a/app/controllers/campus_solutions/degree_progress_controller.rb
+++ b/app/controllers/campus_solutions/degree_progress_controller.rb
@@ -1,0 +1,11 @@
+module CampusSolutions
+  class DegreeProgressController < CampusSolutionsController
+
+    include AllowDelegateViewAs
+
+    def get
+      render json: MyAcademics::DegreeProgress.from_session(session).get_feed_as_json
+    end
+
+  end
+end
--- a/app/models/berkeley/grad_milestones.rb
+++ b/app/models/berkeley/grad_milestones.rb
@@ -1,0 +1,37 @@
+module Berkeley
+  class GradMilestones
+
+    def self.get_status(status_code)
+      statuses[status_code.strip.upcase] unless status_code.blank?
+    end
+
+    def self.get_description(milestone_code)
+      milestones[milestone_code.strip.upcase] unless milestone_code.blank?
+    end
+
+    def self.get_merged_description
+      'Advancement to Candidacy Plan I or Plan II'
+    end
+
+    def self.milestones
+      @milestones ||= {
+        'AAGADVMAS1' => 'Advancement to Candidacy Plan I',
+        'AAGADVMAS2' => 'Advancement to Candidacy Plan II',
+        'AAGFINALCK' => 'Department Final Recommendations',
+        'AAGACADP1' => 'Thesis File Date',
+        'AAGQEAPRV' => 'Approval for Qualifying Exam',
+        'AAGQERESLT' => 'Qualifying Exam Results',
+        'AAGADVPHD' => 'Advancement to Candidacy',
+        'AAGFINALCK' => 'Department Final Recommendations',
+        'AAGDISSERT' => 'Dissertation File Date'
+      }
+    end
+
+    def self.statuses
+      @statuses ||= {
+        'Y' => 'Completed',
+        'N' => 'Not Satisfied'
+      }
+    end
+  end
+end
--- a/app/models/campus_solutions/degree_progress.rb
+++ b/app/models/campus_solutions/degree_progress.rb
@@ -1,0 +1,19 @@
+module CampusSolutions
+  class DegreeProgress < Proxy
+
+    include CampusSolutionsIdRequired
+
+    def initialize(options = {})
+      super options
+      initialize_mocks if @fake
+    end
+
+    def xml_filename
+      'degree_progress.xml'
+    end
+
+    def url
+      "#{@settings.base_url}/UC_AA_PROGRESS_GET.v1/UC_AA_PROGRESS_GET?EMPLID=#{@campus_solutions_id}"
+    end
+  end
+end
--- a/app/models/my_academics/degree_progress.rb
+++ b/app/models/my_academics/degree_progress.rb
@@ -1,0 +1,116 @@
+module MyAcademics
+  class DegreeProgress < UserSpecificModel
+    include Cache::CachedFeed
+    include Cache::JsonifiedFeed
+    include Cache::UserCacheExpiry
+    include CampusSolutions::DegreeProgressFeatureFlagged
+
+    CAREER_LAW = 'LAW'
+    ACAD_PROG_CODE_LACAD = 'LACAD'
+    LINKS_CONFIG = [
+      { feed_key: :advancement_form_submit, cs_link_key: 'UC_CX_GT_AAQEAPPLIC_ADD' },
+      { feed_key: :advancement_form_view, cs_link_key: 'UC_CX_GT_AAQEAPPLIC_VIEW' }
+    ]
+
+    def get_feed_internal
+      return {} unless is_feature_enabled
+      response = CampusSolutions::DegreeProgress.new(user_id: @uid).get
+      degree_progress = response.try(:[], :feed).try(:[], :ucAaProgress).try(:[], :progresses)
+      response[:feed] = HashConverter.camelize({
+        degree_progress: massage_progresses(degree_progress),
+        links: get_links
+      })
+      response
+    end
+
+    def get_links
+      links = {}
+      LINKS_CONFIG.each do |setting|
+        link = fetch_link setting[:cs_link_key]
+        links[setting[:feed_key]] = link unless link.blank?
+      end
+      links
+    end
+
+    def fetch_link(link_key)
+      if (link_feed = CampusSolutions::Link.new.get_url link_key)
+        link = link_feed.try(:[], :link)
+      end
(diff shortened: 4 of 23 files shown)


app/models/my_academics/degree_progress.rb
```diff
@@ -0,0 +1,118 @@
+module MyAcademics
+  class DegreeProgress < UserSpecificModel
+    include Cache::CachedFeed
+    include Cache::JsonifiedFeed
+    include Cache::UserCacheExpiry
+    include CampusSolutions::DegreeProgressFeatureFlagged
+
+    CAREER_LAW = 'LAW'
```
davidlimcheng: General question:  Does it not matter whether or not we have these declarations in a `initialize` method instead?

src/assets/javascripts/angular/controllers/widgets/academics/degreeProgressController.js
```diff
@@ -0,0 +1,44 @@
...
+    _.each(progresses, transformRequirements);
+    return progresses;
+  };
+
+  var loadDegreeProgress = function() {
+    degreeProgressFactory.getDegreeProgress().then(function(data) {
+      angular.extend($scope.degreeProgress, transform(data));
+      $scope.links = _.get(data, 'data.feed.links');
+      $scope.errored = _.get(data, 'data.errored');
+      $scope.isLoading = false;
```
davidlimcheng: @lyttam It might be better to have `$scope.isLoading = false;` in a .finally() promise, as that will ensure we don't have an infinite spinner.

lyttam: @davidlimcheng good call!  I made that change.

app/models/campus_solutions/degree_progress.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    def xml_filename
+      'degree_progress.xml'
+    end
+
+    def url
+      "#{@settings.base_url}/UC_AA_PROGRESS_GET.v1/UC_AA_PROGRESS_GET?EMPLID=#{@campus_solutions_id}"
+    end
+
+
```
johncrossman: @lyttam extra line break

app/models/my_academics/degree_progress.rb
```diff
@@ -0,0 +1,118 @@
...
+      merged_requirements = merge(requirements)
+      progress[:requirements] = merged_requirements
+    end
+
+    def should_exclude(progress)
+      CAREER_LAW == progress[:acadCareer] && ACAD_PROG_CODE_LACAD != progress[:acadProgCode]
+    end
+
+    def normalize(requirements)
+      requirements.map do |requirement|
```
johncrossman: @lyttam do you need exclamation point here?  I.e., `requirements.map!`

app/models/my_academics/degree_progress.rb
```diff
@@ -0,0 +1,118 @@
...
+        first[:name] = Berkeley::GradMilestones.get_merged_description
+        merged_requirements.unshift(first)
+      elsif merge_candidates.length === 1
+        merged_requirements.unshift(merge_candidates.first)
+      end
+      merged_requirements
+    end
+
+    def is_merge_candidate(requirement)
+      is_advancement_to_candidacy = REQUIREMENT_CODE_PLAN_1 == requirement[:code] || REQUIREMENT_CODE_PLAN_2 == requirement[:code]
```
johncrossman: @lyttam what do you think about replacing the line above with:
```
is_advancement_to_candidacy = %w(AAGADVMAS1 AAGADVMAS2).include? requirement[:code]
```
and then ditching REQUIREMENT_CODE_PLAN_1 and REQUIREMENT_CODE_PLAN_2 at top? Same logic, less code?

config/settings/test.yml
```diff
@@ -95,6 +95,7 @@ features:
   cs_advising_scheduler_view: true
   cs_advisor_student_lookup: true
   cs_billing: true
+  cs_degree_progress: true
```
johncrossman: @lyttam  this looks good. I believe `testext.yml` needs it, too.

src/assets/javascripts/angular/controllers/widgets/academics/degreeProgressController.js
```diff
@@ -0,0 +1,47 @@
...
+  $scope.isLoading = true;
+
+  var transformRequirements = function(progress) {
+    _.each(progress.requirements, function(requirement) {
+      requirement.hasForm = _.get(config, '[' + requirement.code + '][hasForm]', false);
+    });
+  };
+
+  var transform = function(data) {
+    var progresses = _.get(data, 'data.feed.degree_progress');
```
johncrossman: @lyttam The server-side feed should use camelCase keys. If it's not too much work, please change the above to `degreeProgress `.

src/assets/javascripts/angular/controllers/widgets/academics/degreeProgressController.js
```diff
@@ -0,0 +1,47 @@
...
+    },
+    AAGQEAPRV: {
+      hasForm: true
+    },
+    AAGADVPHD: {
+      hasForm: true
+    }
+  };
+  $scope.degreeProgress = {};
+  $scope.isLoading = true;
```
johncrossman: @lyttam I recommend changing the above to:
```
$scope.degreeProgress = {
  isLoading: true
}
```
Then, change the related `isLoading` references below. The main benefit is in the template:  `degreeProgress.isLoading` is more readable & precise.

src/assets/templates/widgets/academics/degree_progress.html
```diff
@@ -0,0 +1,34 @@
...
+        <div>
+          <a
+            data-ng-href="{{links.advancement_form_view.url}}"
+            data-ng-bind="links.advancement_form_view.name"
+          ></a>
+        </div>
+      </li>
+    </ul>
+    <ul>
+      <li data-ng-repeat="plan in degreeProgress">
```
johncrossman: @lyttam what happens if grad or law student has an empty `degreeProgress` array?

johncrossman:
Nice work, @lyttam!  My review is done; I posted a few comments/question.

app/models/my_academics/degree_progress.rb
lyttam: Good info.  Another benefit of constants is if the values have to change, you only need to edit in one place and it's at the top of the file.  I'm curious if people prefer other alternatives.

app/controllers/campus_solutions/degree_progress_controller.rb
```diff
@@ -0,0 +1,9 @@
+module CampusSolutions
+  class DegreeProgressController < CampusSolutionsController
+
```
lyttam: @johncrossman Yes, as long as the delegate can see My Academics they should see the Degree Progress card.  I made the change.

app/models/campus_solutions/degree_progress.rb
lyttam: @johncrossman removed.

app/models/berkeley/grad_milestones.rb
```diff
@@ -0,0 +1,37 @@
+module Berkeley
+  class GradMilestones
+
+    def self.get_status(status_code)
+      statuses[status_code.strip.upcase] || nil
```
lyttam: @johncrossman nope.  I changed the logic, also added a spec for this class.

app/models/berkeley/grad_milestones.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+
+    def self.get_description(milestone_code)
+      milestones[milestone_code.strip.upcase] || nil
```
lyttam: @johncrossman fixed.

app/models/my_academics/degree_progress.rb
lyttam: @johncrossman I like that better, since the constants were only used in this one spot anyway. I replaced that line with your suggestion.

config/settings/test.yml
lyttam: @johncrossman Added to testext.yml.

src/assets/javascripts/angular/controllers/widgets/academics/degreeProgressController.js
lyttam: @johncrossman fixed in all of the places.

src/assets/javascripts/angular/controllers/widgets/academics/degreeProgressController.js
lyttam: @johncrossman I updated this to wrap all of the scope variables in a degreeProgress object.

lyttam:
@johncrossman thank you for all of the excellent feedback!

app/controllers/campus_solutions/degree_progress_controller.rb
johncrossman: @lyttam whoops. I see it. Nevermind. :)

johncrossman:
Thank you, @lyttam   Looks really good.

PR #5990, opened 2016-10-11
SISRP-26727 - Final exams graduate courses should not show up in interim

tiffanyqi:
https://jira.berkeley.edu/browse/SISRP-26727

Replaced method to completely remove letters from the course code. Things like 'C242C' would become '0' instead of '242', so it was replaced.


davidlimcheng:
LGTM, thanks @tiffanyqi.

PR #5994, opened 2016-10-11
SISRP-26478 - 7.1 - Refactor My Finances > CARS Billing Summary given deprecation of CARS

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-26478

### DIFF
--- a/src/assets/templates/cars_summary.html
+++ b/src/assets/templates/cars_summary.html
@@ -12,7 +12,7 @@
   <div data-cc-spinner-directive>
     <div data-ng-if="api.user.profile.features.csBilling" class="cc-page-myfinances-account-summary-alert">
       <i class="fa fa-exclamation-triangle cc-icon-gold"></i>
-      <strong>Alert: </strong>Due to a service transition, the billing details for transactions <strong>prior to Fall 2016 </strong>are displayed here.
+      <strong>Alert: </strong>If you had an unpaid balance on your CARS account, that balance has been transferred to your new Fall 2016 account. Submit online payments by clicking the "Make Payment for Fall 2016" button above.
     </div>
     <ul class="cc-widget-list cc-page-myfinances-account-summary-list" data-ng-show="myfinances.summary">
       <li class="cc-page-myfinances-account-summary-item cc-clearfix-container cc-page-myfinances-dpp" data-ng-if="myfinances.summary.isOnDPP">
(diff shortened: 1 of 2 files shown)


src/assets/templates/cars_summary.html
```diff
@@ -12,7 +12,7 @@ <h2 class="cc-left">
   <div data-cc-spinner-directive>
     <div data-ng-if="api.user.profile.features.csBilling" class="cc-page-myfinances-account-summary-alert">
       <i class="fa fa-exclamation-triangle cc-icon-gold"></i>
-      <strong>Alert: </strong>Due to a service transition, the billing details for transactions <strong>prior to Fall 2016 </strong>are displayed here.
+      <strong>Alert: </strong>If you had an unpaid balance on your CARS account, that balance has been transferred to your new Fall 2016 account. Submit online payments by clicking the "Make Payment for Fall 2016" button above.
```
johncrossman: @davidlimcheng where is the logic that displays `"Make Payment for Fall 2016"` button?

davidlimcheng: @johncrossman That would be in [billing_summary.html](https://github.com/ets-berkeley-edu/calcentral/blob/master/src/assets/templates/widgets/billing_summary.html#L46).

johncrossman: Found it in `billing_summary.html`. The button is disabled when `!api.user.profile.isDirectlyAuthenticated && !api.user.profile.canActOnFinances`.  Do we need to wrap the phrase, "Submit online payments by..." in the same conditional?

davidlimcheng: Nice - that would be very clean and makes sense to me.  I'm also looking at the CARS "Make Payment" button and there's no delegate conditional on there, so I wonder if delegates would expect to be able to make a payment on `billing_summary`.  I'll confer with Adrianne.

davidlimcheng: @johncrossman - speaking with Adrianne, I realize that the `!api.user.profile.canActOnFinances` condition does let delegates with financial privileges submit online payments.  Those that don't will not have the Financials tab at all (due to `feed[:hasFinancialsTab] = false unless view_as_privileges[:financial]` in `User::Api`).  

So, the only people that would see the text and not be able to pay would be regular "view-as," which we don't worry about since they can see everything anyways.

To conclude this essay, I think we're OK in not wrapping that text in the same conditionals - especially since they'll still be able to see the button, it'll just be disabled.

johncrossman: great, thanks @davidlimcheng

PR #5998, opened 2016-10-12
SISRP-26745 - Advising Resources - Advising Assignments link is missing

dfkaye:
https://jira.berkeley.edu/browse/SISRP-26745

Second fix added in template for the multi-year planner link (should open in new window): https://jira.berkeley.edu/browse/SISRP-26746

### DIFF
--- a/src/assets/templates/widgets/advisor/advising_links.html
+++ b/src/assets/templates/widgets/advisor/advising_links.html
@@ -5,10 +5,10 @@
       data-link="csLinks.ucServiceIndicators"
     ></div>
   </li>
-  <li data-ng-if="csLinks.ucStudentAdvisor.url">
+  <li data-ng-if="csLinks.ucAdvisingAssignments.url">
     <div
       data-cc-campus-solutions-link-item-directive
-      data-link="csLinks.ucStudentAdvisor"
+      data-link="csLinks.ucAdvisingAssignments"
     ></div>
   </li>
   <li data-ng-if="csLinks.ucReportingCenter.url">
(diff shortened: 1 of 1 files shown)


src/assets/templates/widgets/advisor/advising_links.html
```diff
@@ -5,10 +5,10 @@
       data-link="csLinks.ucServiceIndicators"
     ></div>
   </li>
-  <li data-ng-if="csLinks.ucStudentAdvisor.url">
+  <li data-ng-if="csLinks.ucAdvisingAssignments.url">
     <div
       data-cc-campus-solutions-link-item-directive
-      data-link="csLinks.ucStudentAdvisor"
+      data-link="csLinks.ucAdvisingAssignments"
```
johncrossman: @dfkaye Are we discontinuing the use of `ucStudentAdvisor`?  If so, you must remove it from `advising_student_controller.rb` and `dummy/json/advising_resources.json`

dfkaye: @johncrossman - how about i roll that back in the template and change the name in the model instead?

johncrossman: @dfkaye with respect to the naming, if `ucAdvisingAssignments` is a better description of the content then +1.

dfkaye: @johncrossman on further investigation it looks like we won't need `dummy/json/advising_resources.json` or `fixtures/xml/campus_solutions/advising_resources.xml` or the call to the `/UC_AA_ADVISING_RESOURCES.v1\/UC_ADVISING_RESOURCES/` campus solutions api.

Shall I go ahead and remove all that and update the spec?

johncrossman:
Thanks for taking time to talk this through, @dfkaye

PR #6000, opened 2016-10-12
SISRP-26676 - Student Success Card needs to suppress current term GPA

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-26676

* The JIRA title is a bit misleading - bottom line, we want to suppress the current term GPA because CS will always return 0.0, as grades have not been entered yet.
* We still want to show the term though
* QA PR to come upon merge.

### DIFF
--- a/app/models/campus_solutions/student_term_gpa.rb
+++ b/app/models/campus_solutions/student_term_gpa.rb
@@ -8,6 +8,17 @@ module CampusSolutions
       "#{@settings.base_url}/UC_AA_STDNT_GPA_TERMS.v1/get?EMPLID=#{@campus_solutions_id}"
     end
 
+    def build_feed(response)
+      # Since CS will always return a zero value for the current term, we want to suppress it
+      current_term = Berkeley::Terms.fetch.current
+      response.try(:[], 'UC_AA_TERM_DATA').try(:[], 'UC_AA_TERM_GPA').each do |term_gpa|
+        if term_gpa.try(:[], 'TERM_ID') == current_term.try(:[], :campus_solutions_id)
+          term_gpa['TERM_CUM_GPA'] = nil
+        end
+      end
+      response.parsed_response
+    end
+
     def xml_filename
       'student_term_gpa.xml'
     end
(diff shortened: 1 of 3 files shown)


davidlimcheng:
#6000!

app/models/campus_solutions/student_term_gpa.rb
```diff
@@ -8,6 +8,17 @@ def url
       "#{@settings.base_url}/UC_AA_STDNT_GPA_TERMS.v1/get?EMPLID=#{@campus_solutions_id}"
     end
 
+    def build_feed(response)
+      # Since CS will always return a zero value for the current term, we want to suppress it
+      current_term = Berkeley::Terms.fetch.current
+      response.try(:[], 'UC_AA_TERM_DATA').try(:[], 'UC_AA_TERM_GPA').each do |term_gpa|
+        if term_gpa.try(:[], 'TERM_ID') == current_term.try(:[], :campus_solutions_id)
```
johncrossman: @davidlimcheng will `current_term` contain `: campus_solutions_id`?

davidlimcheng: @johncrossman Yep, that can be seen defined in [Berkeley::Term](https://github.com/ets-berkeley-edu/calcentral/blob/master/app/models/berkeley/term.rb#L10)

johncrossman: @davidlimcheng Your code references `Berkeley::Terms`.  :)

davidlimcheng: @johncrossman Ah that is true, but I believe Berkeley::Terms [calls](https://github.com/ets-berkeley-edu/calcentral/blob/master/app/models/berkeley/terms.rb#L141) on Berkeley::Term to nicely package all of that data for us and make it available for the specified term.  

To be honest, the `Berkeley::Terms` and `Berkeley::Term` is close to wizardry to me, but I know that its implementation is very convenient and [easy to use](https://github.com/ets-berkeley-edu/calcentral/blob/master/app/models/my_academics/registrations.rb#L37) due to @raydavis's work on it.

redconfetti: Yes. Berkeley::Terms provides the wrapping object that populates the various instances such as #current, #running, #next, #future. Those objects are cast from Berkeley::Term, which [has a campus_solutions property](https://github.com/ets-berkeley-edu/calcentral/blob/master/app/models/berkeley/term.rb#L10).

johncrossman: Nice. Thanks, @davidlimcheng @redconfetti  Wizardry, indeed.

PR #6001, opened 2016-10-12
SISRP-26742 - Student Overview page Advising Card (Refactor and Haas Program Messages) are showing incorrect messages

redconfetti:
Essentially updates the academicsFactory for use on the Student Overview page for Advisors so that the filtered feed can be re-used by the Advising card.

https://jira.berkeley.edu/browse/SISRP-26742

### DIFF
--- a/src/assets/javascripts/angular/controllers/widgets/myAdvisingController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/myAdvisingController.js
@@ -6,7 +6,7 @@ var _ = require('lodash');
 /**
  * My Advising controller
  */
-angular.module('calcentral.controllers').controller('MyAdvisingController', function(academicsFactory, myAdvisingFactory, $scope) {
+angular.module('calcentral.controllers').controller('MyAdvisingController', function(academicsFactory, advisingFactory, myAdvisingFactory, $route, $routeParams, $scope) {
   $scope.myAdvising = {
     feedsLoading: {
       myAdvising: true,
@@ -37,7 +37,11 @@ angular.module('calcentral.controllers').controller('MyAdvisingController', func
   };
 
   var loadFeeds = function() {
-    academicsFactory.getAcademics().then(function(data) {
+    var academicsSource = $route.current.isAdvisingStudentLookup ? advisingFactory.getStudentAcademics : academicsFactory.getAcademics;
+    var options = {
+      uid: $routeParams.uid
+    };
+    academicsSource(options).then(function(data) {
       angular.extend($scope.myAdvising.roles, _.get(data, 'data.collegeAndLevel.roles'));
       $scope.myAdvising.feedsLoading.academicsRoles = false;
     });


src/assets/javascripts/angular/factories/academicsFactory.js
```diff
@@ -5,19 +5,22 @@ var angular = require('angular');
 /**
  * Academics Factory
  */
-angular.module('calcentral.factories').factory('academicsFactory', function(apiService) {
-  var url = '/api/my/academics';
-  // var url = '/dummy/json/academics.json';
+angular.module('calcentral.factories').factory('academicsFactory', function(apiService, $route, $routeParams) {
+  var urlStudentAcademicsInfo = '/api/my/academics';
+  // var urlStudentAcademicsInfo = '/dummy/json/academics.json';
+  var urlAdvisorStudentAcademicsInfo = '/api/advising/academics/';
```
davidlimcheng: @redconfetti As discussed, `userOverviewController.js` is making a similar call to this endpoint on `advisingFactory.js`.  Having the page controller as well as the card controller (`advisingStudentController.js`) can result in two of the same calls on the user overview page.

redconfetti:
So the problem with my approach here is that it duplicates the API call to `/api/advising/academics/` inside of academicsFactory.js in addition to where it currently is already called from advisingFactory.js.

The reason I did this was because the Advising card uses it's own controller (MyAdvisingController) instead of just being a template that utilizes the main page controller (AcademicsController or UserOverviewController).

After discussing this with @davidlimcheng, he noted how we sometimes will load data from an ancestor $scope like `$scope.academics = $scope.$parent.$parent.academics;`. I don't like this approach though because it introduces a race condition where the parent controller might not have it's own `$scope.academics` value loaded from the back-end API in time for the child controller to access it.

I'm suggesting that we rely on the `httpService`, which uses $cacheFactory, to stop multiple API requests from being replicated on the same page.

redconfetti:
I’m thinking in terms of Page Controllers (academicsController, UserOverviewController), widget controllers, and factories. When we load in page controllers, we merge widget logic for 4-5 widgets into the same controller. They are not isolated and modular, self contained.

If they share data sources through the factories, with $http and $cacheFactory ensuring that all requests are not duplicated, that loaded data is shared in memory, we can make the widgets modular. They should have their own widget controller, with it's own logic, with it's own template, and use services if they need to share functionality.

If the widget needs to get the same data, but from one source or the other, then shouldn't the widgetController hold that logic, not the [factory](https://github.com/ets-berkeley-edu/calcentral/blob/master/src/assets/javascripts/angular/factories/myAdvisingFactory.js#L15)? I do understand why putting that logic in the factory reduces mistakes where we forget that the user is an Advisor, and thus should receive a filtered data feed. It just seems like a bad idea because it mixes business logic (security) with what are essentially data providers modules (factories). What happens once we have delegate dashboards, and other user role dashboards?

It seems like the pattern should be that the widget controller itself should be responsible for using $route.current to determine the role of the user, and then choose the factory method it will source data from. In addition, the back-end should obviously restrict the data served to the user. An advisor querying /api/my/academics should get nothing, and a student querying /api/advising/advising should get nothing.

src/assets/javascripts/angular/factories/academicsFactory.js
johncrossman: Ping @redconfetti

redconfetti:
@davidlimcheng - I've updated this to take a different approach, based on my thinking above. The widgets should contain their own business logic that checks the $routeParams for flags such as `isAdvisingStudentLookup`, and then determine which source to use.

This will keep the factories simple, and keep the logic for the card itself inside of the card and template that is related to that widget. The logic is now in the controller instead of the factory.

nkutub:
Thanks @redconfetti for the overview, it looks good.

PR #6003, opened 2016-10-12
SISRP-26743 - My Academics - Advising Card - MBA/Juris Doctor (MBA/JD) program should not display duplicate contact info on the advising card

redconfetti:
https://jira.berkeley.edu/browse/SISRP-26743


davidlimcheng:
*thumb emoji*

PR #6005, opened 2016-10-12
SISRP-26772, My Finances tab if student or has_student_history

johncrossman:
https://jira.berkeley.edu/browse/SISRP-26772

We leverage the existing call to `User::HasStudentHistory.has_student_history?`. Note: relying on `CampusSolutions::Billing` to decide if `has_financials_tab?` is risky w.r.t performance.


davidlimcheng:
Hallelujah!

PR #6012, opened 2016-10-13
SISRP-26257 - Filter out inactive studentPlans and remove Academic Statuses when no plans are present

redconfetti:
https://jira.berkeley.edu/browse/SISRP-26257

### DIFF
--- a/app/models/my_academics/college_and_level.rb
+++ b/app/models/my_academics/college_and_level.rb
@@ -122,7 +122,10 @@ module MyAcademics
         roles: role_booleans
       }
 
-      statuses.each do |status|
+      filtered_statuses = filter_inactive_status_plans(statuses)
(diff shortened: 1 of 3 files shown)


app/models/my_academics/college_and_level.rb
```diff
@@ -155,6 +158,22 @@ def parse_hub_plans(statuses)
       plan_set
     end
 
+    def filter_inactive_status_plans(statuses)
+      statuses.each do |status|
+        status['studentPlans'].select! do |plan|
```
paulkerschen: @redconfetti Is there any risk that the destructive `select!` will unexpectedly remove information that another method might be looking for? For example, if the order of the `parse_` methods was changed?

paulkerschen: Thanks @redconfetti.

PR #6017, opened 2016-10-14
SISRP-26794 - Student Overview(Advisor) - Advising Resources card - Service Indicators link - doesn't prepopulate the fields in the CS page

dfkaye:
https://jira.berkeley.edu/browse/SISRP-26794

### DIFF
--- a/app/models/campus_solutions/link.rb
+++ b/app/models/campus_solutions/link.rb
@@ -43,8 +43,8 @@ module CampusSolutions
           link = entry.slice(:url, :urlId, :description, :hoverOverText)
 
           placeholders.try(:each) do |k, v|
-            if v.blank?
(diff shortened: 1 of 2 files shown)


app/models/campus_solutions/link.rb
```diff
@@ -43,16 +43,16 @@ def get_url(url_id, placeholders = {})
...
+              logger.debug "Could not set placeholder for #{k} on link with url_id #{link[:urlId]}"
               link = nil
               break
             else
               link[:url] = link[:url].gsub("{#{k}}", v)
             end
           end
 
-          if !link.blank?
+          if !link.nil?
```
redconfetti: Is this business logic supposed to be...
> If there is a problem with the link, just don't display it at all, okay!
??

johncrossman:
Thanks, @dfkaye

PR #6020, opened 2016-10-16
SISRP-21639 - Adds empty YAML configurations for Class Enrollment deeplinks

redconfetti:
https://jira.berkeley.edu/browse/SISRP-21639

@johncrossman pointed out these missing placeholders in the configuration during v83 release


config/settings.yml
```diff
@@ -447,6 +447,9 @@ campus_solutions_links:
   academics:
     enrollment_verification_google: ''
     enrollment_Verification: ''
+  class_enrollment:
+    concurrent_apply_to_class: ''
+    concurrent_submitted_class_applications: ''
```
johncrossman: @redconfetti  How are these used? A grep of code gets no match

redconfetti: I investigated further, and I've found that we were both wrong. These should have been removed from the configs for DEV, QA, and PROD in the release notes for GL 7.1. See #5906.

PR #6021, opened 2016-10-17
SISRP-26257 - Implement proper fix to filter out inactive academic statuses

redconfetti:


app/models/my_academics/college_and_level.rb
```diff
@@ -123,9 +124,8 @@ def parse_hub_plans(statuses)
       }
 
       filtered_statuses = filter_inactive_status_plans(statuses)
-      active_statuses = active_academic_statuses(filtered_statuses)
```
redconfetti: I essentially removed this filtering of inactive academic statuses from this level within #parse_hub_plans, up to the point right after we obtain the academic statuses, thus forcing the only statuses that are processed are the ones that the filter function does return (active statuses).

spec/models/my_academics/college_and_level_spec.rb
```diff
@@ -377,106 +379,142 @@
...
-
-      # Graduate Statuses - Current Registrations
-      let(:current_registration_secondary) do
-        {
-          "academicCareer" => current_registration_academic_career_secondary,
-          "academicLevel" => current_registration_academic_level_secondary,
-          "term" => current_registration_term_secondary,
-        }
-      end
+    context 'graduate with graduate academic status' do
```
redconfetti: I created a new wrapping context that sets up the student as a graduate instead of an undergraduate. Then I expand the students states within the inner contexts to add secondary plans, such as a Law-based academic status, or an inactive academic status (undergraduate), to aid with the ease of testing the integrated functionality.

PR #6022, opened 2016-10-17
SISRP-25956 - Residency exception from unknown message code - revised

dfkaye:

Change the return value `{}` to `nil` in  `residency_message_code.rb`


dfkaye:
Bad commit - closing

PR #6023, opened 2016-10-17
SISRP-25956 - Residency exception from unknown message code - return value fix

dfkaye:


app/models/berkeley/residency_message_code.rb
```diff
@@ -46,7 +46,7 @@ def residency_message_code(slr_status, residency_status, tuition_exception)
       end
       if message_code.nil?
         logger.warn "Cannot determine message code for residency '#{residency_status}', SLR '#{slr_status}', tuition exception '#{tuition_exception}'"
-        {}
+        nil
```
johncrossman: Thanks, @dfkaye  In `residency_message_code_spec`, please update the tests in `describe 'no match' to expect nil.  The expectation to `be_blank` is not enough, as we've seen. :)

johncrossman:
Thanks, @dfkaye

PR #6024, opened 2016-10-17
SISRP-25362 - adds handling for CS API error responses with status=200

lyttam:
https://jira.berkeley.edu/browse/SISRP-25362 (ticket not related; I ran into this issue working on the degree progress API but it was a preexisting issue)

If we try to hit a CS endpoint that doesn't exist, we'll get a string response but a successful status code.  This fix will flag it as an error so we can properly deal with it and avoid `TypeError`s down the line.

### DIFF
--- a/app/models/campus_solutions/proxy.rb
+++ b/app/models/campus_solutions/proxy.rb
@@ -92,7 +92,7 @@ module CampusSolutions
     end
 
     def is_errored?(feed)
-      feed.is_a?(Hash) && feed[:errmsgtext].present?
+      !feed.is_a?(Hash) || feed[:errmsgtext].present?
     end
 
     def request_options
--- a/spec/models/campus_solutions/proxy_spec.rb
+++ b/spec/models/campus_solutions/proxy_spec.rb
@@ -9,13 +9,22 @@ describe CampusSolutions::Proxy do
       allow(response).to receive(:body).and_return double force_encoding: nil
       allow(response).to receive(:parsed_response).and_return feed
     }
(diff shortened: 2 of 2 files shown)


johncrossman:
@lyttam Thanks for the fix. This needs an associated Jira so the QA Dept can verify. I might have violated this rule in the past: NOJIRA PRs should be reserved for bumping the application version and similar. Changes to logic need a Jira.

lyttam:
@johncrossman Thanks for clarifying.  I'll add it to my Jira-associated branch.  Is it cool to still submit this change in its own PR, separate from the code for the feature I'm working on?

johncrossman:
@lyttam it would be good enough to add the associated Jira number to summary and description of this PR. I don't think you need to close this.

lyttam:
@johncrossman great. PR updated.

app/models/campus_solutions/proxy.rb
```diff
@@ -92,7 +92,7 @@ def url
     end
 
     def is_errored?(feed)
-      feed.is_a?(Hash) && feed[:errmsgtext].present?
+      !feed.is_a?(Hash) || (feed.is_a?(Hash) && feed[:errmsgtext].present?)
```
paulkerschen: @lyttam I think this can be condensed to `!feed.is_a?(Hash) || feed[:errmsgtext].present?` - if it doesn't short-circuit out before the `||`, we know it's got to be a Hash.

johncrossman:
Thanks, @lyttam

PR #6025, opened 2016-10-17
SISRP-26964, avoid crash in MyAcademics::Exams; introduce FinalExamsFeatureFlagged 

johncrossman:
https://jira.berkeley.edu/browse/SISRP-26963
* do nothing when `FinalExamsFeatureFlagged` is false
* add proper nil checks
* update all YAMLs
* move the `sort_by` out of `assign_exams()`
* code style and spec tidying

### DIFF
--- a/app/models/my_academics/exams.rb
+++ b/app/models/my_academics/exams.rb
@@ -1,16 +1,18 @@
-require 'date'
-
 module MyAcademics
   class Exams
     include AcademicsModule
     include ClassLogger
     include DatedFeed
     include User::Student
+    include FinalExamsFeatureFlagged
 
     # Merges the final result into the academics feed
     def merge(data = {})
-      academics_data = parse_academic_data(data[:semesters])
-      data[:examSchedule] = assign_exams(academics_data, Berkeley::FinalExamSchedule.fetch)
+      if is_feature_enabled && (academics_data = parse_academic_data data[:semesters])
+        assign_exams academics_data
+        # Sort by semester, current first
+        data[:examSchedule] = academics_data.sort_by { |semester| semester[:timeBucket] == 'current' ? 0 : 1 }
+      end
     end
 
     # Parses the my academics semesters feed, and prepares data to be populated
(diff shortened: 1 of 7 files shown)


config/settings.yml
```diff
@@ -481,7 +481,7 @@ features:
   cs_transfer_credit: false
   enrollment_verification: false
   enrollment_verification_deeplink: false
-  final_exam_schedule: true
+  final_exam_schedule: false
```
johncrossman: Per convention, the default setting matches what we have in production (false). Other YAMLs have true.

app/models/my_academics/exams.rb
```diff
@@ -1,16 +1,19 @@
...
     include ClassLogger
     include DatedFeed
     include User::Student
+    include FinalExamsFeatureFlagged
 
     # Merges the final result into the academics feed
     def merge(data = {})
-      academics_data = parse_academic_data(data[:semesters])
-      data[:examSchedule] = assign_exams(academics_data, Berkeley::FinalExamSchedule.fetch)
+      return data unless is_feature_enabled
```
paulkerschen: @johncrossman Minor, but I think it may be misleading to specify a return value for the `merge` method, since what matters is not the return value but the mutation to `data`. (If the merge succeeds, then the method will return `data[:examSchedule]`, but that return value is ignored.)

tiffanyqi:
Thanks for improving code style @johncrossman!

app/models/my_academics/exams.rb
johncrossman: Thanks @paulkerschen  The `return` has been ditched.

spec/models/my_academics/exams_spec.rb
```diff
@@ -478,22 +555,31 @@
...
         }
       end
 
+      context 'when the final_exam_schedule feature is disabled' do
+        let(:feature_flag) { false }
+        it 'should do nothing' do
+          expect(merged_result).to be feed
+        end
+      end
+
```
paulkerschen: @johncrossman Looks like a legit Travis failure without a return value from `merge`. We could redefine `merged_result` as `subject.merge(feed); feed` and then check that no `:examSchedule` key was added.

johncrossman: thanks @paulkerschen  I was just applying that kind of fix.

paulkerschen:
Looks good! Thanks @johncrossman.

PR #6027, opened 2016-10-18
SISRP-25362 - Refactors server-side Degree Progress code 

lyttam:

1.  Creates separate models for student and advisor.  The student model (`MyGradMilestones`) injects links from the Link API into the response.  The advisor doesn't need to see these links.

1.  Prepares for the addition of new models for undergrad data (student and advisor) by appending 'Grad' to the name of the CS proxy class.  Will add these new models under a different Jira.

1.  Moves models into a `DegreeProgress` module since we'll be adding more of them.

Will submit a separate PR for front-end work.


johncrossman:
@lyttam do you mind adding Jira URL to the description?

app/models/campus_solutions/degree_progress_grad.rb
```diff
@@ -1,5 +1,5 @@
 module CampusSolutions
-  class DegreeProgress < Proxy
+  class DegreeProgressGrad < Proxy
```
johncrossman: @lyttam for lazy thinkers like myself, this class name might be interpreted as "progress towards graduation", which could apply to undergrads. What do you think about a name change to `GraduateDegreeProgress`?  That should make the purpose clear.

app/models/degree_progress/grad_milestones.rb
```diff
@@ -0,0 +1,18 @@
+module DegreeProgress
+  class GradMilestones < UserSpecificModel
```
johncrossman: @lyttam `GradSchoolMilestones`?

lyttam: @johncrossman this is for GRAD and LAW careers, so does that still make sense? Or `GraduateMilestones`, `GradCareerMilestones`, `GradAndLawMilestones`... what do we call these things?  I appreciate your help with the nomenclature.

app/models/campus_solutions/degree_progress_grad.rb
redconfetti: I like when they line up next to each other. Perhaps?
- DegreeProgressGraduate
- DegreeProgressUndergraduate

This would also help the following line up well:
- DegreeProgressLaw
- DegreeProgressConcurrent

app/models/degree_progress/grad_milestones.rb
redconfetti: Aw man. Maybe we should just take name spacing to the next level and do:

- CampusSolutions::Milestones::Graduate
- CampusSolutions::DegreeProgram::Graduate
- CampusSolutions::Milestones::Undergraduate
- CampusSolutions::DegreeProgram::Undergraduate
- CampusSolutions::Milestones::Law
- CampusSolutions::DegreeProgram::Law

app/models/campus_solutions/degree_progress_grad.rb
johncrossman: @lyttam I think @redconfetti's suggestion is good. Use `Undergrad` instead of `Undergraduate`; conciseness, when possible.

app/models/degree_progress/grad_milestones.rb
johncrossman: @redconfetti maybe you're right. I'm all about readability.  cc: @lyttam

app/models/degree_progress/graduate_milestones.rb
```diff
@@ -0,0 +1,18 @@
+module DegreeProgress
+  class GraduateMilestones < UserSpecificModel
```
paulkerschen: @lyttam Is there a reason we're providing both `DegreeProgress::GraduateMilestones` and `DegreeProgress::MyGraduateMilestones`? It looks like the former is much the same as the latter, but doesn't include links and doesn't currently talk to a controller.

Also, do the `DegreeProgress` models source their data from anywhere other than Campus Solutions? If not, it might make sense to store them under the `CampusSolutions` namespace, as we do for other CS-sourced `My...` models. (It's true that `app/models/campus_solutions` probably has way too many files in it by now, but that might be best addressed in a future refactor.)

paulkerschen:
@lyttam - Never mind, I reread your explanation in the PR and understand what's going on. One request: could you perhaps add comments in `DegreeProgress::GraduateMilestones` and `DegreeProgress::MyGraduateMilestones` explaining that these are parallel student- and advisor-flavored versions of the model, so that hasty readers like me don't get confused?

lyttam:
OK @johncrossman and @redconfetti, take another look if you could and let me know how this looks.  The new nam(espace)ing is as follows.

```
app
|-- controllers
|    |-- campus_solutions
|    |    |-- degree_progress_controller.rb
|-- models
|    |-- campus_solutions
|    |    |-- degree_progress
|    |    |    |-- graduate_milestones.rb
|    |-- degree_progress
|    |    |-- graduate_milestones.rb
|    |    |-- milestones_module.rb
|    |    |-- my_graduate_milestones.rb
```

I'll add the undergraduate versions of each model to the existing directories (will name them UndergradRequirements).  There will be no LAW or UCBX version.

lyttam:
@paulkerschen Good idea, I added some comments.

lyttam:
@johncrossman @redconfetti @paulkerschen I think this PR is ready to merge, unless you see anything I failed to address?

paulkerschen:
Looks good to me, thanks @lyttam.

PR #6028, opened 2016-10-18
CLC-6464 CLC-6465 New tables and models for mailing lists; adapt sync logic

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6464
https://jira.ets.berkeley.edu/jira/browse/CLC-6465

We handle the split between old Calmail lists and new Mailgun lists by creating two subclasses for MailingLists::SiteMailingList, and using Rails single-table inheritance to associate those subclasses with a new `type` column. For now, lists created through the bCourses admin tool still default to Calmail.

Membership logic for Mailgun lists is a bit simpler than that for Calmail, since we're using a local database rather than an external API. However, we do now need to track "is sender?" permissions, since no external service is managing those.

Bamboo build https://bamboo.ets.berkeley.edu/bamboo/browse/CPB-CPT389-2 has no mailing list failures.

### DIFF
--- a/app/controllers/canvas_mailing_lists_controller.rb
+++ b/app/controllers/canvas_mailing_lists_controller.rb
@@ -23,10 +23,17 @@ class CanvasMailingListsController < ApplicationController
   end
 
   # POST /api/academics/canvas/mailing_lists/:canvas_course_id/create
+  # Unlike other CRUD actions, list creation must specify which type of list is being created.
 
   def create
-    list = MailingLists::SiteMailingList.create canvas_site_id: canvas_course_id.to_s, list_name: params['listName']
-    render json: list.to_json
+    list_type = params['listType'] || 'CalmailList'
+    list_class = MailingLists.const_get(list_type) rescue nil
+    if MailingLists::SiteMailingList.subclasses.include? list_class
+      list = list_class.create canvas_site_id: canvas_course_id.to_s, list_name: params['listName']
+      render json: list.to_json
+    else
+      raise Errors::BadRequestError, "Unrecognized list type '#{list_type}'"
+    end
   end
 
   # POST /api/academics/canvas/mailing_lists/:canvas_course_id/populate
--- a/app/models/mailing_lists/calmail_list.rb
+++ b/app/models/mailing_lists/calmail_list.rb
@@ -1,0 +1,109 @@
+module MailingLists
+  class CalmailList < SiteMailingList
+
+    # Lists with membership stored in Calmail enter a 'pending' state on first save, and are not
+    # considered 'created' until we've verified creation on the Calmail side.
+    before_create { self.state = 'pending' }
+    after_find { check_for_creation if self.state == 'pending' }
+
+    def self.domain
+      Settings.calmail_proxy.domain
+    end
+
+    private
+
+    def add_list_urls(feed)
+      feed[:mailingList][:creationUrl] = build_creation_url if self.state == 'pending'
+      feed[:mailingList][:administrationUrl] = build_administration_url if self.state == 'created'
+    end
+
+    def build_administration_url
+      Settings.calmail_proxy.base_url.sub(
+        /api1\Z/,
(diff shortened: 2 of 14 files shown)


app/controllers/canvas_mailing_lists_controller.rb
```diff
@@ -23,9 +23,11 @@ def show
   end
 
   # POST /api/academics/canvas/mailing_lists/:canvas_course_id/create
+  # Unlike other CRUD actions, list creation must specify which type of list is being created.
 
   def create
-    list = MailingLists::SiteMailingList.create canvas_site_id: canvas_course_id.to_s, list_name: params['listName']
+    list_class = "MailingLists::#{params['listType'] || 'CalmailList'}".constantize
```
lyttam: @paulkerschen any concern about `params['listType']` containing an invalid name (or the name of a class you don't want to expose)?  Or will `constantize` throw a `NameError` if the class is not defined?

paulkerschen: @lyttam It should throw a `NameError`, but you're right; we should wrap and raise `BadRequestError` in this case.

app/models/mailing_lists/calmail_list.rb
```diff
@@ -0,0 +1,109 @@
+module MailingLists
+  class CalmailList < SiteMailingList
```
johncrossman: @paulkerschen do we care that the official spelling of "CalMail" has a capital em?

johncrossman: Feel free to disregard this concern. Renaming `module Calmail` would mean a lot of file renaming.

app/models/mailing_lists/calmail_list.rb
```diff
@@ -0,0 +1,109 @@
...
+        population_results[:add][:success] = population_results[:add][:total] - population_results[:add][:failure].count
+        population_results[:remove][:success] = population_results[:remove][:total] - population_results[:remove][:failure].count
+        self.members_count = addresses_after_update.count
+      else
+        self.members_count = population_results[:initial_count] + population_results[:add][:success] - population_results[:remove][:success]
+      end
+
+      logger.info "Added #{population_results[:add][:success]} of #{population_results[:add][:total]} new site members."
+      if population_results[:add][:failure].any?
+        logger.error "Failed to add #{population_results[:add][:failure].count} addresses to #{self.list_name}: #{population_results[:add][:failure].join(' , ')}"
```
johncrossman: @paulkerschen is `join(' , ')` right? Or remove space: `join(', ')`?  (Another instance is below.)

paulkerschen: @johncrossman What was I thinking?! Commas fixed.

db/developer-seed-data.sql
```diff
@@ -55,13 +61,86 @@ DROP SEQUENCE public.link_categories_id_seq;
...
+
+ALTER SEQUENCE canvas_site_mailing_list_members_id_seq OWNED BY canvas_site_mailing_list_members.id;
+
+--
+-- Name: canvas_site_mailing_lists; Type: TABLE; Schema: public; Owner: -; Tablespace:
+--
+
+CREATE TABLE canvas_site_mailing_lists (
+  id integer NOT NULL,
+  canvas_site_id character varying(255),
```
johncrossman: @paulkerschen is `canvas_site_id` nullable?

paulkerschen: @johncrossman Seems it shouldn't be, but this table has existed in production for a while (this part of the SQL is just bringing the seed data into sync). Do you think it's worth another migration to add the constraint?

johncrossman: @paulkerschen nah, migration not needed. We'll let future generations do our dirty work.

johncrossman:
@paulkerschen nice work. The squeaky wheels will be pleased.

PR #6031, opened 2016-10-19
SISRP-21920 - Downgrade enrollment limit calculation error to DEBUG level

redconfetti:
https://jira.berkeley.edu/browse/SISRP-21920

Just in case this fix doesn't make it into GoLive 7.2, we should stop logging in Production.

### DIFF
--- a/app/models/rosters/campus.rb
+++ b/app/models/rosters/campus.rb
@@ -59,11 +59,11 @@ module Rosters
             section_enrollments_open = section_enrollment_limit - section_enrolled_count
             section_waitlisted_open = section_waitlist_limit - section_waitlisted_count
             if (section_enrollments_open < 0)
-              logger.error "Section Enrollment limit exceeded in Section ID #{section[:ccn]}; Enrollment Count: #{section_enrolled_count}; Limit: #{section_enrollment_limit}"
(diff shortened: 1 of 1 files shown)


app/models/rosters/campus.rb
```diff
@@ -59,11 +59,11 @@ def get_feed_internal
             section_enrollments_open = section_enrollment_limit - section_enrolled_count
             section_waitlisted_open = section_waitlist_limit - section_waitlisted_count
             if (section_enrollments_open < 0)
-              logger.error "Section Enrollment limit exceeded in Section ID #{section[:ccn]}; Enrollment Count: #{section_enrolled_count}; Limit: #{section_enrollment_limit}"
+              logger.debug "Section Enrollment limit exceeded in Section ID #{section[:ccn]}; Enrollment Count: #{section_enrolled_count}; Limit: #{section_enrollment_limit}"
```
johncrossman: @redconfetti is there any value in seeing these log statements in prod?  If so, I recommend we log at WARN level.  Ops won't mind.

redconfetti: @johncrossman - I'm not sure which release this JIRA is going to make it in for, but currently in production the issue in SISRP-21920 is causing a bunch of noise in the logs. Currently it is expected that GSIs will have no `section_enrolled_count` value present because they are not associated with a primary section that has the count calculated. This results in a negative value seen by GSIs.

I'm making this PR to suppress those logs in the next release should this bug not be fixed. We can re-enable the logs once things change and the logged statement is actually an exceptional event.

PR #6035, opened 2016-10-20
SISRP-27043, instructor role-code (IRC) is APRX, not PROXY

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-6474
https://jira.berkeley.edu/browse/SISRP-27043

Misunderstanding in official code. Short name is `5-PROXY` but value received by CalCentral is `APRX`. This wrongness in CalCentral is causing Canvas refresh script to fail. See JIRAs for more info.

### DIFF
--- a/app/models/edo_oracle/adapters/common.rb
+++ b/app/models/edo_oracle/adapters/common.rb
@@ -1,6 +1,7 @@
 module EdoOracle
   module Adapters
     module Common
+      include ClassLogger
 
       def adapt_course_cntl_num(row)
         row['course_cntl_num'] = row['section_id'].to_s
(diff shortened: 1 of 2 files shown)


app/models/edo_oracle/adapters/common.rb
```diff
@@ -21,7 +21,7 @@ def adapt_instructor_func(row)
                                    when 'TNIC'  then '2'  # Teaching but Not in Charge, equivalent of Teaching but not Instructor of Record (2)
                                    when 'ICNT'  then '3'  # In Charge but Not Teaching, equivalent of Not teaching but Instructor of Record (3). Instructors coded as 3 must be accompanied by another "teaching" instructor coded as 2.
                                    when 'INVT'  then '4'  # Teaching with Invalid Title, equivalent of No Valid Teaching Title Code (4)
-                                   when 'PROXY' then '5'  # Head GSI, or similar, who assists the instructor(s) in charge with grading (5)
+                                   when 'APRX'  then '5'  # Head GSI, or similar, who assists the instructor(s) in charge with grading (5)
                                    else raise ArgumentError, "Unable to convert to 'instructor_func'. No such role code: '#{code}'"
```
paulkerschen: @johncrossman Do you think raising `ArgumentError` (and so aborting the entire upstream script) is the right thing to do here, or would it be better to log an error for the sharp eyes of Ops and return nil?

paulkerschen:
Looks good, @johncrossman. Will merge at T time.

johncrossman:
@paulkerschen  I just met w/ Ben and Rich.  They are going to talk to Barbara Sowden and/or people responsible for the role-code value.  IF they decide the easiest fix is to change the value from APRX to PROXY then this PR will be amended to ONLY include the "don't throw error" logic.

johncrossman:
@paulkerschen We are going to proceed w/ this PR, as is.

paulkerschen:
Proceeding.

PR #6036, opened 2016-10-20
SISRP-17569 - My Academics [course name] - Faculty Grading

nkutub:
https://jira.berkeley.edu/browse/SISRP-23851

This is to add the grading API call to populate 2 new columns on the teaching semesters card. This required a new model, changes to the myacademics merged feed and frontend changes to the teaching semesters card. Due to the lack of data sources related to the grading period and assistance link, those values are pulled from the YML settings. The code will handle null values for those settings. I also added logging if the date ranges for the grading period were set incorrectly.


app/models/my_academics/academics_module.rb
```diff
@@ -223,5 +223,12 @@ def decode_instruction_format(format)
         else format
       end
     end
+
+    def fetch_link(link_key, placeholders = {})
+      link = CampusSolutions::Link.new.get_url(link_key, placeholders).try(:[], :link)
+      logger.error "Could not retrieve CS link #{link_key} for Class #{self.class.name} feed, uid = #{@uid}" unless link
```
redconfetti: @nkutub We should change this to logger.debug. It is expected that sometimes the CS team will remove a link from the API with the expectation that the link will simply not display within the CalCentral user interface (front-end).

We should ensure that these links show up as `nil` when they do not match, and that the code expects that and appropriately displays nothing instead of raising a Javascript error.

spec/models/my_academics/grading_spec.rb
```diff
@@ -0,0 +1,199 @@
...
+      expect(subject.unexpected_cs_status?('GRD')).to eq false
+    end
+
+    it 'it should invalidate bad CS status' do
+      expect(subject.unexpected_cs_status?'BAD').to eq true
+    end
+  end
+
+  context 'when parsing CS status codes from grading API' do
+    let(:fake) { true }
```
redconfetti: Is there a situation where `fake` becomes `false`? Couldn't we move this one up to the top of the spec so it applies to them all, and remove the other one that duplicates it down on line 174?

src/assets/templates/academics_semester_classes.html
```diff
@@ -73,18 +73,28 @@ <h3 data-ng-if="selectedSemester.timeBucket === 'current'">Current</h3>
     <div class="cc-academics-subtitle">
       <h3>Teaching</h3>
     </div>
+    <div class="cc-academics-course-grading-header cc-widget-padding show-for-medium-up">
+      <span><a data-ng-href="{{selectedTeachingSemester.gradingAssistanceLink}}">■ Assistance with Grading</a></span>
```
redconfetti: @nkutub - Should we perhaps use `&#9632;` here instead of directly embedding the character? I don't have a lot of experience with Unicode characters, but I get the impression that they're more consistent when URL encoded like this. See [UTF-8 Geometric shapes](http://www.w3schools.com/charsets/ref_utf_geometric.asp)

redconfetti:
Other than my misc notes, it looks great.

app/models/my_academics/grading.rb
```diff
@@ -0,0 +1,219 @@
...
+          gradingAssistanceLink:  Settings.grading_period.assistance_link
+        })
+    end
+
+    def add_period_to_semester(semester, term_code)
+      # This is a temp fix for Fall 2016 hardcoded from settings
+      if term_code == '2168' && valid_grading_period?
+        semester.merge!(
+          {
+            gradingPeriodStart: Settings.grading_period.start.to_date.strftime("%b %d"),
```
nkutub: based on my discussion with Rachel this will be significantly different by next term's grading and we should not worry about that.

src/assets/templates/academics_semester_classes.html
nkutub: ya I will do that, I meant to look up the code

nkutub: done!

spec/models/my_academics/grading_spec.rb
nkutub: done!

app/models/my_academics/academics_module.rb
nkutub: done!

PR #6037, opened 2016-10-20
CLC-6467 Junction API endpoint to accept incoming messages from Mailgun

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6467

The controller's job is to: 
  - verify that incoming requests are signed with our Mailgun API key;
  - extract the params we care about and forward them to a model, which will check sender permissions and forward or bounce as needed.


app/controllers/mailing_lists_message_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+  # Verify Mailgun signature per https://documentation.mailgun.com/user_manual.html#webhooks
+  def verify_signature
+    digest = OpenSSL::Digest::SHA256.new
+    nonce = params.values_at('timestamp', 'token').join
+    params['signature'] == OpenSSL::HMAC.hexdigest(digest, Settings.mailgun_proxy.api_key, nonce)
+  end
+
+  # Timestamp must be reasonably close to the current time.
+  def verify_timestamp
+    (Time.now.to_i - params['timestamp'].to_i).abs <= Settings.canvas_mailing_lists.timestamp_tolerance
```
johncrossman: @paulkerschen what does `timestamp` represent?

app/models/mailing_lists/incoming_message.rb
```diff
@@ -0,0 +1,14 @@
+module MailingLists
+  class IncomingMessage
+
+    def initialize(opts)
```
johncrossman: @paulkerschen sanity check: we don't want `opts={}`?

app/controllers/mailing_lists_message_controller.rb
paulkerschen: @johncrossman It's the Unix timestamp of the request (see fixture below). Mailgun includes it as part of the generated signature so that we know it's a real signature generated now and not someone else sneakily reusing an old signature.

app/models/mailing_lists/incoming_message.rb
paulkerschen: @johncrossman Good call. Missing-opts handling will come in the next PR that fleshes out this class. :)

johncrossman:
Thanks, @paulkerschen  Btw, the NRA endorses your choice of Mailgun

PR #6038, opened 2016-10-20
[QA] SISRP-27043, instructor role-code (IRC) is APRX, not PROXY

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-6474
https://jira.berkeley.edu/browse/SISRP-27043

master PR: #6035 

The Canvas refresh scripts (in prod) are failing due to the `raise ArgumentError`. We might be workaround the problem by running scripts from an alternate dir. Worst case, this QA PR leads to a patch deploy.


paulkerschen:
Thanks @johncrossman.

PR #6041, opened 2016-10-21
CLC-6475, we risk OutOfMemoryError in scripts not using standard JVM arg setup

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-6475

New `script/standard-calcentral-JVM-OPTS-profile` to be sourced by scripts needing our default JVM/JRUBY args. In this JIRA, the OutOfMemoryError was hit by `sync-canvas-new-users.sh` but it could happen in other Canvas related scripts running nothing more than `export JRUBY_OPTS="--dev"`

In addition to standardization in many of our shell scripts, this PR does tidying which makes the scripts more readable and reliable.

### DIFF
--- a/script/calcentral-update.sh
+++ b/script/calcentral-update.sh
@@ -1,18 +1,28 @@
 #!/bin/bash
-# Script to run capistrano
+
+######################################################
+#
+# Run Capistrano
+#
+######################################################
 
 cd $( dirname "${BASH_SOURCE[0]}" )/..
 
 # Enable rvm and use the correct Ruby version and gem set.
-[[ -s "$HOME/.rvm/scripts/rvm" ]] && . "$HOME/.rvm/scripts/rvm"
-source .rvmrc
+[[ -s "${HOME}/.rvm/scripts/rvm" ]] && source "${HOME}/.rvm/scripts/rvm"
+source "${PWD}/.rvmrc"
 
 export RAILS_ENV=${RAILS_ENV:-production}
 export LOGGER_STDOUT=only
-export JRUBY_OPTS="--dev"
+
+# JVM args per CalCentral convention
+source "${PWD}/script/standard-calcentral-JVM-OPTS-profile"
 
 echo "------------------------------------------"
-echo "`date`: Redeploying CalCentral on app nodes..."
+echo "$(date): Redeploying CalCentral on app nodes..."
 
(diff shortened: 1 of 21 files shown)


script/calcentral-update.sh
```diff
@@ -1,18 +1,28 @@
...
-source .rvmrc
+[[ -s "${HOME}/.rvm/scripts/rvm" ]] && source "${HOME}/.rvm/scripts/rvm"
+source "${PWD}/.rvmrc"
 
 export RAILS_ENV=${RAILS_ENV:-production}
 export LOGGER_STDOUT=only
-export JRUBY_OPTS="--dev"
+
+# JVM args per CalCentral convention
+source "${PWD}/script/standard-calcentral-JVM-OPTS-profile"
```
johncrossman: The four lines above give an example of a stale `JRUBY_OPTS` assignment that I replaced with "profile" of our best practices.

script/start-torquebox.sh
```diff
@@ -27,24 +28,31 @@ export RAILS_ENV=${RAILS_ENV:-production}
...
-JBOSS_HOME=`bundle exec torquebox env jboss_home`
+JBOSS_HOME=$(bundle exec torquebox env jboss_home)
+
 cp ~/.calcentral_config/standalone-ha.xml ${JBOSS_HOME}/standalone/configuration/
 
-nohup bundle exec torquebox run -b ${IP_ADDR} -p=3000 --jvm-options="${JVM_OPTS}" --clustered --max-threads=${MAX_THREADS} < /dev/null >> ${TORQUEBOX_LOG} 2>> ${LOG} &
+# Set JVM args per CalCentral convention
+source "${PWD}/script/standard-calcentral-JVM-OPTS-profile"
+
+nohup bundle exec torquebox run -b ${IP_ADDR} -p=3000 --jvm-options="\-server \-verbose:gc ${ESCAPED_JVM_OPTS}" --clustered --max-threads=${MAX_THREADS} < /dev/null >> ${TORQUEBOX_LOG} 2>> ${LOG} &
```
johncrossman: JVM args passed via TorqueBox command line switch require the escaping of dashes. Thus we have `ESCAPED_JVM_OPTS` which inherits from standard JVM_OPTS.

paulkerschen:
Thanks @johncrossman. I always learn something about Bash reviewing these.

PR #6042, opened 2016-10-21
SISRP-26828 - 7.2 - Academic Summary on My Academics (Student)

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-26828
Design JIRA:  https://jira.berkeley.edu/browse/SISRP-26096

* Unfortunately, I had to add even more complexity to the front-end of `academic_semesters.html`.  Definitely needs some clean-up in more peaceful times.
* There is a lot of transition term verbiage on the student profile card that I think is related to bug [SISRP-26943](https://jira.berkeley.edu/browse/SISRP-26943).  I left it in in this PR but I think they should be pulled out in this current sprint (on my to-do list).


redconfetti:
@davidlimcheng - It really seems like we have to let these templates become complex until we see what kind of pattern emerges so that we can better organize them.

LGTM

PR #6044, opened 2016-10-21
SISRP-27057 - CalCentral Prod: EFT status & EFT account info not displaying

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-27057

* `#get_internal` was getting hit, and returning `{}` because `@campus_solutions_id` would not be available yet.  `@campus_solutions_id` becomes available when `#get_parsed_response` calls `#lookup_student_id`
* So, we were not making the API call, and sending the front-end an empty object.  I've added some logging to help diagnose this, and moved the call to `#lookup_student_id` to the correct place.  Also added a nil-check security blanket for the front-end.
* It's perplexing because I'm not sure why this was previously working.  It shouldn't have - the only thing I can think of that might have changed are @raydavis's recent changes to the caching mechanisms.  But even tracing those thru their histories, I couldn't find any bypass of `#get_internal` or any trace of `@campus_solutions_id` being made available earlier.  If only I could ask younger, less experienced me.

### DIFF
--- a/app/models/eft/my_eft_enrollment.rb
+++ b/app/models/eft/my_eft_enrollment.rb
@@ -20,19 +20,19 @@ module Eft
       "Failed to connect with EFT system"
     end
 
-    def lookup_student_id
-      if @uid
-        @campus_solutions_id = CalnetCrosswalk::ByUid.new(user_id: @uid).lookup_campus_solutions_id
-      end
+    def campus_solutions_id
+      @campus_solutions_id ||= CalnetCrosswalk::ByUid.new(user_id: @uid).lookup_campus_solutions_id
     end
 
     def get_feed_internal
-      return {} if @campus_solutions_id.nil? || !Settings.features.cs_billing
+      if campus_solutions_id.nil?
+        logger.warn "No Campus Solutions ID found for UID #{@uid}"
+        return {}
+      end
       get_parsed_response
     end
 
     def get_parsed_response
-      lookup_student_id
       url = "#{@settings.base_url}eft_status?studentId=#{@campus_solutions_id}"
       logger.info "get_parsed_response: Fake = #{@fake}; Making request to #{url} on behalf of user #{@uid}; cache expiration #{self.class.expires_in}"
       response = get_response(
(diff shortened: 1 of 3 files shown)


app/models/eft/my_eft_enrollment.rb
```diff
@@ -21,18 +21,19 @@ def default_message_on_exception
     end
 
     def lookup_student_id
-      if @uid
-        @campus_solutions_id = CalnetCrosswalk::ByUid.new(user_id: @uid).lookup_campus_solutions_id
-      end
+      @campus_solutions_id = CalnetCrosswalk::ByUid.new(user_id: @uid).lookup_campus_solutions_id
```
redconfetti: What if we just made this lazy load the ID and return it directly?
```
def campus_solutions_id
  @cs_user_id ||= CalnetCrosswalk::ByUid.new(user_id: @uid).lookup_campus_solutions_id
end
```

Then further down in #get_feed_internal you could just use:
```
if campus_solutions_id.nil?
  ...
end
```

davidlimcheng: @redconfetti  Good stuff!  Will amend.  Love comments like these that teach me how to code a bit more elegantly.

app/models/eft/my_eft_enrollment.rb
```diff
@@ -20,19 +20,19 @@ def default_message_on_exception
...
-      end
+    def campus_solutions_id
+      @campus_solutions_id ||= CalnetCrosswalk::ByUid.new(user_id: @uid).lookup_campus_solutions_id
     end
 
     def get_feed_internal
-      return {} if @campus_solutions_id.nil? || !Settings.features.cs_billing
+      if campus_solutions_id.nil?
+        logger.warn "No Campus Solutions ID found for UID #{@uid}"
+        return {}
```
redconfetti: No. I don't think you need `errorMessage: `. It already shows up in the logs like:
```
[2016-10-24 09:56:34] [WARN] [CalCentral] No Campus Solutions ID found for UID 12345
```

That's enough information, no need to prefix it.

PR #6045, opened 2016-10-21
SISRP-22142 - AA/UX: Advising Card (Enhancement)

redconfetti:
https://jira.berkeley.edu/browse/SISRP-26566

### DIFF
--- a/app/models/advising/my_advising.rb
+++ b/app/models/advising/my_advising.rb
@@ -7,6 +7,17 @@ module Advising
     include Cache::UserCacheExpiry
     include DatedFeed
 
+    RELATIONSHIP_ORDER = [
+      'GSAO', # Graduate Student Affairs Offcr
+      'HGA',  # Head Graduate Advisor
+      'CHR',  # Department Chair
+      'GEA',  # Graduate Equity Advisor
+      'GRAD', # Grad Div Representative
+      'COLL', # College Advisor
+      'MAJ',  # Major Advisor
+      'MIN',  # Minor Advisor
+    ]
+
     def get_feed_internal
       advising_feed = {
         feed: {},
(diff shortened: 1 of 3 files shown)


davidlimcheng:
LGTM, @redconfetti

app/models/advising/my_advising.rb
```diff
@@ -41,8 +52,21 @@ def merge_action_items(advising_feed)
...
+      end
+    end
+
+    def sort_advisors(advisors)
+      return if advisors.nil?
+      sorted_advisors = []
+      grouped_advisors = advisors.group_by { |advisor| advisor[:assignedAdvisorTypeCode] }
+      RELATIONSHIP_ORDER.each do |type|
+        matched_advisors = grouped_advisors.delete(type).to_a
+        sorted_advisors.concat(matched_advisors)
```
lyttam: @davidlimcheng Maybe it's superstitious but deleting things from arrays makes me a little nervous.  You could do it like this (and replace two lines with one):

`sorted_advisors.concat(grouped_advisors[type].to_a)`

spec/models/advising/my_advising_spec.rb
```diff
@@ -98,6 +98,182 @@
       end
     end
 
+    context 'graduate advisor relationships' do
+      let(:graduate_advisor_relationships_xml) do
+        <<-XML
```
lyttam: @davidlimcheng why not move this XML to /fixtures?

app/models/advising/my_advising.rb
redconfetti: @lyttam - This was my PR, not @davidlimcheng 's.

I certainly wouldn't use `delete` on an array that I'm iterating through. In this case I'm deleting each group that is matched from the array, and concatenating them to my sorted list. I do this so that the ones that are grouped but not matched with the types in `RELATIONSHIP_ORDER` can still be included in the sorted array, in whatever order at the end of the sorted array.

spec/models/advising/my_advising_spec.rb
redconfetti: @lyttam - That's a good question. I wonder if we can introduce an Rspec helper to obtain fixtures for use in the tests. I created an article in Confluence on [Mocking](https://confluence.ets.berkeley.edu/confluence/display/MYB/Misc+Notes#MiscNotes-Mocking), which explains that the fixtures are pulled in by the proxies themselves from the fixtures when the proxy is operating in fake mode (i.e. `{fake: true, user_id: '12345'}`).

There is a one-to-one relationship between the proxy and it's default fixture in fake mode, but we can use the `#set_response` method on a proxy to mock a different response OTHER than the default fixture. That is what I'm doing here. I don't like bringing all this XML into the spec, so maybe I should look for a way to move this into `/fixtures` and load it by filename.

app/models/advising/my_advising.rb
lyttam: Ohhh whoops, sorry for the mixup!  So many PRs have crossed my desk as of late.

spec/models/advising/my_advising_spec.rb
redconfetti: @lyttam - I've applied a fix for this in #6066

PR #6046, opened 2016-10-21
SISRP-25362 - Adds Degree Progress (GRAD) card to Student Overview

lyttam:

Advisors will see the Degree Progress card when they do Student Lookup on a GRAD or LAW student.
I added a separate feature flag for the advisor version of the card.  The advisor version is planned for 7.2 and the student version is planned for 7.3+ at the moment.

This is the UI side of the work; the backend work was reviewed under https://github.com/ets-berkeley-edu/calcentral/pull/6027.

### DIFF
--- a/app/models/degree_progress/graduate_milestones.rb
+++ b/app/models/degree_progress/graduate_milestones.rb
@@ -5,16 +5,20 @@ module DegreeProgress
     include Cache::CachedFeed
     include Cache::JsonifiedFeed
     include Cache::UserCacheExpiry
-    include CampusSolutions::DegreeProgressFeatureFlagged
+    include CampusSolutions::DegreeProgressGradAdvisingFeatureFlagged
     include MilestonesModule
 
     def get_feed_internal
-      return {} unless is_feature_enabled
+      return {} unless is_feature_enabled && authorized?
       response = CampusSolutions::DegreeProgress::GraduateMilestones.new(user_id: @uid).get
       response[:feed] = HashConverter.camelize({
         degree_progress: process(response),
       })
       response
+    end
+
+    def authorized?
+      authentication_state.policy.can_view_as?
     end
   end
 end
--- a/app/models/degree_progress/my_graduate_milestones.rb
+++ b/app/models/degree_progress/my_graduate_milestones.rb
@@ -5,7 +5,7 @@ module DegreeProgress
     include Cache::CachedFeed
     include Cache::JsonifiedFeed
     include Cache::UserCacheExpiry
-    include CampusSolutions::DegreeProgressFeatureFlagged
+    include CampusSolutions::DegreeProgressGradStudentFeatureFlagged
     include MilestonesModule
 
     LINKS_CONFIG = [
@@ -14,7 +14,7 @@ module DegreeProgress
     ]
 
     def get_feed_internal
-      return {} unless is_feature_enabled
+      return {} unless is_feature_enabled && authorized?
       response = CampusSolutions::DegreeProgress::GraduateMilestones.new(user_id: @uid).get
       response[:feed] = HashConverter.camelize({
         degree_progress: process(response),
@@ -38,6 +38,10 @@ module DegreeProgress
       end
       logger.error "Could not retrieve CS link #{link_key} for Degree Progress feed, uid = #{@uid}" unless link
       link
+    end
+
+    def authorized?
+      authentication_state.policy.can_view_as? || authentication_state.policy.graduate_student? || authentication_state.policy.law_student?
     end
   end
 end
--- a/app/policies/authentication_state_policy.rb
+++ b/app/policies/authentication_state_policy.rb
@@ -87,4 +87,15 @@ class AuthenticationStatePolicy
     real_auth.is_superuser? || has_advisor_role
   end
 
+  def graduate_student?
+    return false unless @user.real_user_auth.active?
+    attributes = User::AggregatedAttributes.new(@user.user_id).get_feed
+    attributes.try(:[], :roles).try(:[], :graduate)
+  end
+
+  def law_student?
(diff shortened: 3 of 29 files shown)


script/sis/verify_sis_api_endpoints.sh
```diff
@@ -277,7 +277,10 @@ verify_cs 'cs_advisor_student_lookup' "${yml_features_cs_advisor_student_lookup}
 verify_cs 'cs_profile_emergency_contacts' "${yml_features_cs_profile_emergency_contacts}" \
   "/UcApiEmergencyContactGet.v1/?EMPLID=${CAMPUS_SOLUTIONS_ID}"
 
-verify_cs 'cs_degree_progress' "${yml_features_cs_degree_progress}" \
+verify_cs 'cs_degree_progress_advising' "${yml_features_cs_degree_progress_advising}" \
```
redconfetti: 👍

redconfetti:
@lyttam - Does the card only show up for GRAD and LAW students because the API only returns data for those students? I expected to see a check in the code to restrict this to only GRAD or LAW students somehow.

lyttam:
@redconfetti Logic in the HTML prevents the card from being displayed if the student is not GRAD or LAW - but I will add a check in the server-side code as well.

app/models/degree_progress/graduate_milestones.rb
```diff
@@ -5,16 +5,20 @@ class GraduateMilestones < UserSpecificModel
     include Cache::CachedFeed
     include Cache::JsonifiedFeed
     include Cache::UserCacheExpiry
-    include CampusSolutions::DegreeProgressFeatureFlagged
+    include CampusSolutions::DegreeProgressAdvisingFeatureFlagged
     include MilestonesModule
 
     def get_feed_internal
-      return {} unless is_feature_enabled
+      return {} unless is_feature_enabled && authorized?
```
lyttam: @redconfetti I added auth checks here...

app/models/degree_progress/my_graduate_milestones.rb
```diff
@@ -14,7 +14,7 @@ class MyGraduateMilestones < UserSpecificModel
     ]
 
     def get_feed_internal
-      return {} unless is_feature_enabled
+      return {} unless is_feature_enabled && authorized?
```
lyttam: @redconfetti ...and here

app/policies/authentication_state_policy.rb
```diff
@@ -87,4 +87,12 @@ def can_view_other_user_photo?
     real_auth.is_superuser? || has_advisor_role
   end
 
+  def doctoral_student?
```
davidlimcheng: @lyttam If a student pursuing an M.S. is also classified as a `graduate` student, is the name of this method a bit misleading?  I would imagine `doctoral_student` would only want students pursuing a PhD.

lyttam: @davidlimcheng Yeah totally. I've been trying to think of a name that includes PhD, MS, and LAW career. Should I just call it `graduate_student` then?

davidlimcheng: @lyttam Yeah, I think that could reduce some confusion.  I could totally see myself using `#doctoral_student?` in the future for the wrong reasons.

redconfetti: @davidlimcheng , would you agree with the use of `post_undergraduate_student?` as a method name?

redconfetti:
NOTE: I think we need to merge this before #6052

app/policies/authentication_state_policy.rb
davidlimcheng: Yeah that works too - I'd be happy with either.

lyttam: @redconfetti @davidlimcheng Or should I just split it into two functions?  `graduate_student?` and `law_student?`.  Might be more reusable that way.

davidlimcheng: @lyttam I like it.  No confusion here.

lyttam: @davidlimcheng @redconfetti done and done!

config/routes.rb
```diff
@@ -159,6 +159,7 @@
   get '/api/advising/student_attributes/:student_uid' => 'advising_student#student_attributes', :defaults => { :format => 'json' }
   get '/api/advising/advising/:student_uid' => 'advising_student#advising', :defaults => { :format => 'json' }
   get '/api/advising/student_success/:student_uid' => 'advising_student#student_success', :defaults => { :format => 'json' }
+  get '/api/advising/degree_progress/grad/:student_uid' => 'advising_student#degree_progress_graduate', :defaults => { :format => 'json' }
```
lyttam: Changed grd to grad for @davidlimcheng :)

app/models/degree_progress/my_graduate_milestones.rb
```diff
@@ -39,5 +39,9 @@ def fetch_link(link_key)
       logger.error "Could not retrieve CS link #{link_key} for Degree Progress feed, uid = #{@uid}" unless link
       link
     end
+
+    def authorized?
+      authentication_state.policy.can_view_as? || authentication_state.policy.graduate_student? || authentication_state.policy.law_student?
```
raydavis: @lyttam , @redconfetti : A tardy reminder that advisors will fail the the ``can_view_as?`` check.

More generally, an undergraduate calling this API on themselves doesn't really seem like a permissions issue. And permission checks should usually be confined to the controller level if at all possible. This is more of a filter.

lyttam: @raydavis That's the behavior we want in this case, I think - only the student or delegate should see the card on My Academics.  Advisors can see a slightly different version of the card on Student Overview.

PR #6047, opened 2016-10-21
CLC-6475, revert to 'source' on relative path in calcentral-update

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-6475

Prior to PR #6041, we had `source .rvmrc`. In my preference for absolute paths, I changed it to `source "${PWD}/.rvmrc"`.  That might have made Bamboo unhappy: https://bamboo.ets.berkeley.edu/bamboo/browse/CLCD-CQDC

After this merge, if Bamboo passes then I'll restore `source .rvmrc` (relative path) on all scripts.


paulkerschen:
Pwn'd by PWD!

PR #6054, opened 2016-10-25
CLC-6480 bCourses mailing lists: tweak admin tool to handle Mailgun lists

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6480

The Mailgun workflow is simpler than the CalMail workflow. CalMail lists are first "registered" in the Junction database and then created in CalMail; but Mailgun lists live only in Junction and can be created in one shot.

A couple of methods are moved into the MailingLists::SiteMailingList superclass to avoid errors when we first instantiate a placeholder list without yet knowing what subclass it will be.

### DIFF
--- a/app/models/mailing_lists/mailgun_list.rb
+++ b/app/models/mailing_lists/mailgun_list.rb
@@ -9,10 +9,6 @@ module MailingLists
       Settings.mailgun_proxy.domain
     end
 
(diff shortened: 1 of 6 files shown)


johncrossman:
Pulling the trigger on Mailgun. Thanks @paulkerschen

PR #6056, opened 2016-10-25
CLC-6467 bCourses mailing lists: stay out of ActionController's vocabulary

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6467

End-to-end testing immediately proves its worth! The Rails controller has its own uses for `dispatch` and doesn't like having its toes stepped on.


johncrossman:
dispatched `dispatch`   Thx @paulkerschen

PR #6058, opened 2016-10-25
SISRP-27220 - My Academics (Student) - Advising -"Access and Manage Appointments" link is not working

redconfetti:
https://jira.berkeley.edu/browse/SISRP-27220

* Fixes broken link for "Access and Manage My Appointments"
* Fixes missing `ucFromText` issue with links
* Fixes broken spinner for Advising card

### DIFF
--- a/src/assets/templates/widgets/my_advising.html
+++ b/src/assets/templates/widgets/my_advising.html
@@ -5,7 +5,7 @@
       data-ng-if="api.user.profile.features.csAdvisingLinks && myAdvising.links && myAdvising.roles.ugrdUrbanStudies"
       data-cc-campus-solutions-link-directive="myAdvising.links.newAppointment.url"
       data-cc-campus-solutions-link-directive-enabled="{{myAdvising.links.newAppointment.isCsLink}}"
-      data-cc-campus-solutions-link-directive-text="myAdvising.links.newAppointment.backToText"
+      data-cc-campus-solutions-link-directive-text="myAdvising.backToText"
       data-cc-campus-solutions-link-directive-cache="'myAdvising'"
       data-cc-outbound-enabled="false"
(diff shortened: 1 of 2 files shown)


src/assets/templates/widgets/my_advising.html
```diff
@@ -15,7 +15,7 @@ <h2 class="cc-left">Advising</h2>
     </a>
   </div>
 
-  <div data-cc-spinner-directive="myAdvising.isLoading()">
+  <div data-cc-spinner-directive="myAdvising.isLoading">
```
redconfetti: It turns out that our spinner directive is designed to pass a value by reference, not watch the returned value of a function. If someone can recommend a better approach with AngularJS, I'm open.

redconfetti: I tried to update the spinnerDirective to support either a variable with a value, or a function, and it just wasn't working as I expected. I tried many different variations

paulkerschen: @redconfetti If you need to change the `isLoading` value once all requests in the controller have completed, would the `$q.all(...).then` pattern work for you?

https://github.com/ets-berkeley-edu/calcentral/blob/3735e2358836fb6aa08bbe513c9a49b5bbf952a8/src/assets/javascripts/angular/controllers/widgets/statusController.js#L222

redconfetti: Aaah @paulkerschen, thank you for pointing that out. I could group my requests, then use a single 'then' call to mark isLoading as `false`.

redconfetti: @paulkerschen - I applied the change you suggested. It works great, and I'm happy that the code is more concise and easier to maintain now. Thanks again for that tip.

PR #6059, opened 2016-10-25
CLC-6481 Exclude inactive accounts from bCourses DB queries

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6481

Filtering them after the SQL query leads to bogus null results or peculiar result counts.

### DIFF
--- a/app/models/campus_oracle/queries.rb
+++ b/app/models/campus_oracle/queries.rb
@@ -87,6 +87,7 @@ module CampusOracle
                   count(*) over() result_count
           from calcentral_person_info_vw pi
           where lower( concat(concat(trim(pi.last_name), ','), trim(pi.first_name)) ) like '#{clean_search_string.downcase}%'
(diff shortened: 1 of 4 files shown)


app/models/campus_oracle/queries.rb
```diff
@@ -116,6 +117,7 @@ def self.find_people_by_email(email_search_string, limit = 0)
                   count(*) over() result_count
           from calcentral_person_info_vw pi
           where lower(pi.email_address) like '%#{clean_search_string.downcase}%'
+            and (affiliations LIKE '%-TYPE-%') and (person_type != 'Z')
```
johncrossman: @raydavis this looks fine to me. No need for `pi.` prefix on `affiliations` and `person_type`.

paulkerschen:
Looks good to me too, @raydavis, thanks for tightening the valve.

PR #6064, opened 2016-10-26
SISRP-26969 - Continued refactoring as response to EDO Hub Academic Status API changes

redconfetti:

This is an update I made to resolve a bug that was occurring with the iHub Academic Status API response. I made this just in case we'd need to filter out inactive academic statuses because the bug hadn't been fixed.

I'm not sure if this should go in with the release. @raydavis, perhaps we should discuss this. It's a defensive update, which I did complete. I'm just wanting to give this work a chance.


redconfetti:
This PR was a continuation of work done in #6012, thus bringing the filtering of "inactive" academic statuses out of the `#parse_hub_plans` method into the `#hub_college_and_level` status loop, right after they've been extracted from the iHub response.

We're going to close this out because the iHub Academic Status API was updated to no longer return these statuses. I will make a follow up PR to remove the `#active_academic_statuses` method, and the call to it from within `#parse_hub_plans`.

PR #6065, opened 2016-10-26
SISRP-26579 Limit size of advisor student lookups

raydavis:
https://jira.berkeley.edu/browse/SISRP-26579

Unconstrained name searches can take more than a half-hour to complete, mostly due to the poor performance of data fetches from CS Hub.


paulkerschen:
"Davis... I think the student's name was Davis. Hmm."

PR #6068, opened 2016-10-26
SISRP-26969 - Remove inactive status filtering from MyAcademics::CollegeAndLevel

redconfetti:

Replacement for #6064. Removes application of code that filters out inactive academic statuses now that the iHub Academic Status API should no longer return academic statuses that have no active plans present.

### DIFF
--- a/spec/models/my_academics/college_and_level_spec.rb
+++ b/spec/models/my_academics/college_and_level_spec.rb
@@ -718,24 +718,6 @@ describe MyAcademics::CollegeAndLevel do
       })
     end
 
-    context 'when filtering out inactive academic statuses' do
-      let(:hub_academic_statuses) { [hub_academic_status, hub_academic_status_secondary] }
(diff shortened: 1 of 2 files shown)


spec/models/my_academics/college_and_level_spec.rb
```diff
@@ -718,24 +718,6 @@
...
-      end
-      it 'returns academic statuses that have active plans present' do
-        expect(active_statuses.count).to eq 1
-        expect(active_statuses[0]['studentCareer']['academicCareer']['code']).to eq 'UGRD'
-      end
-    end
-
     context 'when filtering out inactive plans from statuses' do
       let(:filtered_academic_statuses) { subject.filter_inactive_status_plans(hub_academic_statuses) }
       it 'removes inactive plans from each status' do
```
davidlimcheng: @redconfetti Do we still need this test?

PR #6071, opened 2016-10-27
CLC-6478, GoogleAuthController variant to manage OEC-specific Google tokens

johncrossman:

This is the first half of the CLC-6478 story. The logic of `GoogleAuthController` was moved to a the reusable model `GoogleApps::Oauth2TokensGrant`. Next, a dedicated OEC controller was added, with stricter access rules, that allows the same kind of token management.

A lot more test coverage.

The next step will be removal OEC's client_id and client_secret from YAML.


config/routes.rb
```diff
@@ -99,9 +99,13 @@
   post '/api/academics/canvas/mailing_lists/:canvas_course_id/delete' => 'canvas_mailing_lists#destroy', :defaults => { :format => 'json' }
 
   # OEC endpoints
-  get '/api/oec_tasks' => 'oec_tasks#index', :defaults => { :format => 'json' }
-  post '/api/oec_tasks/:task_name' => 'oec_tasks#run', :defaults => { :format => 'json' }
-  get '/api/oec_tasks/status/:task_id' => 'oec_tasks#task_status',  :defaults => { :format => 'json' }
+  get '/api/oec/google/request_authorization'=> 'oec_google_auth#refresh_tokens', :defaults => { :format => 'json' }
+  get '/api/oec/google/handle_callback' => 'oec_google_auth#handle_callback', :defaults => { :format => 'json' }
+  get '/api/oec/google/current_scope' => 'oec_google_auth#current_scope', :defaults => { :format => 'json' }
+  get '/api/google/remove_authorization' => 'oec_google_auth#remove_authorization', :via => :post
```
paulkerschen: @johncrossman 1) should this be `/api/oec/google/remove_authorization`? 2) should it start with `post`?

app/controllers/oec_tasks_controller.rb
```diff
@@ -37,7 +37,7 @@ def run
     }
   end
 
-  # GET /api/oec_tasks/status/:task_id
+  # GET /api/oec/tasks/status/:task_id
```
paulkerschen: +1 on route changes

app/controllers/oec_google_auth_controller.rb
```diff
@@ -0,0 +1,53 @@
+class OecGoogleAuthController < ApplicationController
```
paulkerschen: @johncrossman If I'm reading this class right, it seems to share most of its logic with GoogleAuthController, apart from authentication and a couple of variant settings. Would it make sense to just have it inherit from GoogleAuthController?

johncrossman: @paulkerschen Yes, agreed.

config/routes.rb
johncrossman: Good catch, @paulkerschen

johncrossman:
Thank @paulkerschen   I accidentally pushed a redundant commit to this PR so I'm closing.  A cleaner PR is at #6073

PR #6072, opened 2016-10-27
CLC-6471 bCourses mailing lists: fix attachment handling

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6471

Another try at attachment handling. Local testing had encouraged a bad assumption about how Rails represents multipart uploads (custom objects, not hashes).


johncrossman:
Looks like a fix to me. Thx, @paulkerschen

PR #6073, opened 2016-10-27
CLC-6478, GoogleAuthController variant to manage OEC-specific Google tokens

johncrossman:
This PR is a cleaner version of #6071 (closed)

This is the first half of the CLC-6478 story. The logic of `GoogleAuthController` was moved to a the reusable model `GoogleApps::Oauth2TokensGrant`. Next, a dedicated OEC controller was added, with stricter access rules, that allows the same kind of token management.

A lot more test coverage.

The next step will be removal OEC's client_id and client_secret from YAML.

### DIFF
--- a/app/controllers/google_auth_controller.rb
+++ b/app/controllers/google_auth_controller.rb
@@ -1,119 +1,63 @@
-require 'google/api_client'
-
 class GoogleAuthController < ApplicationController
-  include ClassLogger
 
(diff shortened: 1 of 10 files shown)


johncrossman:
@paulkerschen's comments in #6071 have been addressed.

paulkerschen:
Thanks @johncrossman, looking good.

PR #6075, opened 2016-10-27
SISRP-25377 - Retires the University Requirements card

lyttam:
https://jira.berkeley.edu/browse/SISRP-25377


davidlimcheng:
@lyttam I'm a bit confused by the timing of this - Wouldn't we want to retire the University Requirements card at the same time we introduce the Degree Progress card, which is in 7.3?  Or is the BearFacts data obsolete, so this card already provides no value?

lyttam:
@davidlimcheng Good question.  The advising team wants University Requirements retired because the data wouldn't be accurate.  Advisors will get the Degree Progress card in 7.2, and students will be left high and dry for now I guess.

PR #6080, opened 2016-10-28
CLC-6478, eliminate need for Google access- and refresh-tokens in YAML file (OEC)

johncrossman:

Core classes in GoogleApps module now rely on `app_id` to determine which client_id, etc. to use.

### DIFF
--- a/app/models/google_apps/client.rb
+++ b/app/models/google_apps/client.rb
@@ -7,7 +7,12 @@ module GoogleApps
 
     class << self
       def client
-        @client ||= Google::APIClient.new(options={:application_name => "CalCentral", :application_version => "v1", :auto_refresh_token => true, :retries => 3})
+        @client ||= Google::APIClient.new(options={
+          application_name: 'CalCentral',
+          application_version: 'v1',
+          auto_refresh_token: true,
+          retries: 3
+        })
       end
 
       def discover_resource_method(api, api_version, resource, method)
@@ -19,12 +24,12 @@ module GoogleApps
         end
       end
 
-      def new_fake_auth
-        new_auth("fake_access_token")
+      def new_fake_auth(app_id)
+        new_auth(app_id, 'fake_access_token')
       end
 
-      def new_client_auth(token_hash)
-        new_auth(token_hash["access_token"], token_hash)
+      def new_client_auth(app_id, options={})
+        new_auth(app_id, options[:access_token], options)
       end
 
       def generate_request_hash(page_params)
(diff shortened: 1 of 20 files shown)


app/models/google_apps/client.rb
```diff
@@ -19,12 +24,12 @@ def discover_resource_method(api, api_version, resource, method)
...
-      def new_fake_auth
-        new_auth("fake_access_token")
+      def new_fake_auth(app_id)
+        new_auth(app_id, 'fake_access_token')
       end
 
-      def new_client_auth(token_hash)
-        new_auth(token_hash["access_token"], token_hash)
+      def new_client_auth(app_id, options={})
+        new_auth(app_id, options[:access_token], options)
```
johncrossman: @paulkerschen no more stringified keys is intentional. You'll notice `symbolize_keys` added in `User::Oauth2Data.get`

app/models/google_apps/credential_store.rb
```diff
@@ -1,74 +1,59 @@
...
-          logger.debug "OAuth tokens in OAuth2 instance, where UID=#{@uid}"
-          credentials = update_tokens(auth.access_token, auth.refresh_token, auth.issued_at, auth.expires_in, @options)
-        end
-      end
-      logger.warn "Google OAuth tokens not found during attempted database update. UID=#{@uid}; auth.class #{auth.class}" if credentials.nil?
-      credentials
+    # Do not change the signature of this method because it is invoked by Google::APIClient
+    def write_credentials(credentials)
+      credentials.symbolize_keys!
+      write(credentials[:access_token], credentials[:refresh_token], credentials)
```
johncrossman: @paulkerschen I need to amend the above to add in `auth.is_a?(Signet::OAuth2::Client)` check.  Stand by.

app/models/google_apps/credential_store.rb
```diff
@@ -2,73 +2,70 @@ module GoogleApps
...
-      expiration_time = issued_at.blank? || expires_in.blank? ? 0 : issued_at.to_i + expires_in.to_i
-      logger.debug "Put OAuth tokens to db where UID=#{@uid}"
-      User::Oauth2Data.new_or_update(@uid.to_s, GoogleApps::Proxy::APP_ID, access_token, refresh_token,
-                                     expiration_time, options)
     end
 
-    def load_credentials_from_db
-      client_id = @options[:client_id] || Settings.google_proxy.client_id
-      client_secret = @options[:client_secret] || Settings.google_proxy.client_secret
-      scope = @options[:scope] || Settings.google_proxy.scope
```
johncrossman: @paulkerschen we can rely on`app_id` to take us to the proper `client_id` and `client_secret`; we no longer need to pass them via `@options`. That said, we _should_ continue to support a custom `:expiration_time` (as seen below in write() method. I'll restore the passing of `@opts` for that purpose. Amendment coming.

app/models/google_apps/credential_store.rb
```diff
@@ -2,73 +2,73 @@ module GoogleApps
...
+        write(auth[:access_token], auth[:refresh_token], opts)
+      elsif auth.is_a?(Signet::OAuth2::Client) && auth.access_token && auth.refresh_token
+        logger.debug "OAuth tokens in #{auth.class} (app_id: #{@app_id}; uid: #{@uid})"
+        opts = @opts.merge({
+          issued_at: auth.issued_at,
+          expires_in: auth.expires_in
+        })
+        write(auth.access_token, auth.refresh_token, opts)
+      else
+        raise ArgumentError, "Signet::OAuth2 is missing tokens OR we have unsupported type of OAuth2 client auth: #{auth}"
```
johncrossman: I added: `... else raise error...` when valid auth properties not found.

app/models/google_apps/proxy.rb
```diff
@@ -12,23 +11,21 @@ class Proxy < BaseProxy
...
+        when APP_ID then Settings.google_proxy
+        else nil
       end
+    end
+
+    def initialize(options = {})
+      @app_id = options[:app_id] || APP_ID
+      super(Proxy.config_of(@app_id), options)
 
+      @authorization = load_authorization options
```
johncrossman: Our `initialize()` is made simpler by moving the big `if ... elsif ... else ...` (above) to the private `load_authorization()` method below. No logic change here.

paulkerschen:
Thanks for the explanations, @johncrossman. Let's see how it does!

PR #6082, opened 2016-10-28
CLC-6489, DriveManager might pass non-symbol keys to CredentialStore

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-6489

Two things:
1. We guarantee symbolized hash keys prior to `write_credentials`.
1. Fix 'real' rspec: `load_credentials` should not expect  custom override of expiration_time. It's purely a read operation.


johncrossman:
@paulkerschen would you mind reviewing this follow-up PR w/ slight mod.

paulkerschen:
Sneaky keys. Looks good @johncrossman.

PR #6089, opened 2016-10-31
CLC-6493, UI should report error if no Google OAuth tokens registered to UID of OEC

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-6493

UI will report, "...features are unavailable because no Google OAuth tokens are registered with UID 12004714. Please report this problem." Thus, less guesswork for our beleaguered OEC admin.


paulkerschen:
Friendly! Thanks, @johncrossman, waiting on T.

PR #6090, opened 2016-10-31
SISRP-26895 - AA/UX: Student Success Card (Future Term)

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-26895

* Spoke with the API team, and had to change the API itself (breaking contract) to accept new values.
* Now we use the term-specific GPA (as opposed to running cumulative).
* Since a student can take all P/NP or Withdraw from all classes in a term, we don't want those terms affecting the GPA trendline - so we `nil` them out.

QA PR to come upon merge.


redconfetti:
@davidlimcheng Looks good to me

PR #6092, opened 2016-10-31
SISRP-26980 SISRP-26578 Trust CS on active Student affiliation; CalNet changes

raydavis:
https://jira.berkeley.edu/browse/SISRP-26980
https://jira.berkeley.edu/browse/SISRP-26578

- CS is now processing Student affiliation changes in a timely fashion.
- CalNet is switching from "XYZ-STATUS-EXPIRED" to "FORMER-XYZ" affiliations.
- CalNet no longer routinely returns apparent conflicts. The old expdate logic is unneeded.
- Revised rspec mocks to reflect the sparse CS role hashes.

### DIFF
--- a/app/models/berkeley/user_roles.rb
+++ b/app/models/berkeley/user_roles.rb
@@ -1,13 +1,15 @@
 module Berkeley
   module UserRoles
     extend self
+    include ClassLogger
 
     def roles_from_affiliations(affiliations)
       affiliations ||= []
       {
         :student => affiliations.index {|a| (a.start_with? 'STUDENT-TYPE-')}.present?,
         :registered => affiliations.include?('STUDENT-TYPE-REGISTERED'),
-        :exStudent => affiliations.include?('STUDENT-STATUS-EXPIRED'),
+        # TODO Remove '-STATUS-EXPIRED' logic once CalNet transition is complete.
+        :exStudent => (affiliations & ['STUDENT-STATUS-EXPIRED', 'FORMER-STUDENT']).present?,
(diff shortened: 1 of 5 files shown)


raydavis:
If this passes initial review & basic testing on Master, it should go into the QA branch so that our code is ready for the CalNet affiliation changes.

app/models/berkeley/user_roles.rb
```diff
@@ -18,36 +20,24 @@ def roles_from_affiliations(affiliations)
...
-            affiliations = affiliations - expired_aff
-          else
-            exp_date = ldap_record["berkeleyedu#{expdate_substring}expdate".to_sym].first
-            if exp_date.blank? || DateTime.parse(exp_date) > DateTime.now
-              affiliations = affiliations - expired_aff
-            else
-              affiliations = affiliations -  active_aff
-            end
-          end
+          logger.warn "UID #{ldap_record[:uid]} has conflicting CalNet affiliations #{affiliations}"
```
davidlimcheng: Would it help to log which affiliation we're removing?  Or would that be too noisy?

raydavis: Since it's predictable from the code, I didn't think it was worth making explicit. Would you rather have it in?

davidlimcheng: Nah I think it's OK - if we find it useful in the future we can always add it in then.

davidlimcheng:
LGTM @raydavis.  Hello again, expired students.

PR #6093, opened 2016-11-01
CLC-6494, Course Capture LTI script: when Canvas Tab API call fails, log error and return nil

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-6494

See `context 'Canvas reports an error'` below to understand purpose. We're not sure why it failed but at least the nightly cron won't abort. More QA will help us understand severity.

### DIFF
--- a/spec/models/canvas_lti/webcast_lti_refresh_spec.rb
+++ b/spec/models/canvas_lti/webcast_lti_refresh_spec.rb
@@ -1,41 +1,80 @@
 describe CanvasLti::WebcastLtiRefresh do
 
-  let(:webcast_tool_id) { rand(999999) }
+  let(:course_capture_tool_id) { random_id }
   let(:term_yr) { 2015 }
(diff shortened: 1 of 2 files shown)


spec/models/canvas_lti/webcast_lti_refresh_spec.rb
```diff
@@ -1,41 +1,80 @@
...
+          eligible_without_recordings
+        ]
+      }
       allow_any_instance_of(Webcast::SignUpEligible).to receive(:request_internal).and_return eligible
     end
 
-    context 'sign-up phase is open' do
-      context 'course site has webcast' do
+    context 'sign-up phase is active' do
+      context 'Canvas reports an error' do
```
johncrossman: Fyi, this spec is the new code. Every other change in `webcast_lti_refresh_spec.rb` is syntax and naming cleanup.

PR #6097, opened 2016-11-01
SISRP-27417 - WorkCenter Link working intermittently for Advisors

redconfetti:
https://jira.berkeley.edu/browse/SISRP-27417


nkutub:
thanks for the clarification. @redconfetti . looks good.

PR #6104, opened 2016-11-01
SISRP-27467 - Many "Unexpected CS Grading Status" errors logged

nkutub:
https://jira.berkeley.edu/browse/SISRP-27467

* updated grading status error handling to default to noCSdata for blank status codes. 
* debug vs error logging for unexpected grading status codes other than blank 
* bug fix for handling feed which returns single hashes vs array with single element. 
* minor UI change.


raydavis:
Thanks, @nkutub . The JIRA ticket doesn't say how to interpret the blank/empty status, though. Did you get any explanation from the data owners?

nkutub:
yes discussed it with Rachel and she confirmed to treat it as if there was not CS data returned at all

PR #6109, opened 2016-11-01
SISRP-26755 - Schedule Planner link - Campus Solutions page header needs to be removed

redconfetti:
https://jira.berkeley.edu/browse/SISRP-26755

See QA PR #6110

The issue that was occurring was that the `isCsLink` flag is being forced in [CampusSolutions::AdvisingResources](https://github.com/ets-berkeley-edu/calcentral/blob/ed63d211/app/models/campus_solutions/advising_resources.rb#L131), which results in the link not opening in a new window... and the "ucFrom" parameter (and other related parameters) being included.

So instead of using the `ccCampusSolutionsLinkItemDirective` like we do with these links, I just broke this out into using the `ccCampusSolutionsLinkDirective` to force it to behave as specified in the JIRA (open in a new window, do not use ucFrom parameters to make CalCentral banner appear).

### DIFF
--- a/src/assets/templates/widgets/toolbox/user_preview.html
+++ b/src/assets/templates/widgets/toolbox/user_preview.html
@@ -146,11 +146,7 @@
                 <div data-ng-if="ucAdvisingResources.links.schedulePlannerStudentSpecific.url">
                   <a
                     data-ng-bind="ucAdvisingResources.links.schedulePlannerStudentSpecific.name"
-                    data-cc-campus-solutions-link-directive="ucAdvisingResources.links.schedulePlannerStudentSpecific.url"
-                    data-cc-campus-solutions-link-directive-enabled="{{false}}"
-                    data-cc-campus-solutions-link-directive-text="ucAdvisingResources.links.schedulePlannerStudentSpecific.backToText"
-                    data-cc-campus-solutions-link-directive-cache="cache"
-                    data-cc-outbound-enabled="{{true}}"
+                    data-ng-href="{{ucAdvisingResources.links.schedulePlannerStudentSpecific.url}}"
                   ></a>
                 </div>
                 <div data-ng-if="ucAdvisingResources.csLinks.studentWebnowDocuments.url">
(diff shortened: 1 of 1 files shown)


src/assets/templates/widgets/toolbox/user_preview.html
```diff
@@ -144,10 +144,14 @@ <h3 data-ng-bind="targetUser.error.summary"></h3>
...
                 </div>
                 <div data-ng-if="ucAdvisingResources.links.schedulePlannerStudentSpecific.url">
-                  <div
-                    data-cc-campus-solutions-link-item-directive
-                    data-link="ucAdvisingResources.links.schedulePlannerStudentSpecific"
-                  ></div>
+                  <a
+                    data-ng-bind="ucAdvisingResources.links.schedulePlannerStudentSpecific.name"
+                    data-cc-campus-solutions-link-directive="ucAdvisingResources.links.schedulePlannerStudentSpecific.url"
+                    data-cc-campus-solutions-link-directive-enabled="{{false}}"
```
redconfetti: @davidlimcheng - Good catch, but I'm aware that if the `isCsLink` value was FALSE for this link, we'd have the desired behavior without having to bypass the campusSolutionsLinkItemDirective. I didn't want to have to go into the model and modify it's behavior just to get the desired behavior. I felt that doing it at the template level is better, especially given that we're already planning on cleaning this whole thing up in GoLive 7.3 with [SISRP-27494](https://jira.berkeley.edu/browse/SISRP-27494) and [SISRP-27468](https://jira.berkeley.edu/browse/SISRP-27468)

redconfetti: In summary, I'm trying to get this into QA ASAP with the exact behavior we're looking for.

NOTE: This link isn't coming from the CS Link API, it's the only one remaining that is configured. In GL 7.3 it will be obtained from a new API that is being developed in CS, because it's a special link for the College Scheduler application (non-CS app) that will represent an Advisor viewing a students information.

davidlimcheng: @redconfetti Understood, but won't we have the same outcome if we just use `data-cc-outbound-enabled` (to open in a new tab) and remove the `data-cc-campus-solutions-link-directive` attributes entirely?

PR #6123, opened 2016-11-02
CLC-6499, remove Google::APIClient::InstalledAppFlow from DriveManager class

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-6499

Notes:
* read Jira above on `Google::APIClient::InstalledAppFlow` removal
* `google_api` and `drive_api` methods lazy-load API proxy (only one `discovered_api('drive', 'v2')` to maintain)
* `drive_manager_spec` no longer makes real Google API calls; it simply confirms our error handling and handling of multiple rows in Google response 
* hash syntax is cleaner

### DIFF
--- a/spec/models/google_apps/drive_manager_spec.rb
+++ b/spec/models/google_apps/drive_manager_spec.rb
@@ -1,74 +1,108 @@
 describe GoogleApps::DriveManager do
 
-  context '#real', testext: true, order: :defined do
+  subject(:drive) { GoogleApps::DriveManager.new(random_id, random_id) }
 
-    before(:all) do
-      settings = Settings.oec.google.marshal_dump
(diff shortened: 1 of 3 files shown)


spec/models/google_apps/drive_manager_spec.rb
```diff
@@ -1,74 +1,108 @@
 describe GoogleApps::DriveManager do
 
-  context '#real', testext: true, order: :defined do
+  subject(:drive) { GoogleApps::DriveManager.new(random_id, random_id) }
```
johncrossman: The "realness" of this spec is not worth the cost of maintenance. We now have the coverage we need. More info: https://jira.ets.berkeley.edu/jira/browse/CLC-6489

paulkerschen: We couldn't keep it real!

paulkerschen:
Looks good @johncrossman. Waiting for King T.

PR #6124, opened 2016-11-02
SISRP-27492 - Advising Dashboard - Remove Academic Progress Link and What If Report Link

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-27492

Succinct explanation from Liz (functional owner) so you don't have to dig through the JIRA:

> We will remove these two links from the Advising Resources card from the Advisor Dashboard for GL 7.2. Both links exist on the Student Overview Page > Personal Summary card and has a better experience as the EMPLID of the student is passed in they land exactly where they need to in Campus Solutions without having extra clicks and a page view that we were trying to avoid them having.

* QA PR to come upon merge.

### DIFF
--- a/app/models/campus_solutions/advising_resources.rb
+++ b/app/models/campus_solutions/advising_resources.rb
@@ -44,7 +44,6 @@ module CampusSolutions
 
       advising_link_settings = [
         # advisors see these on advisor-specific views
-        { feed_key: :uc_academic_progress_report, cs_link_key: 'UC_CX_APR_RPT'},
         { feed_key: :uc_administrative_transcript, cs_link_key: 'UC_CX_ADM_TRANSCRIPT' },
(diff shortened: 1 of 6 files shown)


lyttam:
@davidlimcheng Do we still need to fetch that link from the API, or can we remove this key from the config in `CampusSolutions::AdvisingResources`?

```
      advising_link_settings = [
        # advisors see these on advisor-specific views
        { feed_key: :uc_academic_progress_report, cs_link_key: 'UC_CX_APR_RPT'},
```

davidlimcheng:
Great catch @lyttam, we will no longer need the non-student-specific links, so I removed both the Academic Progress Report link and the What If Report link from `CampusSolutions::AdvisingResources`.  Should be good now.

PR #6125, opened 2016-11-02
SISRP-27641, User::StoredUsers: avoid destroy call on nil:NilClass

johncrossman:
https://jira.berkeley.edu/browse/SISRP-27641

Oddly, we had `found.size > 0` and `found.first` nil. Race condition? This change _should_ avoid recurrence.

### DIFF
--- a/app/models/user/stored_users.rb
+++ b/app/models/user/stored_users.rb
@@ -4,8 +4,8 @@ module User
 
     def self.get(uid)
       users = {
-        :saved => [],
-        :recent => []
+        saved: [],
+        recent: []
       }
 
       uid_entries = get_stored_uid_entries(uid)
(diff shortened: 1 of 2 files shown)


app/models/user/stored_users.rb
```diff
@@ -110,18 +110,18 @@ def self.get_stored_uid_entries(uid)
...
       end
     end
 
     def self.delete(model, uid, uid_to_delete)
-      found = model.where(:stored_uid => uid_to_delete.to_s)
-      if (found.size > 0)
-        found.first.destroy
+      found = model.where(stored_uid: uid_to_delete.to_s)
+      if found && (first = found.first)
+        first.destroy
```
paulkerschen: @johncrossman FYI the idiomatic one-liner would be 

```
if (found = model.find_by stored_uid: uid_to_delete.to_s)
  found.destroy
```

redconfetti:
A race condition? I don't think this code is asynchronous. Maybe ActiveRecord was returning a value from #size even when there wasn't one present?

app/models/user/stored_users.rb
raydavis: @johncrossman I think @paulkerschen 's comment still applies. If we suspect a race condition, it does seem like a good idea to eliminate the extra ``where`` step. In fact, I wonder if ``model.destroy_all(stored_uid: uid_to_delete.to_s)`` might come even closer to an atomic operation...

johncrossman: Thanks @raydavis  I'll make the fix.

johncrossman:
Thanks for the tip, @paulkerschen  Amendment is in. Of course, I couldn't resist tidying up the spec.

paulkerschen:
Thanks @johncrossman, looks good.

PR #6126, opened 2016-11-02
SISRP-24743 - Filtering empty academic career and level descriptions

redconfetti:
https://jira.berkeley.edu/browse/SISRP-24743

### DIFF
--- a/app/models/my_academics/college_and_level.rb
+++ b/app/models/my_academics/college_and_level.rb
@@ -104,13 +104,13 @@ module MyAcademics
             careers << career
           end
         end
-      end.uniq.compact
+      end.uniq.reject { |level| level.to_s.empty? }
     end
 
     def parse_hub_level(statuses)
       level = statuses.collect do |status|
         status['currentRegistration'].try(:[], 'academicLevel').try(:[], 'level').try(:[], 'description')
-      end.uniq.compact.to_sentence
+      end.uniq.reject { |level| level.to_s.empty? }.to_sentence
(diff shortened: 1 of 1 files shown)


app/models/my_academics/college_and_level.rb
```diff
@@ -104,13 +104,13 @@ def parse_hub_careers(statuses)
...
         end
-      end.uniq.compact
+      end.uniq.reject { |level| level.empty? }.compact
     end
 
     def parse_hub_level(statuses)
       level = statuses.collect do |status|
         status['currentRegistration'].try(:[], 'academicLevel').try(:[], 'level').try(:[], 'description')
-      end.uniq.compact.to_sentence
+      end.uniq.compact.reject { |level| level.empty? }.to_sentence
```
redconfetti: @davidlimcheng - Nope. It only removes the nil's, not the empty strings. Maybe I should remove the #compact and it will do the same thing, because even #empty? detects nils.

redconfetti:
```
>> ['', nil, 'abc', nil, '', 'xyz'].compact
=> ["", "abc", "", "xyz"]
>> ['', nil, 'abc', nil, '', 'xyz'].reject { |e| e.empty? }
NoMethodError: undefined method `empty?' for nil:NilClass
>> ['', nil, 'abc', nil, '', 'xyz'].reject { |e| e.to_s.empty? }
=> ["abc", "xyz"]
```

PR #6128, opened 2016-11-03
CLC-6489, OEC-related testext uses Bamboo-only refresh_token

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-6489

Notes:
* The #real test in `credential_store_spec` was of little value; it is gone.
* But, the #realness of `sheets_manager_spec` and `remote_drive_spec` is valuable. I followed the pattern of `google_proxy` configs on Bamboo and add new setting: `oec.google. testext.refresh_token`

I updated `testext.local.yml` on Bamboo (docker), ran a personal build and it's green: https://bamboo.ets.berkeley.edu/bamboo/browse/CPB-CJT86-4

### DIFF
--- a/config/settings/testext.yml
+++ b/config/settings/testext.yml
@@ -60,6 +60,18 @@ cache:
 calnet_crosswalk_proxy:
   fake: true
 
+# Bamboo (testext) only:
+# - OEC-related spec relies on oec.google.testext.refresh_token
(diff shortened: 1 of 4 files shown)


paulkerschen:
If Bamboo is happy, so am I.

PR #6145, opened 2016-11-03
NOJIRA Bumping release version number to 85

raydavis:


johncrossman:
Thanks for remembering the `db/migrate`, @raydavis

PR #6151, opened 2016-11-04
CLC-6505, Canvas data refresh must handle nil from Canvas::Report::Sections.get_csv

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-6505

Steven is contacting Instructure about `ErrorReport:140435606` (as seen in JIRA above). Regardless of what we learn, protecting ourselves from `nil` is a good thing.


paulkerschen:
Bueno @johncrossman.

PR #6153, opened 2016-11-07
SISRP-27723 - Committees card showing inactive members for both Faculty as well as Grad Students

nkutub:
https://jira.berkeley.edu/browse/SISRP-27723

This is the PR for the emergency release to fix a blocker where the inactive members of an active committee were being displayed. Add logic to remove inactive members from active committees. 

This also fixed bug with photo check logic

Added text to display for faculty committees if there are no committees to display. 

Added API endpoints to the testing script.


app/controllers/my_committees_controller.rb
```diff
@@ -22,7 +21,7 @@ def student_photo
   def member_photo
     member_id = Integer(params['member_id'], 10)
     my_committees = MyCommittees::Merged.from_session(session)
-    if my_committees.can_user_see_member_photo?(member_id)
+    if  my_committees.get_feed_as_json.include? "member/#{ member_id }"
```
raydavis: I feel dumb pointing this double-space out but hey...

raydavis:
Looks good otherwise, @nkutub !

app/controllers/my_committees_controller.rb
nkutub: @raydavis done!

PR #6159, opened 2016-11-08
SISRP-27258 - 7.3 Class Enrollment Card - Add Class Schedule Link

redconfetti:
https://jira.berkeley.edu/browse/SISRP-27258


nkutub:
@redconfetti looks good

PR #6163, opened 2016-11-09
SISRP-27781 - CalCentral header links and navigation links during doc upload is not working 

redconfetti:
https://jira.berkeley.edu/browse/SISRP-27781

### DIFF
--- a/src/assets/javascripts/angular/controllers/widgets/finaid/finaidCommunicationsController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/finaid/finaidCommunicationsController.js
@@ -6,15 +6,17 @@ var _ = require('lodash');
 /**
  * Finaid Communications controller
  */
-angular.module('calcentral.controllers').controller('FinaidCommunicationsController', function($q, $scope, activityFactory, finaidFactory, finaidService, tasksFactory) {
+angular.module('calcentral.controllers').controller('FinaidCommunicationsController', function($q, $scope, activityFactory, finaidFactory, finaidService, linkService, tasksFactory) {
+  var backToText = 'Financial Aid and Scholarships';
   $scope.communicationsInfo = {
-    isLoading: true,
     aidYear: '',
-    taskStatus: '!completed',
+    backToText: backToText,
+    isLoading: true,
     counts: {
       completed: 0,
       uncompleted: 0
-    }
+    },
+    taskStatus: '!completed'
   };
 
   $scope.toggleCompletedTasks = function() {
@@ -29,6 +31,14 @@ angular.module('calcentral.controllers').controller('FinaidCommunicationsControl
     };
     return activityFactory.getFinaidActivity(options).then(function(data) {
       angular.extend($scope, data);
+      if (_.get($scope, 'list')) {
+        linkService.addBackToTextToResources($scope.list, backToText);
+        _.forEach($scope.list, function(value) {
+          if (_.get(value, 'elements')) {
+            linkService.addBackToTextToResources(value.elements, backToText);
+          }
+        });
+      }
       $scope.activityInfo.isLoading = false;
     });
   };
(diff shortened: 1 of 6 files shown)


redconfetti:
Hold off on merging this. I still need to fix the ucFromText for the links.

src/assets/javascripts/angular/controllers/widgets/finaid/finaidCommunicationsController.js
```diff
@@ -6,15 +6,17 @@ var _ = require('lodash');
...
     aidYear: '',
-    taskStatus: '!completed',
+    backToText: backToText,
+    isLoading: true,
     counts: {
       completed: 0,
       uncompleted: 0
-    }
+    },
+    taskStatus: '!completed'
```
nkutub: @redconfetti , not sure why you are using a string here vs a boolean? I don't see taskStatus in the code being other than completed or not completed should it not be better to just do something like isCompleted: true

src/assets/javascripts/angular/controllers/widgets/finaid/finaidCommunicationsController.js
```diff
@@ -29,6 +31,14 @@ angular.module('calcentral.controllers').controller('FinaidCommunicationsControl
     };
     return activityFactory.getFinaidActivity(options).then(function(data) {
       angular.extend($scope, data);
+      if (_.get($scope, 'list')) {
+        linkService.addBackToTextToResources($scope.list, backToText);
+        _.forEach($scope.list, function(value) {
```
davidlimcheng: It seems like we are expecting $scope.list to be an array of objects - is that true?  If so, would it make more sense to use `object` instead of `value`?

src/assets/javascripts/angular/controllers/widgets/finaid/finaidCommunicationsController.js
redconfetti: @nkutub - I didn't add that. I only re-arranged the list of properties so that the keys are in alphabetical order. It's not within the scope of my task to investigate and refactor that.

src/assets/javascripts/angular/controllers/widgets/finaid/finaidCommunicationsController.js
redconfetti: @davidlimcheng - Sure!

src/assets/javascripts/angular/controllers/widgets/finaid/finaidCommunicationsController.js
nkutub: OK, as long as it works we can change it later when we have more time.

src/assets/javascripts/angular/controllers/widgets/finaid/finaidCommunicationsController.js
nkutub: @davidlimcheng , It is such a small function, it is clear couple lines down that it is an obj type. Technically, objects are passed as pointers in memory which are values so not gonna worry too much about this one :)

src/assets/javascripts/angular/controllers/widgets/finaid/finaidCommunicationsController.js
```diff
@@ -29,6 +31,14 @@ angular.module('calcentral.controllers').controller('FinaidCommunicationsControl
     };
     return activityFactory.getFinaidActivity(options).then(function(data) {
       angular.extend($scope, data);
+      if (_.get($scope, 'list')) {
+        linkService.addBackToTextToResources($scope.list, backToText);
```
redconfetti: @davidlimcheng I'll need to discuss this with you further. I'm a little confused.

src/assets/javascripts/angular/controllers/widgets/finaid/finaidCommunicationsController.js
redconfetti: @nkutub - I'm not sure [that Javascript passes variables by reference](https://repl.it/EXlC/1)

Variable references do become bound to variables of the same name found outside of the function scope though, as far as I know.

PR #6165, opened 2016-11-09
SISRP-28013 - 7.3: Refactor front-end to support backToText for new CS Link API

redconfetti:
https://jira.berkeley.edu/browse/SISRP-28013

### DIFF
--- a/src/assets/javascripts/angular/services/linkService.js
+++ b/src/assets/javascripts/angular/services/linkService.js
@@ -1,0 +1,29 @@
+'use strict';
+
+var angular = require('angular');
+var _ = require('lodash');
+
+angular.module('calcentral.services').factory('linkService', function() {
+  /**
+   * Add the back to text for external link objects
+   */
+  var addBackToTextToLink = function(link, backToText) {
+    link.backToText = backToText;
+    return link;
+  };
+
+  /**
+   * Adds the back to text to each object in a resource collection
+   */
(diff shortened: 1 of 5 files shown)


src/assets/javascripts/angular/services/linkService.js
```diff
@@ -0,0 +1,29 @@
...
+  var addBackToTextToLink = function(link, backToText) {
+    link.backToText = backToText;
+    return link;
+  };
+
+  /**
+   * Adds the back to text to each object in a resource collection
+   */
+  var addBackToTextToResources = function(resources, backToText) {
+    var addBackToTextLinkWrapper = function(resource) {
```
lyttam: @redconfetti would it be more efficient to move this function def outside of `addBackToTextToResources` so that it doesn't get redefined every time `addBackToTextToResources` is called?

redconfetti: @lyttam - It wouldn't work in that scenario because it wouldn't be able to reference the `backToText` variable that was passed into the function as an argument.

The reason I'm adding this wrapper is because the _.mapValues method by convention passes a single argument to the callback function (addBackToTextLinkWrapper), and I'm wanting it to be able to reference `backToText` and have Javascript find that variable present immediately within the scope/context outside of the function.

Wrapping these calls in this function was a way for that to be accomplished.

lyttam: Cool

PR #6166, opened 2016-11-09
SISRP-27912 Do not throw exceptions from Degree Progress if no CS ID

raydavis:
https://jira.berkeley.edu/browse/SISRP-27912


raydavis:
FYI, @lyttam

lyttam:
@raydavis ahhh good to know. Next time I'll try and remember to write a test for this case!

PR #6168, opened 2016-11-13
SISRP-28019 - Committees card need to show Insider/outsider details for committee members

nkutub:
https://jira.berkeley.edu/browse/SISRP-28019

update 11/15: 
* after discussing with @raydavis, I removed the static UID's and returned them to random ones.
* added mocks to fake the CrossWalk calls 

-------------------------------------------------
* fix testext failures due to random uid. 

This was a tough one. I had to configure my local machine to testext, thank @redconfetti for your help on that.  It was difficult since it was an intermittent error. Some randomly generated uid's being used in the CS proxy calls for the committees API's were sometimes valid and sometimes errored the crosswalk calls. At the end, I'm not sure why the proxy calls are still requiring a valid uid to be passed even though the fake flag is set to true.

### DIFF
--- a/spec/models/campus_solutions/faculty_committees_spec.rb
+++ b/spec/models/campus_solutions/faculty_committees_spec.rb
@@ -1,6 +1,7 @@
 describe CampusSolutions::FacultyCommittees do
 
-  let(:user_id) { '10634814' }
+  let(:user_id) { random_id }
+  let(:user_cs_id) { random_id }
 
   shared_examples 'a proxy that gets data' do
     subject { proxy.get }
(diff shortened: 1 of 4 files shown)


raydavis:
> I'm not sure why the proxy calls are still requiring a valid uid to be passed even though the fake flag is set to true.

@nkutub , I'm afraid the top priority is to figure that out rather than to work around it with a hope-it-never-matches UID. I know you probably have more pressing issues to deal with in this sprint, though, so that fix may have to be put off for a while. In the meantime, I'd vote for leaving the bug open & the tests quarantined. What do you think?

raydavis:
Actually, I think I see one big issue. Proxy faking is handled by matching a HTTP request to a fake payload; if no match is found, then the fake-out fails and you find yourself pushed into cold reality.

By default, the match is on URI. By leaving the ``user_id`` property out of ``CampusSolutions::FacultyCommittees.new(fake: true, user_id: uid)``, you made a global URI which doesn't match the mock.

Could you try some scenarios with the random UID and the user_id argument both included?

nkutub:
@raydavis , I just made the fixes to this PR, please review when you have a chance

raydavis:
Thanks, @nkutub !

PR #6172, opened 2016-11-15
SISRP-23919 SISRP-25478 Restrict access to confidential students

raydavis:
https://jira.berkeley.edu/browse/SISRP-23919
https://jira.berkeley.edu/browse/SISRP-25478

Reviewing one commit at a time will probably be easiest. The "NOJIRA" commit carries a particularly long comment for those who enjoy such things.


app/models/degree_progress/my_graduate_milestones.rb
```diff
@@ -14,7 +14,7 @@ class MyGraduateMilestones < UserSpecificModel
     ]
 
     def get_feed_internal
-      return {} unless is_feature_enabled && authorized?
+      return {} unless is_feature_enabled && target_audience?
```
nkutub: @raydavis you mentioned in the comments that the filter for the specific audience should never be at the get_internal_feed level yet you are still limiting access to a specific audience here. Is that not the same thing sorta. If this feed needed to be used in another filtered way for not graduates/law we would have to refactor this.

app/policies/authentication_state_policy.rb
```diff
@@ -86,16 +86,4 @@ def can_view_other_user_photo?
...
-    attributes = User::AggregatedAttributes.new(@user.user_id).get_feed
-    attributes.try(:[], :roles).try(:[], :graduate)
-  end
-
-  def law_student?
-    return false unless @user.real_user_auth.active?
-    attributes = User::AggregatedAttributes.new(@user.user_id).get_feed
-    attributes.try(:[], :roles).try(:[], :law)
-  end
 end
```
nkutub: totally makes sense about the "can_" usage in authorization. So technically we could have kept it here if we changed it to can_view_graduate_milestones

nkutub: but since it is such a specific use case makes no sense to put it at a top level authorization.

app/models/degree_progress/my_graduate_milestones.rb
raydavis: Thanks, @nkutub , good question. As I understand it, this isn't being treated as a security issue but as a functional issue, where the return value for undergrads is currently intended to be an empty hash (rather than throwing an alarming log message). And yes, it's true that when/if undergrads get a different payload, then this code would be changed to match.

But I hope there will be a refactoring before then to eliminate the MyGraduateMilestones / "GraduateMilestones" (really meaning AdvisorViewGraduateMilestones) split, replacing them with just one shared cached endpoint.

nkutub:
looks good to me

app/models/degree_progress/my_graduate_milestones.rb
raydavis: The main reason, though, why a view-as check _always_ needs to happen outside the cache-writing-method is that we want the view-as folks to get a different response from what looks like the same feed. In other words, what counts is that "the specific audience" differs but the same cache key is being fetched.

The MyGraduateMilestones payload is identified by the student's UID. So no matter _who_ asks for a MyGraduateMilestones object, the cached data will depend on the Graduate/Law status of that UID and always be the same value. If a view-as viewer requests it before the real student does, there's no harm: the viewer and the real-student should see the same data.

Instead, consider something like MyUpNext, where a view-as viewer should not see any Google Calendar events but the real-student should. If the view-as viewer happened to look at that student's dashboard before the student logged in, they could store an emptied-out feed in the cache, and sabotage the student's view. So we filter above the cache to deal with that.

lyttam:
Thanks @raydavis for the walkthrough and the helpful comments!

PR #6177, opened 2016-11-17
SISRP-28019 - Committees card need to show Insider/outsider details for committee members

nkutub:
https://jira.berkeley.edu/browse/SISRP-28019

* fix bug with faculty committee api returning nil generates javascript error due to empty feed.

### DIFF
--- a/app/models/my_committees/faculty_committees.rb
+++ b/app/models/my_committees/faculty_committees.rb
@@ -9,22 +9,23 @@ module MyCommittees
     end
 
     def get_feed
-      result = {}
+      result = {
+        facultyCommittees: {
+          active: [],
+          completed: []
+        }
+      }
       feed = CampusSolutions::FacultyCommittees.new(user_id: @uid).get[:feed]
       if feed && (cs_committees = feed[:ucSrFacultyCommittee].try(:[], :facultyCommittees))
         # Service range will be parsed from nested member data with match on emplid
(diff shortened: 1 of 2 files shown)


davidlimcheng:
LGTM, @nkutub - but we're going to need to modify the JIRA.  Usually, when we fix bugs, they're with a new bug ticket, which we mark as `Resolved` upon PR merge.  This lets the QA team know that the bug is ready to be tested again, and if it passes, they'll mark it as `Closed`.  Since the JIRA you linked to seems to be an issue separate from this specific bug, it'll be hard/impossible for QA to know that this particular bug is ready to be retested.  I recommend `Reopening` that ticket, and marking it as `Resolved` once this PR is merged, with a comment describing what the bug was and, ideally, a screenshot of what would happen (I'm assuming you caught this new bug before QA did).

nkutub:
@davidlimcheng , thanks for the review. This bug was found by me testing other work with a UID which resulted in an edge case returned from the API I was not anticipating in code. So technically there was no QA involved that caught that. Would we still open a new ticket even though we are still in Dev/test lifecycle for this work? I just lean on reduced paperwork route. I do understand, I did do a change to code base and technically all changes should be re-QA'd so you might still have a point.

davidlimcheng:
@nkutub Makes sense, I think we're good since the whole feature will be tested again anyways.  But in the future, IMO it's good to open tickets for self-discovered bugs so that QA will know what to test after fixing.

PR #6179, opened 2016-11-17
SISRP-27695 - Adds sub-plan to the Profile card

lyttam:

Students can have a sub-plan associated with their plan (major or minor) which will now display on the Profile card.

So far it looks like there will be a 1-1 relationship between plan and sub-plan.  I'm checking to see if we'll ever have to handle a list of sub-plans and will make changes accordingly if that's the case.

ideal

### DIFF
--- a/src/assets/stylesheets/_profile.scss
+++ b/src/assets/stylesheets/_profile.scss
@@ -22,6 +22,9 @@
       width: 92px;
     }
   }
+  .cc-widget-profile-indent {
+    margin-left: 10px;
+  }
   .cc-widget-profile-content-fullname {
     font-size: 16px;
   }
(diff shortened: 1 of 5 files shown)


src/assets/stylesheets/_profile.scss
```diff
@@ -22,6 +22,9 @@
       width: 92px;
     }
   }
+  .cc-widget-profile-indent {
```
nkutub: @lyttam , this css class in not in correct order.

redconfetti: Although the css-lint doesn't require the styles to be defined in alphabetical order, I think that's a nice thing for us to do when we can. It helps a person find styles in the list quicker if we know which order they're in.

What do you think @lyttam ?

lyttam: @nkutub @redconfetti Fine by me, except none of the classes in this file are in correct order.  It doesn't seem prudent to reorder the classes right now.  If you feel strongly about it, I could do it in my next PR - I'll be submitting a few PRs touching this code.

PR #6183, opened 2016-11-18
SISRP-18802 Retire or replace "Bearfacts" event notifications

raydavis:
https://jira.berkeley.edu/browse/SISRP-18802

NOTE: There is not yet a replacement for Final Grades notifications. If one becomes available, this removed code might be of interest as a reference, since it transforms a JMS message about a class section into locally-stored "Notifications" items for every student in the section.


nkutub:
@raydavis , nice delete!. So, in the future, if we got a new Final Grades notification replacement, would we create it the same way as subclass of AbstractEventProcessor and add it back into PROCESSOR_CLASSES?

raydavis:
Depends on what the replacement service looks like, @nkutub -- it might send an entirely different payload, like a separate message for each student. But since there aren't other working examples for using CalCentral's Postgres "Notifications" table, I wanted to highlight the use case. It's potentially a handy thing to know about.

nkutub:
Looks good!

PR #6184, opened 2016-11-18
SISRP-26741 - 7.3: Academic Records on My Academics (Student)

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-26741
Design JIRA:  https://jira.berkeley.edu/browse/SISRP-21544

### DIFF
--- a/src/assets/javascripts/angular/controllers/widgets/academics/academicRecordsController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/academics/academicRecordsController.js
@@ -3,7 +3,7 @@
 var angular = require('angular');
 var _ = require('lodash');
 
-angular.module('calcentral.controllers').controller('AcademicRecordsController', function(academicRecordsFactory, apiService, $scope) {
+angular.module('calcentral.controllers').controller('AcademicRecordsController', function(academicRecordsFactory, apiService, csLinkFactory, $scope) {
 
   $scope.officialTranscript = {
     postParams: {},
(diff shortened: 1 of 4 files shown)


src/assets/javascripts/angular/controllers/widgets/academics/academicRecordsController.js
```diff
@@ -3,14 +3,19 @@
...
+angular.module('calcentral.controllers').controller('AcademicRecordsController', function(academicRecordsFactory, apiService, csLinkFactory, $scope) {
 
   $scope.officialTranscript = {
     postParams: {},
     postUrl: '',
     postUrlHover: 'Request Transcript',
     isLoading: true
   };
+  $scope.lawTranscriptLink = {
+    link: 'http://www.law.berkeley.edu/php-programs/registrar/forms/transcriptrequestform.php',
```
nkutub: @davidlimcheng , I thought that we are moving away from hard coding links in the front end. Is this something that can be added to the feed returned from the academic record factory? I know we have used the backend to either pull links from the Link API or get it from settings yml.

nkutub:
I will merge this PR since the comments are simply design best practice questions that need to be addressed with the entire team.

src/assets/javascripts/angular/controllers/widgets/academics/academicRecordsController.js
raydavis: @nkutub , I suspect code-embedded URLs remain the most efficient approach when links don't embed personal data, links don't vary by server environment, and links don't come from CS. Seems like the other existing approaches might create more maintenance work than they solve. I'm also interested in other opinions, though!

src/assets/javascripts/angular/controllers/widgets/academics/academicRecordsController.js
```diff
@@ -52,11 +57,21 @@ angular.module('calcentral.controllers').controller('AcademicRecordsController',
     });
   };
 
+  var fetchLawUnofficialTranscriptLink = function() {
+    csLinkFactory.getLink({
```
davidlimcheng: @nkutub After playing around with the Link API, I totally agree.  This case is kinda weird though, as the TranscriptData API call is a completely different feature from this link to CS, and the decision to have both of these links on the same card was purely a design coincidence.  In other words, these features should be exclusive of each other, which is why I decided to separate its call.

PR #6185, opened 2016-11-18
SISRP-27695 - Adds designated emphasis to Profile card

lyttam:

Designated emphasis is to grad students as a minor is to undergrads.  

We'll display it alongside majors and minors.  Following the "rule of 3", since we're now repeating the same code three times in the template I moved it into a directive.  We'll likely be adding a fourth type (certificates) in the future.


nkutub:
@lyttam, looks good. I guess I would like to ask the team if using a directive for a specific implementation vs one for a generic use across the code base is acceptable.

PR #6189, opened 2016-11-21
SISRP-27695 - Adds conferred (awarded) degree(s) to Profile card

lyttam:

If a student has completed degrees, display them on the profile card.  

Still missing some of the data points (academic program) so I may be submitting a follow-up PR.

image


davidlimcheng:
Looks great, @lyttam.

PR #6190, opened 2016-11-21
SISRP-24177 Fix Hub Terms API verification between terms

raydavis:
https://jira.berkeley.edu/browse/SISRP-24177

Prevents unwelcome distraction when we're between academic terms.


davidlimcheng:
ezpz lemon squeezy

PR #6194, opened 2016-11-21
CLC-6515, OEC Report Diff task should log level WARN when row is blank

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-6515

See JIRA description.

cc: @paulkerschen


paulkerschen:
Agreed, @johncrossman, this is an improvement.

PR #6197, opened 2016-11-22
SISRP-23756 - Financials - waivers appearing with item type W

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-23756


nkutub:
looks good

PR #6200, opened 2016-11-23
SISRP-27722 - Standardizes <h*> styles sitewide for mac/win compatibility

lyttam:
https://jira.berkeley.edu/browse/SISRP-27722

After complaints that our headers looked wimpy on Windows 7, I updated the h3 and h4 styles* so that they appear similar on both Windows (Chrome) and Mac (Chrome).  I will continue spot-checking in other browsers as time permits.  Wanted to get this out first in case there are any concerns, since this affects the entire site.

If you want to see the before/after, view screenshots here:  https://drive.google.com/open?id=0BzCFJacXKzFBdGxWWWhBN2xQdmM

*I'm a little concerned about not being able to test every single header on the site.

### DIFF
--- a/src/assets/stylesheets/_academics.scss
+++ b/src/assets/stylesheets/_academics.scss
@@ -13,7 +13,6 @@
     }
   }
   h4 {
-    font-size: 14px;
     &:not(:first-of-type) {
       margin-top: 20px;
     }
--- a/src/assets/stylesheets/_base.scss
+++ b/src/assets/stylesheets/_base.scss
@@ -81,7 +81,12 @@ h2 {
   font-size: 18px;
 }
 h3 {
-  font-size: 16px;
+  font-size: 15px;
+  font-weight: 600;
+}
+h4 {
+  font-size: 13px;
+  font-weight: 600;
 }
 
 // Inset list items
--- a/src/assets/stylesheets/_student_success.scss
+++ b/src/assets/stylesheets/_student_success.scss
@@ -4,7 +4,7 @@
     margin: 15px 15px 12px;
     span {
       display: inline;
-      font-size: 16px;
+      font-size: 15px;
       font-weight: normal;
     }
   }


davidlimcheng:
Thanks for the screenshots, @lyttam.  The "after" headers do look more hearty.  Can you speak a bit as to why these changes "fix" the Windows views?  Does it have something to do with using `font-weight` vs. `<strong>`?  Is there anything we can do to move away from wimpy Windows headers in the future while developing new features?

lyttam:
@davidlimcheng Yes, the font-weight of all h* headers was set to 500 (_base.scss line 43), which looked fine and bold on a Mac but completely non-bold on Windows. I'm certainly not a type expert but I suspect it has something to do with this (see "Fallback"):  https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight

I'm kind of surprised given that we're using websafe fonts Helvetica/Arial.  Maybe there is a way to figure out exactly which font is getting used in a given browser and then check if that font has 500 weight available.  Quicker solution is to switch to using a more standard weight, I think.

redconfetti:
Perhaps we should define standard weights that we know work, then use SCSS variables throughout the stylesheets to establish a convention.

```
// _base.css ?
/* Consistent with Mac and Windows */
$bold-font-weight: 600;

// some_other.css
font-weight: $bold-font-weight;
```

davidlimcheng:
Thanks for the explanation, @lyttam.  LGTM.  

@redconfetti, I think that's a great idea.  For now, since we mostly are depending on attribute tags to inherit these properties, I think we should be in the clear.

lyttam:
@redconfetti great idea, I will look at our current use of font-weight and then perhaps submit another PR for that.

redconfetti:
Yeah. I would just keep it in mind if this continues to come up. Right now I think you have it taken care of with the header styles.

PR #6203, opened 2016-11-28
SISRP-28337 - My Academics Page JavaScript error "Cannot read property 'haasMastersFinEng' of undefined"

nkutub:
https://jira.berkeley.edu/browse/SISRP-28337

* needed to filter out call for non-students since roles object is not populated.


raydavis:
Thanks for the fix, @nkutub !

PR #6204, opened 2016-11-29
SISRP-28316 Fix Delegate authorization

raydavis:
https://jira.berkeley.edu/browse/SISRP-28316

If this tests out on Master, I'll open a QA PR to match.

### DIFF
--- a/app/controllers/act_as_controller.rb
+++ b/app/controllers/act_as_controller.rb
@@ -10,8 +10,8 @@ class ActAsController < ApplicationController
 
   def start
     uid_param = params['uid']
-    act_as_authorization uid_param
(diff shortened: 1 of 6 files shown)


raydavis:
The two main points of the code changes are:

1. Adding a Delegate Act-As Controller test to catch such problems.

2. Centralizing the viewer-mode-specific authorization check on target user up into the controller-specific ``act_as_authorization`` method. Part of the target-user-specific logic had previously been living in the shared ``valid_params?`` method. But this logic wasn't correct for Delegate authorization. And it caused redundant work for Advisor authorization, since it made ``fetch_another_users_attributes`` show up in two places. Now the attributes-fetch is only checked once, and only when it's needed

PR #6215, opened 2016-11-29
CLC-6520, 'Create a Site' button after Canvas dashboard_sidebar renders

johncrossman:

With Instructure's Nov 19 Canvas release came the removal of `class="rs-margin-lr"` in the div that wraps `<button id="start_new_course" .../>`. The button is now wrapped in a simple `<div>`.

The new logic uses jQuery `replaceWith()`. (jQuery is valid in context of Canvas customization.)


public/canvas/canvas-customization.js
```diff
@@ -111,9 +111,8 @@
               });
 
               // Add the 'Create a Site' button to the Dashboard page
-              waitUntilAvailable('#right-side .rs-margin-lr', false, function($container) {
-                $('#start_new_course').remove();
-                $container.prepend($createSiteButton);
+              waitUntilAvailable('#start_new_course', false, function($container) {
```
paulkerschen: @johncrossman This makes sense. Do we still need to pass `$container` into this function?

paulkerschen: @johncrossman Or is `$container` now equivalent to `$('#start_new_course')`, meaning we could use it below instead of doing a second DOM search for the element?

johncrossman: A good catch, @paulkerschen  Our friend `jshint` also dislikes the never used var.  And yes, `$container` is the button.  Amendment is in.

johncrossman:
I need to do one more verification before we merge this.  Closing for a moment.

PR #6216, opened 2016-11-30
SISRP-28284 Reduce performance impact of EDO DB crosslistings

raydavis:
https://jira.berkeley.edu/browse/SISRP-28284

Also increases cache lifespan of detailed section info.

### DIFF
--- a/app/models/edo_oracle/queries.rb
+++ b/app/models/edo_oracle/queries.rb
@@ -278,7 +278,7 @@ module EdoOracle
           TRIM(crs."transcriptTitle") AS course_title_short
         FROM SISEDO.API_CROSSLISTINGSV00_VW xlist
         JOIN SISEDO.API_COURSEV00_VW crs ON
-          (xlist."cms-version-independent-id" = crs."cms-version-independent-id")
+          (xlist."cms-id" = crs."cms-id")
         WHERE xlist."displayName" = '#{course_code}'
       SQL
       result.first if result
--- a/app/models/edo_oracle/user_courses/base.rb
+++ b/app/models/edo_oracle/user_courses/base.rb
@@ -189,7 +189,10 @@ module EdoOracle
 
       def merge_cross_listed_titles(course)
         if (course[:catid].start_with? 'C') && !course[:name]
-          if (title_results = EdoOracle::Queries.get_cross_listed_course_title course[:course_code])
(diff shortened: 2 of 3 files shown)


paulkerschen:
I approve this work! but will let someone on the SIS side review and hit the button.

app/models/edo_oracle/user_courses/base.rb
```diff
@@ -189,7 +189,10 @@ def row_to_section_data(row, cross_listing_tracker=nil)
 
       def merge_cross_listed_titles(course)
         if (course[:catid].start_with? 'C') && !course[:name]
-          if (title_results = EdoOracle::Queries.get_cross_listed_course_title course[:course_code])
+          title_results = self.class.fetch_from_cache "cross_listed_title-#{course[:course_code]}" do
+            EdoOracle::Queries.get_cross_listed_course_title course[:course_code]
+          end
+          if title_results.present?
             course[:name] = title_results['course_title'] || title_results['course_title_short']
```
redconfetti: @raydavis - Okay, so even though we get this data and cache it within [EdoOracle::UserCourses::All](https://github.com/raydavis/calcentral/blob/31d9f348/app/models/edo_oracle/user_courses/all.rb) and [EdoOracle::UserCourses::SelectedSections] as a larger computed data set, this will make it so that at least this expensive crosslisted courses lookup is stored in a shared course specific cache store.

For a course with course code 'CHEM101', wouldn't it have the same data cached twice in both `EdoOracle::UserCourses::All/CHEM101` and `EdoOracle::UserCourses::SelectedSections/CHEM101`... but this is fine because it's really only cacheing two course title strings... small data set, but expensive to obtain.

app/models/edo_oracle/queries.rb
```diff
@@ -278,7 +278,7 @@ def self.get_cross_listed_course_title(course_code)
           TRIM(crs."transcriptTitle") AS course_title_short
         FROM SISEDO.API_CROSSLISTINGSV00_VW xlist
         JOIN SISEDO.API_COURSEV00_VW crs ON
-          (xlist."cms-version-independent-id" = crs."cms-version-independent-id")
+          (xlist."cms-id" = crs."cms-id")
         WHERE xlist."displayName" = '#{course_code}'
```
redconfetti: @raydavis - I take it that `cms-id` was a better index to use than `cms-version-independent-id`?

PR #6217, opened 2016-11-30
CLC-6520, 'Create a Site' button after Canvas dashboard_sidebar renders

johncrossman:

Info on jQuery selectors in play:
* http://api.jquery.com/nth-last-child-selector/
* http://api.jquery.com/prepend/

My initial solution (PR #6215) was faulty. We cannot rely on `waitUntilAvailable('#start_new_course', ...)` because that button does not always render (see default Canvas logic). Therefore, UC Berkeley's custom rules (e.g., staff can 'Create a Site') won't work; the replace logic will never fire.

### DIFF
--- a/public/canvas/canvas-customization.js
+++ b/public/canvas/canvas-customization.js
@@ -111,7 +111,7 @@
               });
 
               // Add the 'Create a Site' button to the Dashboard page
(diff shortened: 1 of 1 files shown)


raydavis:
Thanks, @johncrossman

PR #6221, opened 2016-11-30
CLC-6520, select #right-side child w/o class for prepend 'Create a Site'

johncrossman:

PR #6217 did not do the job. See Steven Williams comments in Jira above about unwanted buttons in "To Do" and "Recent Activity" items. This PR targets the no-class `<div>` within immediate children of `#right-side`.

DO NOT MERGE until Steven Williams does initial validation at https://ucberkeley.beta.instructure.com


raydavis:
Steven has verified the fix.

PR #6222, opened 2016-11-30
[QA] CLC-6520, select #right-side child w/o class for prepend 'Create a Site'

johncrossman:

master PR #6221 

cc: @paulkerschen


paulkerschen:
Bueno.

PR #6239, opened 2016-12-01
SISRP-28333 - MyAcademics feed taking over 90 sec to load for faculty user in CC-QA

nkutub:
https://jira.berkeley.edu/browse/SISRP-28333

* Added instance level variable for grading log issue. 
* used cached proxy from FacultyDelegate API calls, this will help for instructors with the same set of sections
* added logic to only apply FacultyDelegate to semester for student users and only teaching semesters for non-student users, this should improve performance by at least 50%

### DIFF
--- a/app/models/campus_solutions/faculty_delegate.rb
+++ b/app/models/campus_solutions/faculty_delegate.rb
@@ -1,5 +1,5 @@
 module CampusSolutions
-  class FacultyDelegate < Proxy
+  class FacultyDelegate < CachedProxy
 
     def initialize(options = {})
       super options
--- a/app/models/my_academics/faculty_delegate.rb
+++ b/app/models/my_academics/faculty_delegate.rb
@@ -2,14 +2,15 @@ module MyAcademics
   class FacultyDelegate < UserSpecificModel
 
     def merge(data)
-      teaching_semesters = data[:teachingSemesters]
-      if teaching_semesters
-        add_grd_access_to_semesters(teaching_semesters)
-      end
-      semesters = data[:semesters]
+      semesters = is_user_student? ? data[:semesters] : data[:teachingSemesters]
       if semesters
         add_grd_access_to_semesters(semesters)
       end
+    end
+
+
+    def is_user_student?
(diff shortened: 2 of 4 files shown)


app/models/campus_solutions/faculty_delegate.rb
```diff
@@ -1,5 +1,5 @@
 module CampusSolutions
-  class FacultyDelegate < Proxy
+  class FacultyDelegate < CachedProxy
```
raydavis: @nkutub , did you check with DEBUG level to see how the feed value will be cached? To me it looks like the feed value depends on ``term_id`` and ``course_id``, but all feed values will end up with the same global cache key, causing the wrong results to be read from cache. What I expected to see is an implementation of ``instance_key``, as seen in [AddressLabel](https://github.com/ets-berkeley-edu/calcentral/blob/v84.1/app/models/campus_solutions/address_label.rb).

app/models/my_academics/faculty_delegate.rb
```diff
@@ -2,16 +2,17 @@ module MyAcademics
   class FacultyDelegate < UserSpecificModel
 
     def merge(data)
-      teaching_semesters = data[:teachingSemesters]
-      if teaching_semesters
-        add_grd_access_to_semesters(teaching_semesters)
-      end
-      semesters = data[:semesters]
+      semesters = is_user_student? ? data[:semesters] : data[:teachingSemesters]
```
raydavis: I need to acquaint myself with the current UX design, but I'm curious about whether this either-or split will work correctly for GSIs.

raydavis: Also, since the ``is_user_student?`` check is specific to "now" rather than covering past academic terms, I'm curious about what the desired UX is for past terms, for both students and instructors.

raydavis: Also, since the is_user_student? check is specific to "now" rather than covering past academic terms, I'm curious about what the desired UX is for past terms, for both students and instructors.

PR #6255, opened 2016-12-02
SISRP-28450 - Fixing hover-over text on Class page Grading tab

redconfetti:
https://jira.berkeley.edu/browse/SISRP-28450

See QA PR #6256


redconfetti:
The primary change in this is simply adding the 'title' attributes to the anchor tags. I'm only expanding the formatting so it's easier to read otherwise.

nkutub:
good, catch @redconfetti

redconfetti:
It was actually Deepthika that caught it.

PR #6258, opened 2016-12-05
SISRP-28333 - MyAcademics feed taking over 90 sec to load for faculty user in CC-QA

nkutub:
https://jira.berkeley.edu/browse/SISRP-28333

* removed dependency on faculty_delegate API 
* deleted model and spec for faculty_delegate feed
* added gradeRosterAccess column to edoDB instructor query 
* changed MyAccademic::FacultyDelegate to depend on edoDB query and not API 
* small spelling corrections. 
* used instance level var for MyAcademic::Grading to reduce log noise.


redconfetti:
LGTM

davidlimcheng:
LGTM.  nice @nkutub

PR #6276, opened 2016-12-12
NOJIRA Clarify variable names for section IDs

raydavis:
This kept confusing me a couple of weeks back, so I finally lost my temper and fixed it.


paulkerschen:
Thanks @raydavis. I wonder why I put those trousers on backward in the first place.

PR #6270, opened 2016-12-09
SISRP-27468 - Student Overview - Personal Summary card - Update 'Schedule Planer' link

redconfetti:
https://jira.berkeley.edu/browse/SISRP-27468

PR #6283, opened 2016-12-13
SISRP-28705 - Adds ADVISORID to all college scheduler API requests

redconfetti:
https://jira.berkeley.edu/browse/SISRP-28705

See QA PR #6284. NOTE: The QA PR is much different due to changes not being present from #6270 in the QA branch

### DIFF
--- a/app/models/campus_solutions/college_scheduler_url.rb
+++ b/app/models/campus_solutions/college_scheduler_url.rb
@@ -25,11 +25,12 @@ module CampusSolutions
     def url
       if @student_uid.present? && student_campus_solutions_id = CalnetCrosswalk::ByUid.new(user_id: @student_uid).lookup_campus_solutions_id
         emplid = student_campus_solutions_id
-        advisor_id_param = "&ADVISORID=#{@campus_solutions_id}"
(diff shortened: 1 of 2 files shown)


redconfetti:
FYI: The API requires that the ADVISORID parameter be last in the request URL or else there is an HTTP 500 error.

app/models/campus_solutions/college_scheduler_url.rb
```diff
@@ -25,11 +25,12 @@ def xml_filename
     def url
       if @student_uid.present? && student_campus_solutions_id = CalnetCrosswalk::ByUid.new(user_id: @student_uid).lookup_campus_solutions_id
         emplid = student_campus_solutions_id
-        advisor_id_param = "&ADVISORID=#{@campus_solutions_id}"
+        advisor_id = campus_solutions_id
```
lyttam: Also, what will the url look like if `advisor_id` is `nil`?  If you have time you could add a unit test for that case.

redconfetti: @lyttam - Great catch. Yes. I accidentally forgot the `@`

PR #6284, opened 2016-12-13
[QA] SISRP-28705 - Adds ADVISORID to all college scheduler API requests

redconfetti:
https://jira.berkeley.edu/browse/SISRP-28705

See DEV PR #6283

### DIFF
--- a/app/models/campus_solutions/college_scheduler_url.rb
+++ b/app/models/campus_solutions/college_scheduler_url.rb
@@ -22,7 +22,7 @@ module CampusSolutions
     end
 
     def url
-      "#{@settings.base_url}/UC_SR_COLLEGE_SCHDLR_URL.v1/get?EMPLID=#{@campus_solutions_id}&STRM=#{@term_id}&ACAD_CAREER=#{@acad_career}&INSTITUTION=UCB01"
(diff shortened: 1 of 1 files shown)


redconfetti:
Background: An update was made to the CS API for College Scheduler URL where the `ADVISORID` parameter was added to support providing the single sign on (SSO) link for Advisors (see [SISRP-19662](https://jira.berkeley.edu/browse/SISRP-19662)).

It wasn't clear that the new parameter is required and must be at the end of the HTTP request URL, and thus the existing student link in the Class Enrollment card was broken because the API has been returning an HTTP 500 error in production. This should resolve the error, thus bringing the functionality back for Students in v85.2.

davidlimcheng:
@redconfetti.  Makes sense.  So I assume, from the code/JIRAs, that the link is only currently exposed to students?  And, once we cut the new QA branch for v86, the new code in Master that currently handles the advisor SSO will take over.  Am I understanding that correctly?

redconfetti:
@davidlimcheng - Yes. This model obtains the College Scheduler URL for a student from the CS API. This seeks to fix that request so that the CS API returns the data instead of an HTTP 500 error.

The DEV PR updates a more complex version of this model which is affected by changes made in #6270, which updates the model to handle both Student and Advisor SSO.

Essentially CS updated their API in the last sprint, but we didn't release the Advisor functionality in v85, which is how this bug slipped through the cracks into production.

PR #6285, opened 2016-12-13
SISRP-27699 - Implements strategy for identifying summer visiting students

lyttam:
This is the backend work required to support https://jira.berkeley.edu/browse/SISRP-27699.

We needed a way for every page in calcentral to be aware of the roles that are assigned to a student based on their academic status (plan or career).  Prior to this, that information was added to the feed in `CollegeAndLevel` and only available on the My Academics page.

I refactored the roles logic into a new model called `MyAcademicStatus`, which takes the feed from `AcademicStatus` and inserts the roles into it.  The roles will now be included in the `CollegeAndLevel` feed and in `HubEdoController#academic_status`.  `GpaUnits` and `AdvisingStudentController` are still calling `AcademicStatus` directly unless anyone sees a reason to switch those over too.

I also added a new role that will be assigned to students in a "summer visitor" plan (there are a number of plans that have this designation).

### DIFF
--- a/app/models/berkeley/academic_roles.rb
+++ b/app/models/berkeley/academic_roles.rb
@@ -1,0 +1,56 @@
+module Berkeley
+  module AcademicRoles
+    extend self
+
+    ACADEMIC_PLAN_ROLES = [
+      {role_code: 'fpf', match: ['25000FPFU'], types: [:enrollment]},
+      {role_code: 'haasFullTimeMba', match: ['70141MBAG'], types: []},
+      {role_code: 'haasEveningWeekendMba', match: ['701E1MBAG'], types: []},
+      {role_code: 'haasExecMba', match: ['70364MBAG'], types: []},
+      {role_code: 'haasMastersFinEng', match: ['701F1MFEG'], types: []},
+      {role_code: 'haasMbaPublicHealth', match: ['70141BAPHG'], types: []},
+      {role_code: 'haasMbaJurisDoctor', match: ['70141BAJDG'], types: []},
+      {role_code: 'ugrdUrbanStudies', match: ['19912U'], types: []},
+      {
+        role_code: 'summerVisitor',
+        match: %w(99000U 99000INTU 99000G 99000INTG 99V03U 99V04U 99V05U 99V09U 99V03G 99V05G 99V06G 99V07G 99V08G 99V10G 99V06U 99V07U 99V08U 99V10U 99V02G 99V04G 99V09G),
+        types: []
+      },
+    ]
+
+    ACADEMIC_CAREER_ROLES = [
+      {role_code: 'law', match: ['LAW'], types: [:enrollment]},
+      {role_code: 'concurrent', match: ['UCBX'], types: [:enrollment]}
+    ]
+
+    def get_academic_plan_role_code(plan_code, type = nil)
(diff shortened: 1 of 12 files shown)


redconfetti:
I'm in the process of reviewing this, but must focus on facilitating QA for emergency release. I will certainly review this and sign off sometime today.

app/models/berkeley/academic_roles.rb
```diff
@@ -0,0 +1,56 @@
...
+      role_codes = roles.select do |matcher|
+        is_match = matcher[:match].include?(academic_code)
+        is_type = type.blank? || matcher[:types].include?(type.to_sym)
+        is_match && is_type
+      end
+      role_codes.empty? ? nil : role_codes.first[:role_code]
+    end
+
+    def role_codes
+      (ACADEMIC_PLAN_ROLES | ACADEMIC_CAREER_ROLES).collect do |matcher|
```
redconfetti: @lyttam - Hm... I didn't know about the pipe operator and it's use with arrays.

app/models/my_academics/college_and_level.rb
```diff
@@ -102,6 +76,10 @@ def parse_hub_holds(response)
       holds
     end
 
+    def parse_hub_roles(response)
+      response.try(:[], :feed).try(:[], 'student').try(:[], 'roles')
+    end
```
redconfetti: Do you think roles grow on trees? You can just reach up and pick them off the tree?!?!

Yes mother! Now you can!!!

redconfetti:
@lyttam - I like that you're moving the cacheing to this model that will extract, transform, and merge useful data into the same structure that will then be cached. This is a step in the direction we discussed where one model handles the extracting/transforming, and the other acts as a simple proxy. I really like that you put the Berkeley specific business logic in a module.

I am noticing that you're still leaving dependencies that rely on the non-cached HubEdos::AcademicStatus proxy such as:

* [app/controllers/hub_edo_controller.rb](https://github.com/lyttam/calcentral/blob/SISRP-27699/app/controllers/hub_edo_controller.rb#L5) - There is front-end code that still relies on this, and without the cacheing it would affect performance
* [app/models/my_academics/gpa_units.rb](https://github.com/lyttam/calcentral/blob/SISRP-27699/app/models/my_academics/gpa_units.rb#L20) - I know that this data is cached by MyAcademics::Merged, but there could be a slight affect on performance if it's reading from a non-cached data source when when already exists (the new HubEdos::MyAcademicStatus).

I'm not sure if you've evaluated whether or not [app/models/notifications/sis_expiry_processor.rb](https://github.com/lyttam/calcentral/blob/SISRP-27699/app/models/notifications/sis_expiry_processor.rb#L66) needs to be updated to expire HubEdos::MyAcademicStatus instead.

lyttam:
Ah, @redconfetti, that's the reason to switch `GpaUnits` and `AdvisingStudentController` over to use `MyAcademicStatus`.  Thanks for pointing that out (`HubEdoController` is already calling the new model).

Looks like I need to update `SisExpiryProcessor` as well. Will push some more changes to this PR shortly.

lyttam:
Updated with @redconfetti's suggestions.  Nobody is calling `AcademicStatus` directly anymore.

PR #6291, opened 2016-12-16
SISRP-28993 - Not Enrolled status for students who are enrolled in classes - CC Status and Holds (INC0376834)

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-28993

* The value of `unitsEnrolled` coming from the Registrations API has more of an "units in progress" definition in CS.  So, at the end of the term, `unitsEnrolled` drops to zero.  `unitsTaken` then becomes the value we need to look at.
* Also tried to clean up the code to be a bit more DRY.


raydavis:
Sneaky!

PR #6297, opened 2016-12-21
SISRP-25362 - 7.4 - Student Overview (Advisor) Degree Progress card for Grad and Law Careers

davidlimcheng:

* Biggest change on our end (lot of stuff going on in that ticket) is opening this up to students who have *any* Degree Progress data - not just current students.
* Also reformatted front-end to use one spinner for both graduate and undergraduate data.  There's a small population who might have both a graduate and an undergraduate Degree Progress card, but for the most part, people will only have one.
* This brought up a design question though - as we are introducing yet another spinner to the User Overview page.  Do we go the My Academics (Student) route and use one spinner for everything?  But that would delay user's views of ready-to-go data.  I discussed this with @lyttam and Chara and while it is a larger question that should include both designers and developers, I think we are getting to the point  where the data we are gathering from different sources is becoming so large and fractured that we need to start looking for a solution.  The open-endedness of this question is ultimately why we decided not to have this merged with `MyAcademics::FilteredForAdvisors` at this time.

### DIFF
--- a/src/assets/templates/widgets/academics/degree_progress_graduate.html
+++ b/src/assets/templates/widgets/academics/degree_progress_graduate.html
@@ -2,25 +2,24 @@
   <div class="cc-widget-title">
     <h2>Degree Progress</h2>
   </div>
-  <div class="cc-widget-padding" data-cc-spinner-directive="degreeProgressGraduate.isLoading">
-    <div data-ng-if="degreeProgressGraduate.errored">There was an error retrieving degree progress data.</div>
-    <div data-ng-if="!degreeProgressGraduate.errored && !degreeProgressGraduate.progresses.length">No degree progress data is available.</div>
-    <ul data-ng-if="!degreeProgressGraduate.errored && degreeProgressGraduate.progresses" class="cc-list-links">
-      <li data-ng-if="degreeProgressGraduate.links.advancementFormSubmit.url">
+  <div class="cc-widget-padding">
+    <div data-ng-if="degreeProgress.graduate.errored">There was an error retrieving graduate degree progress data.</div>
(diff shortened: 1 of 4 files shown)


src/assets/templates/widgets/academics/degree_progress_graduate.html
```diff
@@ -2,25 +2,24 @@
   <div class="cc-widget-title">
     <h2>Degree Progress</h2>
   </div>
-  <div class="cc-widget-padding" data-cc-spinner-directive="degreeProgressGraduate.isLoading">
-    <div data-ng-if="degreeProgressGraduate.errored">There was an error retrieving degree progress data.</div>
-    <div data-ng-if="!degreeProgressGraduate.errored && !degreeProgressGraduate.progresses.length">No degree progress data is available.</div>
```
lyttam: @davidlimcheng If there's no degree progress data, don't we still want this message?  Without it, we'd just be showing an empty card. It's possible for a student to have neither grad nor undergrad degree progress data.

davidlimcheng: @lyttam I actually put that check on `user_overview.html`:
```
data-ng-if="api.user.profile.features.csDegreeProgressGradAdvising && degreeProgress.graduate.progresses.length"
```
So instead of showing two cards for everybody, we'd hide the card that does not apply to them.  And if they have no data for either, we'd hide both cards.

lyttam: @davidlimcheng cool, that definitely sounds better than showing one card with a "no data" message and the other card with data.

raydavis:
@lyttam Is this ready to merge so far as you're concerned?

lyttam:
@raydavis @davidlimcheng yep good to merge.

PR #6299, opened 2016-12-22
SISRP-27699 - Customized UX for Summer Visitor students

lyttam:
https://jira.berkeley.edu/browse/SISRP-27699

Follow up to the backend work to create a new role to assign to students who have a "summer visitor" academic plan (#6285)

We want to hide stuff from summer visitors that isn't relevant to them.  I plan to do some further refactoring to perhaps get rid of `academicStatusFactory` and use `academicsService` to handle the holds data (will cover this in a separate PR).

Here's an explanation of all the UI changes:  https://docs.google.com/document/d/1eX43bf4WbrHuOQTPq0_QOrJ_kUmBgd8bYPPlhBEJzGk/edit#

### DIFF
--- a/src/assets/javascripts/angular/controllers/pages/dashboardController.js
+++ b/src/assets/javascripts/angular/controllers/pages/dashboardController.js
@@ -1,26 +1,17 @@
 'use strict';
 
-var _ = require('lodash');
 var angular = require('angular');
 
 /**
  * Dashboard controller
  */
-angular.module('calcentral.controllers').controller('DashboardController', function($scope, apiService, academicStatusFactory, userService) {
-  $scope.academicStatus = {
-    roles: {}
-  };
-
-  var parseAcademicStatusRoles = function(data) {
-    _.extend($scope.academicStatus.roles, _.get(data, 'data.feed.student.roles'));
-  };
+angular.module('calcentral.controllers').controller('DashboardController', function($scope, apiService) {
 
   var init = function() {
     if (apiService.user.profile.hasDashboardTab) {
       apiService.util.setTitle('Dashboard');
-      academicStatusFactory.getAcademicStatus().then(parseAcademicStatusRoles);
     } else {
-      userService.redirectToHome();
+      apiService.user.redirectToHome();
     }
   };
 
--- a/src/assets/javascripts/angular/controllers/pages/dashboardController.js
+++ b/src/assets/javascripts/angular/controllers/pages/dashboardController.js
@@ -1,14 +1,24 @@
 'use strict';
 
+var _ = require('lodash');
 var angular = require('angular');
 
 /**
  * Dashboard controller
  */
-angular.module('calcentral.controllers').controller('DashboardController', function($scope, apiService, userService) {
+angular.module('calcentral.controllers').controller('DashboardController', function($scope, apiService, academicStatusFactory, userService) {
+  $scope.academicStatus = {
+    roles: {}
+  };
+
+  var parseAcademicStatusRoles = function(data) {
+    _.extend($scope.academicStatus.roles, _.get(data, 'data.feed.student.roles'));
+  };
+
   var init = function() {
     if (apiService.user.profile.hasDashboardTab) {
       apiService.util.setTitle('Dashboard');
+      academicStatusFactory.getAcademicStatus($scope).then(parseAcademicStatusRoles);
     } else {
       userService.redirectToHome();
     }
--- a/src/assets/javascripts/angular/controllers/pages/dashboardController.js
+++ b/src/assets/javascripts/angular/controllers/pages/dashboardController.js
@@ -18,7 +18,7 @@ angular.module('calcentral.controllers').controller('DashboardController', funct
   var init = function() {
     if (apiService.user.profile.hasDashboardTab) {
       apiService.util.setTitle('Dashboard');
-      academicStatusFactory.getAcademicStatus($scope).then(parseAcademicStatusRoles);
+      academicStatusFactory.getAcademicStatus().then(parseAcademicStatusRoles);
     } else {
       userService.redirectToHome();
     }
--- a/src/assets/javascripts/angular/factories/campusLinksFactory.js
+++ b/src/assets/javascripts/angular/factories/campusLinksFactory.js
@@ -1,5 +1,6 @@
 'use strict';
 
+var _ = require('lodash');
 var angular = require('angular');
 
 /**
@@ -31,20 +32,22 @@ angular.module('calcentral.factories').factory('campusLinksFactory', function(ap
     }
   };
 
-  /**
-   * Depending on the roles, determine whether the current user should be able to view the link
-   * @param {Object} link Link object
-   * @return {Boolean} Whether the user should be able to view the link
-   */
-  var canViewLink = function(link) {
-    for (var i in link.roles) {
-      if (link.roles.hasOwnProperty(i) &&
-          link.roles[i] === true &&
-          link.roles[i] === apiService.user.profile.roles[i]) {
-        return true;
-      }
-    }
-    return false;
+  var hasWhitelistedRole = function(linkRoles) {
+    var userRoles = apiService.user.profile.roles;
+    return _.some(linkRoles, function(value, key) {
+      return userRoles[key] === value;
+    });
+  };
+
+  var hasBlacklistedRole = function(linkRoles) {
+    var academicRoles = apiService.academics.roles;
+    return _.some(academicRoles, function(hasRole, role) {
+      return hasRole && linkRoles[role] === false;
+    });
+  };
+
+  var userCanView = function(link) {
+    return hasWhitelistedRole(link.roles) && !hasBlacklistedRole(link.roles);
   };
 
   /**
@@ -74,14 +77,14 @@ angular.module('calcentral.factories').factory('campusLinksFactory', function(ap
       topcategories: {}
     };
     angular.forEach(links, function(link) {
-      var canUserViewLink = canViewLink(link);
-      if (canUserViewLink) {
+      if (userCanView(link)) {
         addToTopCategories(link, response.topcategories);
-      }
-      if (canUserViewLink && isLinkInCategory(link, currentTopCategory)) {
-        response.links.push(link);
-        for (var i = 0; i < link.subCategories.length; i++) {
-          addToSubcategories(link.subCategories[i], response.subcategories);
+
+        if (isLinkInCategory(link, currentTopCategory)) {
+          response.links.push(link);
+          for (var i = 0; i < link.subCategories.length; i++) {
+            addToSubcategories(link.subCategories[i], response.subcategories);
+          }
         }
       }
     });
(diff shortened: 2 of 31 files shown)


src/assets/javascripts/angular/factories/campusLinksFactory.js
```diff
@@ -31,20 +32,22 @@ angular.module('calcentral.factories').factory('campusLinksFactory', function(ap
...
-    return false;
+  var hasWhitelistedRole = function(linkRoles) {
+    var userRoles = apiService.user.profile.roles;
+    return _.some(linkRoles, function(value, key) {
+      return userRoles[key] === value;
+    });
+  };
+
+  var hasBlacklistedRole = function(linkRoles) {
+    var academicRoles = apiService.academics.roles;
```
raydavis: If I'm reading the PR correctly, the big functional difference between normal "profile" roles and these "academics" roles is that the latter are treated as a blacklist in links. Since a lot of people might consider "student" or "teacher" or "advisor" to be academic roles, this could set us up for unexpected disappearances for GSIs, or for advisors who are instructors, etc. Would it be possible to put a big bold blinking warning to that effect at the code which fills in the roles list?

src/assets/javascripts/angular/controllers/pages/dashboardController.js
```diff
@@ -1,14 +1,24 @@
...
+  };
+
+  var parseAcademicStatusRoles = function(data) {
+    _.extend($scope.academicStatus.roles, _.get(data, 'data.feed.student.roles'));
+  };
+
   var init = function() {
     if (apiService.user.profile.hasDashboardTab) {
       apiService.util.setTitle('Dashboard');
+      academicStatusFactory.getAcademicStatus().then(parseAcademicStatusRoles);
```
raydavis: This looks like it might further delay the already sometimes disconcerting wait to get to a laid-out web page when a person first logs in. Is that accurate? If so, would there be any way to narrow down the impact? (I won't be at all surprised by a "No" to that question, by the way!)

raydavis:
I like the approach you've taken here, @lyttam , and I hope you had a good regex search-and-replace on hand when you edited campuslinks.json! Just had a couple of queries...

src/assets/templates/widgets/status_and_holds.html
```diff
@@ -96,9 +96,9 @@ <h4 data-ng-bind="registration.name"></h4>
       </div>
     </div>
 
-    <div data-cc-spinner-directive="residency.isLoading">
+    <div data-cc-spinner-directive="residency.isLoading" data-ng-if="showResidency">
```
davidlimcheng: @lyttam We've had problems in the past putting an `ngIf` in the same node as `data-cc-spinner-directive`.  I believe what happened was the `ngIf` negated the spinner, and the card would just appear.  Did this test OK with the expected behavior?

src/assets/javascripts/angular/services/academicsService.js
```diff
@@ -4,7 +4,12 @@
 var _ = require('lodash');
 var angular = require('angular');
 
-angular.module('calcentral.services').service('academicsService', function() {
+angular.module('calcentral.services').service('academicsService', function($http, $route, $routeParams) {
+  // var urlAcademicStatus = '/dummy/json/hub_academic_status.json';
+  var urlAcademicStatus = '/api/edos/academic_status';
```
davidlimcheng: @lyttam In keeping with our front-end organization, `academicsFactory.js` seems like it would be a better place for this ajax call to live.  Is there a particular reason you're making the call here instead?

src/assets/javascripts/angular/services/academicsService.js
```diff
@@ -290,8 +295,23 @@ angular.module('calcentral.services').service('academicsService', function() {
     }
   };
 
+  var handleAcademicStatus = function(data) {
+    angular.extend(roles, data.feed.student.roles);
+  };
+
+  var fetch = function() {
+    var url = $route.current.isAdvisingStudentLookup ? urlAdvisingAcademicStatus + $routeParams.uid : urlAcademicStatus;
+    return $http.get(url, {
```
davidlimcheng: @lyttam If we do choose to leave this call here, would we want to use `apiService.http.request` instead, to preserve front-end cache?

src/assets/javascripts/angular/services/profileMenuService.js
```diff
@@ -228,8 +241,13 @@ angular.module('calcentral.services').factory('profileMenuService', function(api
     });
   };
 
+  var getAcademics = function() {
+    return apiService.academics.fetch();
+  };
+
   var getNavigation = function() {
     return apiService.user.fetch()
+      .then(getAcademics)
```
davidlimcheng: Just a general question for me to clarify my understanding of this - does this call open up `var academicRoles = apiService.academics.roles;` for use?

src/assets/javascripts/angular/factories/campusLinksFactory.js
lyttam: @raydavis Yes, definitely.  I realized somewhat late in the game that we need to treat the academic role list as a blacklist, which is different than user roles (that's why I had to avoid combining them into a single list of roles).

src/assets/javascripts/angular/controllers/pages/dashboardController.js
lyttam: @raydavis yes, unfortunately it will make that initial page load even slower.  I might be able to improve the experience by moving this call to the controller for the card that needs the data, rather than doing it in the top-level dashboard controller (though if other parts of the page wind up needing the data I'll have to move it back).  

In general, I would like to pursue a strategy of rendering whatever parts of the page that we can while the data is being fetched (i.e., the parts of the page that don't rely on that data), since that can give the illusion that our site is faster than it is.

src/assets/javascripts/angular/services/academicsService.js
lyttam: @davidlimcheng yeah, I had it only in `academicStatusFactory` but then had to create this service to use in `profileMenuService` once I realized the roles were needed there (I have a feeling that a service should not be calling a factory - at least that seems convoluted to me).  I was planning to refactor and get rid of `academicStatusFactory` entirely, and make `academicsService` responsible for doing the ajax call.  Does that go against the pattern of the rest of the code?  I spent some time going back and forth on it so I'd like to know your thoughts.

src/assets/javascripts/angular/services/academicsService.js
lyttam: @davidlimcheng I made `academicsService` call `$http` directly instead of calling `apiService.http` because `apiService` exposes both `academicsService` and `httpService` in the same object, and it felt too much like a circular dependency to have one of those calling the other.  I believe this will still do caching the same way `httpService` does.

src/assets/javascripts/angular/services/academicsService.js
davidlimcheng: @lyttam Hmm I see what you mean.  The only reason why I would think it'd be better to put it in its own Factory (and I do agree that calling a Factory from a Service is a bit convoluted) is if other features need to access the same data.  In my opinion, having this service as the only access point from the front-end strongly couples it with Academics Data, and if a Financial Aid feature in the future requires Academic Status data for whatever reason, it might be even more convoluted to call `AcademicsService.fetch`, when every other pattern in our repository points to Factories for these calls.

That being said I think this is very preferential, and would love to see what the other devs think as well.

src/assets/javascripts/angular/controllers/pages/dashboardController.js
raydavis: We could also restrict it to "student" roles, but of course those are the bulk of our users anyway... Like I say, no surprise; if the lag becomes unendurable, we can tackle it as a follow-up.

src/assets/javascripts/angular/factories/campusLinksFactory.js
raydavis: Thanks, @lyttam ! As long as that b.b.b. warning gets into the code somehow, I'm OK with @davidlimcheng handling the merge.

src/assets/templates/widgets/status_and_holds.html
lyttam: @davidlimcheng Yes, I didn't notice any issues like that.  The spinner shows until the data is present if the `ngIf` evaluates to true.

This is one of those problem spots where we have 2 spinners on the same card for different sets of data. I guess I'll leave it alone for now though.

src/assets/javascripts/angular/services/academicsService.js
lyttam: @davidlimcheng yeah, I would benefit from a discussion on when to use services vs. factories.  From what I've read, they are sort of interchangeable, but it looks like we're using them in a somewhat opinionated way and I would like to understand the opinion (and decide if we want to stick with it going forward).

To me, it doesn't seem like a service is coupling this with academics data any more than a factory would - either one needs to make the same academic-status API call and then parse the roles from the response.  I noticed that our factories are usually setting stuff in $scope, while our services typically expose data so that it can be referenced like `apiService.academics.roles` which seems more flexible to me.  I'm not sure why it was done that way, since I don't think Angular enforces that pattern at all.  I'd like to know what you and @nkutub, @raydavis, and @redconfetti think.

As far as Financial Aid feature requiring academics data, it appears that's the new paradigm... Dashboard and Campus Links also depend on it now.  The academic status data is how the academic roles are derived, and those roles are now needed on every single page.

src/assets/javascripts/angular/services/profileMenuService.js
lyttam: @davidlimcheng any service, controller, or factory that includes `apiService` can access `apiService.academics.roles`, as long as `apiService.academics.fetch` has been called and the promise object that it returns has been resolved. 

The `var academicRoles` is only accessible inside the scope of function `hasBlacklistedRole` though.

raydavis:
@lyttam , Travis jshint failed on ``dashboard.js`` -- looks legit.

src/assets/javascripts/angular/services/academicsService.js
raydavis: The convention in CalCentral from early on was to use factories to interact with the server (kind of the browser-side equivalent of our server-side ``Proxy`` convention) and services for other code shared across controllers. Like most conventions, it's mainly useful so we don't have to think so much. "Where is this page actually getting that data?" "Cherchez la factory!"

lyttam: @raydavis good to know; I've gone the wrong direction then. I can refactor this back into a factory, time-permitting.

So it seems it would not be too weird for a service to call a factory?  Say if the factory were responsible for fetching data from the server, and the service were responsible for manipulating it in a way that's needed by multiple controllers.

lyttam:
@raydavis fixed Travis; all good to merge now.

src/assets/javascripts/angular/services/academicsService.js
lyttam: @raydavis Agreed, I definitely don't want to leave both the service and the factory in there. The factory is hanging around until I refactor the holds-related bit (which is unrelated to the work in this PR, but needs to be cleaned up).

It would seem there are People on the Internet who have Strong Opinions about service vs factory in angular.  We're using them kind of as if they were interchangeable (having our services expose an API via an object, a la revealing module pattern) - and in fact, some of our "services" are actually factories (see `profileMenuService`).  Nothing wrong with this approach at all; I like the idea of having a convention to make it easier to find stuff in the code.

While 'service' sounds to me like a thing that fetches data, and 'factory' sounds like a thing that generates instances of an object, I want to stick with our convention to avoid confusion.  I'd like to get this merged so folks can start testing, but I can submit another PR with some refactoring work.  cc @davidlimcheng

raydavis: A follow-up PR sounds fine to me, @lyttam

davidlimcheng: Great discussion, @lyttam and @raydavis!  Clarified a lot of things for me as well.

PR #6300, opened 2016-12-22
SISRP-28848 - My Academics(Grad) - Higher Degree Committees - Qualifing Exam Approval data needs to be fixed.

nkutub:
https://jira.berkeley.edu/browse/SISRP-28848

* changed committee type mapping 
* enabled non-active members for non-active committees on student view.


app/models/my_committees/committees_module.rb
```diff
@@ -58,22 +58,22 @@ def committee_student_photo_url (cs_committee)
   end
 
   def committee_status_title (cs_committee)
-    if cs_committee[:committeeType].to_s == 'OUA_FNLZN'
+    if cs_committee[:committeeType].to_s == 'QE'
```
lyttam: @nkutub you could make 'QE' a constant (e.g. `COMMITTEE_TYPE_EXAM_DATE = 'QE`) and then you'd only have to change it in one spot next time.

nkutub: @lyttam I meant to do that and keep forgetting, good call.

davidlimcheng:
LGTM, @nkutub.

PR #6302, opened 2016-12-22
SISRP-29182 Remove obsolete APIs before they cause more breakage

raydavis:
https://jira.berkeley.edu/browse/SISRP-29182

"18 additions and 6,431 deletions" -- now, _that_ feels like a good day's work!


raydavis:
Amended to fix jscs fail.

app/models/finaid/fin_aid_year.rb
```diff
@@ -1,4 +1,5 @@
 module Finaid
+  # TODO : DROP THIS TABLE ONE RELEASE AFTER REMOVAL OF LEGACY FINAID SUPPORT.
```
lyttam: @raydavis do we have a corresponding Jira to remember this by, or some other way of tracking/scheduling this work?

raydavis: Yes, @lyttam , but thank you for asking: https://jira.berkeley.edu/browse/SISRP-29183

raydavis:
Unshockingly, @lyttam 's massive PR conflicts with my massive PR. I'll amend ASAP

raydavis:
Amended.

lyttam:
Wow, @raydavis, this is a lot of cleanup!  So, just so I understand, regstatus and myfinaid were giving us pre-Fall16 "current and future" terms - and, now that we're past Fall 2016, there would be no more terms for those APIs to give?

raydavis:
Right, @lyttam . None of those APIs were reliable for past history (although they failed in different and sometimes misleading ways).

src/assets/templates/myfinances.html
```diff
@@ -12,7 +12,6 @@ <h1 class="cc-heading-page-title">My Finances</h1>
 
     <div class="medium-6 large-4 cc-column-2 columns">
       <div data-ng-include="'widgets/finaid_summary.html'" data-ng-if="api.user.profile.features.csFinAid && !academicStatus.roles.summerVisitor"></div>
-      <div data-ng-include="'widgets/finaid_activity.html'" data-ng-if="!api.user.profile.isApplicantOnly"></div>
```
nkutub: @raydavis, so reading this I'm assuming that there is no alternative source for finaid activity and that it is just not a feature we are providing anymore?

raydavis: @nkutub , the old FinAid API & UX has been replaced by new CS APIs & UX. I wasn't much involved in the transition so I can't say item-by-item what functionality might have been added or dropped. Either way, even in the pre-CS world, the finaid_activity card for the 2015-2016 academic year would be gone by now. Does that clarify?

nkutub: @raydavis good enough for now, I will eventually get my arms around all the functions CC provides and its data sources

raydavis: Be careful not to tear a ligament!

davidlimcheng:
LGTM @lyttam @nkutub.  Good stuff @raydavis !

PR #6304, opened 2016-12-28
SISRP-29104 - CalCentral - Class Information page missing lecture date, time and location for Sp17 UGBA 118 (INC0367091 & INC0376337)

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-29104

* Not shown here:  Updated EDODB query that should resolve existing CS data not being returned.
* The `IS NOT NULL` query parameter doesn't actually need to be there, even when the value is empty in CS, the query will return a value.  That is, there are no "`mtg."location-code" IS NOT NULL` rows.  I'm guessing they send us an empty string instead.

### DIFF
--- a/app/models/edo_oracle/queries.rb
+++ b/app/models/edo_oracle/queries.rb
@@ -162,8 +162,7 @@ module EdoOracle
         )
         WHERE
           sec."term-id" = '#{term_id}' AND
-          sec."id" = '#{section_id}' AND
-          mtg."location-code" IS NOT NULL
+          sec."id" = '#{section_id}'
         ORDER BY meeting_start_date, meeting_start_time
       SQL
     end
--- a/src/assets/templates/academics_classinfo.html
+++ b/src/assets/templates/academics_classinfo.html
@@ -101,7 +101,8 @@
               <div class="small-9 columns">
                 <div data-ng-repeat="schedule in section.schedules.recurring" data-ng-if="section.schedules.recurring" class="cc-academics-schedule-room">
                   <span data-ng-bind="schedule.schedule"></span> |
-                  <span data-ng-bind="schedule.roomNumber"></span> <span data-ng-bind="schedule.buildingName"></span> <br>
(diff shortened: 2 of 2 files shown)


nkutub:
@davidlimcheng I'm assuming you meant to say instead of "That is, there are no "mtg."location-code" IS NOT NULL rows. I'm guessing they send us an empty string instead." 

you really mean:  mtg."location-code" will never be NULL? correct or am I reading this incorrectly?

raydavis:
I corrected the JIRA link.

raydavis:
@davidlimcheng , the original "IS NOT NULL" filter might have been there for a reason. Do you know what the reason would be?

davidlimcheng:
Thanks @raydavis.  Luckily, it looks like @redconfetti created the query in question in #5116.  Unfortunately, that was all the way back in March.  Jason's comments in the [JIRA](https://jira.berkeley.edu/browse/SISRP-16710) ticket suggests that it's an artifact of the parallel `CampusOracle::Queries.rb` query:

The current query in CampusOracle::Queries has:
and sched.BUILDING_NAME is NOT NULL
I guess the building name would be null, and there was no reason to display it. Right now there are no meetings with "location-code" being NULL. I'm going to omit this for now. We will have to adjust based on the changes made to imported data.

But, it also suggests that Jason removed it, and re-inserted the line back into the code for some reason.  @redconfetti  - any chance of this ringing a bell for you?

raydavis:
I'm guessing that clause came in as an EDO DB equivalent of a [legacy Oracle check](https://github.com/ets-berkeley-edu/calcentral/blame/v85.3/app/models/campus_oracle/queries.rb#L362), done to prevent the same dangling schedule lines shown in SISRP-29104.

Oracle DBs don't support empty strings, so string columns may contain NULLs or filler whitespace. If that's what's going on here, then those could be filtered with the WHERE clause ``TRIM(whatever) IS NOT NULL``. But the logic you have in the PR looks fine, so I'll go ahead and merge it.

PR #6306, opened 2016-12-30
SISRP-28787 - Add Cache Expiration for Outgoing/Incoming Academic Plan Link

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-28787

* Adds a cache expiry function when user clicks the Multi-Year Planner link, and another cache expiry function when the user clicks "Return to Student Overview" while in CS.
* Also added to the Multi-Year Planner link in `user_preview.html`.  If we're showing the same link in two different places on the same page, we should have them behave the same way.
* This PR is a bit larger than I expected to be - so feel free to ping me on any questions.  I do intend to open a QA PR for this as well.

### DIFF
--- a/app/controllers/bootstrap_controller.rb
+++ b/app/controllers/bootstrap_controller.rb
@@ -46,6 +46,7 @@ class BootstrapController < ApplicationController
     # Sent from Campus Solutions Fluid UI when returning to CalCentral from
     # a Fluid activity that may have changed some data.
     flag = params['ucUpdateCache']
+    url = params['url']
     case flag
       when 'finaid'
         CampusSolutions::FinancialAidExpiry.expire current_user.user_id
@@ -53,6 +54,11 @@ class BootstrapController < ApplicationController
         CampusSolutions::PersonDataExpiry.expire current_user.user_id
       when 'enrollment'
         CampusSolutions::EnrollmentTermExpiry.expire current_user.user_id
+      when 'advisingAcademics'
+        # Since current_user does not know the student-overview UID the advisor is looking-up, we strip it from the URL
+        if (student_uid = url[/[0-9]+/])
+          MyAcademics::FilteredForAdvisor.expire student_uid
+        end
       else
         # no-op
     end
--- a/app/models/my_academics/academic_plan.rb
+++ b/app/models/my_academics/academic_plan.rb
@@ -1,11 +1,12 @@
 module MyAcademics
   class AcademicPlan < UserSpecificModel
     include AdvisingAcademicPlannerFeatureFlagged
+    include User::Student
 
     def merge(data)
       if is_feature_enabled
-        update_url_proxy = CampusSolutions::AcademicPlan.new(user_id: @uid).get
-        data[:updatePlanUrl] = update_url_proxy.try(:[], :feed).try(:[], :updateAcademicPlanner).try(:[], :url)
+        update_url_proxy = CampusSolutions::Link.new.get_url('UC_CX_PLANNER_ADV_STDNT', {:EMPLID => campus_solutions_id}).try(:[], :link)
+        data[:updatePlanUrl] = update_url_proxy
         data[:planSemesters] = get_plan_semesters(data[:semesters])
       end
     end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -169,6 +169,7 @@ Calcentral::Application.routes.draw do
   get '/api/advising/student_success/:student_uid' => 'advising_student#student_success', :defaults => { :format => 'json' }
   get '/api/advising/degree_progress/grad/:student_uid' => 'advising_student#degree_progress_graduate', :defaults => { :format => 'json' }
   get '/api/advising/degree_progress/ugrd/:student_uid' => 'advising_student#degree_progress_undergrad', :defaults => { :format => 'json' }
+  get '/api/advising/cache_expiry/academics/:student_uid' => 'advising_student#academics_cache_expiry', :defaults => { :format => 'json' }
   post '/advisor_act_as' => 'advisor_act_as#start'
   post '/stop_advisor_act_as' => 'advisor_act_as#stop'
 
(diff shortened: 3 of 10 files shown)


nkutub:
@davidlimcheng looks like a failed Travis

app/controllers/bootstrap_controller.rb
```diff
@@ -46,13 +46,19 @@ def check_cache_clear_flag
...
     case flag
       when 'finaid'
         CampusSolutions::FinancialAidExpiry.expire current_user.user_id
       when 'profile'
         CampusSolutions::PersonDataExpiry.expire current_user.user_id
       when 'enrollment'
         CampusSolutions::EnrollmentTermExpiry.expire current_user.user_id
+      when 'advisingAcademics'
+        # Since current_user does not know the student-overview UID the advisor is looking-up, we strip it from the URL
+        if (student_uid = url[/[0-9]+/])
```
nkutub: @davidlimcheng cool did not know you can pass regx to an array. This is a little brittle since it assumes the only number passed in the url will be for the student_id. there are ways to parse using URI/CGI which are more explicit. up to you if you feel this is good enough.

src/assets/javascripts/angular/controllers/pages/userOverviewController.js
```diff
@@ -6,7 +6,7 @@ var _ = require('lodash');
 /**
  * Preview of user profile prior to viewing-as
  */
-angular.module('calcentral.controllers').controller('UserOverviewController', function(academicsService, adminService, advisingFactory, academicStatusFactory, apiService, enrollmentVerificationFactory, linkService, statusHoldsService, studentAttributesFactory, $route, $routeParams, $scope) {
+angular.module('calcentral.controllers').controller('UserOverviewController', function(academicsService, adminService, advisingFactory, academicStatusFactory, apiService, enrollmentVerificationFactory, linkService, statusHoldsService, studentAttributesFactory, $http, $route, $routeParams, $scope) {
```
nkutub: @davidlimcheng , I see you added $http, but you did not use it, am I missing something?

config/routes.rb
```diff
@@ -169,6 +169,7 @@
   get '/api/advising/student_success/:student_uid' => 'advising_student#student_success', :defaults => { :format => 'json' }
   get '/api/advising/degree_progress/grad/:student_uid' => 'advising_student#degree_progress_graduate', :defaults => { :format => 'json' }
   get '/api/advising/degree_progress/ugrd/:student_uid' => 'advising_student#degree_progress_undergrad', :defaults => { :format => 'json' }
+  post '/api/advising/cache_expiry/academics/:student_uid' => 'advising_student#academics_cache_expiry', :via => :post, :defaults => { :format => 'json' }
```
nkutub: @davidlimcheng, are you sure this should not be a get?

src/assets/javascripts/angular/factories/advisingFactory.js
```diff
@@ -55,6 +56,10 @@ angular.module('calcentral.factories').factory('advisingFactory', function(apiSe
     return apiService.http.request(options, urlAdvisingDegreeProgressUndergrad + options.uid);
   };
 
+  var expireAcademicsCache = function(options) {
+    $http.post(urlAdvisingAcademicsCacheExpiry + options.uid);
```
nkutub: @davidlimcheng, are you sure this should not be a get? looks like you are appending param to URI not as a parameter to the function "post".

app/controllers/bootstrap_controller.rb
davidlimcheng: @nkutub Very interesting, the URI parsing seems like it'd work perfectly here.  I'll take a closer look at it.  Thanks for the pointer!

src/assets/javascripts/angular/controllers/pages/userOverviewController.js
davidlimcheng: @nkutub Great catch, I added it for testing purposes.  Will remove.

src/assets/javascripts/angular/factories/advisingFactory.js
davidlimcheng: @nkutub I think you are correct - a Get might work better.  I originally had it as a `post` and sent the `student_uid` param, but our front-end linter settings are a bit strict, plus it would stray away from the rest of the camel-cased front-end code.  Will amend.

app/controllers/bootstrap_controller.rb
davidlimcheng: @nkutub - after looking a bit closer into Ruby URI/CGI parsing, it looks like those won't be applicable because the URL in our case is a root-relative URL.  I think this should be fine for now.

davidlimcheng:
Should be good for follow-up now @nkutub, thanks for the review.

app/models/my_academics/academic_plan.rb
```diff
@@ -4,12 +4,19 @@ class AcademicPlan < UserSpecificModel
...
-        update_url_proxy = CampusSolutions::AcademicPlan.new(user_id: @uid).get
-        data[:updatePlanUrl] = update_url_proxy.try(:[], :feed).try(:[], :updateAcademicPlanner).try(:[], :url)
+        lookup_student_id
+        update_url_proxy = CampusSolutions::Link.new.get_url('UC_CX_PLANNER_ADV_STDNT', {:EMPLID => @cs_id}).try(:[], :link)
+        data[:updatePlanUrl] = update_url_proxy
         data[:planSemesters] = get_plan_semesters(data[:semesters])
       end
     end
 
+    def lookup_student_id
```
raydavis: FWIW other MyAcademics classes which need explicit access to ``campus_solutions_id`` or student ID include the User::Student module. If only to avoid inadvertent double-copies in the future, it might be worth considering here.

davidlimcheng: Makes sense @raydavis, I'll amend.  Thanks for the pointer.

davidlimcheng:
@raydavis @nkutub Amended

PR #6307, opened 2016-12-30
SISRP-29039 - Change order of committee members

nkutub:
https://jira.berkeley.edu/browse/SISRP-29039

* sorted committee members by last then first name
* sorted committees by service end then start dates.

### DIFF
--- a/app/models/my_committees/faculty_committees.rb
+++ b/app/models/my_committees/faculty_committees.rb
@@ -31,9 +31,12 @@ module MyCommittees
         faculty_committee = parse_cs_committee(cs_committee, is_completed)
         if cs_committee.present?
           # Add additional pieces of data needed faculty committees
+          cs_faculty_committee_svc = parse_cs_faculty_committee_svc(cs_committee)
           faculty_committee.merge!(
             student: parse_cs_committee_student(cs_committee),
-            serviceRange: parse_cs_faculty_committee_svc(cs_committee)
+            serviceRange: cs_faculty_committee_svc[:serviceRange],
+            csMemberEndDate: cs_faculty_committee_svc[:memberEndDate],
+            csMemberStartDate: cs_faculty_committee_svc[:memberStartDate]
           )
           if is_completed
             committees_result[:completed] << faculty_committee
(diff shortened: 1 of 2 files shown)


app/models/my_committees/faculty_committees.rb
```diff
@@ -42,17 +45,29 @@ def parse_cs_faculty_committees (cs_committees, committees_result)
...
+      cs_committees.try(:sort_by) do |committee|
+        [committee[:csMemberEndDate] || '2038-01-19', committee[:csMemberStartDate] || '']
+      end.try(:reverse)
+    end
+
     def parse_cs_faculty_committee_svc (cs_committee)
       committee_service_range_result = ''
       user_committee = cs_committee[:committeeMembers].try(:select) do |mem|
         mem[:memberEmplid].present? &&  mem[:memberEmplid] == @emplid
       end.first
```
redconfetti: @nkutub - Could we update this to use [Enumerable#find](http://devdocs.io/ruby~2.2/enumerable#method-i-find) instead of using #select to collect all matches before using #first on the returned Array?

app/models/my_committees/faculty_committees.rb
```diff
@@ -42,17 +45,29 @@ def parse_cs_faculty_committees (cs_committees, committees_result)
...
         end
       end
+      committees_result[:completed] = sort_committees_by_svc(committees_result[:completed])
+      committees_result[:active] = sort_committees_by_svc(committees_result[:active])
       committees_result
     end
 
+    def sort_committees_by_svc(cs_committees)
+      cs_committees.try(:sort_by) do |committee|
+        [committee[:csMemberEndDate] || '2038-01-19', committee[:csMemberStartDate] || '']
```
redconfetti: @nkutub - Where is this `2038-01-19` date default coming from? Why this date?

nkutub: @redconfetti , the repl is for select vs find. I was basing my decision because this is a string compare which is usually a more expensive compare:
http://brandon.dimcheff.com/2009/11/18/rubys-sort-vs-sort-by.html

app/models/my_committees/faculty_committees.rb
nkutub: @redconfetti, yup this should be changed, not sure why I did this the in the first place. I will change.

app/models/my_committees/faculty_committees.rb
nkutub: as for the 2038-01-19, I think this is the largest representable date in ruby not using the DateTime class. I set it to this just in case down the road someone tries to convert the string to a date and don't want an error to occur. https://makandracards.com/makandra/1017-maximum-representable-value-for-a-ruby-time-object

nkutub: thinking about it now, I think I will change this to 9999-99-99 since it is a date sort and I need to represent and open end date

raydavis: Yikes, @nkutub , that's a pretty obscure use of magic numbers. Maybe it would be better to just let the would-be converter know that the value might be blank?

nkutub: @raydavis, the problem here is that I'm using sort_by vs sort. In sort_by, the code block returns the value to be sorted on. In sort, the code block returns the sorting comparison result of . 1, 0, -1, this allows for us to add logic to check for nil and apply to the comparison result in the block. On the other hand, sort_by only allows us to pass the field value to be used for sorting beyond the code block. Becuase of this and that sort_by would be faster for string comparison, I'm going to pass back the value of "9999-99-99" for nil end date values. I may be missing something.

raydavis: Got it, @nkutub . When you amend the PR, could you explain the wacko date in code comments? Thanks!

redconfetti:
I reviewed the docs on #sort_by as well as the [Schwartzian transform](https://en.wikipedia.org/wiki/Schwartzian_transform) used by the method. I see that it performs the memoization that converts the input collection to a set of tuples no matter what, whereas the #sort method doesn't do this. #sort must sort in place, with optional argument that returns -1, 0, or 1.

I created a [REPL](https://repl.it/FAKX/4) just to see if there was a big difference with a small array.

PR #6308, opened 2017-01-03
SISRP-29098 - Registration and CNP warnings are not displayed in the popover

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-29098

* Opens up status popover to advising view-as.
* QA PR to come.


lyttam:
@davidlimcheng looks good, and this change makes sense to me (assuming the popover shows the same data as the Status/Holds card)

PR #6313, opened 2017-01-05
SISRP-27698 - re-styles <hr> to make it visible

lyttam:
https://jira.berkeley.edu/browse/SISRP-27698

The testing team pointed out that the horizontal separator was missing from the "Net Cost" section on the Financial Aid and Scholarships card.  In fact, it was just blending into its surroundings as we changed the background color of the section.  I overrode the sitewide \<hr\> style to make this one darker so it stands out.

QA PR forthcoming.


nkutub:
@lyttam  LGTM

PR #6317, opened 2017-01-06
CLC-6517, CalCentral no longer expects iTunes/RSS from Course Capture feed

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-6517

As of Jan 2, the production Course Capture feed neither includes RSS URLs nor iTunes info. Thus we can trim the fat in CalCentral's code-base: no more "audio" support.


src/assets/templates/widgets/webcast.html
```diff
@@ -4,7 +4,7 @@
   <div data-ng-if="!proxyErrorMessage">
```
johncrossman: Hot tip: use `?w=1` to make this particular file diff more readable.

raydavis:
Less code, hurrah!

PR #6318, opened 2017-01-06
CLC-6522 Update font in local bCourses LTI tools to Lato

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6522

Since Lato is a web font not shared across iframes, we need to fetch it from Google ourselves for bCourses pages only.

A couple of sizes and weights are tweaked to play better with the new font.

This should apply to all LTI tools except Roster Photos, which continues to use plain old CalCentral Arial for shared-template reasons.


johncrossman:
Better `Lato` than never. Thx, @paulkerschen

PR #6319, opened 2017-01-08
SISRP-29434 - Makes link handling null-safe on Student Overview

lyttam:
https://jira.berkeley.edu/browse/SISRP-29434

While smoke-testing the 7.4 release we noticed a JS error on Student Overview.  Seems pretty minor and doesn't impact functionality, and it is only occurring when the advising_academic_planner feature flag is set to false.


nkutub:
LGTM

PR #6321, opened 2017-01-11
CLC-6532 Add ASSOCIATED_PRIMARY column to SIS import sheets

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6532

This data is queried with an Oracle `listagg`, since in some cases more than one associated primary section may be returned.

The new column will also appear in merged course confirmation sheets, since they use the same heading definition as SIS import sheets.


johncrossman:
Thanks for talking me through this, @paulkerschen  Looks good.

PR #6322, opened 2017-01-11
SISRP-27494 - Replace 'backToText' with currentPage loading convention

redconfetti:
https://jira.berkeley.edu/browse/SISRP-27494

This is my first pass at a refactor for the CS Link API. 

After this update, the string we use for the `ucFromText` parameter that is included with CS links will originate from the AngularJS route configuration, with the convention being that it is usually merged with the controller level $scope (via the linkService.addCurrentRouteSettings method), or sometimes merged with link objects (via linkService.addBackToTextToResources). 

Next we can next refactor the code to include the string when the 'ucFromText' boolean is indicated as true, via the campusSolutionsLinkItemDirective. The same is true of the ucFromLink boolean behavior.


raydavis:
That does look tidier!

I guess I might as well take my merge conflict sooner rather than later... :)

PR #6323, opened 2017-01-11
SISRP-29001 - ENF grade-roster events fail on EDO DB ConnectionTimeoutError

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-29001

* See JIRA comments for logic behind this PR.

### DIFF
--- a/app/models/notifications/sis_expiry_processor.rb
+++ b/app/models/notifications/sis_expiry_processor.rb
@@ -4,13 +4,17 @@ module Notifications
 
     def process(event, timestamp)
       return false unless accept? event
-      logger.debug "Processing event: #{event}; timestamp = #{timestamp}"
-      if (expiry_module = get_expiry event)
-        if (uid = (event['topic'] == 'sis:faculty:grade-roster') ? (get_instructor_uids event) : (get_uid event))
-        expiry_module.expire uid
+      begin
+        logger.debug "Processing event: #{event}; timestamp = #{timestamp}"
+        if (expiry_module = get_expiry event)
+          if (uid = (event['topic'] == 'sis:faculty:grade-roster') ? (get_instructor_uids event) : (get_uid event))
+            expiry_module.expire uid
+          end
+        else
+          logger.warn "Event topic #{event['topic']} not recognized"
         end
-      else
-        logger.warn "Event topic #{event['topic']} not recognized"
+      ensure
+        ActiveRecord::Base.clear_active_connections!
       end
     end
 


paulkerschen:
@davidlimcheng It seems like this should work, but I'm wondering whether it's ideal to create a whole new `manual_pooled_query` method which duplicates a lot of code from `safe_query`, and is basically equivalent to that method plus a connection clear.

I admit I'm curious whether the pattern used in LiveUpdatesWarmer would work here; that is, if it would be enough to just add the connection clear in an `ensure` block at the end of `Notifications::SisExpiryProcessor#process`. That would involve less code; and if the connection clear lives in the same class that's doing the background job, the reason for that clear might be... er... clearer.

davidlimcheng:
@paulkerschen Let me try that out and see if it will fix the problem.  My initial hunch is that adding a connection clear at the end of `Notifications::SisExpiryProcessor#process` will still give us the same problem of opening too many new connections instead of using available connections in the pool, since the problem lies with `active_connection?` evaluating to `false` in `connection_pool.rb` due to Background Process threads being unrecognizable to ActiveRecord.

The main difference between the newly created `manual_pooled_query` and the classic `safe_query` is the exclusion of the `with_connection(&block)` method (surfaced through `use_pooled_connection`, which is where the troublesome `active_connection?` check lives.

In any case I'm also very curious to see if this would solve our problem, so - experimentation time!

paulkerschen:
Thanks for the extra explanation, @davidlimcheng - on first go-round I actually missed the most important part, which was omitting `use_pooled_connection` in the new query. That makes sense, and I think this is good to merge unless you want to experiment further. 

(Now that prod CC is running so hot, it's always possible that similar issues will crop up with ActiveRecord in other background jobs such as LiveUpdatesWarmer - though that one might connect only to Postgres and not the EDO DB.)

lyttam:
@davidlimcheng Nice work, this is deep!  I really appreciate all your documentation.  High quality comments, too.

Is it true that by avoiding calling `with_connection(&block)`, your new `manual_pooled_query` method assumes that a connection belonging to the calling class' name (`klass.name`) already exists in the pool?

davidlimcheng:
Thanks @paulkerschen & @lyttam! Although, I definitely spoke too soon.

I took @paulkerschen's recommendation and added an `ActiveRecord::Base.clear_active_connections!` method at the end of `Notifications::SisExpiryProcessor#process`.  This solved our problem, and did not add a running count of additional connections to our pool.  Paul and I have some ideas as to what's going on, but as I've already spent 20+ hours on this problem and have feature development to move on to, I will reluctantly amend the commit and move forward without a proper explanation as to why this is behaving the way it is.

The good news is that the code change is very minimal, and should open up queries to be available for all ENF events without clogging up the connection pool.  Again, thanks a million for all your help with this Paul!

davidlimcheng:
@paulkerschen @lyttam Should be good to go after T.  =(

paulkerschen:
@davidlimcheng And on to less frustrating things!

PR #6324, opened 2017-01-12
CLC-6533 OEC: return preferred name from enrollments query

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6533

In cases where a preferred name is not present, we fill in from User::BasicAttributes, as we do for missing email addresses.


lyttam:
Interesting problem; people seem to want a middle name field.  I'm happy with this solution in the meantime.

PR #6325, opened 2017-01-12
SISRP-29308 - Delegate access looses Return to Calcentral header

redconfetti:
https://jira.berkeley.edu/browse/SISRP-29308


lyttam:
LGTM.

PR #6326, opened 2017-01-12
SISRP-29466 Handle Hub Term API feeds with missing session timePeriods

raydavis:
https://jira.berkeley.edu/browse/SISRP-29466

Avoid exact repeats of Tuesday's outage.

### DIFF
--- a/app/models/berkeley/term.rb
+++ b/app/models/berkeley/term.rb
@@ -80,9 +80,11 @@ module Berkeley
         @classes_end = @instruction_end.advance(days: -7)
         @final_exam_week_start = @instruction_end.advance(days: 3)
         @final_exam_cs_data_available = @end.advance(days: -56)
-        session['timePeriods'].each do |timePeriod|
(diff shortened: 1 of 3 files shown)


davidlimcheng:
LGTM

app/models/berkeley/term.rb
```diff
@@ -144,7 +146,7 @@ def grades_entered
     def to_h
       methods = [:campus_solutions_id, :name, :year, :code, :slug, :to_english,
         :start, :end, :classes_start, :classes_end, :instruction_end, :grades_entered,
-        :is_summer, :legacy?, :final_exam_week_start]
+        :is_summer, :legacy?, :final_exam_week_start, :end_drop_add]
```
nkutub: @raydavis, I'm assuming this was just something you caught but not related the fix

raydavis: I wouldn't have noticed it if not for the bug making me want an easy way to see "end_drop_add", but otherwise unrelated, @nkutub

PR #6332, opened 2017-01-19
SISRP-28970, webcast_spec: when fake=false use stubbed term data

johncrossman:
https://jira.berkeley.edu/browse/SISRP-28970

The `fake: false` spec is testing 'connection failure' and term data matters not to that test. (Personal Bamboo build is happy: https://bamboo.ets.berkeley.edu/bamboo/browse/MYB-TRC3-1)


paulkerschen:
Bamboo's better off not knowing about the real 2017. Thanks @johncrossman

PR #6333, opened 2017-01-19
CLC-6547, Course Capture related notifications will expire after 2 days

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-6547

Default `cutoff` is 10 days. Course Capture is 2 days.

### DIFF
--- a/app/models/my_activities/merged.rb
+++ b/app/models/my_activities/merged.rb
@@ -15,8 +15,9 @@ module MyActivities
       ]
     end
 
-    def self.cutoff_date
-      @cutoff_date ||= (Settings.terms.fake_now || Time.zone.today.in_time_zone).to_datetime.advance(days: -10).to_time.to_i
+    def self.cutoff_date(offset = 10)
+      today = Settings.terms.fake_now || Time.zone.today.in_time_zone
+      today.to_datetime.advance(days: -offset).to_time.to_i
     end
 
     def get_feed_internal
--- a/app/models/my_activities/notification_activities.rb
+++ b/app/models/my_activities/notification_activities.rb
@@ -10,7 +10,8 @@ module MyActivities
     def self.append!(uid, activities)
       result = []
       use_pooled_connection {
-        result = Notifications::Notification.where(:uid => uid.to_s, :occurred_at => Time.at(MyActivities::Merged.cutoff_date)..Time.now) || []
+        cutoff_time = Time.at MyActivities::Merged.cutoff_date
+        result = Notifications::Notification.where(:uid => uid.to_s, :occurred_at => cutoff_time..Time.now) || []
       }
       result.each do |notification|
         translator = (self.translators[notification.translator] ||= "Notifications::#{notification.translator}".constantize.new)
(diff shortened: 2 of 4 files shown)


app/models/my_activities/merged.rb
```diff
@@ -15,8 +15,9 @@ def self.providers
       ]
     end
 
-    def self.cutoff_date
-      @cutoff_date ||= (Settings.terms.fake_now || Time.zone.today.in_time_zone).to_datetime.advance(days: -10).to_time.to_i
+    def self.cutoff_date(days_into_the_past = 10)
+      now = Settings.terms.fake_now || Time.zone.today.in_time_zone
+      @cutoff_date ||= now.to_datetime.advance(days: -days_into_the_past).to_time.to_i
```
paulkerschen: @johncrossman, I think this instance variable may complicate the problem.  Since it's inside a class method, it's associated with the whole class and will get cached the first time `MyActivities::Merged.cutoff_date` is called, with or without the custom `days_into_the_past`, returning the same value ever after.

Even more confusingly, it might get stored with different values on different nodes in the cluster.

TBH, I'm not sure this instance variable ought to be here at all. It seems like its effect would be to set the cutoff date on app startup and then keep it fixed until the next app restart. Perhaps I'm missing something?

johncrossman: @paulkerschen Gah, you're right. Amendment needed; I'll avoid generalizing that method.

johncrossman:
Do not merge. @raydavis pointed out very problematic code in `def self.cutoff_date`. A problem that made it to prod many moons ago.

raydavis:
The class-instance bug isn't @johncrossman 's fault but for whatever reason his PR finally made us _notice_ it after a couple of years. Thanks, John, & sorry about that!

app/models/my_activities/merged.rb
```diff
@@ -16,7 +16,8 @@ def self.providers
     end
 
     def self.cutoff_date
-      @cutoff_date ||= (Settings.terms.fake_now || Time.zone.today.in_time_zone).to_datetime.advance(days: -10).to_time.to_i
+      today = Settings.terms.fake_now || Time.zone.today.in_time_zone
+      today.to_datetime.advance(days: -10).to_time.to_i
```
lyttam: @johncrossman A suggestion:   since 10 is the default offset and 2 is a webcasts-specific offset, you could make this function accept an optional `offset` parameter

```
def self.cutoff_date(offset = 10)
  today = Settings.terms.fake_now || Time.zone.today.in_time_zone
  today.to_datetime.advance(days: -offset).to_time.to_i
end
```

and then call it from Webcasts::each_recent_recording

```
today = self.cutoff_date(CUTOFF_DAY_COUNT)
```

johncrossman: Thanks @lyttam A good suggestion. I had same idea but decided to minimize refactoring. Now that we've identified a well worn bug, the code _must_ be addressed. A good excuse to enhance; I'll amend.

raydavis:
Some top-notch hoop-jumping displayed here, @johncrossman

PR #6334, opened 2017-01-20
SISRP-29294 Support CalCentral-only & ETS-services-only configurations

raydavis:
https://jira.berkeley.edu/browse/SISRP-29294

Adds ```providedServices``` to configuration. The default is to continue to provide all functionality.

API and template requests are filtered by providedServices. Background jobs are effectively filtered as well.

Below the surface, server-side logic remains completely shared. Breaking up feeds and upstream dependencies will be done later.

An ETS-only server will almost always be reached via an LTI iframe rather than by directly logging in, & so its "standalone" UX is very primitive.


config/routes.rb
```diff
@@ -157,103 +59,210 @@
...
-  get '/api/advising/student_success/:student_uid' => 'advising_student#student_success', :defaults => { :format => 'json' }
-  get '/api/advising/degree_progress/grad/:student_uid' => 'advising_student#degree_progress_graduate', :defaults => { :format => 'json' }
-  get '/api/advising/degree_progress/ugrd/:student_uid' => 'advising_student#degree_progress_undergrad', :defaults => { :format => 'json' }
-  get '/api/advising/cache_expiry/academics/:student_uid' => 'advising_student#academics_cache_expiry', :defaults => { :format => 'json' }
-  post '/advisor_act_as' => 'advisor_act_as#start'
-  post '/stop_advisor_act_as' => 'advisor_act_as#stop'
+  if provided_services.include? 'calcentral'
+    get '/api/stats' => 'stats#get_stats', :via => :get, :defaults => { :format => 'json' }
+
+    post '/api/my/calendar/opt_in' => 'user_api#calendar_opt_in', :via => :post
```
paulkerschen: @raydavis A note: all the `:via` parameters in our routes are Rails 3 syntax, which is now sheer noise and could have come out back in 9a8eb2473853112d19f3f9b8e1862cc02b90945d. I'll leave it to you whether this PR is a good place for such tedious cleanup.

raydavis: This is the day we put our reality spectacles on, I guess. Thanks, @paulkerschen ; I think I'll amend as a separate commit.

lyttam:
Looks good to me, thanks @raydavis

johncrossman:
@raydavis As mentioned, the OEC stuff looks good. I tested:
* Google OAuth (re-)generation
* term_setup rake task on command line
* sis_import task via browser

cc: @pfarestveit  OEC bDrive test account has been reset. I'm done.

PR #6337, opened 2017-01-20
SISRP-29765 - Surfaces the Capstone milestone on Degree Progress

lyttam:
https://jira.berkeley.edu/browse/SISRP-29765

We weren't showing the Capstone milestone; now we are.

### DIFF
--- a/src/assets/javascripts/angular/controllers/widgets/academics/degreeProgressController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/academics/degreeProgressController.js
@@ -5,19 +5,21 @@ var _ = require('lodash');
 
 angular.module('calcentral.controllers').controller('DegreeProgressController', function(degreeProgressFactory, $scope) {
 
-  $scope.degreeProgressGraduate = {
-    isLoading: true
+  $scope.degreeProgress = {
+    graduate: {
+      isLoading: true
+    }
   };
 
   var loadDegreeProgress = function() {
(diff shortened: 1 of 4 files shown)


src/assets/javascripts/angular/controllers/widgets/academics/degreeProgressController.js
```diff
@@ -5,19 +5,21 @@ var _ = require('lodash');
...
+    }
   };
 
   var loadDegreeProgress = function() {
     degreeProgressFactory.getDegreeProgress()
       .then(function(data) {
-        $scope.degreeProgressGraduate.progresses = _.get(data, 'data.feed.degreeProgress');
-        $scope.degreeProgressGraduate.links = _.get(data, 'data.feed.links');
-        $scope.degreeProgressGraduate.errored = _.get(data, 'errored');
+        $scope.degreeProgress.graduate.progresses = _.get(data, 'data.feed.degreeProgress');
```
nkutub: @lyttam, I see a change on the scope variable hierarchy in the controller but I don't see a change in any template file to reference the new hierarchy. Am I missing something?

lyttam: @nkutub The template is shared by this and another controller, and the change was made to the template (and that other controller) in a prior release.  So this is cleanup; it probably should have been changed at the same time but it wasn't actually being used yet.

PR #6338, opened 2017-01-20
SISRP-29570 - Mid-Point Grades: Instructor-Facing

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-29570
Design JIRA:  https://jira.berkeley.edu/browse/SISRP-29519

* Title says it all
* Also makes this feature `fake_now`-friendly, enabling testing of past/future dates, since that drives most of the UI.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -451,15 +451,31 @@ final_exam_schedule: [
   }
 ]
 
-# This is a temporary setting for grading period which will be set manually for Fall 2016 semester.
+# This is a temporary setting for grading period which will be set manually for Fall 2016 & Spring 2017 semesters.
 grading_period:
-  assistance_link: ''
-  general:
-    start: ''
-    end: ''
-  law:
-    start: ''
-    end: ''
+  links:
+    final: ''
+    midpoint:  ''
+    law:  ''
+  dates:
+    general:
+      fall_2016:
+        start: ''
+        end: ''
+      spring_2017:
+        midpoint:
+          start: ''
+          end: ''
+        final:
+          start: ''
+          end: ''
+    law:
+      fall_2016:
+        start: ''
+        end: ''
+      spring_2017:
+        start: ''
+        end: ''
 
 # A feature will be disabled if the corresponding feature-flag is false or nil.
 features:
--- a/spec/models/my_academics/grading_spec.rb
+++ b/spec/models/my_academics/grading_spec.rb
@@ -7,63 +7,67 @@ describe MyAcademics::Grading do
     MyAcademics::Grading.new(uid)
   end
 
+  before do
+    allow(Settings.terms).to receive(:fake_now).and_return nil
+  end
+
   context 'when grading period from settings is set' do
     before do
-      allow(Settings.grading_period.general).to receive(:start).and_return '2016-12-12 0:0:0'
-      allow(Settings.grading_period.general).to receive(:end).and_return '2016-12-12 23:59:59'
+      allow(Settings.grading_period.dates.general.fall_2016).to receive(:start).and_return '2016-12-12 0:0:0'
+      allow(Settings.grading_period.dates.general.fall_2016).to receive(:end).and_return '2016-12-12 23:59:59'
     end
     it 'it succeeds on grading period validation' do
-      expect(subject.valid_grading_period?(false)).to eq true
+      expect(subject.valid_grading_period?(false, '2168')).to eq true
     end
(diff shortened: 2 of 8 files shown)


spec/models/my_academics/grading_spec.rb
```diff
@@ -188,18 +194,18 @@
...
-      allow(Settings.features).to receive(:hub_term_api).and_return false
+      allow(Settings.features).to receive(:hub_term_api).and_return true
       allow(CampusSolutions::Grading).to receive(:new).and_return(grading_proxy)
       allow(MyAcademics::AcademicsModule).to receive(:fetch_link).and_return(fake_grading_url)
       allow(subject).to receive(:get_grading_period_status).and_return(:gradingPeriodNotSet)
     end
 
     it 'it should return expected values merged into section' do
       subject.merge(feed)
+      pp feed
```
redconfetti: @davidlimcheng - I think we need to remove this `pp` call

spec/models/my_academics/grading_spec.rb
```diff
@@ -140,15 +146,15 @@
     end
 
     it 'it should return nil for missing ccn in sections' do
-      expect(subject.get_cs_status(ccn = nil , term_code ='2168')).to eq nil
+      expect(subject.get_cs_status(ccn = nil, is_law = false, term_code ='2168')).to eq nil
     end
```
redconfetti: @davidlimcheng - Why are we assigning a value to variables (`is_law`, `term_code`) within the function call?
Aren't these variables defined in memory only during the call, and then the value of the variable itself is passed to `#get_cs_status`?

If you were using @is_law, or @term_code, I could understand that these variables are shared between the tests (each code block passed to `it`), but that's not the case here. 

Shouldn't this be:
```
expect(subject.get_cs_status(nil, false, '2168')).to eq nil
```

Or are you doing this so that people know what the values represent? Because if that's what you're wanting to do with your code, then you should make #get_cs_status accept single hash as an argument like `get_cs_status(cnn: nil, is_law: false, term_code: '2168')`. That also makes it possible to specify the arguments in any order. Or alternatively you can use [splats](https://endofline.wordpress.com/2011/01/21/the-strange-ruby-splat/) as an argument.

config/settings.yml
```diff
@@ -451,15 +451,31 @@ final_exam_schedule: [
   }
 ]
 
-# This is a temporary setting for grading period which will be set manually for Fall 2016 semester.
+# This is a temporary setting for grading period which will be set manually for Fall 2016 & Spring 2017 semesters.
```
lyttam: @davidlimcheng so will we have to add something to the settings before every new semester?

spec/models/my_academics/grading_spec.rb
lyttam: @redconfetti @davidlimcheng FWIW, I actually like David's way of writing the method call.  It explicitly calls out each param so you don't have to guess or drill down to find out, but it allows you to enforce a particular method signature which is a safer way to code.  If we start passing Hashes around, it's harder to validate that the parameters are what we expect them to be (and leads to harder-to-trace bugs).

spec/models/my_academics/grading_spec.rb
davidlimcheng: @redconfetti Whoops!  Amended - good catch.

spec/models/my_academics/grading_spec.rb
davidlimcheng: @redconfetti @lyttam Yeah, I actually just kept @nkutub's old stylings here of calling out what each parameter is.  I like it too, it requires less manual checking and is much more clear in what we are trying to test.  I do agree that we shouldn't be doing this in the actual code, but as this is the spec, I think it makes sense to leave it as is.

config/settings.yml
davidlimcheng: Hopefully not!  The only reason we are hardcoding Spring again is because [SISRP-29623, custom tables to hold these dates in CS,](https://jira.berkeley.edu/browse/SISRP-29623) is not ready.  Once it is, we will have to refactor this code once again, but hopefully it will be for the last time.  We'll be able to grab the dates from CS through an API, match up on term, and drive the UX accordingly.

davidlimcheng:
@redconfetti @lyttam Amended with responding comments.  If you think this is good to go, please give me a heads up before merging so that I can edit the corresponding YML configurations on our servers.

PR #6339, opened 2017-01-20
SISRP-28259 Fetch message explaining Confidential Student status from CS

raydavis:
https://jira.berkeley.edu/browse/SISRP-28259

Adding four new files to support two sentences of basically static text reeks of overengineering, but streamlining the logic would take an unjustifiable refactoring effort, and so...


redconfetti:
@raydavis - With the current version both Travis processes failed

raydavis:
Thanks, @redconfetti . It was another one of those merge conflicts in routes.rb -- I hope I'm finally past those now!

raydavis:
Travis is great again.

src/assets/templates/widgets/profile/basic_confidential.html
```diff
@@ -1,13 +1,8 @@
-<div class="cc-page-widget-profile-section">
+<div class="cc-page-widget-profile-section" data-ng-controller="ConfidentialStudentController" data-cc-spinner-directive="confidentialStudent.isLoading">
```
lyttam: @raydavis Hmmm do we need a spinner here?  What if we just show nothing until the message text arrives?

raydavis: Good question, @lyttam . I vacillated but, as with the other bulk, decided to follow existing conventions here.

In general, we don't like to have new sections suddenly pop up without warning, and given that only a very very small audience will ever see the message, I also didn't want to refactor to hide the entire Basic profile display to support it.

The request hits a globally cached endpoint and should normally complete well before the spinner has a chance to appear. I can increase the odds, though, by giving the cached message a 24-hour lifespan. I'll amend the PR for that.

raydavis:
PR amended

davidlimcheng:
@raydavis @lyttam LGTM

raydavis:
@lyttam Did my change make the PR mergeable?

lyttam:
@raydavis  Yes, thanks for the ping

PR #6343, opened 2017-01-23
SISRP-29827 - Midpoint Status Column is missing on My Academics Tab

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-29827

* How I got caught:  The front-end academicsController has an `if` statement to show the most relevant semester on the My Academics tab for instructors, but only if `MyAcademics::Semesters` is empty.  If you're not tunneled to the CampusOracle DB, you'll never hit this part of the statement, because `fake` will always return data for you in `MyAcademics::Semesters`.


redconfetti:
LGTM

PR #6344, opened 2017-01-24
SISRP-29459 - CalCentral CUM Units is only displaying UC Berkeley Units total

nkutub:
https://jira.berkeley.edu/browse/SISRP-29459

lots of analysis for the smallest change :) 
* fix put in to source cumulative units for new field added to academic status API


davidlimcheng:
LGTM

PR #6346, opened 2017-01-25
SISRP-29795 - Create Enrollment API Proxy

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-29795
Some more good information on why I chose to create a "CurrentEnrollments" class instead of a more general, catch-all "Enrollments" class here:
https://jira.berkeley.edu/browse/SISRP-29571

* The Enrollments Proxy portion of the `verify_sis_endpoints` script *WILL* fail for the time being, until v2 of the Enrollments API proxy gets migrated to Hub UAT.

### DIFF
--- a/app/models/hub_enrollments/proxy.rb
+++ b/app/models/hub_enrollments/proxy.rb
@@ -1,0 +1,132 @@
+module HubEnrollments
+  class Proxy < BaseProxy
+    include ClassLogger
+    include Proxies::Mockable
+    include User::Student
+    include SafeJsonParser
+
+    def initialize(options = {})
+      super(Settings.hub_enrollments_proxy, options)
+      if @fake
+        @campus_solutions_id = lookup_campus_solutions_id
+        initialize_mocks
+      end
+    end
+
+    def instance_key
+      @uid
+    end
+
+    def json_filename
+      ''
+    end
+
+    def mock_json
+      read_file('fixtures', 'json', json_filename)
+    end
+
+    def mock_request
+      super.merge(uriMating: url)
+    end
+
+    def get
+      wrapped_response = self.class.handling_exceptions(instance_key) do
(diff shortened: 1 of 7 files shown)


app/models/hub_enrollments/proxy.rb
```diff
@@ -0,0 +1,132 @@
...
+        {
+          feed: empty_feed,
+          noStudentId: true
+        }
+      else
+        logger.info "Fake = #{@fake}; Making request to #{url} on behalf of user #{@uid}; cache expiration #{self.class.expires_in}"
+        opts = opts.merge(request_options)
+        response = get_response(url, opts)
+        logger.debug "Remote server status #{response.code}, Body = #{response.body.force_encoding('UTF-8')}"
+        if response.code == 404
```
lyttam: @davidlimcheng Do you know what it looks like when the API fails (i.e. have you been able to test that scenario)? CS endpoints seem to error rather subtly, so in `CampusSolutions::Proxy` we had to wrap some logic in `is_errored?` to handle the case where the API call fails but still gives us a 200 status.  (In a perfect world, the APIs would use 40x status codes appropriately)

davidlimcheng: @lyttam great question!  Yes, I did test out the error handling.  I tried to keep the same patterns as seen in the `HubEdos::Proxy` model, but if you look at the way the two modules handle errors, you'll see that the API behaves a bit differently.  I believe this is because the Enrollments API has a couple of different ways to be called - you can enter Student ID or a section ID.  Anyhow, this proxy should recognize errors thrown from the Hub and handle them accordingly.

lyttam: @davidlimcheng cool! I think this is good to merge, then.

PR #6347, opened 2017-01-26
SISRP-29512 - 7.5: Refactor coding conventions to match new CS Link API

redconfetti:
https://jira.berkeley.edu/browse/SISRP-29512

* Introduces ccCampusSolutionsLinkV2Directive
* Updates outboundLinkDirective.js code to not apply to ccCampusSolutionsLinkV2Directive
* Updates linkService to share outboundLink methods
* Updates Advisor Resource links in Advisor Resources card (Dashboard) and Personal Summary card (Student Overview)

### DIFF
--- a/src/assets/javascripts/angular/directives/campusSolutionsLinkV2Directive.js
+++ b/src/assets/javascripts/angular/directives/campusSolutionsLinkV2Directive.js
@@ -118,9 +118,9 @@ angular.module('calcentral.directives').directive('ccCampusSolutionsLinkV2Direct
    * Process ucFrom parameter configuration
    */
   var getUcFromParamConfig = function(linkObj, scope, attrs) {
-    var includeUcFrom = !!(scope.$eval(attrs.ccCampusSolutionsLinkV2DirectiveUcFrom) || _.get(linkObj, 'ucfrom')) || false;
-    var includeUcFromLink = !!(scope.$eval(attrs.ccCampusSolutionsLinkV2DirectiveUcFromLink) || _.get(linkObj, 'ucfromlink')) || false;
-    var includeUcFromText = !!(scope.$eval(attrs.ccCampusSolutionsLinkV2DirectiveUcFromText) || _.get(linkObj, 'ucfromtext')) || false;
+    var includeUcFrom = !!(scope.$eval(attrs.ccCampusSolutionsLinkV2DirectiveUcFrom) || _.get(linkObj, 'ucfrom') || _.get(linkObj, 'ucFrom')) || false;
+    var includeUcFromLink = !!(scope.$eval(attrs.ccCampusSolutionsLinkV2DirectiveUcFromLink) || _.get(linkObj, 'ucfromlink') || _.get(linkObj, 'ucFromLink')) || false;
+    var includeUcFromText = !!(scope.$eval(attrs.ccCampusSolutionsLinkV2DirectiveUcFromText) || _.get(linkObj, 'ucfromtext') || _.get(linkObj, 'ucFromText')) || false;
     return {
       includeUcFrom: includeUcFrom,
       includeUcFromLink: includeUcFromLink,
--- a/src/assets/javascripts/angular/directives/campusSolutionsLinkV2Directive.js
+++ b/src/assets/javascripts/angular/directives/campusSolutionsLinkV2Directive.js
@@ -72,7 +72,6 @@ var angular = require('angular');
  *
  */
 angular.module('calcentral.directives').directive('ccCampusSolutionsLinkV2Directive', function(linkService) {
-
   /**
    * Sets the text for the anchor tag if not already present
    */
@@ -96,13 +95,11 @@ angular.module('calcentral.directives').directive('ccCampusSolutionsLinkV2Direct
    */
   var getLinkConfig = function(linkObj, scope, attrs) {
     var baseLinkUrl = scope.$eval(attrs.ccCampusSolutionsLinkV2DirectiveUrl) || _.get(linkObj, 'url');
-    var includeUcFrom = !!(scope.$eval(attrs.ccCampusSolutionsLinkV2DirectiveUcFrom) || _.get(linkObj, 'ucfrom'));
-    var includeUcFromLink = !!(scope.$eval(attrs.ccCampusSolutionsLinkV2DirectiveUcFromLink) || _.get(linkObj, 'ucfromlink'));
-    var includeUcFromText = !!(scope.$eval(attrs.ccCampusSolutionsLinkV2DirectiveUcFromText) || _.get(linkObj, 'ucfromtext'));
     var ccCacheString = attrs.ccCampusSolutionsLinkV2DirectiveCcCache || _.get(linkObj, 'ccCache');
     var ccPageName = scope.$eval(attrs.ccCampusSolutionsLinkV2DirectiveCcPageName) || _.get(linkObj, 'ccPageName') || 'CalCentral';
     var ccPageUrl = scope.$eval(attrs.ccCampusSolutionsLinkV2DirectiveCcPageUrl) || _.get(linkObj, 'ccPageUrl');
-    var decoratedLinkUrl = decorateLink(baseLinkUrl, includeUcFrom, includeUcFromLink, includeUcFromText, ccCacheString, ccPageName, ccPageUrl);
+    var ucFromParamsConfig = getUcFromParamConfig(linkObj, scope, attrs);
+    var decoratedLinkUrl = decorateLink(baseLinkUrl, ucFromParamsConfig.includeUcFrom, ucFromParamsConfig.includeUcFromLink, ucFromParamsConfig.includeUcFromText, ccCacheString, ccPageName, ccPageUrl);
 
     // TODO: Discover why link objects either include 'showNewwindow' or 'shownewwindow' property
     var showNewWindow = scope.$eval(attrs.ccCampusSolutionsLinkV2DirectiveShowNewWindow) || _.get(linkObj, 'shownewwindow') || _.get(linkObj, 'showNewWindow') || false;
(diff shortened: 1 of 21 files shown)


src/assets/javascripts/angular/directives/campusSolutionsLinkV2Directive.js
```diff
@@ -0,0 +1,164 @@
...
+ *
+ * Intended for use with objects that include Campus Solutions Link API properties
+ *
+ * Example:
+ *   {
+ *     "name": "Academic Progress Report",
+ *     "title": "View this student's Academic Progress Report",
+ *     "ccPageName": 'My Academics',
+ *     "ccPageUrl": 'http://calcentral.berkeley.edu/academics/',
+ *     "ucfrom": true,
```
davidlimcheng: @redconfetti Extremely minor, but some of these are not camel-cased.

davidlimcheng: Ahh, or is that how it's presented from the Link API?

src/assets/javascripts/angular/directives/campusSolutionsLinkV2Directive.js
```diff
@@ -0,0 +1,164 @@
...
+ */
+angular.module('calcentral.directives').directive('ccCampusSolutionsLinkV2Directive', function(linkService) {
+
+  /**
+   * Sets the text for the anchor tag if not already present
+   */
+  var setLinkBody = function(linkBody, element) {
+    if (linkBody) {
+      var currentLinkBody = element.text();
+      if (currentLinkBody === '' || currentLinkBody === undefined) {
```
davidlimcheng: @redconfetti Do you think we need a `currentLinkBody === null` check here as well??

I'd also suggest checking for `!currentLinkBody`, but unfortunately that returns true if `currentLinkBody === 0` as well, so that it's not really air-tight in this context.

src/assets/javascripts/angular/directives/campusSolutionsLinkV2Directive.js
```diff
@@ -0,0 +1,164 @@
...
+   */
+  var getLinkConfig = function(linkObj, scope, attrs) {
+    var baseLinkUrl = scope.$eval(attrs.ccCampusSolutionsLinkV2DirectiveUrl) || linkObj.url;
+    var includeUcFrom = !!(attrs.ccCampusSolutionsLinkV2DirectiveUcFrom || linkObj.ucfrom);
+    var includeUcFromLink = !!(attrs.ccCampusSolutionsLinkV2DirectiveUcFromLink || linkObj.ucfromlink);
+    var includeUcFromText = !!(attrs.ccCampusSolutionsLinkV2DirectiveUcFromText || linkObj.ucfromtext);
+    var ccCacheString = attrs.ccCampusSolutionsLinkV2DirectiveCcCache || linkObj.ccCache;
+    var ccPageName = scope.$eval(attrs.ccCampusSolutionsLinkV2DirectiveCcPageName) || linkObj.ccPageName || 'CalCentral';
+    var ccPageUrl = scope.$eval(attrs.ccCampusSolutionsLinkV2DirectiveCcPageUrl) || linkObj.ccPageUrl;
+    var showNewWindow = scope.$eval(attrs.ccCampusSolutionsLinkV2DirectiveShowNewWindow) || linkObj.shownewwindow || false;
```
davidlimcheng: @redconfetti Would `false` be a default value for `showNewWindow` and `linkHoverText`?  If so, would it be helpful to add it to the already-awesome documentation at the top?  Also applies to "CalCentral" being a default value for `ccPageName`

davidlimcheng:
Great job @redconfetti, just left a couple of questions for ya.

src/assets/javascripts/angular/directives/campusSolutionsLinkV2Directive.js
redconfetti: @davidlimcheng True, I've added some mention of this in the docs. showNewWindow is a boolean argument. linkHoverText is only returned in my code as false to stop an empty string from being applied as the `title` attribute.

src/assets/javascripts/angular/directives/campusSolutionsLinkV2Directive.js
redconfetti: Yeah. I should use `lodash.isEmpty`

```
_.isEmpty(null);
// => true
 
_.isEmpty(true);
// => true
 
_.isEmpty(1);
// => true

_.isEmpty('');
// => true

_.isEmpty(undefined);
// => true

_.isEmpty(NaN);
// => true

_.isEmpty(null);
// => true

_.isEmpty({something:'something'});
// => false

_.isEmpty('dsfdsf');
// => false
```

src/assets/javascripts/angular/directives/campusSolutionsLinkV2Directive.js
redconfetti: Yeah. We will have to change this convention once all the links across the board are migrated, then we can safely update the back-end to use camelcase. The `shownewwindow` and `showNewWindow` both show up differently in different places. We'll have to fix that later.

davidlimcheng: @redconfetti Yikes.  In that case, would it make sense to pass the link object to a function that will iterate `String.tolowerCase()` over the key names before moving forward?  Or is that overkill?

lyttam: +1, seems worth it. @redconfetti @davidlimcheng

lyttam:
@redconfetti is the idea to use the V2 directive going forward, and leave the original for backwards compatibility?  If so, I would rename the old one (e.g. `campusSolutionsLinkV1Directive`) and give the new one a permanent name (`campusSolutionsLinkDirective`) so that nothing has to change when we eventually remove the old one.

src/assets/javascripts/angular/directives/campusSolutionsLinkV2Directive.js
```diff
@@ -0,0 +1,177 @@
...
+   */
+  var getLinkConfig = function(linkObj, scope, attrs) {
+    var baseLinkUrl = scope.$eval(attrs.ccCampusSolutionsLinkV2DirectiveUrl) || _.get(linkObj, 'url');
+    var ccCacheString = attrs.ccCampusSolutionsLinkV2DirectiveCcCache || _.get(linkObj, 'ccCache');
+    var ccPageName = scope.$eval(attrs.ccCampusSolutionsLinkV2DirectiveCcPageName) || _.get(linkObj, 'ccPageName') || 'CalCentral';
+    var ccPageUrl = scope.$eval(attrs.ccCampusSolutionsLinkV2DirectiveCcPageUrl) || _.get(linkObj, 'ccPageUrl');
+    var ucFromParamsConfig = getUcFromParamConfig(linkObj, scope, attrs);
+    var decoratedLinkUrl = decorateLink(baseLinkUrl, ucFromParamsConfig.includeUcFrom, ucFromParamsConfig.includeUcFromLink, ucFromParamsConfig.includeUcFromText, ccCacheString, ccPageName, ccPageUrl);
+
+    // TODO: Discover why link objects either include 'showNewwindow' or 'shownewwindow' property
```
lyttam: @redconfetti OK to leave this TODO, or should this be taken care of before release?

redconfetti: @lyttam - Yes. I'm wanting to avoid renaming the other directive across all the code, as this would extend my efforts.

Also, this specific key issue would require back-end changes, which I'm trying to avoid with GL 7.5 because this work might overlap with TST/QAT so much. I've created [SISRP-29923](https://jira.berkeley.edu/browse/SISRP-29923) to handle back-end updates to complete the migration to this new directive. I've added a note to the comments in that JIRA to focus on this during the continued refactor.

src/assets/javascripts/angular/directives/campusSolutionsLinkV2Directive.js
redconfetti: @davidlimcheng @lyttam - the CS Link API provides us with the key as 'UCFROM', 'UCFROMLINK', and 'UCFROMTEXT'. It's not possible to convert these to camelcase, unless we hard-code the mapping in the code.

I can do this, but I'm wanting to avoid making changes to the back-end coding conventions with GoLive 7.5. This is certainly something I can promise we will do to make things complete in [SISRP-29923](https://jira.berkeley.edu/browse/SISRP-29923) during GL 8.0 (or another release).

lyttam: @redconfetti seems like you could make them all uppercase or all lowercase for consistency, unless I am missing something.

src/assets/javascripts/angular/directives/campusSolutionsLinkV2Directive.js
lyttam: @redconfetti cool thanks for the explanation!

src/assets/javascripts/angular/directives/campusSolutionsLinkV2Directive.js
davidlimcheng: @redconfetti @lyttam Yeah, what I'm worried about is errors popping up due to the link object giving us something like `showNewWindow` when we're looking for `shownewwindow`.  Since the keys that CS is handing us are inconsistent, should we create a function to lowercase all of the key names across the board to avoid these type of errors from popping up?

redconfetti:
@davidlimcheng @lyttam @nkutub @raydavis

Just FYI, the new ccCampusSolutionsLinkV2Directive is supposed to replace the old one. It follows the specifications of how CalCentral should render links based on CS Link API configuration. I know it sounds ideal to rename the old directive, but I'd rather let SISRP-29923 handle the refactoring in a following release.

src/assets/javascripts/angular/directives/campusSolutionsLinkV2Directive.js
davidlimcheng: @redconfetti has noted that the JIRA is tracking this problem, the ongoing refactor work includes the casing problem, and that the changed links so far have been validated.  I think this should be good to merge.

PR #6348, opened 2017-01-26
SISRP-29777 - When no degree progress data for grad student, displays the card with a message

lyttam:
https://jira.berkeley.edu/browse/SISRP-29777

I changed the message slightly from what was requested (removed the word "completely" because it doesn't make sense).


nkutub:
@lyttam, LGTM

PR #6349, opened 2017-01-26
SISRP-29571 - Mid-Point Grades: Student-Facing

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-29571

* Need to rename "Current Enrollments", opened up [SISRP-29960](https://jira.berkeley.edu/browse/SISRP-29960) to do that
* Surfaces midpoint grades for the current semester to enrollments where the user has a role of "Student"
* Removed `jsonifiedfeed` from Enrollments API because I realized we only use it on server-side.


nkutub:
@davidlimcheng, nice work!

PR #6350, opened 2017-01-27
SISRP-29960 - Rename Current Enrollments API

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-29960

* Current Enrollments is a very limiting name when you can pass any term ID for that term's enrollments...let's help this API spread its wings and live to its full potential.


raydavis:
Fly proud, fly free, TermEnrollments

PR #6351, opened 2017-01-27
SISRP-29841 - CalCentral/My Academics shows instructor names that should not print

nkutub:
https://jira.berkeley.edu/browse/SISRP-29841

* Consumed flag already in EDODB instructor view to hide instructors names from class info page if flag is not set to 'Y' and a student is viewing the page. 
* correct logic to allow GSI's to also see grading delegate information.


raydavis:
Both of the Travis twins are unhappy, @nkutub

nkutub:
@raydavis, twins happy again :)

raydavis:
Now you see 'em, now you don't...

PR #6352, opened 2017-01-27
SISRP-26723 - Look for summer-2016 course data/enrollments in both legacy and EDO

lyttam:
https://jira.berkeley.edu/browse/SISRP-26723

Some students have two different SIDs on our system - they were issued one SID when they enrolled in summer courses, and a second SID when they came back the following fall.

These unlucky students have summer 2016 course enrollment data on the legacy Campus Oracle DB under the first SID, and the same data on EDO DB under the second SID.

The effect of this PR is to check in both datasources for summer 2016 enrollments.  That way, if a student logs in with the first SID they will see their summer 2016 enrollments, and if they log in with the second SID they will see those same summer 2016 enrollments as well as fall 2016 and later.

### DIFF
--- a/app/models/edo_oracle/user_courses/base.rb
+++ b/app/models/edo_oracle/user_courses/base.rb
@@ -5,8 +5,7 @@ module EdoOracle
       def initialize(options = {})
         super(Settings.edodb, options)
         @uid = @settings.fake_user_id if @fake
-        # Non-legacy terms are those after Settings.terms.legacy_cutoff.
-        @non_legacy_academic_terms = Berkeley::Terms.fetch.campus.values.reject &:legacy?
+        @non_legacy_academic_terms = Berkeley::Summer16EnrollmentTerms.non_legacy_terms
       end
 
       def self.access_granted?(uid)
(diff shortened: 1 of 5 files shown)


app/models/edo_oracle/user_courses/base.rb
```diff
@@ -16,7 +15,8 @@ def self.access_granted?(uid)
       def merge_enrollments(campus_classes)
         return if @non_legacy_academic_terms.empty?
         previous_item = {}
-        EdoOracle::Queries.get_enrolled_sections(@uid, @non_legacy_academic_terms).each do |row|
+
+          EdoOracle::Queries.get_enrolled_sections(@uid, @non_legacy_academic_terms).each do |row|
```
redconfetti: @lyttam - Why is this indented? The `end` for this loop is on line 28 unindented.

lyttam: @redconfetti good catch!  Amended.

raydavis:
@lyttam , the need for Summer 2016 terms which can be either "legacy" or "CS" is specific to student enrollments. What do instructors with Summer 2016 classes see after your change?

raydavis:
Also of interest: students who took Spring 2016, Summer 2016, _and_ Fall 2016 classes, since their SID will stay the same.

lyttam:
Good questions, @raydavis.  I verified that instructors can see the summer 2016 courses they are teaching, but I'm having trouble finding summer 2016 courses with students enrolled.  Waiting for the class sections view to be repopulated after we lost some data over the weekend.

Students who took courses in those three terms look good (I compared with CC-QA).

raydavis:
Thanks for checking, @lyttam !

PR #6353, opened 2017-01-27
SISRP-29026 Enable concurrent request handling in development mode

raydavis:
https://jira.berkeley.edu/browse/SISRP-29026

Just makes it easier to do more realistic testing when ``RAILS_ENV=development`` -- should be a no-op unless you change your settings.


redconfetti:
LGTM. Now we can config per each environment!

PR #6362, opened 2017-01-31
SISRP-30021 - Show checkmark icon next to completed milestone

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-30021

### DIFF
--- a/src/assets/templates/widgets/academics/degree_progress_graduate.html
+++ b/src/assets/templates/widgets/academics/degree_progress_graduate.html
@@ -23,8 +23,13 @@
       <li data-ng-repeat="plan in degreeProgress.graduate.progresses">
         <h3 data-ng-bind="plan.acadPlan ? plan.acadPlan + ' &mdash; Graduate Division Milestones' : 'Graduate Division Milestones'"></h3>
         <div data-ng-repeat="requirement in plan.requirements">
-          <h4 data-ng-bind="requirement.name"></h4>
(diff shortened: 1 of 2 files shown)


src/assets/templates/widgets/academics/degree_progress_graduate.html
```diff
@@ -23,8 +23,13 @@
       <li data-ng-repeat="plan in degreeProgress.graduate.progresses">
         <h3 data-ng-bind="plan.acadPlan ? plan.acadPlan + ' &mdash; Graduate Division Milestones' : 'Graduate Division Milestones'"></h3>
         <div data-ng-repeat="requirement in plan.requirements">
-          <h4 data-ng-bind="requirement.name"></h4>
-          <span data-ng-bind="requirement.statusDescr"></span>&nbsp;<span data-ng-bind="requirement.dateFormatted.dateString"></span>&nbsp;<span data-ng-bind="requirement.formNotification"></span>
+          <div class="cc-left cc-degree-progress-graduate-checkmark">
+            <i class="fa fa-check cc-icon-green" data-ng-if="requirement.status === 'Y' || requirement.status === 'P'"></i>&nbsp;
```
davidlimcheng: Why this `&nbsp;` here?

We want this `div` to appear and take up space even if it's empty to preserve the formatting of the list.  This is why `width: 10px;` is defined in the SCSS.  Interestingly enough, a `div` must have either a `height` or a space to take up width, even if the width is explicitly defined: [Stack Overflow](http://stackoverflow.com/questions/4171286/how-to-make-a-div-with-no-content-have-a-width)

Just thought it was an interesting tidbit that I learned today.

redconfetti:
Looks good to me

PR #6366, opened 2017-02-01
SISRP-30081 - Removes active/completed committee distinction

lyttam:
https://jira.berkeley.edu/browse/SISRP-30081

This PR unfortunately contains the changes from my previous one, #6364, plus some additional cleanup that I missed the first time (completely removing references to the active/completed separation of committees).


app/models/my_committees/committees_module.rb
```diff
@@ -84,32 +84,20 @@ def committee_status_icon (cs_committee)
     end
   end
 
-  def parse_cs_committee_members (cs_committee, include_inactive)
+  def parse_cs_committee_members (cs_committee)
     committee_members_result = get_empty_committee_members
-    cs_committee_members = filter_members(cs_committee[:committeeMembers], include_inactive)
+    cs_committee_members = cs_committee[:committeeMembers]
```
davidlimcheng: Is it possible for `cs_committee[:committeeMembers]` to be empty?  Should we fortify this with `cs_committee.try(:[], :committeeMembers)`?

app/models/my_committees/faculty_committees.rb
```diff
@@ -10,10 +10,7 @@ def merge(feed)
 
     def get_feed
       result = {
-        facultyCommittees: {
-          active: [],
-          completed: []
-        }
+        facultyCommittees: []
       }
       feed = CampusSolutions::FacultyCommittees.new(user_id: @uid).get[:feed]
```
davidlimcheng: Looking at `CampusSolutions::Proxy`, it looks like `:feed` is not made available if the lookup of `campus_solutions_id` fails for the user, and we'll get `{noStudentId: true}` returned instead.  Because of this, should we fortify the `[:feed]` call with a `.try` as well?

app/models/my_committees/committees_module.rb
lyttam: @davidlimcheng good question!  It should be an empty array rather than nil, but I made the change anyway just to protect against any nuance from the API.

app/models/my_committees/faculty_committees.rb
lyttam: @davidlimcheng yes.  I made the change.

davidlimcheng:
Thanks @lyttam!  LGTM once Travis is greened.

davidlimcheng:
Oops, looks like two failing Travis' @lyttam

lyttam:
@davidlimcheng Travis has been appeased

PR #6368, opened 2017-02-01
CLC-6562, ditch the KB link on bCourses site_creation page

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-6562

Per Steven's recent comment in Jira above.


paulkerschen:
The right link for CollaboTools. Thanks @johncrossman

PR #6372, opened 2017-02-02
SISRP-29452 - My Academics and Student Overview cards displaying Incorrect GPA and Unit Data - combining UG+G

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-29452

* This fix should really be done in the back-end, but doing so will require some non-trivial code changes to important feeds that I don't feel comfortable doing so close to 7.5
* Will open a JIRA to follow-up and refactor this fix (bring it to server-side) for GL8.0.
  * Opened [SISRP-30187](https://jira.berkeley.edu/browse/SISRP-30187)
* QA PR to come upon merge

### DIFF
--- a/src/assets/javascripts/angular/controllers/pages/userOverviewController.js
+++ b/src/assets/javascripts/angular/controllers/pages/userOverviewController.js
@@ -41,6 +41,7 @@ angular.module('calcentral.controllers').controller('UserOverviewController', fu
     dataSeries: []
   };
   $scope.studentSuccess = {
+    activeCareers: null,
     gpaChart: {
       series: {
         className: 'cc-student-success-color-blue'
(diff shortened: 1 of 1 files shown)


src/assets/javascripts/angular/controllers/pages/userOverviewController.js
```diff
@@ -303,6 +308,22 @@ angular.module('calcentral.controllers').controller('UserOverviewController', fu
       });
   };
 
+  /**
+   * This should be done in back-end, but requires a refactoring of MyAcademics::FilteredForAdvisor and StudentSuccess::Merged.
+   * This is a temporary fix aimed for GoLive 7.5, but should be refactored for GoLive 8.
+   */
+  var filterInactiveCareers = function() {
+    return _.remove($scope.studentSuccess.termGpa, function(term) {
```
lyttam: @davidlimcheng it looks like this might be modifying `$scope.studentSuccess.termGpa` unnecessarily, since you are also modifying it on line 269.  I would prefer `_.filter` instead of `_.remove`, unless the latter is somehow more efficient.

davidlimcheng: @lyttam Great catch.  I changed the code to remove the unnecessary definition in line 269.  I actually do think `_.remove` is more efficient in this context, because we don't really care about preserving the original array.  So, instead of creating another array object, I opted to modify the existing one and just get rid of the data we don't care about.  Let me know this looks OK.

PR #6381, opened 2017-02-04
SISRP-30239 - Only show GRAD Degree Progress card if UGRD is hidden

lyttam:
https://jira.berkeley.edu/browse/SISRP-30239

If the student has undergrad degree progress, we were showing that plus an empty graduate degree progress card.  This hides the empty card in that case.  QA PR forthcoming

### DIFF
--- a/src/assets/javascripts/angular/controllers/pages/userOverviewController.js
+++ b/src/assets/javascripts/angular/controllers/pages/userOverviewController.js
@@ -260,6 +260,8 @@ angular.module('calcentral.controllers').controller('UserOverviewController', fu
         $scope.degreeProgress.undergraduate.progresses = _.get(data, 'data.feed.degreeProgress.progresses');
         $scope.degreeProgress.undergraduate.errored = _.get(data, 'errored');
       }).finally(function() {
+        $scope.degreeProgress.undergraduate.showCard = apiService.user.profile.features.csDegreeProgressUgrdAdvising && $scope.degreeProgress.undergraduate.progresses.length;
+        $scope.degreeProgress.graduate.showCard = apiService.user.profile.features.csDegreeProgressGradAdvising && $scope.degreeProgress.graduate.progresses.length;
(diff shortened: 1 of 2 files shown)


src/assets/javascripts/angular/controllers/pages/userOverviewController.js
```diff
@@ -260,6 +260,8 @@ angular.module('calcentral.controllers').controller('UserOverviewController', fu
         $scope.degreeProgress.undergraduate.progresses = _.get(data, 'data.feed.degreeProgress.progresses');
         $scope.degreeProgress.undergraduate.errored = _.get(data, 'errored');
       }).finally(function() {
+        $scope.degreeProgress.undergraduate.showCard = apiService.user.profile.features.csDegreeProgressUgrdAdvising && $scope.degreeProgress.undergraduate.progresses.length;
+        $scope.degreeProgress.graduate.showCard = apiService.user.profile.features.csDegreeProgressGradAdvising && !$scope.degreeProgress.undergraduate.progresses.length;
```
davidlimcheng: @lyttam Shouldn't we show both cards, if the student is at the tail-end of the undergrad and beginning their graduate career?  I'm not sure if that's an actual use-case, but it seems like it could realistically be one.

Is there a reason why we don't want to key `graduate.showCard` off of `degreeProgress.graduate.progresses.length` instead?

lyttam: @davidlimcheng I did it that way assuming we should always show one or the other, but not both... but you make a good point, I think we should rely on the existence or non-existence of the data coming from CS to determine whether to display the card.  I will change the second flag to look at `degreeProgress.graduate.progresses.length`.

lyttam: @davidlimcheng Amended.

davidlimcheng:
Looks good, thanks @lyttam !

PR #6384, opened 2017-02-04
SISRP-29961 - Ensure proper ordering of milestones - Reopened

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-29961

* Peppered in some `#try`'s in some unrelated methods just in case.
* The `order_number` definition is kind of weird because both Master and PhD tracks share a milestone, "Department Final Recommendations."  I was limited to assigning this milestone only one order number, despite it showing up in different places in each respective track.  I staggered the Masters track to start at 2 to make up for this.
* QA PR to come.

### DIFF
--- a/app/models/berkeley/degree_progress_graduate.rb
+++ b/app/models/berkeley/degree_progress_graduate.rb
@@ -2,11 +2,15 @@ module Berkeley
   class DegreeProgressGraduate
 
     def self.get_status(status_code)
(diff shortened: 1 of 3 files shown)


lyttam:
@Looks good. Hopefully in a future iteration (when we figure out how to tell if milestones are for a masters or a doctoral program), we can get around that ordering weirdness with the shared milestone.

PR #6390, opened 2017-02-07
SISRP-30252 - 'Request Official Verification' - Google form link showing in addition to CS link

redconfetti:
https://jira.berkeley.edu/browse/SISRP-30252

In [SISRP-29535](https://jira.berkeley.edu/browse/SISRP-29535) a CS link that was configured in the application YAML was migrated to the CS Link API, and the `enrollment_verification_deeplink` feature flag was removed. A static link (Google form) was left in as a fallback, displayed when the previously configured link was not present, resulting in two links shown.

This update retains the original logic, however clarification is still being requested from the SR team (and may result in a follow up PR).

* Updates logic for fallback to static Google link for Enrollment Verification.
* Removes unused scope settings.


davidlimcheng:
lgtm @redconfetti

PR #6404, opened 2017-02-08
SISRP-30189 - Grading in CalCentral Accessibility Issues

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-30189

* `aria-hidden="true"` makes the content invisible to screen readers.
* I thought there wasn't really a need to have a screenreader slog through the legend when there's no use for it, since the status themselves will be dictated as speech already.


redconfetti:
Yes. I remember the italic element that is used to display an icon tripping up the screen readers, so this is a good update.

PR #6410, opened 2017-02-09
SISRP-30250 - 'Assistance with Midpoint Grading: General' needs to be added to the Grading tab on Class Roster

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-30250

QA PR to come.


lyttam:
Looks good.

PR #6413, opened 2017-02-09
[QA] SISRP-30250 - 'Assistance with Midpoint Grading: General' needs to be added to the Grading tab on Class Roster

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-30250

QA PR of #6410


redconfetti:
Looks the same to me, only QA! 🍎  👏

PR #6418, opened 2017-02-10
SISRP-30364 - Hides link on Committees card until the e-form is ready

lyttam:
https://jira.berkeley.edu/browse/SISRP-30364

Prevents the link URL from being handed to the front end, but leaves the code/markup to handle it in place since we expect to be able to surface this link in the next release.


lyttam:
Cancelled... will disable the link on the CS side instead.

PR #6417, opened 2017-02-10
CLC-6575, Canvas People tool: custom js for both React-based UI and the legacy UI

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-6575

Instructure tells us the new React-based UI will go live on Feb 18, the same day as our scheduled CalCentral/Junction release. This PR prepares the Add People tool for the new Canvas UI and preserves legacy support in case these future releases do not happen as scheduled.

PR #6419, opened 2017-02-10
[QA] CLC-6575, Canvas People tool: custom js for both React-based UI and the legacy UI

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-6575

master PR is #6417


johncrossman:
@pfarestveit wants to verify this on canvas-test today, before the legacy UI is gone from Instructure's test tiers. _Note:_ He did a quick check of the fix on master, using canvas-beta.

PR #6428, opened 2017-02-14
CLC-6575, Canvas React-based UI: customize error in Add People

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-6575

The times they are a changin'... Dynamically tweak the 'Add People' error dialog:
* We attach listeners to the 'Cancel', 'Back' and 'Next' buttons because, as long as you are on the Add People page, they are not re-drawn.
* Unlike the old UI, the bad email addresses (or IDs) are not rendered in an easy-to-isolate list. We must extract the data from a table and then remove the table. We are removing the option to create new users using unrecognized input.

### DIFF
--- a/public/canvas/canvas-customization.js
+++ b/public/canvas/canvas-customization.js
@@ -215,6 +215,10 @@
     };
   };
 
+  var modifiedMarker = function() {
+    return 'calcentral-modified';
+  };
+
   var constructFindPersonLink = function(toolId) {
     if (toolId) {
       var linkUrl = window.ENV.COURSE_ROOT_URL + '/external_tools/' + toolId;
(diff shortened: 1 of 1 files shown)


public/canvas/canvas-customization.js
```diff
@@ -329,20 +332,88 @@
...
+  };
+
+  var listenersForAddPeopleError = function(backButton) {
+    addPeopleNextClickable(false);
+    // The 'Next' button will be re-enabled when user clicks 'Back' or 'Cancel'.
+    backButton[0].addEventListener('click', enableNextButton);
+    var cancelButton = backButton.siblings(':button').first();
+    if (cancelButton.length > 0) {
+      cancelButton[0].addEventListener('click', enableNextButton);
+    }
```
johncrossman: Note: I just amended the line above such that `siblings()` is filtered by ':button'. This is our best attempt at reaching the Cancel button

johncrossman: See `Cancel` button in screenshot below: 
image

paulkerschen:
Thanks for the PeopleTools, @johncrossman.

PR #6430, opened 2017-02-14
CLC-6579, Help icon ('How do I add people to my site?') on tab of Add People tool

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-6579

Fortunately, both old and new UIs have same id for tab in question. Notice that I ditched `modifiedMarker()` function in favor of local vars.

### DIFF
--- a/public/canvas/canvas-customization.js
+++ b/public/canvas/canvas-customization.js
@@ -215,10 +215,6 @@
     };
   };
 
-  var modifiedMarker = function() {
-    return 'calcentral-modified';
-  };
-
   var constructFindPersonLink = function(toolId) {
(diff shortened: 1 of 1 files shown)


public/canvas/canvas-customization.js
```diff
@@ -215,10 +215,6 @@
     };
   };
 
-  var modifiedMarker = function() {
-    return 'calcentral-modified';
-  };
```
johncrossman: Function begone. 😜  cc: @paulkerschen

paulkerschen: Modified!

PR #6431, opened 2017-02-14
SISRP-30254 - Add FERPA management link to CS Link API

redconfetti:
https://jira.berkeley.edu/browse/SISRP-30254

Migrates from a CS API that provided the FERPA Restriction Management link, to the CS Link API


lyttam:
@redconfetti Nice to be removing all that code.  Feels good man.

PR #6444, opened 2017-02-16
SISRP-30513 - Grading My Academics merged model reporting errors in QA

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-30513

* Really glad this was caught, but a bit worried that it was caught during load testing and not beforehand.
* Fixed an error check that was only checking for Final grade status.
* Needed to fortify the code a bit more for nil values from APIs, and fudged on an error object that needed to include both final and midpoint values.
* QA PR to come upon merge.

### DIFF
--- a/app/models/my_academics/grading.rb
+++ b/app/models/my_academics/grading.rb
@@ -179,7 +179,7 @@ module MyAcademics
     def get_cs_status(ccn, is_law, term_code)
       cnn_status = nil
       if (grading_feed = get_grading_data)
-        grading_statuses = grading_feed[:feed].try(:[],:ucSrClassGrading).try(:[],:classGradingStatuses)
+        grading_statuses = grading_feed.try(:[],:feed).try(:[],:ucSrClassGrading).try(:[],:classGradingStatuses)
         cnn_status = find_ccn_grading_statuses(grading_statuses, ccn, is_law, term_code)
       end
       cnn_status
@@ -220,7 +220,7 @@ module MyAcademics
     end
 
     def parse_cs_grading_status(cs_grading_status, is_law)
-      return :noCsData if unexpected_cs_status?(cs_grading_status, is_law)
+      return {finalStatus: :noCsData, midpointStatus: :noCsData} if unexpected_cs_status?(cs_grading_status, is_law)
       cs_grading_status[:finalStatus] = case cs_grading_status[:finalStatus]
         when 'GRD'
           :GRD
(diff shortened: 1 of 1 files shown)


app/models/my_academics/grading.rb
```diff
@@ -250,12 +250,16 @@ def parse_cc_grading_status(cs_grading_status, is_law, is_midpoint, term_code)
...
 
     def unexpected_cs_status?(cs_grading_status, is_law)
-      return false if !!%w{GRD RDY APPR POST}.find { |s| s == cs_grading_status[:finalStatus] } || cs_grading_status[:finalStatus].blank?
-      if !is_law
-        return false if !!%w{APPR NRVW RDY}.find { |s| s == cs_grading_status[:midpointStatus] } || cs_grading_status[:midpointStatus].blank?
+      has_error = false
+      if cs_grading_status.nil? ||
+         !(!!%w{GRD RDY APPR POST}.find { |s| s == cs_grading_status.try(:[],:finalStatus) } || cs_grading_status.try(:[],:finalStatus).blank?) ||
+         (!is_law && !(!!%w{APPR NRVW RDY}.find { |s| s == cs_grading_status.try(:[],:midpointStatus) } || cs_grading_status.try(:[],:midpointStatus).blank?))
+        has_error = true
```
lyttam: @davidlimcheng since it's late in the game, this is more a future suggestion than anything - but it would be nice to have this logic encapsulated in functions with useful names that describe the meaning of the logic.  For example, 

```
def unexpected_cs_status?(cs_grading_status, is_law)
    has_error = false
    if cs_grading_status.nil? || !special?(cs_grading_status) || . . . 
end

def special?(cs_grading_status)
    (!!%w{GRD RDY APPR POST}.find { |s| s == cs_grading_status.try(:[],:finalStatus) }
end
```

Replacing 'special' with some word that describes what `[GRD, RDY, APPR, POST]` means in plain English.

davidlimcheng: @lyttam That's a great suggestion and makes this much, much more readable.  I'd actually like to change this now, it shouldn't take too long.  I'll amend.

app/models/my_academics/grading.rb
```diff
@@ -250,12 +250,16 @@ def parse_cc_grading_status(cs_grading_status, is_law, is_midpoint, term_code)
...
       end
-      logger.debug "Unexpected CS Final Grading Status Received (Final: #{cs_grading_status[:finalStatus]}#{', Midpoint: ' + cs_grading_status[:midpointStatus] if cs_grading_status.key?(:midpointStatus)}) for Class #{self.class.name} feed, uid = #{@uid}"
-      true
+      if has_error && cs_grading_status.present?
+        logger.debug "Unexpected CS Final Grading Status Received (Final: #{cs_grading_status[:finalStatus]}#{', Midpoint: ' + cs_grading_status[:midpointStatus] if cs_grading_status.key?(:midpointStatus)}) for Class #{self.class.name} feed, uid = #{@uid}"
```
redconfetti: Did you test this logger statement in some manner?

redconfetti: @davidlimcheng - Actually this works, I just have never used a conditional statement within an interpolated statement like this, I wasn't sure how it would work - https://repl.it/Fm8A

PR #6445, opened 2017-02-16
SISRP-30543 - Hides the committee status for QE committees

lyttam:
https://jira.berkeley.edu/browse/SISRP-30543

It was determined that more complex logic needs to be involved with displaying the committee status for Qualifying Exam committees, so we're going to hide this line for 7.5.


redconfetti:
LGTM

PR #6451, opened 2017-02-17
CLC-6586 OEC: Remove one-year-ago cutoff for old instructor data

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6586

From this day forward, we wish to preserve instructor data for all time.


redconfetti:
Looks beautiful to me

PR #6452, opened 2017-02-21
CLC-6564 Add term name to initial course site name

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6564


johncrossman:
Looks right to me, @raydavis  Waiting on T.

PR #6455, opened 2017-02-22
CLC-6588 OEC: Highlight salient differences in diff report

paulkerschen:
Per requirements in https://jira.ets.berkeley.edu/jira/browse/CLC-6588:
- The symbolic '+/-' column is replaced with a more verbose 'REASON' column.
- Rows with multiple discrepancies will appear multiple times in the diff, each time with a different value in the 'REASON' column.
- Values without discrepancies are omitted from the report, apart from 'COURSE_NAME', which enables identification.

### DIFF
--- a/app/models/oec/report_diff_task.rb
+++ b/app/models/oec/report_diff_task.rb
@@ -5,6 +5,8 @@ module Oec
 
     attr_accessor :diff_report
 
+    COLUMNS_TO_COMPARE = %w(COURSE_NAME FIRST_NAME LAST_NAME EMAIL_ADDRESS DEPT_FORM EVALUATION_TYPE MODULAR_COURSE START_DATE END_DATE)
+
     def run_internal
       @confirmations_folder = @remote_drive.find_nested([@term_code, Oec::Folder.confirmations])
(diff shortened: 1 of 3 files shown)


app/models/oec/report_diff_task.rb
```diff
@@ -98,32 +100,49 @@ def add_rows_to_diff(diff_report, dept_code, sis_data, dept_data, keys)
...
-          diff_row[column] = dept_row ? dept_row[column] : nil
+
+        if !sis_row
+          # Add a single row with values from the department sheet.
+          diff_report[diff_key] = diff_row.merge(dept_row.slice('LDAP_UID', *COLUMNS_TO_COMPARE)).merge({
+            'REASON' => 'Not in SIS'
+          })
+
+        elsif !dept_row
+          #Add a single row with values from the SIS import sheet.
```
johncrossman: @paulkerschen Space needed before "Add"

johncrossman:
Looks good, @paulkerschen  I suppose I'll wait for the trivial fix... then merge.

paulkerschen:
Thanks @johncrossman, whitespace fixed.

PR #6456, opened 2017-02-23
SISRP-28700 - Adds expiry handler for undergrad Degree Progress

lyttam:
https://jira.berkeley.edu/browse/SISRP-28700

The Degree Progress card for undergrads gets its data from the Academic Progress Report.  Advisors want to be able to run the APR, which generates a new set of data, and see that data reflected in CalCentral right away rather than wait for the cache to expire.

An event will fire when the APR is run for a single student.  The event notification allows for an array of student IDs in case we want to have the event fire when the APR is run in batch (for many students).

Payload for the event notification is detailed here:  https://docs.google.com/document/d/10uOE80vkeDXpf-7jv_I0vfnIlSpZUsUTvKIa1zLkZ1k/edit#heading=h.qa4txi13e9gh

### DIFF
--- a/app/models/campus_solutions/degree_progress/undergrad_requirements_expiry.rb
+++ b/app/models/campus_solutions/degree_progress/undergrad_requirements_expiry.rb
@@ -1,0 +1,13 @@
+module CampusSolutions
+  module DegreeProgress
+    module UndergradRequirementsExpiry
+      def self.expire(uids=[])
+        uids.each do |uid|
+          [::DegreeProgress::UndergradRequirements].each do |klass|
+            klass.expire uid
+          end
+        end
+      end
+    end
+  end
+end
--- a/app/models/notifications/sis_expiry_processor.rb
+++ b/app/models/notifications/sis_expiry_processor.rb
@@ -2,13 +2,26 @@ module Notifications
   class SisExpiryProcessor
     include ClassLogger
 
+    PROVIDERS = Hash[
+      'student' => SisExpirySingleStudentProvider.new,
+      'students' => SisExpiryStudentsProvider.new
+    ]
+
     def process(event, timestamp)
       return false unless accept? event
       begin
         logger.debug "Processing event: #{event}; timestamp = #{timestamp}"
         if (expiry_module = get_expiry event)
-          if (uid = (event['topic'] == 'sis:faculty:grade-roster') ? (get_instructor_uids event) : (get_uid event))
-            expiry_module.expire uid
+
+          if event['topic'] == 'sis:faculty:grade-roster'
+            uids = get_instructor_uids event
+            expiry_module.expire uids
+          else
+            payload = event.try(:[], 'payload')
+            payload.keys.each do |key|
+              uids = PROVIDERS[key].get_uids(event)
+              expiry_module.expire uids if uids
+            end
           end
         else
           logger.warn "Event topic #{event['topic']} not recognized"
@@ -26,16 +39,6 @@ module Notifications
 
     def get_expiry(event)
       EXPIRY_BY_TOPIC[event['topic']]
-    end
-
-    def get_uid(event)
-      if (campus_solutions_id = event['payload'] && event['payload']['student'] && event['payload']['student']['StudentId'])
-        uid = CalnetCrosswalk::ByCsId.new(user_id: campus_solutions_id).lookup_ldap_uid
-        logger.error "No UID found for Campus Solutions ID #{campus_solutions_id}" unless uid
-      else
-        logger.error "Could not parse Campus Solutions ID from event #{event}"
-      end
-      uid
     end
 
     def get_instructor_uids(event)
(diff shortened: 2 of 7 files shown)


lyttam:
I'm hoping @davidlimcheng and/or @raydavis can review this and tell me if I'm doing it wrong, since I can't really test it locally.

lyttam:
Update:  as @davidlimcheng  pointed out, the payload structure that we're expecting has diverged - actually into three different versions.  

I have added handling for two of the versions - student (single ID) and students (list of IDs).  In the future, all events using the single ID payload will be replaced by a list of IDs that contains one item, at which point we can retire `SisExpiryStudentProvider`.

I'll leave it to David to add handling for the third type of payload (sis:faculty:grade-roster).

app/models/notifications/sis_expiry_students_provider.rb
```diff
@@ -0,0 +1,17 @@
+module Notifications
+  class SisExpiryStudentsProvider
+    include ClassLogger
+
+    def get_uids(student_id)
```
davidlimcheng: @lyttam Looking at line 7, is the input param supposed to be `ids`?

app/models/campus_solutions/degree_progress/undergrad_requirements_expiry.rb
```diff
@@ -0,0 +1,13 @@
+module CampusSolutions
+  module DegreeProgress
+    module UndergradRequirementsExpiry
+      def self.expire(uids=[])
+        uids.each do |uid|
+          [::DegreeProgress::UndergradRequirements].each do |klass|
```
davidlimcheng: @lyttam Is there a reason why there are two colons prepending the module/class?

davidlimcheng:
@lyttam I really like the approach you took.  It makes a lot of sense and provides an easy way to implement any further custom parsing definitions.  

The only concern I have is that `SisExpiryStudentsProvider` assumes the input to be an array.  We asked Tamer for this behavior, but he was a bit hesitant about committing to an 8.0 release.  He asked for us to check-in on Monday for an update, so would you mind if we wait until Monday afternoon to merge this code?

app/models/campus_solutions/degree_progress/undergrad_requirements_expiry.rb
lyttam: @davidlimcheng Yes!  I'm glad you asked.  The `::` represents the root level module.  If I didn't include it here, line 6 would refer to `CampusSolutions::DegreeProgress:: UndergradRequirements` since we're in the `CampusSolutions` module already.

app/models/notifications/sis_expiry_processor.rb
```diff
@@ -2,13 +2,26 @@ module Notifications
   class SisExpiryProcessor
     include ClassLogger
 
+    PROVIDERS = Hash[
+      'student' => SisExpiryStudentProvider.new,
+      'students' => SisExpiryStudentsProvider.new
```
lyttam: Not sure if `new`ing here is the properest or most beautiful way of doing things.  Anyone prefer a different way?

app/models/notifications/sis_expiry_students_provider.rb
lyttam: @davidlimcheng Hmm yes.  I will fix that and maybe write a unit test that will catch it.

app/models/campus_solutions/degree_progress/undergrad_requirements_expiry.rb
davidlimcheng: Very good to know!

app/models/notifications/sis_expiry_single_student_provider.rb
```diff
@@ -0,0 +1,18 @@
...
+
+    def get_uids(event)
+      if (campus_solutions_id = event.try(:[], 'payload').try(:[], 'student').try(:[], 'StudentId'))
+        uid = CalnetCrosswalk::ByCsId.new(user_id: campus_solutions_id).lookup_ldap_uid
+        logger.error "No UID found for Campus Solutions ID #{campus_solutions_id}" unless uid
+      else
+        logger.error "Could not parse Campus Solutions ID from event #{event}"
+      end
+      uid
+    end
```
redconfetti: @lyttam, I notice that this returns a single UID and uses the plural `get_uids` instead of singular `get_uid`? Should it return the single UID in an array, to match the interface where SisExpiryStudentsProvider#get_uids returns an array?

lyttam: @redconfetti Yes, it would be nice if the return type matched.  But then I'd have to update every `CampusSolutions::*Expiry` to accept an array... and I wanted to avoid touching those.  Presumably, `SisExpirySingleStudentProvider` will be going away after we coordinate with Tamer to get each of the event notifications refactored to always send an array.

lyttam:
@davidlimcheng I just discussed array vs. single ID with Tamer.  It turns out that Marc is the one who would be able to assure us that `<students />` will always contain an array, and `<student />` will always contain a single node.

Can we merge for now and discuss with him when we meet tomorrow?

davidlimcheng:
@lyttam Yep sounds good to me.  I'll also get my piece in, under the assumption that the potentially-multiple-ID events will always contain an array.

PR #6461, opened 2017-02-28
SISRP-30491 - 8.0 - Grading - Advisor Facing Midterm Grades

davidlimcheng:
SISRP-30491 - 8.0 - Grading - Advisor Facing Midterm Grades

* Surfaces student midpoint grades to advisors.
* No work needed in the back-end thankfully, due to `AcademicPlan` pulling directly from `Semesters`.

### DIFF
--- a/src/assets/stylesheets/_academics.scss
+++ b/src/assets/stylesheets/_academics.scss
@@ -220,10 +220,6 @@
   .cc-academics-link-disabled {
     color: $cc-color-light-medium-grey;
   }
-  .cc-academics-midpoint-show-hide {
-    color: $cc-color-link;
-    cursor: pointer;
-  }
   .cc-academics-nocontent-container {
     margin-top: 15px;
   }
--- a/src/assets/templates/widgets/academic_plan.html
+++ b/src/assets/templates/widgets/academic_plan.html
@@ -17,7 +17,7 @@
     <ul class="cc-widget-list cc-widget-list-border-top">
         <li
           data-ng-repeat="semester in planSemesters"
-          class="cc-academics-semester-list cc-widget-list-hover cc-enrollment-card-section cc-table"
+          class="cc-widget-list-hover cc-table"
           data-ng-class="{'cc-widget-list-hover-opened':(semester.show)}"
           data-cc-accessible-focus-directive
           data-ng-click="api.widget.toggleShow($event,  null, semester, 'Student Overview - Academic Plan')">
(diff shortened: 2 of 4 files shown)


src/assets/templates/widgets/academic_plan.html
```diff
@@ -1,4 +1,4 @@
-<div class="cc-widget">
+<div class="cc-widget cc-page-academics">
```
lyttam: @davidlimcheng seems a bit odd to apply this class to a card.  It contains overall styles for the page (e.g. header styles) as well as styles that apply to specific cards like Enrollment.

I think our scss could stand a bit of refactoring.  This isn't on you for this release, but in the future we should consider keeping page-level styles in a place like `_academics.scss`, and putting card-specific styles like `cc-class-enrollment-formatted-content` in a place like `_enrollment_card.scss`.

src/assets/templates/widgets/academic_plan.html
```diff
@@ -80,17 +80,38 @@ <h3 class="cc-left">
                     <tr>
                       <th width="60%" scope="col">Enrolled</th>
                       <th width="25%" scope="col" title="Units">Units</th>
-                      <th data-ng-if="api.user.profile.canViewGrades" width="15%" scope="col"></th>
+                      <th data-ng-if="api.user.profile.canViewGrades" width="15%" scope="col" title="Grades">Grade</th>
                     </tr>
                   </thead>
                   <tbody data-ng-repeat="class in semester.enrolledClasses">
-                    <tr data-ng-if="!class.transcript && section.is_primary_section" data-ng-repeat="section in class.sections">
+                    <tr data-ng-if="!class.transcript && section.is_primary_section" data-ng-repeat-start="section in class.sections">
```
lyttam: @davidlimcheng I didn't know about `ngRepeatStart` - cool!

lyttam:
@davidlimcheng Just one concern about the use of `cc-page-academics` class. Otherwise, looks good.

src/assets/templates/widgets/academic_plan.html
davidlimcheng: @lyttam I totally agree - the reasoning behind adding the `cc-page-academics` style was because of the existing styles on line 20, as well as the plan to eventually add this card to the My Academics page when we start surfacing it to students.  I do think we should definitely avoid using page-level styles on specific cards, and `_academics.scss` is a melting pot of card-level stylings.  I created [SISRP-30977](https://jira.berkeley.edu/browse/SISRP-30977) so that this doesn't fall through the cracks.

davidlimcheng: In the meantime, I'll go ahead split up this card from `_academics.scss`.  Amendment to come.

src/assets/stylesheets/_academics.scss
```diff
@@ -220,10 +220,6 @@
   .cc-academics-link-disabled {
     color: $cc-color-light-medium-grey;
   }
-  .cc-academics-midpoint-show-hide {
```
davidlimcheng: @lyttam Since this style needed to be shared on both My Academics and Student Overview, I moved it to `_useful.scss` as a more general "non-anchored-link" style.  I can get more into why we needed to implement this, if interested.

src/assets/templates/widgets/academic_plan.html
```diff
@@ -17,7 +17,7 @@ <h2 class="cc-left">Academic Plan</h2>
     <ul class="cc-widget-list cc-widget-list-border-top">
         <li
           data-ng-repeat="semester in planSemesters"
-          class="cc-academics-semester-list cc-widget-list-hover cc-enrollment-card-section cc-table"
```
davidlimcheng: @lyttam It turns out we didn't even need to use these styles, since `cc-widget-list-hover` and `cc-table` were enough, so I just got rid of the `academics` parent style altogether.

PR #6463, opened 2017-03-01
SISRP-30489 - Develop CalCentral Consumption of Class Grade-Roster JMS

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-30489

### DIFF
--- a/app/models/notifications/sis_expiry_class_provider.rb
+++ b/app/models/notifications/sis_expiry_class_provider.rb
@@ -1,0 +1,25 @@
+module Notifications
+  class SisExpiryClassProvider
+    include ClassLogger
+
+    def get_uids(event)
+      if event && ids = event.try(:[], 'payload').try(:[], 'class').try(:[], 'instructors').try(:[], 'id')
+        uids = []
+        # 'id' will contain a single value if there is one ID, and an array if there are multiple,
+        # so we always wrap it in an array for consistency.
+        ids = ([] << ids).flatten
+        ids.each do |instructor_id|
+          if (uid = CalnetCrosswalk::ByCsId.new(user_id: instructor_id).lookup_ldap_uid)
+            uids << uid
+          else
+            logger.error "No UID found for Campus Solutions ID #{instructor_id}"
+          end
+        end
+        uids
+      else
+        logger.error "Could not parse Campus Solutions ID from event #{event}"
+      end
+    end
+
+  end
+end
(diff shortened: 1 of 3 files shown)


davidlimcheng:
Paging @lyttam for review

app/models/notifications/sis_expiry_processor.rb
```diff
@@ -3,31 +3,22 @@ class SisExpiryProcessor
...
-          logger.warn "Event topic #{event['topic']} not recognized"
+      logger.debug "Processing event: #{event}; timestamp = #{timestamp}"
+      if (expiry_module = get_expiry event)
+        payload = event.try(:[], 'payload')
+        payload.keys.each do |key|
+          uids = PROVIDERS[key].get_uids(event)
+          expiry_module.expire uids if uids
         end
-      ensure
-        ActiveRecord::Base.clear_active_connections!
```
davidlimcheng: Since we're no longer opening a database connection for any of our events, I thought it was safe to get rid of this line.

app/models/notifications/sis_expiry_class_provider.rb
```diff
@@ -0,0 +1,25 @@
+module Notifications
+  class SisExpiryClassProvider
+    include ClassLogger
+
+    def get_uids(event)
+      if event && ids = event.try(:[], 'payload').try(:[], 'class').try(:[], 'instructors').try(:[], 'id')
+        uids = []
+        # 'id' will contain a single value if there is one ID, and an array if there are multiple,
+        # so we always wrap it in an array for consistency.
+        ids = ([] << ids).flatten
```
lyttam: @davidlimcheng neat way of casting to an array; you could also use `Array(ids).each ...`

app/models/notifications/sis_expiry_processor.rb
```diff
@@ -41,22 +32,6 @@ def get_expiry(event)
       EXPIRY_BY_TOPIC[event['topic']]
     end
 
-    def get_instructor_uids(event)
-      uids = []
-      if (section_data = event['payload'] && event['payload']['student'] && event['payload']['student']['StudentId'].to_s.split('|'))
-        if (term_id = section_data[0]) && (section_id = section_data[1])
-          instructors = EdoOracle::Queries.get_section_instructors(term_id, section_id)
```
lyttam: @davidlimcheng just curious, was there a reason for getting the instructor's UID from EDO DB previously?  Now you'll be getting it from CalNet.

davidlimcheng: @lyttam Yeah, we were querying the EDODB because the event would only send us term ID and section ID.  We probably could have called the Class API as well, but EDODB seemed like the better choice to get all instructors for a certain section.

PR #6464, opened 2017-03-01
SISRP-28717 Update SOR to CS for CalCentral

nkutub:
https://jira.berkeley.edu/browse/SISRP-28717

added warning for legacy database use

data source flow gates controlled by new flag

added allow_legacy_fallback flag to semesters model

added allow_legacy_fallback flag to all legacy? logic

### DIFF
--- a/app/models/campus_oracle/queries.rb
+++ b/app/models/campus_oracle/queries.rb
@@ -186,6 +186,7 @@ module CampusOracle
     end
 
     def self.get_enrolled_students(ccn, term_yr, term_cd)
+      logger.warn 'Calling get_enrolled_students on campus_oracle when allow_legacy_fallback flag set to false' unless Settings.features.allow_legacy_fallback
       result = []
       use_pooled_connection {
         sql = <<-SQL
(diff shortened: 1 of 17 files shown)


app/models/campus_oracle/queries.rb
```diff
@@ -186,6 +186,7 @@ def self.get_reg_status(person_id, term_yr, term_cd)
     end
 
     def self.get_enrolled_students(ccn, term_yr, term_cd)
+      logger.warn 'Calling get_enrolled_students on campus_oracle when allow_legacy_fallback flag set to false' unless Settings.features.allow_legacy_fallback
```
nkutub: @raydavis, I was thinking of that, but I figured it should not run these queries at all if a did the job right. How have you done this in the past to log only once a day?

raydavis: If the flag is enabled at all, you're guaranteed a flood of noise until the next deploy. And if the flag is never enabled, then there was no point to having it. :)

For something so absolutely core to CalCentral, bCourses, and OEC services, I think logging once at startup is probably sufficient. But looking for a very long-lived cached query would be a possibility too. (That's how we occasionally-but-not-for-every-student-every-time logged warnings about impending obsolescence of the Financial Aid dates.) ``Berkeley::Terms`` is an example in this case.

raydavis: In chat, @nkutub pointed out that I might want to scroll the GitHub preview window over a bit so I could see that the warnings only happened if the flag was supposed to block the queries. Duh.

It was a nice chat, anyway.

PR #6467, opened 2017-03-03
SISRP-30588 - CalCentral Code Fix: Transfer Credit Report Links

redconfetti:
https://jira.berkeley.edu/browse/SISRP-30588

A conflict arose where a link intended only for student was embedded in the students Transfer Credits model. Passing links down the chain to the model to determine if it should provide the 'Student' links, or the 'Advisor' links would be inappropriate.

After some analysis I see that the Student Overview page is sharing a filtered version of the My Academics feed, which doesn't include links intended for Advisors. It's pure student data, yet only the data Advisors should see for the student.

The trend thus far has been to retrieve links for Advisors from the Advising Resources feed shared between the 'Advising Resources' card on the My Dashboard page, and the Personal Summary card on the Student Overview (containing student specific links that include the EMPLID).

Because this is the first time that we are placing a student specific link in a card, intended only for an Advisor, I've decided to implement two new My Academics modules specifically for Advisor Links and Student Links, each merged into the data feed respectively based on which type of user is viewing the data.

Moving forward we should add CS links to these models if they're going to be shown on the My Academics page (MyAcademics::StudentLinks) or Student Overview page (MyAcademics::AdvisorLinks).

### DIFF
--- a/app/models/my_academics/advisor_links.rb
+++ b/app/models/my_academics/advisor_links.rb
@@ -1,0 +1,18 @@
+module MyAcademics
+  class AdvisorLinks
+    include AcademicsModule
+    include ClassLogger
+
+    def merge(data)
+      data[:advisorLinks] = links
+    end
+
+    def links
+      campus_solutions_id = CalnetCrosswalk::ByUid.new(user_id: @uid).lookup_campus_solutions_id
+      tcReportLink = fetch_link('UC_CX_XFER_CREDIT_REPORT_ADVSR', { :EMPLID => campus_solutions_id })
+      {
+        tcReportLink: tcReportLink
+      }
+    end
+  end
+end
--- a/app/models/my_academics/advisor_links.rb
+++ b/app/models/my_academics/advisor_links.rb
@@ -8,10 +8,12 @@ module MyAcademics
     end
 
     def links
-      campus_solutions_id = CalnetCrosswalk::ByUid.new(user_id: @uid).lookup_campus_solutions_id
+      campus_solutions_id = get_campus_solutions_id(@uid)
       tcReportLink = fetch_link('UC_CX_XFER_CREDIT_REPORT_ADVSR', { :EMPLID => campus_solutions_id })
+      updatePlanUrl = fetch_link('UC_CX_PLANNER_ADV_STDNT', {:EMPLID => campus_solutions_id})
       {
-        tcReportLink: tcReportLink
+        tcReportLink: tcReportLink,
+        updatePlanUrl: updatePlanUrl
       }
     end
   end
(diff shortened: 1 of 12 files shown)


app/models/my_academics/advisor_links.rb
```diff
@@ -0,0 +1,18 @@
...
+  class AdvisorLinks
+    include AcademicsModule
+    include ClassLogger
+
+    def merge(data)
+      data[:advisorLinks] = links
+    end
+
+    def links
+      campus_solutions_id = CalnetCrosswalk::ByUid.new(user_id: @uid).lookup_campus_solutions_id
```
davidlimcheng: @redconfetti I don't think it's necessary to change in this PR, but for future use, we can always use the `CampusSolutionsIdRequired` mixin to get rid of this line.  That will ensure that the call is made at the proxy level, in `CampusSolutions::Proxy`.

spec/models/my_academics/advisor_links_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  subject do
+    {}.tap {|x| MyAcademics::AdvisorLinks.new(uid).merge(x)}
+  end
+
+  it 'merges student links into feed' do
+    expect(subject[:advisorLinks]).to be
+    expect(subject[:advisorLinks][:tcReportLink]).to be
+    expect(subject[:advisorLinks][:tcReportLink][:name]).to eq 'Transfer Credit Report'
+    expect(subject[:advisorLinks][:tcReportLink][:url]).to eq "https://bcswebqat.is.berkeley.edu/psp/bcsqat/EMPLOYEE/HRMS/c/SSR_ADVISEE_OVRD.CSU_DA_TRN_CDT.GBL?EMPLID=#{user_cs_id}"
```
davidlimcheng: @redconfetti Since this is `{fake: true}`, this shouldn't fail if we were pointing to another environment and ran the test, right?

redconfetti: As far as I know the CS Link proxy is configured to be fake in the TEST environment, so it has to get the fixture data, and never the live data.

I wonder if this might fail in the `testext` environment.

redconfetti: @davidlimcheng - After the tests pass, this will be good. I realize that I should have stubbed the CS Link models responses to make this test safe, even in TESTEXT env

redconfetti: @davidlimcheng - If you look you'll see that the two Rspec tests now stub the response from CampusSolutions::Link to ensure that the tests run consistently in `test` and `testext` Rails environments.

davidlimcheng:
Looks good @redconfetti!  Just need some branch conflicts to be resolved, but LGTM.

davidlimcheng:
Thanks @redconfetti

PR #6475, opened 2017-03-07
SISRP-30927 - 8.0 - Academic Summary - Student Profile Section

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-30927

* What's the team's feeling on having half-done features hanging out in DEV/QAT environments before release?  I didn't want to add a feature flag because we don't really need it, but if I leave this feature for one giant PR, it's going to be huge.  So I'm developing this piece-wise, but that means piece-wise functionality in our development environments.  What do you guys think?  Also very interested in PDG's opinion as well (@raydavis @lyttam @redconfetti @nkutub)
* Pretty big PR, so happy to go over this in person as well.
* I was debating for a very long time to use the existing `academics_student_profile.html` and `college_and_level.html` templates, but the designs are different enough that the dynamic classes would have been, in my opinion, way more complex than creating a whole new template.  I actually tried it, and it just became very messy, very quickly.
* A lot of changes are just fixed indentation on existing templates.

### DIFF
--- a/app/models/my_academics/gpa_units.rb
+++ b/app/models/my_academics/gpa_units.rb
@@ -36,9 +36,12 @@ module MyAcademics
           result = result.merge(totalUnitsForGpa)
           unit_sum += result[:totalUnitsForGpa]
         end
-        if (totalUnitsNotForGpa = parse_hub_total_units_not_for_gpa status) && totalUnitsNotForGpa.present?
-          result[:totalUnitsNotForGpa] = totalUnitsNotForGpa
-          unit_sum += result[:totalUnitsNotForGpa]
+        if (totalUnitsTakenNotForGpa = parse_hub_total_units_taken_not_for_gpa status) && totalUnitsTakenNotForGpa.present?
+          result[:totalUnitsTakenNotForGpa] = totalUnitsTakenNotForGpa
+        end
+        if (totalUnitsPassedNotForGpa = parse_hub_total_units_passed_not_for_gpa status) && totalUnitsPassedNotForGpa.present?
+          result[:totalUnitsPassedNotForGpa] = totalUnitsPassedNotForGpa
+          unit_sum += result[:totalUnitsPassedNotForGpa]
         end
         if (unit_total != unit_sum)
           logger.warn("Hub unit conflict for UID #{@uid}: Total units (#{unit_total}) does not match summed units (#{unit_sum}).")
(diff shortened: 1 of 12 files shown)


davidlimcheng:
Need to make a quick amendment tomorrow morning, based on new comments on the JIRA ticket.  Please hold off on merging until I comment once again.

davidlimcheng:
OK, PR has been updated and is ready to go.

src/assets/templates/directives/academic_plans.html
```diff
@@ -1,16 +1,16 @@
...
+        <th data-ng-pluralize count="plans.length" when="options.labels"></th>
+        <td>
+          <div data-ng-repeat="plan in plans">
+            <div class="cc-text-light" data-ng-if="plan.college" data-ng-bind="plan.college"></div>
+            <div><strong data-ng-bind="plan.{{type}}"></strong></div>
+            <div class="cc-widget-profile-indent" data-ng-bind="plan.subPlan"></div>
+          </div>
+        </td>
+      </tr>
     </tbody>
```
redconfetti: 👏 👏 👏 👏 👏

app/models/my_academics/gpa_units.rb
```diff
@@ -73,7 +76,13 @@ def parse_hub_total_units_for_gpa(status)
...
 
-    def parse_hub_total_units_not_for_gpa(status)
+    def parse_hub_total_units_taken_not_for_gpa(status)
+      if (units = status['cumulativeUnits']) && (total_units = units.find { |u| u['type'] && u['type']['code'] == 'Not For GPA'})
+        total_units['unitsTaken'].to_f
+      end
+    end
+
+    def parse_hub_total_units_passed_not_for_gpa(status)
       if (units = status['cumulativeUnits']) && (total_units = units.find { |u| u['type'] && u['type']['code'] == 'Not For GPA'})
```
redconfetti: @davidlimcheng - Was there no corresponding update to be made to `spec/models/my_academics/gpa_units_spec.rb`?

redconfetti:
IMHO, as long as the "half-done" features don't break existing functionality then I don't have a problem with making PR's piecemeal.

If possible, I think it's best if the feature isn't revealed in any obvious fashion when possible if the feature isn't completed. Like this is a new page, so if the page isn't linked to from the My Academics page, that would keep the release clean.

app/models/my_academics/gpa_units.rb
davidlimcheng: @redconfetti I took a look at the spec, but I felt there were enough checks on the data structure already, and all the existing tests were still valid.

davidlimcheng:
Thanks @redconfetti.  This page is currently being linked to from My Academics, but as I was telling @lyttam, if push comes to shove, I can always feature flag it before release.  But, if you think I should do that now before merging, I can add that in, if it'll alleviate any lingering worries.

lyttam:
As I already articulated to @davidlimcheng IRL, I'm fine with having half-finished features be _visible_ in DEV and QA.  I want to strive for showing features to functionals in a partial state so we can get important feedback from them earlier in the sprint, rather than wait until we think a feature is production-ready before letting them have a look at it.

PR #6477, opened 2017-03-07
SISRP-31075 - 8.0 - Academic Summary - Transfer Credit Section

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-31075

* 2nd piece to Academic Summary
* That's a beautiful +/-, but a lot of it is just indentation once again.

### DIFF
--- a/src/assets/javascripts/angular/controllers/pages/academicSummaryController.js
+++ b/src/assets/javascripts/angular/controllers/pages/academicSummaryController.js
@@ -3,20 +3,46 @@
 var angular = require('angular');
 var _ = require('lodash');
 
-angular.module('calcentral.controllers').controller('AcademicSummaryController', function(academicsFactory, academicsService, apiService, profileFactory, userService, $route, $scope) {
+angular.module('calcentral.controllers').controller('AcademicSummaryController', function(academicsFactory, academicsService, apiService, linkService, profileFactory, userService, $route, $scope) {
   apiService.util.setTitle('Academic Summary');
+  linkService.addCurrentRouteSettings($scope);
   $scope.isAcademicSummary = $route.current.isAcademicSummary;
   $scope.academicSummary = {
     isLoading: true
   };
   $scope.expectedGradTerm = academicsService.expectedGradTerm;
 
-  var parseAcademics = function(data) {
-    angular.extend($scope, _.get(data, 'data'));
+  var showTransferCredit = function() {
+    return !!(_.get($scope, 'transferCredit.ucTransferCrseSch.unitsTotal') || _.get('transferCredit.ucTestComponent.totalTestUnits'));
+  };
+
+  var showTransferCreditPoints = function(creditItems) {
+    // If one transfer credit institution includes a 'ucbGradePoints' value, show the Points column.  Otherwise, hide it.
+    return !!_.find(creditItems, function(credit) {
+      return (_.get(credit, 'ucbGradePoints') > 0);
+    });
+  };
+
+  var parseGpaUnits = function() {
     // Testing units are lumped in with Transfer Units on the academic summary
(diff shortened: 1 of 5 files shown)


src/assets/javascripts/angular/controllers/pages/academicSummaryController.js
```diff
@@ -3,22 +3,48 @@
...
   $scope.isAcademicSummary = $route.current.isAcademicSummary;
   $scope.academicSummary = {
     isLoading: true
   };
   $scope.expectedGradTerm = academicsService.expectedGradTerm;
 
-  var parseAcademics = function(data) {
-    angular.extend($scope, _.get(data, 'data'));
+  var showTransferCredit = function() {
+    return !!(_.get($scope, 'transferCredit.ucTransferCrseSch.unitsTotal') || _.get('transferCredit.ucTestComponent.totalTestUnits'));
```
redconfetti: @davidlimcheng

I see that using the `!!` causes a count to be converted to a boolean value.
```
>> !!0
=> false
>> !!1
=> true
```

Why are you using this against the first count, but not the second?

Perhaps you meant to wrap both in parenthesis to group them?

davidlimcheng: @redconfetti They are both actually wrapped in a parenthesis - the first `(` after the `!!` wraps around both `_.get`'s.

src/assets/stylesheets/_academic_summary.scss
```diff
@@ -31,6 +39,9 @@
   }
   .cc-table {
     margin: 0 15px;
+    div {
```
redconfetti: @davidlimcheng - I recall Christian Vuerings holding a standard of avoiding selectors that use elements or IDs when possible because element and ID selectors have much stronger precedence over class selectors, and using them when it's not needed can lead to conflicts that require complex code refactoring in future updates.

You're specifying this within .cc-table. So if I run into a table in the future that has a cell containing a DIV that I want to have padding, this style might will take precedence over my class such as `.cc-table .cc-table-cell-value-container { padding-bottom: 10px; }`.

It's recommended that we use classes as much as possible because it's easy to use selectors specifying 2 to 3 nested elements to over-ride conflicting styles, without having to re-arrange and refactor the styles.

I'd also recommend checking out this article on [Object Oriented CSS](https://www.smashingmagazine.com/2011/12/an-introduction-to-object-oriented-css-oocss/) that emphasizes the re-usability of classes over other types of style definitions.

src/assets/javascripts/angular/controllers/pages/academicSummaryController.js
redconfetti: Oh derp. You're right.

src/assets/stylesheets/_academic_summary.scss
davidlimcheng: @redconfetti Good article!  I would argue that using `cc-table` as an example is something that would happen extremely rarely due to the nesting of the custom stylings under `.cc-academic-summary`.  I do agree that it is better to use classes as much as possible over element selectors in most cases.  I'll take another look at this particular CSS, and make an amendment.

davidlimcheng:
@redconfetti I replaced the general `cc-table` style with `cc-academic-summary-table` and used custom stylings in tandem with `cc-table`.  I think this is definitely a cleaner solution, and provides an easy way to edit styling without breaking the whole table as well.  

Also get rid of the `div` selector, but I did leave some of the more minor ones, such as `h3`, `h4`, and `td`.  Let me know if you don't feel comfortable with those in the code, although I don't think it should be a big problem.

redconfetti:
@davidlimcheng - Yeah. I can't expect you to fix every other bad practice that got into the code when you make a PR. Looks good, thanks.

PR #6479, opened 2017-03-08
CLC-6501, /api/oec/google/current_scope must use proper app_id in GoogleProxy call

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-6501

The call to get OEC `current_scope` (Google OAuth) had been using default app_id therefore finding no match in the db. Furthermore, OEC-mode is not compatible with the Google API call for 'self' so we ride the coattails of `can_administer_oec?` check.


paulkerschen:
Looks good.

PR #6481, opened 2017-03-08
NOJIRA Remove very obsolete "sandbox" from Capistrano script

raydavis:
This script is used in Bamboo deploy jobs. The old "sandbox" environment is long gone, but the code isn't just harmless noise. Its lack of a nil check led us to proliferate unnecessary lines in our ``.calcentral_config/server_config.yml`` files.


johncrossman:
Sandbox begone! Thx, @raydavis

PR #6483, opened 2017-03-08
SISRP-28447 Remove CalCentral Release Notes dependency on ETS website

raydavis:
https://jira.berkeley.edu/browse/SISRP-28447

Please don't merge until we have OK from Ben, @nkutub , and CC QA for this to go into GL8.

### DIFF
--- a/app/models/service_alerts/merged.rb
+++ b/app/models/service_alerts/merged.rb
@@ -5,7 +5,11 @@ module ServiceAlerts
 
     def get_feed_internal
       feed = {}
-      feed[:releaseNote] = EtsBlog::ReleaseNotes.new.get_latest
+      feed[:releaseNote] = if (latest_splash = ServiceAlerts::Alert.get_latest_splash)
+        latest_splash.to_feed
+      else
+        EtsBlog::ReleaseNotes.new.get_latest
+      end
       if (latest_alert = ServiceAlerts::Alert.get_latest)
         feed[:alert] = latest_alert.to_feed
       end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170228210646) do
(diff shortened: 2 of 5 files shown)


app/models/service_alerts/merged.rb
```diff
@@ -5,7 +5,11 @@ class Merged
 
     def get_feed_internal
       feed = {}
-      feed[:releaseNote] = EtsBlog::ReleaseNotes.new.get_latest
+      feed[:releaseNote] = if (latest_splash = ServiceAlerts::Alert.get_latest_splash)
```
johncrossman: @raydavis this really should be named `last_splash` #the-breeders  ...Okay, nevermind.

johncrossman:
Looks good, @raydavis  Waiting on T

app/models/service_alerts/merged.rb
redconfetti: > damn shame.... damn shaaame

johncrossman:
On second thought, I'll wait for "OK from Ben" as description asks.

db/schema.rb
```diff
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170228210646) do
+ActiveRecord::Schema.define(version: 20170307233607) do
```
redconfetti: @lyttam - FYI this PR includes a migration, so the release management instructions on migrations may apply with CLC v88

redconfetti:
@raydavis - Oh. So we're not getting rid of ETS Release notes altogether, just implementing a CalCentral-only release note mechanism.

db/schema.rb
raydavis: BTW, @redconfetti , the next release already has some other DB migrations.

raydavis:
@redconfetti , you would be getting rid of ETS Release notes whenever you want to. But I wanted to allow a conservative fallback for the first release at least. Does that make sense?

raydavis:
@nkutub & Janhavi have OKed this for GL8. Thank you both for checking!

PR #6486, opened 2017-03-09
SISRP-25357 - Show Degree Progress (Undergrad) on My Academics

redconfetti:
https://jira.berkeley.edu/browse/SISRP-25357

* Back-End
  * Adds DegreeProgress::MyUndergradRequirements model for Student data feed
  * Renames CampusSolutions::DegreeProgressController as MyDegreeProgressController to serve Undergraduate and Graduate feeds for Students (My Academics)
  * Moves logic for unnecessary feature flag modules into each dependent model
* Front-End
  * Updates degreeProgressFactory to serve new student endpoints for each Degree Progress card
  * Renames front-end `DegreeProgressController` as `GraduateDegreeProgressController`, and adds front-end `UndergraduateDegreeProgressController`


src/assets/templates/academics.html
```diff
@@ -23,7 +23,16 @@ <h1 class="cc-heading-page-title">My Academics</h1>
       data-ng-class="{'medium-6':(api.user.profile.roles.student || api.user.profile.roles.exStudent), 'medium-4':(!(api.user.profile.roles.student || api.user.profile.roles.exStudent))}"
       class="large-4 columns cc-column-2 cc-academics-row-margin">
       <div data-ng-include src="'academics_semesters.html'" data-ng-if="api.user.profile.hasStudentHistory && semesters.length"></div>
-      <div data-ng-include src="'widgets/academics/degree_progress_graduate.html'" data-ng-controller="DegreeProgressController"  data-ng-if="api.user.profile.features.csDegreeProgressGradStudent && (api.user.profile.roles.graduate || api.user.profile.roles.law)"></div>
+      <div
+        data-ng-if="api.user.profile.features.csDegreeProgressUgrdStudent && (api.user.profile.roles.undergrad)"
```
lyttam: @redconfetti you'll also want to check `$scope.degreeProgress.undergraduate.progresses.length` here.  The undergrad card should not show to undergrad students if they have no requirements yet.  This is different than the grad card, which shows to grad students whether there are milestones or not.

lyttam:
@redconfetti Very nice!  The way you structured everything makes sense.  Thanks for your work on this.  Only one small change requested.

app/models/degree_progress/graduate_milestones.rb
```diff
@@ -21,5 +20,11 @@ def get_feed_internal
       end
       response
     end
+
+    private
+
+    def is_feature_enabled?
```
davidlimcheng: @redconfetti Is there a reason why we're no longer using a mixin for this value?  I don't have a problem doing it this way at all, just curious why.

src/assets/javascripts/angular/controllers/widgets/academics/undergraduateDegreeProgressController.js
```diff
@@ -0,0 +1,25 @@
...
+    undergraduate: {
+      isLoading: true
+    }
+  };
+
+  var loadDegreeProgress = function() {
+    degreeProgressFactory.getUndergraduateRequirements()
+      .then(function(data) {
+        $scope.degreeProgress.undergraduate.progresses = _.get(data, 'data.feed.degreeProgress.progresses');
+        $scope.degreeProgress.undergraduate.errored = _.get(data, 'errored');
```
davidlimcheng: @redconfetti Looking at `CampusSolutions::Proxy` it looks like `errored` is on the same level as `feed`:

```
          {
            statusCode: 400,
            errored: true,
            feed: feed
          }
```
So, I believe it will have to unwrap the data similarly:
`$scope.degreeProgress.undergraduate.errored = _.get(data, 'data.errored');`

src/assets/templates/academics.html
lyttam: Actually, @redconfetti, you can leave this the way you have it - the logic for hiding/showing the card is going to be updated in https://jira.berkeley.edu/browse/SISRP-30643.

redconfetti: @lyttam - You verbally reminded me that this requirement is something that will be added in another sub-task of this feature development work.

app/models/degree_progress/graduate_milestones.rb
redconfetti: @davidlimcheng - The mix-ins were created by @lyttam as a matter of convention that she saw in other similar code. I noticed that there was a one-to-one relationship between the models and the modules. Since no other models were using the modules, I thought it would be best to clean up by merging the logic into the model itself.

If another model relies on the same feature flag code in the future, then we should move that logic into a shared module then.

src/assets/javascripts/angular/controllers/widgets/academics/undergraduateDegreeProgressController.js
redconfetti: @davidlimcheng - Great catch.

app/models/degree_progress/graduate_milestones.rb
davidlimcheng: @redconfetti Makes complete sense, I'll keep this in mind for future feature flags.

PR #6491, opened 2017-03-10
SISRP-30877 - Shows inactive committee members on completed committees

lyttam:
Parent JIRA:  https://jira.berkeley.edu/browse/SISRP-30713

https://jira.berkeley.edu/browse/SISRP-SISRP-30877

If the committee is active but some of its members are inactive, we assume they're no longer on the committee, and we don't show them.  If the committee is completed, we show all the members who were on it at some point.


davidlimcheng:
LGTM

PR #6494, opened 2017-03-10
SISRP-25357 - Surface formatted last report date and adds 'In Progress' status

redconfetti:
Parent JIRA - https://jira.berkeley.edu/browse/SISRP-25357

https://jira.berkeley.edu/browse/SISRP-30966
https://jira.berkeley.edu/browse/SISRP-30634

### DIFF
--- a/app/models/berkeley/degree_progress_undergrad.rb
+++ b/app/models/berkeley/degree_progress_undergrad.rb
@@ -2,8 +2,15 @@ module Berkeley
   class DegreeProgressUndergrad
     include ClassLogger
 
-    def self.get_status(status_code)
-      statuses[status_code.strip.upcase] unless status_code.blank?
+    def self.get_status(status_code, in_progress_value)
+      return nil if status_code.blank?
+      uniform_status_code = status_code.strip.upcase
+      in_progress = uniform_status_code == 'COMP' && in_progress_boolean(in_progress_value)
+      in_progress ? 'In Progress' : statuses[uniform_status_code]
+    end
+
+    def self.in_progress_boolean(in_progress_value)
+      in_progress_value == 'Y'
     end
 
     def self.requirements_whitelist
--- a/app/models/degree_progress/requirements_module.rb
+++ b/app/models/degree_progress/requirements_module.rb
@@ -1,5 +1,6 @@
 module DegreeProgress
   module RequirementsModule
+    include DatedFeed
 
     def process(response)
       degree_progress = response.try(:[], :feed).try(:[], :ucAaProgress)
@@ -16,6 +17,7 @@ module DegreeProgress
             next
           end
           result.push(progress).last.tap do |prog|
+            prog[:reportDate] = format_report_date prog.delete(:rptDate)
             prog[:requirements] = requirements
           end
         end
@@ -32,6 +34,12 @@ module DegreeProgress
       sort result
     end
 
+    def format_report_date(report_date_unformatted)
+      return nil if report_date_unformatted.blank?
+      report_date_object = strptime_in_time_zone(report_date_unformatted, '%Y-%m-%d')
+      format_date(report_date_object, '%b %e, %Y').try(:[], :dateString).to_s.squish
+    end
+
     def should_include(requirement)
       Berkeley::DegreeProgressUndergrad.requirements_whitelist.include?(Integer(requirement[:code], 10)) unless requirement[:code].blank?
     rescue ArgumentError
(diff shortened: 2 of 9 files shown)


src/assets/stylesheets/_degree_progress.scss
```diff
@@ -1,3 +1,11 @@
+.cc-undergrad-degree-progress-card {
+  .cc-degree-progress-report-date {
+    color: $cc-color-lighter-medium-grey;
+    font-size: 13px;
+    margin-bottom: 10px;
+  }
+}
+
```
redconfetti: @lyttam - I don't really see the two cards sharing styles. I decided to go the route of defining a separate namespace for the undergraduate card.

src/assets/templates/widgets/academics/degree_progress_undergrad.html
```diff
@@ -1,31 +1,36 @@
...
-                  <i class="fa" data-ng-class="{'cc-icon fa fa-check cc-icon-green': requirement.status === 'Satisfied', 'cc-icon fa fa-exclamation-triangle cc-icon-gold': requirement.status === 'Not Satisfied'}"></i>
-                  <span data-ng-bind="requirement.status"></span>
-                </td>
-              </tr>
-            </tbody>
-          </table>
-        </div>
-      </li>
-    </ul>
+    <div data-ng-if="!degreeProgress.undergraduate.errored && degreeProgress.undergraduate.progresses">
```
redconfetti: I made sure to retain your style changes, as well as the new 'University Requirement' header title. This simply indented the code and wrapped it in a DIV with the data-ng-if. It also adds the 'Degree Progress as of ...' with report date.

app/models/degree_progress/requirements_module.rb
```diff
@@ -32,6 +34,26 @@ def massage_requirements(progress)
       sort result
     end
 
+    def format_report_date(report_date_unformatted)
+      return '' if report_date_unformatted.blank?
+      format_report_date_object strptime_in_time_zone(report_date_unformatted, '%Y-%m-%d')
+    end
+
+    # Needed because '%e' (blank padded day of month) does not strip returned single digit day of month
```
redconfetti: When I was running the RSpec test it was returning `March  3, 2017` instead of `March 3, 2017`. I know the front-end HTML probably would have displayed without the extra space, but I still didn't like it. I decided to loop through each item (month, day, year) and use #strip to ensure that the value has no spaces before or after.

app/models/berkeley/degree_progress_undergrad.rb
```diff
@@ -2,8 +2,19 @@ module Berkeley
...
     include ClassLogger
 
-    def self.get_status(status_code)
-      statuses[status_code.strip.upcase] unless status_code.blank?
+    def self.get_status(status_code, in_progress_value)
+      status = nil
+      if status_code.present?
+        status = statuses[status_code.strip.upcase]
+        if (in_progress_boolean(in_progress_value))
+          status = 'In Progress'
```
lyttam: @redconfetti My understanding of the requirements is that we only want to show 'In Progress' if `status_code = 'COMP'`.  If `in_progress_value = 'Y'` but `status_code = 'FAIL'`, we should treat that as a fail.  So you just need to add checking for that.  See SISRP-30553.

app/models/degree_progress/requirements_module.rb
paulkerschen: @redconfetti Do you know `squish`? http://www.rubydoc.info/docs/rails/4.0.0/String:squish

app/models/degree_progress/requirements_module.rb
```diff
@@ -32,6 +34,26 @@ def massage_requirements(progress)
...
+    def format_report_date_object(report_date_object)
+      format_hash = {
+        year: '%Y',
+        month: '%b',
+        day: '%e'
+      }
```
lyttam: @redconfetti  Why not use `'%b %d, %Y'` format?  That's what Grad Degree Progress and Committees are using.

app/models/degree_progress/requirements_module.rb
redconfetti: image

app/models/degree_progress/requirements_module.rb
redconfetti: @lyttam - '%e' is the day without a leading zero. It would read better for students without the leading zero in that format of date.

app/models/berkeley/degree_progress_undergrad.rb
redconfetti: Great catch @lyttam . I need to include that in the test coverage too.

lyttam:
Looks good!

PR #6497, opened 2017-03-13
SISRP-30636 - Adds APR link to UGRD Degree Progress card for advisor only

lyttam:
Parent:  https://jira.berkeley.edu/browse/SISRP-30580

Advisors will see a message with a link to run the APR.  Students will see a message with no link. (https://jira.berkeley.edu/browse/SISRP-30636)

I also wrapped the report run date and label in `ngIf` because it seems the API is not sending that date.  I will investigate that further, but at least wanted to make sure we don't ever display the label with the date missing.

### DIFF
--- a/app/models/degree_progress/undergrad_requirements.rb
+++ b/app/models/degree_progress/undergrad_requirements.rb
@@ -7,6 +7,10 @@ module DegreeProgress
     include Cache::UserCacheExpiry
     include RequirementsModule
 
+    LINKS_CONFIG = [
+      { feed_key: :academic_progress_report, cs_link_key: 'UC_CX_APR_RPT_STDNT' }
+    ]
+
     def get_feed_internal
       return {} unless is_feature_enabled?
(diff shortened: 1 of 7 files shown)


app/models/degree_progress/undergrad_requirements.rb
```diff
@@ -15,13 +19,31 @@ def get_feed_internal
...
           degree_progress: process(response),
+          links: get_links
         })
       end
       response
     end
 
     private
 
+    def get_links
```
redconfetti: @lyttam - I don't want to block you here with a refactoring task that goes beyond the scope of this task, but I'm noticing that we're duplicating code by implementing pattern of retrieving a collection of links and merging it into a specific data feed. I've created [SISRP-31252](https://jira.berkeley.edu/browse/SISRP-31252) as a suggested future refactoring task.

lyttam: @redconfetti I agree that we should refactor this link-fetching behavior a shared module.  I didn't want to do it for this release since we're so close to code freeze, but maybe in 8.1.

PR #6500, opened 2017-03-14
SISRP-30003 Shift Canvas support from CalCentral to ETS-Junction servers

raydavis:
https://jira.berkeley.edu/browse/SISRP-30003

This file is used to set up JS and CSS customization and front-end URLs.


johncrossman:
+1  Thx, @raydavis

PR #6501, opened 2017-03-14
SISRP-30618 - Authorization errors in CC Production logs related to committees

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-30618

* Full disclosure, was not able to test this locally due to lack of test data, but we should have test cases in time for QA. 
* Removing the template from `academics.html` with a conditional should stop `AcademicsHigherDegreeCommitteesController` from instantiating, preventing the call to `api/my/committees`.
* If a delegate did manually hit the `api/my/comittees` endpoint, the `Pundit::NotAuthorizedError` will still be raised in the logs, notifying us of anything fishy.  For this reason, I did not add anything to the existing restrictions on `my_committees_controller.rb`.
* Changing `#empty?` to `#blank?` ensures no no error is raised if CS sends us `nil` for that particular field.

### DIFF
--- a/app/models/my_committees/committees_module.rb
+++ b/app/models/my_committees/committees_module.rb
@@ -98,7 +98,7 @@ module MyCommittees::CommitteesModule
   end
 
   def determine_qualifying_exam_status_message(cs_committee)
-    if cs_committee[:studentMilestoneAttempts].empty?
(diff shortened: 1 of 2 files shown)


nkutub:
LGTM @davidlimcheng

PR #6506, opened 2017-03-15
SISRP-31239 - Remove ScheduleBuilder Link from CalCentral Academic Planning

redconfetti:
https://jira.berkeley.edu/browse/SISRP-31239


lyttam:
A+

PR #6509, opened 2017-03-16
SISRP-25357 - Resolve tests failing in TestExt env

redconfetti:
https://jira.berkeley.edu/browse/SISRP-25357

### DIFF
--- a/spec/models/degree_progress/my_undergrad_requirements_spec.rb
+++ b/spec/models/degree_progress/my_undergrad_requirements_spec.rb
@@ -2,6 +2,12 @@ describe DegreeProgress::MyUndergradRequirements do
 
   let(:model) { described_class.new(user_id) }
   let(:user_id) { '12345' }
+  let(:emplid) { '12345678' }
+  before do
+    proxy_class = CampusSolutions::DegreeProgress::UndergradRequirements
+    fake_proxy = proxy_class.new(user_id: user_id, fake: true)
+    allow(proxy_class).to receive(:new).and_return fake_proxy
+  end
 
   describe '#get_feed_internal' do
     let(:flag) { :cs_degree_progress_ugrd_student }
@@ -11,6 +17,7 @@ describe DegreeProgress::MyUndergradRequirements do
     it_behaves_like 'a proxy that returns undergraduate milestone data'
 
     it 'does not include the Academic Progress Report link in the response' do
+      allow(Settings.features).to receive(flag).and_return(true)
       expect(subject[:feed][:links]).not_to be
     end
   end


davidlimcheng:
Nice @redconfetti, looks like you figured it out.  Can you talk a bit about why these testext tests were failing, so that we can avoid this situation in the future?

lyttam:
@redconfetti Looks like the change is to stub the CS proxy called by our `described_class` to force it to return fake data.  Is this necessary because the testext environment uses real data by default?  Just trying to wrap my head around this too.  CC @davidlimcheng

redconfetti:
@davidlimcheng - Thanks for asking. 

Yes. In the test environment the proxies are configured with `fake: true`, as apparent in the many proxies configured in [config/settings/test.yml](https://github.com/ets-berkeley-edu/calcentral/blob/997b28c4/config/settings/test.yml). Within [config/settings/testext.yml](https://github.com/ets-berkeley-edu/calcentral/blob/997b28c4/config/settings/testext.yml) this isn't the case, because the Bamboo server that is located behind campus firewalls, and is whitelisted to access campus official data sources, is supposed to run ALL the tests, including the `testext` tests. It's supposed to be able to actually connect to actual external data sources and verify that the contract between CalCentral and those data sources is not broken.

This is made clear by the [Rspec configuration](https://github.com/ets-berkeley-edu/calcentral/blob/997b28c4/spec/spec_helper.rb#L104,L108) which excludes tests that include the `testext: true` argument passed to the `it` method in Rspec tests.

I want to stay away from configuring this proxy (DegreeProgress::MyUndergradRequirements) in config/settings/testext.yml as `fake: true`, just so that the tests pass, because this will then force the test to always use fixture data, confusing developers and establishing a bad practice.

At the same time I didn't want to duplicate and mock the entire data feed that DegreeProgress::MyUndergradRequirements returns in this Rspec file, when the same type of feed exists in a fixture. That's why I wanted to mock this model to run in fake mode for this specific model. 

Really, it is CampusSolutions::DegreeProgress::UndergradRequirements that is the proxy that should optionally perform a testext test to verify the contract. DegreeProgress::MyUndergradRequirements is a cacheing model with it's own logic. So I'm fine with it mocking the data it receives from CampusSolutions::DegreeProgress::UndergradRequirements during these tests.

We would probably be writing `testext` tests for Campus Solutions APIs if we had a Campus Solutions environment that we were guaranteed would ALWAYS serve data for specific UIDs or EMPLIDs. That is not the case, so we are not even attempting to do this.

If that was the case, we would probably verify the behavior of the coded logic in our Rspec tests for proxies using mocked data, and in a separate context defining the testext tests without the mocking of fixture data being implemented.

## Example

```
describe CampusSolutions::DegreeProgress::UndergradRequirements do
  context 'mocked scenarios' do
    let(:user_id) { '12345' }
    let(:proxy) { described_class.new(fake: true, user_id: user_id) }

    describe '#get' do
      subject { proxy.get }
      it_behaves_like 'a simple proxy that returns errors'
      it_behaves_like 'a proxy that got data successfully'
    end
  end

  context 'live api contract validation' do
    let(:user_id) { '61889' } # Oski Bear
    let(:proxy) { described_class.new(fake: false, user_id: user_id) }

    it 'returns requirements', :testext => true do
      result = proxy.get
      expect(result[:feed][:requirements]).to be
    end
  end
end
```

/cc @lyttam @nkutub

redconfetti:
I've documented the gist of the above comment in [Confluence](https://confluence.ets.berkeley.edu/confluence/display/MYB/Testing#Testing-TestvsTestextConflicts)

davidlimcheng:
Makes total and complete sense @redconfetti, thanks for the in-depth explanation and for uncovering the black box that is the testext environment.  @lyttam, thumbs up to merge from me when ready.

PR #6510, opened 2017-03-16
SISRP-31364 - Allows ex-student to see GRAD Degree Progress

lyttam:
Parent:  https://jira.berkeley.edu/browse/SISRP-25708

If user is an ex-student, we don't know whether they were grad or undergrad.  We only want to show them the card if they were grad.  

To get around this problem, we will only show the card to ex-students IF the API returns data for the card (effectively indicating that the person was a grad student). (https://jira.berkeley.edu/browse/SISRP-31364)


davidlimcheng:
LGTM, @lyttam

PR #6514, opened 2017-03-17
NOJIRA Fix Capistrano deploy script

raydavis:
The shell scripts assume they will be started from the top "calcentral" directory & should write output to the "./log/" directory.
Capistrano instead started them from "calcentral/scripts", leading to errors for all logging.
This is an old bug but the Bamboo agent is the only one who would have seen it in action.


johncrossman:
Looks right to me. Thanks, @raydavis

PR #6518, opened 2017-03-19
SISRP-15392 - Delegated Access: ‘not found’ redirects for inaccessible pages need improvement

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-15392

Adds "Access Denied" pages for inaccessible pages as a delegate, with redirect to either Academics, Finances, or Campus Links (whichever is available).

Also makes Campus Links available to delegates by way of making Academic Status (and the dependent Roles) available.  The caveat is that delegates must be authorized for enrollments, so delegates who are only authorized for financials will have a broken experience still.

To fix this, we need to decouple Roles from Academic Status.  That requires a larger code change, so I think we should put in a fix in a future release instead of cramming it in to 8.0.

QA PR to come.

### DIFF
--- a/src/assets/templates/delegate_access_denied.html
+++ b/src/assets/templates/delegate_access_denied.html
@@ -1,0 +1,24 @@
+<div class="cc-page-status">
+
+  <h1 class="cc-visuallyhidden">Access denied</h1>
+
+
+  <div class="cc-widget cc-page-status-container">
+    <div class="cc-widget-title">
+      <h2>You are not authorized to access this page.</h2>
+    </div>
+
+    <div class="cc-widget-padding">
+      <div class="cc-left cc-page-status-container-text">
+        <h3>Access Denied</h3>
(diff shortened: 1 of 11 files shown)


davidlimcheng:
Please hold off on merging - encountered some bugs after I had opened this.

nkutub:
@davidlimcheng , looks good to me

src/assets/templates/delegate_access_denied.html
```diff
@@ -0,0 +1,24 @@
...
+
+
+  <div class="cc-widget cc-page-status-container">
+    <div class="cc-widget-title">
+      <h2>You are not authorized to access this page.</h2>
+    </div>
+
+    <div class="cc-widget-padding">
+      <div class="cc-left cc-page-status-container-text">
+        <h3>Access Denied</h3>
```
redconfetti: image

PR #6519, opened 2017-03-20
SISRP-31399 - Delegates Seeing Duplicate Classes in Semesters

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-31399

* If you're not a delegate, your `semesters` feed will be based off of `EdoOracle::UserCourses::All#get_all_campus_courses`, which includes `EdoOracle::UserCourses::Base#merge_detailed_section_data`.  This function uses `#uniq!` to get rid of duplicate section data. 
* If you are a delegate, your `semesters` feed will be based off of `EdoOracle::UserCourses::All#get_enrollments_summary`, which does not have the `#uniq!` function called.
* This PR adds the call to `#uniq!` to `#get_enrollments_summary`.  

* Why did we only start seeing this bug now?  Either we missed it in multiple QA attempts on the delegate experience, or, more likely, the EDODB query started returning duplicate sections for each class.  Maybe this has something to do with the refactoring/optimization going on?  We might have to expect to see some weirdness for a little bit while we iron out details for the query optimizations.  cc @nkutub @lyttam @redconfetti 

* QA PR to come.

### DIFF
--- a/app/models/edo_oracle/user_courses/all.rb
+++ b/app/models/edo_oracle/user_courses/all.rb
@@ -29,7 +29,9 @@ module EdoOracle
           campus_classes = {}
           merge_enrollments campus_classes
           sort_courses campus_classes
-          Hash[campus_classes.sort.reverse]
+          campus_classes = Hash[campus_classes.sort.reverse]
+          remove_duplicate_sections(campus_classes)
+          campus_classes
         end
       end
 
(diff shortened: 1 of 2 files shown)


nkutub:
@davidlimcheng, Greg Snow has mentioned the issues related to duplicates in the views and has addressed them with the new SOR views to come in 8.1. We should validate these dups once Greg and DBAs have migrated the new views in EDODB for QA

PR #6526, opened 2017-03-21
SISRP-28770 - Update references to  success and failure methods to meet requirements of AngularJS 1.6.0 API

redconfetti:
https://jira.berkeley.edu/browse/SISRP-28770

* Updates references to success and failure methods to use then(), meeting requirements of AngularJS 1.6.0 API
* Updates callbacks to use 'response', rather than 'data' for callback argument reference
* Configure $qProvider to suppress errors with unhandled promise rejections

### DIFF
--- a/src/assets/javascripts/angular/configuration/qConfiguration.js
+++ b/src/assets/javascripts/angular/configuration/qConfiguration.js
@@ -1,0 +1,10 @@
+'use strict';
+
+var angular = require('angular');
+
+/**
+ * Configure $qProvider to suppress errors on unhandled promise rejections
+ */
+angular.module('calcentral.config').config(function($qProvider) {
+  $qProvider.errorOnUnhandledRejections(false);
+});
--- a/src/assets/javascripts/angular/controllers/pages/academicsController.js
+++ b/src/assets/javascripts/angular/controllers/pages/academicsController.js
@@ -167,34 +167,35 @@ angular.module('calcentral.controllers').controller('AcademicsController', funct
   };
 
   var loadNumberOfHolds = function() {
-    return academicStatusFactory.getHolds()
-      .then(function(data) {
-        $scope.numberOfHolds = _.get(data, 'holds.length');
-      });
+    return academicStatusFactory.getHolds().then(
+      function(parsedHolds) {
+        $scope.numberOfHolds = _.get(parsedHolds, 'holds.length');
+      }
+    );
   };
 
-  var loadRegistrations = function(data) {
-    var registrations = _.get(data, 'registrations');
+  var loadRegistrations = function(response) {
+    var registrations = _.get(response, 'data.registrations');
     $scope.hasRegStatus = !_.isEmpty(registrations);
   };
 
-  var parseAcademics = function(data) {
-    angular.extend($scope, data);
+  var parseAcademics = function(response) {
+    angular.extend($scope, _.get(response, 'data'));
 
     $scope.isLSStudent = academicsService.isLSStudent($scope.collegeAndLevel);
     $scope.isUndergraduate = _.includes(_.get($scope.collegeAndLevel, 'careers'), 'Undergraduate');
-    $scope.hasTeachingClasses = academicsService.hasTeachingClasses(data.teachingSemesters);
+    $scope.hasTeachingClasses = academicsService.hasTeachingClasses(_.get(response, 'data.teachingSemesters'));
     $scope.canViewFinalExamSchedule = $scope.api.user.profile.roles.student && !$scope.api.user.profile.delegateActingAsUid && !$scope.academicStatus.roles.summerVisitor;
 
     // Get selected semester from URL params and extract data from semesters array
     var semesterSlug = ($routeParams.semesterSlug || $routeParams.teachingSemesterSlug);
     if (semesterSlug) {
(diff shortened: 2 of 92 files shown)


src/assets/javascripts/angular/controllers/pages/academicsController.js
```diff
@@ -167,34 +167,35 @@ angular.module('calcentral.controllers').controller('AcademicsController', funct
...
+      function(parsedHolds) {
+        $scope.numberOfHolds = _.get(parsedHolds, 'holds.length');
+      }
+    );
   };
 
-  var loadRegistrations = function(data) {
-    var registrations = _.get(data, 'registrations');
+  var loadRegistrations = function(response) {
+    var registrations = _.get(response.data, 'registrations');
```
davidlimcheng: @redconfetti In this and similar cases, can we change this to 

`var registrations = _.get(response, 'data.registrations');`

instead?  Reason being, `response.data` will still throw an error if `response` doesn't have a `data` property, which we know happens from time to time due to API timeouts.  Of course, this will probably throw new errors if we don't handle the case of `registrations === null`, but that's a much more useful error that will lead us to better fixes.

redconfetti: @davidlimcheng - Good catch, I missed  this.

src/assets/javascripts/angular/configuration/qConfiguration.js
```diff
@@ -0,0 +1,10 @@
+'use strict';
+
+var angular = require('angular');
+
+/**
+ * Configure $qProvider to suppress errors on unhandled promise rejections
+ */
+angular.module('calcentral.config').config(function($qProvider) {
+  $qProvider.errorOnUnhandledRejections(false);
```
davidlimcheng: @redconfetti Interesting!  I didn't know this could be configured, very cool.  I am questioning its need though - were we getting errors on certain controllers?  I would imagine any errors we were getting from unhandled rejections were useful in that it was highlighting areas of weak code.  But it could be I'm misunderstanding the use cases.

Can you talk a bit about when these errors popped up, and why we would want to ignore them?

redconfetti: @davidlimcheng - We need applications to alert us of errors in exceptional situations, meaning that something has occurred that was not expected. With the case of the [sirController](https://github.com/ets-berkeley-edu/calcentral/blob/91f375af/src/assets/javascripts/angular/controllers/widgets/sir/sirController.js#L117), it is coded to reject (and thus halt execution of the chain of functions) in two situations. These were not HTTP requests that resulted in 400 or 500 errors (thus exceptions), but these are expected from what I can see.

These are situations like the user doesn't have any checklist items, so it doesn't show the card.

With Angular 1.5.8 there were no errors raised in this situation. Angular was designed to not throw an error if you choose to not handle a rejection. With Angular 1.6.0+, it throws the error, which isn't good. I talked to Nader and he recommended that we suppress these errors to obtain the same framework specifications that our code was designed to run within.

davidlimcheng: Makes complete sense, thanks for the explanation @redconfetti

davidlimcheng:
This all looks great to me, @redconfetti.  Great job with this.

One thing I'd like to mention is that in addition to an error-handling function injected right after the success function, we can also use `.catch()` to handle errors.  IMO, `.catch()` is a bit more clean because you can chain multiple `.then()` functions, and stick a `.catch()` at the end of a long chain, with the catch-function being activated whenever a Javascript error is surfaced on any part of the chain.

It doesn't look like there are any immediate areas where we would benefit by changing from a `success, error` pattern to a `.catch()` pattern, though.

I'll leave this open for a bit so others can take a look at it, but will merge it by EOD or tomorrow morning.

redconfetti:
Thanks for that note @davidlimcheng. I structured most of the updates so that one could easily add a comma after the first function, and then define an error handling function right below it.

I like that the `catch()` function call exists so that we can define a single default handler for displaying error messages.

src/assets/javascripts/angular/controllers/pages/canvasCreateCourseSiteController.js
```diff
@@ -58,18 +58,20 @@ angular.module('calcentral.controllers').controller('CanvasCreateCourseSiteContr
   var timeoutPromise;
   var jobStatusLoader = function() {
     timeoutPromise = $timeout(function() {
```
nkutub: @redconfetti , I'm little confused here, you are setting this var but not using it

redconfetti: @nkutub - The `timeoutPromise` variable referenced on line 60 is the same one defined in the outer scope on line 58. I wanted to make sure that the timeout is accessible to `statusProcessor()` on [line 39](https://github.com/redconfetti/calcentral/blob/2d3380a4/src/assets/javascripts/angular/controllers/pages/canvasCreateCourseSiteController.js#L39).

I didn't add this in the PR though, so I'm not sure why you're referencing it here.

nkutub:
@redconfetti wow, that hurt my head, big PR!. I did my scan and looks good, with @davidlimcheng and me doing our review, I will merge it. We will have to be very good at testing this sprint since it touches a lot of code.

PR #6531, opened 2017-03-22
SISRP-31555 Clarify meaning of "Splash" in Service Alert / Release Note

raydavis:
https://jira.berkeley.edu/browse/SISRP-31555

As SIS QA pointed out, the original label is too ambiguous.


lyttam:
Makes sense to me!

PR #6540, opened 2017-03-23
SISRP-30643 - GL 8.1 - Show card with message when there is no data

redconfetti:
https://jira.berkeley.edu/browse/SISRP-30643


davidlimcheng:
LGTM

PR #6541, opened 2017-03-23
SISRP-31292 - Terms and Conditions checklist should not have the Terms and Conditions button active once a student agrees

nkutub:
https://jira.berkeley.edu/browse/SISRP-31292

* disabled buttons on approved


davidlimcheng:
Looks good @nkutub

PR #6547, opened 2017-03-28
SISRP-31609 - Faculty cannot view and communicate via email with visiting summer students - Enrollment page

nkutub:
https://jira.berkeley.edu/browse/SISRP-31609

* joined the enrollments get_roster EDODB query to PERSON_EMAILV00_VW
* updated enrollment/roster email with EDODB email if email obtained from CALDAP was blank
* added feature flag allow_alt_email_addr_for_enrollments to mitigate any performance issues during ER

### DIFF
--- a/app/models/edo_oracle/queries.rb
+++ b/app/models/edo_oracle/queries.rb
@@ -39,6 +39,12 @@ module EdoOracle
         crs."status-code" = 'ACTIVE')
     SQL
 
+    JOIN_ROSTER_TO_EMAIL = <<-SQL
+       LEFT OUTER JOIN SISEDO.PERSON_EMAILV00_VW email ON (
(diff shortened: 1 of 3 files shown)


app/models/edo_oracle/queries.rb
```diff
@@ -319,6 +325,11 @@ def self.get_enrolled_students(section_id, term_id)
 
     # Extended version of #get_enrolled_students used for rosters
     def self.get_rosters(ccns, term_id)
+      if Settings.features.allow_other_emails_for_enrollments
```
lyttam: @nkutub Just a personal pref, but this would make more sense to me if it were called `allow_alt_email_addr_for_enrollments`.

PR #6550, opened 2017-03-28
SISRP-31605 - Show filing date on Committees card only when finishing milestone is complete

lyttam:
https://jira.berkeley.edu/browse/SISRP-31605

Filing date indicates when the "finishing" milestone for the committee has been completed.  Adds a check to make sure we are displaying the actual complete date, and not just the date of an unsuccessful attempt on the finishing milestone.

Slated for emergency release.


app/models/my_committees/committees_module.rb
```diff
@@ -148,6 +148,10 @@ def get_cs_committee_role_key (role_name)
     end
   end
 
+  def is_active?(cs_committee)
+    cs_committee.try(:[], :committeeFinishingMilestoneComplete) != 'Y'
```
redconfetti: @lyttam - Should this be `!==` ? I think we want to use the strict operation when possible.

lyttam: @redconfetti I don't think `!==` is a valid Ruby operator... `===` is however.

redconfetti: @lyttam , you're right. this is Ruby, not Javascript. Derp!

PR #6553, opened 2017-03-30
SISRP-31605 - Extends 'advanced date' logic to look at milestone attempts

lyttam:
https://jira.berkeley.edu/browse/SISRP-31605

Trying to cover all the tricky cases with a committee, it's associated milestones, and the attempts on those milestones.

When the student made an unsuccessful attempt on an advancement milestone, the API doesn't give us the "advanced date".  In this case we have get that date from the attempt itself.


nkutub:
LGTM @lyttam

PR #6561, opened 2017-04-05
SISRP-31636 - Hard-codes the committee name

lyttam:
https://jira.berkeley.edu/browse/SISRP-31636

The field we were relying upon for the committee name is being repurposed for eForm-related data.  CalCentral needs to be smart enough to map the committee type code to a user-friendly name.


app/models/my_committees/committees_module.rb
```diff
@@ -34,9 +47,17 @@ def parse_cs_committee (cs_committee)
     committee
   end
 
+  def translate_committee_type(cs_committee)
+    COMMITTEE_TYPES[cs_committee[:committeeType].try(:intern)].try(:[], :label)
+  end
+
+  def qualifying_exam?(cs_committee)
+    cs_committee[:committeeType].to_s.strip.upcase === COMMITTEE_TYPES[:QE][:code]
```
nkutub: @lyttam , did you forget the try here? on line 51 you used a try for the same attribute  :committeeType

lyttam: @nkutub the `try` on line 51 is for `intern` and `label`.  On lines 51 and 55, I'm assuming the data from CS will include the COMMITTEE_TYPE element.  Do you think I should add a `try` for that as well?  I would hope that the API contract ensures each data element will be present in the response, even if it's null.

nkutub: @lyttam , sorry for some reason I thought that nil.to_s would error but it does not. It looks good to me

app/models/my_committees/committees_module.rb
```diff
@@ -34,9 +47,17 @@ def parse_cs_committee (cs_committee)
     committee
   end
 
+  def translate_committee_type(cs_committee)
+    COMMITTEE_TYPES[cs_committee[:committeeType].try(:intern)].try(:[], :label)
```
nkutub: @lyttam, ok I don't understand this line :(. Also, I think you are missing ] at end of line. I don't see sample data that shows that you will have { committeeType: { intern: { label: 'value' } } }

app/models/my_committees/committees_module.rb
```diff
@@ -34,9 +47,18 @@ def parse_cs_committee (cs_committee)
     committee
   end
 
+  def translate_committee_type(cs_committee)
+    committee_type_code = cs_committee[:committeeType].try(:intern)
+    COMMITTEE_TYPES[committee_type_code].try(:[], :label)
```
lyttam: @nkutub I separated this into two lines to make it more readable.

PR #6562, opened 2017-04-06
SISRP-31889 - Grading: Prod: Spring Final Grades Not Appearing for Students in CalCentral (SNow 445421)

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-31889

* I want to reiterate as said in the ticket, that this is just a bandage fix so that students can see their grades in the current term.
* [SISRP-32020](https://jira.berkeley.edu/browse/SISRP-32020) will be the larger work of reassessing how `semesters.rb` parses data, and making related changes to the front-end templates.
* QA PR to come since this will be part of Tuesday's Emergency Release.


lyttam:
Looks good.

PR #6567, opened 2017-04-07
SISRP-31628 - Replicates committees status icon logic on Degree Progress

lyttam:
https://jira.berkeley.edu/browse/SISRP-31628

For Qualifying Exam Results milestone only:  yellow icon if one failed attempt; red icon if more than one failed attempt.


davidlimcheng:
easy peasy lemon squeezy

PR #6568, opened 2017-04-07
SISRP-31579 - updates Proposed Exam Date logic on Degree Progress

lyttam:
https://jira.berkeley.edu/browse/SISRP-31579

Degree Progress for graduate students:
We only want to show the proposed exam date between the time when the student is approved to take the qualifying exam and when they first attempt the exam.


nkutub:
LGTM @lyttam

PR #6569, opened 2017-04-11
SISRP-31911 - Refactors Link API calls into a mixin

lyttam:
https://jira.berkeley.edu/browse/SISRP-31911

Original goal was just to change the log level to DEBUG on a couple proxies that were treating missing links as ERRORs, but it felt right to take a step further to promote DRYness.  

Going forward:  for any proxies needing to call the Link API, simply include `LinkFetcher`.

### DIFF
--- a/lib/link_fetcher.rb
+++ b/lib/link_fetcher.rb
@@ -1,0 +1,11 @@
+module LinkFetcher
+  include ClassLogger
+
+  def fetch_link(link_key, placeholders = {})
+    if (link_feed = CampusSolutions::Link.new.get_url link_key, placeholders)
+      link = link_feed.try(:fetch, :link)
+    end
+    logger.debug "Could not parse CS link response for id #{link_key}, uid = #{@uid}" unless link
+    link
+  end
+end
--- a/spec/lib/link_fetcher_spec.rb
+++ b/spec/lib/link_fetcher_spec.rb
@@ -1,0 +1,71 @@
+describe LinkFetcher do
+
+  describe '#fetch_link' do
+    before do
+      allow(proxy_class).to receive(:new).and_return mock_proxy
+      link_fetcher.extend(LinkFetcher)
+      allow(link_fetcher).to receive(:logger).and_return ClassLogger::LogWrapper.new('Object')
+    end
+    let(:link_fetcher) { Object.new }
+    let(:url) { 'fake url' }
+    let(:proxy_class) { CampusSolutions::Link }
+    let(:mock_proxy) { proxy_class.new(fake: true) }
+
+    context 'when proxy returns malformed response' do
+      before do
+        allow(mock_proxy).to receive(:get_url).and_return('bad response')
+      end
+      subject do
+        link_fetcher.fetch_link(link_key)
+      end
+      let(:link_key) { nil }
+
+      it 'calls the proxy with appropriate arguments and logs an error' do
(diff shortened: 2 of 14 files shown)


lib/link_fetcher.rb
```diff
@@ -0,0 +1,10 @@
+module LinkFetcher
+
+  def fetch_link(link_key, placeholders = {})
+    if (link_feed = CampusSolutions::Link.new.get_url link_key, placeholders)
+      link = link_feed.try(:fetch, :link)
+    end
+    Rails.logger.debug "Could not retrieve CS link #{link_key} for #{self.class.name} feed, uid = #{@uid}" unless link
```
davidlimcheng: @lyttam, I see in the code that we usually use the `ClassLogger` mixin, which automatically decorates the message by prepending the class name to the message.  Is that something we can use here, as well?

davidlimcheng:
Awesome!  Good stuff @lyttam, just one question that I had for you.  I also suspect that this will cause merge conflicts with my Summer Grading PR, so I'd like to merge this one first if possible.

lyttam:
@davidlimcheng Amended to use `ClassLogger`, also improved the error message.

spec/lib/link_fetcher_spec.rb
```diff
@@ -0,0 +1,71 @@
+describe LinkFetcher do
+
+  describe '#fetch_link' do
+    before do
+      allow(proxy_class).to receive(:new).and_return mock_proxy
+      link_fetcher.extend(LinkFetcher)
+      allow(link_fetcher).to receive(:logger).and_return ClassLogger::LogWrapper.new('Object')
+    end
+    let(:link_fetcher) { Object.new }
```
davidlimcheng: @lyttam Very minor, but is there a reason we couldn't do 

`let (:link_fetcher) { LinkFetcher }`

And remove the `extend` statement above?

lyttam: @davidlimcheng it's tricky to test a module that you can't make an instance of.  I wanted to keep the module a pure mixin, so in order to test it I had to create an object instance and then mix it in (though I'm `extend`ing rather than using `include` - not sure if it matters).

davidlimcheng: Ahh, makes total sense.

raydavis:
Hey, @lyttam & @davidlimcheng : Hope y'all don't mind a couple of belated side-notes...

* The LinkFetcher module assumes that whatever calls it would have a "@uid" instance variable on hand, but it looks like "placeholders" might be more useful to log when there's an error.

* Just as a FYI, when you've got a pure drop-in function that _doesn't_ make assumptions about being included in a particular sort of object, it can be handy to add ``extend self`` to the module. That lets you avoid any question of "where is this code coming from?" by (optionally) calling
```
LinkFetcher.fetch_link(arga, argb)
```
instead of needing to include.

* When testing pure mixins, we've generally created test-classes in the spec files themselves. You can find lots of examples in "spec/lib/cache". That matches the way the code-under-test actually gets used, and so it's easier (at least for simple minds like me) to tell at a glance what's going on.

lyttam:
@raydavis thanks for the feedback!  I like the idea of logging `placeholders`, and I'll try out your suggestions.  Will submit another PR.

raydavis:
Thanks for your patience, @lyttam ! FWIW, none of my comments were things that would've blocked a merge or required amendment -- just data points which you might not have bumped into already...

PR #6570, opened 2017-04-11
SISRP-30492 - 8.1 - Summer Grading

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-30492

* Introduces Summer Grading to CalCentral, which follows significantly different logic than regular term grading.
* Lots of the logic breakdown can be found here:  https://jira.berkeley.edu/browse/SISRP-30589
* But as always, feel free to ask me any questions.

### DIFF
--- a/app/models/my_academics/grading.rb
+++ b/app/models/my_academics/grading.rb
@@ -1,7 +1,10 @@
 module MyAcademics
   class Grading < UserSpecificModel
-
+    include GradingModule
     include LinkFetcher
+
+    #TODO: Starting Fall 2017, replace this with an array of the current term + the three previous terms.
+    ACTIVE_GRADING_TERMS = ['2168', '2172', '2175']
 
     def merge(data)
       teaching_semesters = data[:teachingSemesters]
@@ -10,88 +13,62 @@ module MyAcademics
       end
     end
 
-    def grading_status_mapping
-      {
-        noCsData: {
-          beforeGradingPeriod: :periodNotStarted,
-          inGradingPeriod: :periodNotStarted,
-          afterGradingPeriod: :periodNotStarted,
-          gradingPeriodNotSet: :periodNotStarted
-        },
-        GRD: {
-          beforeGradingPeriod: :periodNotStarted,
-          inGradingPeriod: :periodStarted,
-          afterGradingPeriod: :gradesOverdue,
-          gradingPeriodNotSet: :periodStarted
-        },
-        POST: {
-          beforeGradingPeriod:  :gradesPosted,
-          inGradingPeriod:  :gradesPosted,
-          afterGradingPeriod:  :gradesPosted,
-          gradingPeriodNotSet:  :gradesPosted
-        },
-        RDY: {
-          beforeGradingPeriod:  :gradesApproved,
-          inGradingPeriod:  :gradesApproved,
-          afterGradingPeriod:  :gradesApproved,
-          gradingPeriodNotSet:  :gradesApproved
-        },
-        NRVW: {
-          beforeGradingPeriod: :periodNotStarted,
-          inGradingPeriod: :periodStarted,
-          afterGradingPeriod: :periodStarted,
-          gradingPeriodNotSet: :periodStarted
-        },
-        # Approved midpoint grades will mimic posted grade behavior on the front-end
-        APPR: {
-          beforeGradingPeriod:  :gradesPosted,
-          inGradingPeriod:  :gradesPosted,
-          afterGradingPeriod:  :gradesPosted,
-          gradingPeriodNotSet:  :gradesPosted
-        }
-      }
-    end
-
     def add_grading_to_semesters(teaching_semesters)
       teaching_semesters.try(:each) do |semester|
-        term_code = Berkeley::TermCodes.slug_to_edo_id(semester[:slug])
-        add_grading_header(semester, term_code)  if has_general?(semester[:classes])
-        add_grading_header_law(semester, term_code) if has_law?(semester[:classes])
-        add_grading_to_classes(semester[:classes], term_code)
-       end
-    end
-
-    def has_law?(semester_classes)
-      !!semester_classes.try(:find) do |semester_class|
-        semester_class[:dept].present? && semester_class[:dept] == 'LAW'
+        term_id = Berkeley::TermCodes.slug_to_edo_id(semester[:slug])
+        add_grading_links semester
+        if ACTIVE_GRADING_TERMS.include? term_id
+          if is_summer_semester? semester
+            add_grading_dates_summer(semester[:classes])
+          else
+            add_grading_dates(semester, term_id)  if has_general?(semester[:classes])
+            add_grading_dates_law(semester, term_id) if has_law?(semester[:classes])
+          end
+        end
+        add_grading_to_classes(semester[:classes], term_id)
       end
     end
 
-    def has_general?(semester_classes)
-      !!semester_classes.try(:find) do |semester_class|
-        semester_class[:dept].present? && semester_class[:dept] != 'LAW'
+    def add_grading_links(semester)
+      if has_general?(semester[:classes])
+        # A termCode of 'C' denotes a summer term.  Every non-law term has midpoint grading, except summer.
+        if is_summer_semester? semester
+          semester.merge!(
+            {
+              gradingAssistanceLink: grading_period.links.final
+            })
+        else
+          semester.merge!(
+            {
+              gradingAssistanceLinkMidpoint: grading_period.links.midpoint,
+              gradingAssistanceLink: grading_period.links.final
+            }
+          )
+        end
+      end
+      if has_law?(semester[:classes])
+        semester.merge!(
+          {
+            gradingAssistanceLinkLaw: grading_period.links.law
+          })
       end
     end
 
-    def add_grading_header(semester, term_code)
+    def add_grading_dates(semester, term_id)
       # This is a temp fix for Fall 2016 & Spring 2017 hardcoded from settings
       fall_dates = grading_period.dates.general.fall_2016
       spring_dates = grading_period.dates.general.spring_2017
-      if (term_code == '2168' || term_code == '2172') && valid_grading_period?(false, term_code)
+      if (term_id == '2168' || term_id == '2172') && valid_grading_period?(false, term_id)
         semester.merge!(
           {
-            gradingAssistanceLinkMidpoint: term_code == '2172' ? grading_period.links.midpoint : nil,
-            gradingAssistanceLink:  grading_period.links.final,
-            gradingPeriodMidpointStart: term_code == '2172' ? format_period_start(spring_dates.midpoint.start) : nil,
-            gradingPeriodMidpointEnd: term_code == '2172' ? format_period_end(spring_dates.midpoint.end) : nil,
-            gradingPeriodFinalStart: term_code == '2172' ? format_period_start(spring_dates.final.start) : format_period_start(fall_dates.start),
-            gradingPeriodFinalEnd:  term_code == '2172' ? format_period_end(spring_dates.final.end) : format_period_end(fall_dates.end)
+            gradingPeriodMidpointStart: term_id == '2172' ? format_period_start(spring_dates.midpoint.start) : nil,
+            gradingPeriodMidpointEnd: term_id == '2172' ? format_period_end(spring_dates.midpoint.end) : nil,
+            gradingPeriodFinalStart: term_id == '2172' ? format_period_start(spring_dates.final.start) : format_period_start(fall_dates.start),
+            gradingPeriodFinalEnd:  term_id == '2172' ? format_period_end(spring_dates.final.end) : format_period_end(fall_dates.end)
           })
       else
         semester.merge!(
           {
-            gradingAssistanceLinkMidpoint: grading_period.links.midpoint,
-            gradingAssistanceLink:  grading_period.links.final,
             gradingPeriodMidpointStart: nil,
             gradingPeriodMidpointEnd: nil,
             gradingPeriodFinalStart: nil,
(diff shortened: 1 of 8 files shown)


davidlimcheng:
Please hold off on merging for now - there is a Grading meeting today (4/11) that may or may not change the grading window for general classes.  If there is a change, it will be very minimal, so I figure it'd be better to submit this PR earlier for review, since it's a big one.

app/models/my_academics/grading_module.rb
```diff
@@ -0,0 +1,257 @@
...
+      rescue
+        logger.error "Grading Period dates errored out for #{datetype} in Settings for Class #{self.class.name} feed, uid = #{@uid}"
+        return true
+      end
+      false
+    end
+
+    # Campus Solutions will send a date string (ex. 23-MAY-16) that Rails incorrectly parses as a UTC Time object.  We want the time casted to Pacific with the same time preserved, so that other time zones will
+    # automatically convert to the correct time relative to Berkeley's Pacific time zone
+    def cast_utc_to_pacific(utc_time)
```
davidlimcheng: @lyttam Do you think this belongs in `DatedFeed`?  I went back and forth on it, and decided that this was a function very specific to Grading - but am open to moving it over to that module for potential future DRY-ness.

davidlimcheng:
OK, PR is good to go.  Formula has been reviewed and signed off by functional partners.  But, I'd still like to merge #6569 and fix any conflicts before merging this PR.

app/models/my_academics/grading_module.rb
```diff
@@ -0,0 +1,257 @@
...
+          return dates_out_of_order?(grading_period.dates.general.fall_2016, 'General Fall 2016')
+        elsif term_id == '2172'
+          return dates_out_of_order?(grading_period.dates.general.spring_2017.midpoint, 'General Midpoint Spring 2017') ||
+            dates_out_of_order?(grading_period.dates.general.spring_2017.final, 'General Final Spring 2017')
+        else
+          return true
+        end
+      end
+    end
+
```
redconfetti: @davidlimcheng - For the above 3 methods, why is it using hardcoded term codes for Fall 2016 and Spring 2017? Is it just these years that data integrity is expected to be low for some reason?

davidlimcheng: @redconfetti, a variety of reasons, but the most important being that Spring 2017 has midpoint grading, and Fall 2016 does not have midpoint grading.  That, in conjunction with the dates being hardcoded in the configuration file, made this (in my opinion) the simplest, easiest solution for a temporary problem.

Moving forward, all non-summer terms will have midpoint grading.  We will also not be hardcoding dates, but grabbing them from CS via API.  Having that much-needed consistency moving forward will really help simplify this code.  For now though, the requirements necessitated this unfortunate code.

app/models/my_academics/grading_module.rb
```diff
@@ -0,0 +1,257 @@
...
+    def format_period_start_summer(start_date)
+      start_date.to_date.strftime('%m/%d')
+    end
+
+    def format_period_end_summer(end_date)
+      return end_date.to_date.strftime('%m/%d') if DateTime.now.year == end_date.to_date.year
+      end_date.to_date.strftime('%m/%d/%Y')
+    end
+
+    def unexpected_cs_status?(cs_grading_status, is_law)
```
lyttam: @davidlimcheng Based on the name, this looks like a function who's job is to perform some logic checks and return a boolean.  So it feels unexpected to see it actually doing more than that, i.e. the `logger.debug` statement on line 121.  

I'm a fan of functions that are as basic as possible and only have one job to do.  You might consider moving the `logger.debug` statement outside of this function (to wherever this function is called, perhaps).

app/models/my_academics/grading_module.rb
```diff
@@ -0,0 +1,257 @@
...
+    def valid_grading_period?(is_law, term_id)
+      # Use class level var to reduce noise in log on invalid grading period
+      return @valid_grading_period if !is_law  && @valid_grading_period.present?
+      return @valid_grading_period_law if is_law && @valid_grading_period_law.present?
+      @valid_grading_period = check_grading_period?(is_law, term_id) unless is_law
+      @valid_grading_period_law = check_grading_period?(is_law, term_id) if is_law
+      is_law ? @valid_grading_period_law : @valid_grading_period
+    end
+
+    def check_grading_period?(is_law, term_id)
```
lyttam: @davidlimcheng I'd name this function `valid_grading_period_dates?` or something to that effect.

app/models/my_academics/grading_module.rb
```diff
@@ -0,0 +1,257 @@
...
+        if term_id == '2168'
+          return dates_are_blank?(grading_period.dates.law.fall_2016, 'Law Fall 2016')
+        elsif term_id == '2172'
+          return dates_are_blank?(grading_period.dates.law.spring_2017, 'Law Spring 2017')
+        else
+          return true
+        end
+      else
+        if term_id == '2168'
+          return dates_are_blank?(grading_period.dates.general.fall_2016, 'General Fall 2016')
```
lyttam: @davidlimcheng Since these strings get repeated, you could make them constants.

app/models/my_academics/grading_module.rb
```diff
@@ -0,0 +1,257 @@
...
+        elsif term_id == '2172'
+          return dates_are_blank?(grading_period.dates.general.spring_2017.midpoint, 'General Midpoint Spring 2017') ||
+            dates_are_blank?(grading_period.dates.general.spring_2017.final, 'General Final Spring 2017')
+        else
+          return true
```
lyttam: @davidlimcheng Just curious why these functions all default to return true... is it the case that terms prior to 2168 all have invalid grading periods?

app/models/my_academics/grading_module.rb
lyttam: @davidlimcheng Hmm, yeah I think it would make sense to have it there, but I can also see the case for not putting it there until there's a need to use it in more than one place.

src/assets/javascripts/angular/services/academicsService.js
```diff
@@ -306,6 +307,15 @@ angular.module('calcentral.services').service('academicsService', function() {
     }
   };
 
+  var isSummerSemester = function(selectedTeachingSemester) {
+    var termCode = _.get(selectedTeachingSemester, 'termCode');
+    if (termCode === 'C') {
```
lyttam: @davidlimcheng you could remove the `if`/`else` and just `return termCode === 'C'`

lyttam:
@davidlimcheng Nice work!  Looks like some really complex business requirements, but you've encapsulated much of it so nicely in `GradingModule`.

app/models/my_academics/grading_module.rb
davidlimcheng: @lyttam yep!  Since Grading was introduced to CalCentral in 2168, we have no information/data for anything before then.

src/assets/javascripts/angular/services/academicsService.js
davidlimcheng: @lyttam Ah, my Javascript-ness must've leaked through here.  That's apparent in the usage of `===` instead of `==` too =)

app/models/my_academics/grading_module.rb
davidlimcheng: Totally - amendment coming.

app/models/my_academics/grading_module.rb
davidlimcheng: Yeah, that's better, since `check_grading_period?` isn't really a question.

app/models/my_academics/grading_module.rb
davidlimcheng: I agree that functions should just have one job.  I'll find a better home for the logger statement.

src/assets/javascripts/angular/services/academicsService.js
davidlimcheng: Oh it is JS!  Whoops - nevertheless, your comment is still valid.

app/models/my_academics/grading.rb
```diff
@@ -100,89 +77,139 @@ def add_grading_header(semester, term_code)
...
+              gradingLink: has_grade_access ? get_grading_link(ccn, term_id, cs_grading_status) : nil
             })
         end
       end
     end
 
     def has_grading_access?(section)
       !!section[:instructors].try(:find) do |instructor|
-        instructor[:uid] == @uid && instructor[:ccGradingAccess] != :noGradeAccess
+        instructor[:uid].try(:to_i) == @uid.try(:to_i) && instructor.try(:[], :ccGradingAccess) != :noGradeAccess
```
redconfetti: @davidlimcheng - Why are we using try here? Don't we want an exception raised, with an error, if instructor UID or @uid is not numeric. Also if they both are not numeric, then `#try` will return nil for both.

app/models/my_academics/grading_module.rb
```diff
@@ -0,0 +1,261 @@
...
+      end
+      false
+    end
+
+    def dates_are_blank?(dates, datetype)
+      dates.each_value do |date|
+        if (date.blank?)
+          logger.error "No date set for Grading Period #{datetype} in Settings for Class #{self.class.name} feed, uid = #{@uid}"
+        end
+        return true
```
redconfetti: @davidlimcheng - I know you just moved this over to another module, but I'm curious how this works properly if the `return true` is not inside of the `if` statement that runs when the date is blank.

app/models/my_academics/grading_module.rb
```diff
@@ -0,0 +1,261 @@
...
+      end
+      false
+    end
+
+    def dates_out_of_order?(dates, datetype)
+      begin
+        if DateTime.parse(dates.start.to_s) >= DateTime.parse(dates.end.to_s)
```
redconfetti: @davidlimcheng - Why are we converting dates to a string, only to use DateTime to parse them again? Are the dates not in DateTime format, but in Date or Time format or something?

spec/models/my_academics/grading_spec.rb
```diff
@@ -159,54 +164,43 @@
...
   end
 
   context 'when grading returns statuses for a teaching semester' do
-    let(:fake_grading_url) { 'http://fake.grading.com' }
     let(:uid) { '238382' }
     let(:grading_proxy) { CampusSolutions::Grading.new(user_id: uid, fake: fake) }
     let(:feed) { {}.tap { |feed| MyAcademics::Teaching.new(uid).merge feed } }
 
     before do
+      @uid = uid
```
redconfetti: @davidlimcheng - Why are you using an instance variable here? I don't see it referenced elsewhere in the methods in this spec.

spec/models/my_academics/grading_spec.rb
```diff
@@ -218,6 +212,29 @@
...
+  end
+
+  context 'when grading for a summer term' do
 
+    let(:uid) { '904715' }
+    let(:grading_proxy) { CampusSolutions::Grading.new(user_id: uid, fake: fake) }
+    let(:feed) { {}.tap { |feed| MyAcademics::Teaching.new(uid).merge feed } }
+
+    before do
+      @uid = uid
```
redconfetti: @davidlimcheng - See above regarding instance variable for `@uid`

app/models/my_academics/grading.rb
davidlimcheng: @redconfetti Hmm, you bring up good points.  To me, the most likely error would be for `instructor[:uid]` to be `nil` for some reason, which probably won't ever happen, but who knows what CS will send us every once in awhile.  In that case I think using a `try` would be correct, because that would mean the instructor does not have grade access for that class.  Because of that scenario being the most likely, I would think keeping the `try` here would be better than raising an error, because we can then say "this instructor should not have grade access for this class", and move on.

app/models/my_academics/grading_module.rb
davidlimcheng: @redconfetti The dates here are taken from the YML configuration file, which IIRC, parse as an `OpenStruct` object.

spec/models/my_academics/grading_spec.rb
davidlimcheng: @redconfetti This is referenced in the `has_grade_access?` method, which is run throughout the course of `subject.merge(feed)` on line 207.

spec/models/my_academics/grading_spec.rb
davidlimcheng: @redconfetti Same as my answer above.

app/models/my_academics/grading_module.rb
redconfetti: Oh I see, they're taken from the YML config in string format, you're just making sure they're converted to a String before passing to DateTime.parse.

app/models/my_academics/grading_module.rb
davidlimcheng: @redconfetti You're totally right, it should be in the `if` clause.  I took a look to see why the rspec was passing despite this function incorrectly handling the dates.   There is another function, `dates_out_of_order?` that the dates are being checked against that compares the `start` date and the `end` date with a `>=` comparator.  This was erroring out due to a `nil` being compared to a `DateTime`.  Since the test was just checking for `valid_grading_period?` to be false, the test passed.

What this tells me is that this particular function is unnecessary.  Great catch - I'll amend now.

spec/models/my_academics/grading_spec.rb
redconfetti: @davidlimcheng - Sure, the MyAcademics::Grading module is included in classes that have `@uid` set, however this is declaring @uid within the Rspec... not the module. It's not injecting it into the module or class being tested using #instance_eval or #class_eval. So within your Rspec tests, I don't see another reference to @uid.

It seems like it's not used here. `uid` exists because of the `let` statement on line 3 of the specs, but the `@uid` isn't used elsewhere it seems.

davidlimcheng: @redconfetti Ah, I see what you mean.  I suppose `@uid` is also set in `UserSpecificModel`'s initialization, which takes the `uid` input and assigns it to @uid.  

Will remove and amend.

PR #6571, opened 2017-04-11
SISRP-31764 Fall back to LDAP for CS ID if no Crosswalk

raydavis:
https://jira.berkeley.edu/browse/SISRP-31764

The concept is simple enough, but the original implementation of Crosswalk caching created a tangle of concerns which needed combing out.

To ease the comb's progress, I deleted a lot of now obsolete code and moved a lot of other code around.


spec/controllers/sessions_controller_spec.rb
```diff
@@ -70,18 +73,21 @@
       end
     end
 
-    context 'campus solutions' do
-      let(:user_id) { '12352' }
-
-      it 'will cache the Campus Solutions IDs' do
```
raydavis: Contrary to this comment's implication, the old RSpec code didn't test SessionController caching. Instead, the test code merely called Crosswalk and confirmed that Crosswalk returned results. SessionController's cache logic only kicks in via the "extra" method on the Omniauth object.

app/models/user/identifiers.rb
```diff
@@ -1,5 +1,5 @@
 module User
-  module Student
+  module Identifiers
```
davidlimcheng: @raydavis, I see that we're also using the User::Student mixin in `MyAcademics::AcademicPlan`, `MyAcademics::CollegeAndLevel`, `MyAcademics::Exams`, and `MyAcademics::TransferCredit`.  Would we have to rename those, as well?

app/models/calnet_crosswalk/by_sid.rb
```diff
@@ -1,13 +0,0 @@
...
-    def url
-      "#{@settings.base_url}/LEGACY_SIS_STUDENT_ID/#{@uid}"
-    end
-
-    def mock_request
-      super.merge(uri_matching: "#{@settings.base_url}/LEGACY_SIS_STUDENT_ID/#{@uid}")
-    end
-
-  end
-end
```
davidlimcheng: @raydavis, very small, but are we also able to remove [this line](https://github.com/ets-berkeley-edu/calcentral/blob/master/config/settings.yml#L309) from `settings.yml`?

raydavis: Not just able, but obliged to! Thanks, @davidlimcheng -- amendment directly...

app/models/calnet_ldap/user_attributes.rb
```diff
@@ -38,5 +47,40 @@ def self.get_attributes_by_name(name, include_guest_users=false)
...
+        logger.warn "Conflicting berkeleyEduStuID #{sid} and berkeleyEduCSID #{cs_id} for UID #{ldap_uid}"
+      end
+      {
+        ldap_uid: ldap_uid,
+        student_id: sid,
+        campus_solutions_id: cs_id
+      }
+    end
+
+    def self.string_attribute(ldap_record, key)
```
davidlimcheng: Very nice, I'm going to have to steal this from you in the future!

davidlimcheng:
Whew, good stuff @raydavis!  Lots of work here.  As always, I'm learning new things just looking at your PR.  I also have to remind myself to look at the whole PR before leaving comments, not just an individual commit.  

I've just got one outstanding question regarding renaming the `User::Student` mixin, otherwise LGTM.  But I'll hold off on merging for now, it'll be good to get another pair of eyes on this as well.

raydavis:
Thanks, @davidlimcheng . What's your question about the renaming?

As background, the pretty-silly User::Student mixin was originally put in to get the Student ID needed to call the BearFact APIs. But once CS IDs came in, there started to be a conceptual overlap between the generic (not-just-Crosswalk) use of the Crosswalk identifiers classes and the old BearFacts code, and the confusion led to duplicate code (lots of User::Student mixers creating their own ``@campus_solution_id`` instances) and inefficiency. Since I needed a place to move the not-just-Crosswalk aspects of Crosswalk code, and I would need to deal with the mixins anyway, I slightly repurposed User::Student.

As always, suggestions are welcome!

davidlimcheng:
Thanks for the explanation, @raydavis.  I'm definitely guilty of causing some of that inefficiency too, so I'm glad this PR is going on.  My question was regarding renaming `User::Student` in several other places (`MyAcademics::AcademicPlan`, `MyAcademics::CollegeAndLevel`, `MyAcademics::Exams`, and `MyAcademics::TransferCredit`), but I see now that you've removed the mixin altogether from those classes.

LGTM!

davidlimcheng:
@raydavis, looks like a merge conflict after I had merged #6569

raydavis:
Always a risk with a big PR, @davidlimcheng . I'll amendagen.

davidlimcheng:
Mergarific, thanks @raydavis !

PR #6577, opened 2017-04-13
SISRP-32137 - Number each attempt on the Qualifying Exam milestone

lyttam:
https://jira.berkeley.edu/browse/SISRP-32137

This brings the Degree Progress card more in line with the Committees card.  We weren't bothering to show "Exam n:" on the second or third exam attempt as long as they passed it on that attempt.  Now we will show the exam number on every attempt.

### DIFF
--- a/app/models/degree_progress/milestones_module.rb
+++ b/app/models/degree_progress/milestones_module.rb
@@ -76,22 +76,13 @@ module DegreeProgress
     end
 
     def parse_milestone_attempt(milestone_attempt)
-      milestone_attempt = {
-        sequenceNumber: milestone_attempt[:attemptNbr].to_i,
-        date: parse_date(milestone_attempt[:attemptDate]),
-        result: Berkeley::GraduateMilestones.get_status(milestone_attempt[:attemptStatus], Berkeley::GraduateMilestones::QE_RESULTS_MILESTONE),
-        statusCode: milestone_attempt[:attemptStatus]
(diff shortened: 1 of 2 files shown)


app/models/degree_progress/milestones_module.rb
```diff
@@ -76,22 +76,13 @@ def parse_milestone_attempts(requirement)
...
-        sequenceNumber: milestone_attempt[:attemptNbr].to_i,
-        date: parse_date(milestone_attempt[:attemptDate]),
-        result: Berkeley::GraduateMilestones.get_status(milestone_attempt[:attemptStatus], Berkeley::GraduateMilestones::QE_RESULTS_MILESTONE),
-        statusCode: milestone_attempt[:attemptStatus]
+      attempt_result = Berkeley::GraduateMilestones.get_status(milestone_attempt[:attemptStatus], Berkeley::GraduateMilestones::QE_RESULTS_MILESTONE)
+      sequence_number = milestone_attempt[:attemptNbr].to_i
+      {
+        sequenceNumber: sequence_number,
+        statusCode: milestone_attempt[:attemptStatus],
+        display: "Exam #{sequence_number}: #{attempt_result} #{parse_date(milestone_attempt[:attemptDate])}"
```
davidlimcheng: @lyttam, with the change in back-end parsing, I would also expect a necessary change in front-end presentation - or is that not the case here?

davidlimcheng: Ah nevermind, I see that milestone_attempt[:display] _was_ in the prior object as well.

davidlimcheng:
Looks good!

PR #6580, opened 2017-04-14
SISRP-31567 - refactors student success

lyttam:

Another blind attempt to solve that mystery bug where the data doesn't load on the card the first time.

The console.log statements seemed to point to some timing issue or race condition involving setting things in `$scope` and then relying on them to be there in other parts of the JS controller.  I hope this will fix that, but we won't know until we can test it in dev.

I'd like this to have @davidlimcheng's blessing before merging.


davidlimcheng:
@lyttam, if anything, this simplifies the code and makes it much more readable.  Fingers crossed it also will squash the pesky little bug too.  Thanks @lyttam!

PR #6582, opened 2017-04-16
SISRP-27686 -  Semesters - Withdrawal Status

nkutub:
https://jira.berkeley.edu/browse/SISRP-27686

* added withdrawal data to semesters feed 
* new EDODB query  get_withdrawal_status
* added withdrawal info to semseters card

### DIFF
--- a/src/assets/templates/academics_semesters.html
+++ b/src/assets/templates/academics_semesters.html
@@ -11,16 +11,18 @@
   <div class="cc-widget-padding cc-flex-space-between-vertical-15">
     <div class="cc-table cc-academics-semester cc-academics-semester-{{semester.timeBucket}}" data-ng-repeat="semester in semesters | limitTo:pastSemestersLimit">
       <div>
-        <h3 data-ng-if="semester.hasEnrollmentData && semester.slug && !isAdvisingStudentLookup"><a class="cc-left" data-ng-href="/academics/semester/{{semester.slug}}" data-ng-bind="semester.name"></a></h3>
-        <h3 class="cc-left" data-ng-if="!semester.hasEnrollmentData || !semester.slug || isAdvisingStudentLookup" data-ng-bind="semester.name"></h3>
+        <h3 data-ng-if="!semester.hasWithdrawalData && semester.hasEnrollmentData && semester.slug && !isAdvisingStudentLookup"><a class="cc-left" data-ng-href="/academics/semester/{{semester.slug}}" data-ng-bind="semester.name"></a></h3>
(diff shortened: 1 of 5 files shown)


src/assets/templates/academics_semesters.html
```diff
@@ -11,16 +11,18 @@ <h2 class="cc-left">Semesters</h2>
   <div class="cc-widget-padding cc-flex-space-between-vertical-15">
     <div class="cc-table cc-academics-semester cc-academics-semester-{{semester.timeBucket}}" data-ng-repeat="semester in semesters | limitTo:pastSemestersLimit">
       <div>
-        <h3 data-ng-if="semester.hasEnrollmentData && semester.slug && !isAdvisingStudentLookup"><a class="cc-left" data-ng-href="/academics/semester/{{semester.slug}}" data-ng-bind="semester.name"></a></h3>
-        <h3 class="cc-left" data-ng-if="!semester.hasEnrollmentData || !semester.slug || isAdvisingStudentLookup" data-ng-bind="semester.name"></h3>
+        <h3 data-ng-if="!semester.hasWithdrawalData && semester.hasEnrollmentData && semester.slug && !isAdvisingStudentLookup"><a class="cc-left" data-ng-href="/academics/semester/{{semester.slug}}" data-ng-bind="semester.name"></a></h3>
+        <h3 class="cc-left" data-ng-if="semester.hasWithdrawalData || !semester.hasEnrollmentData || !semester.slug || isAdvisingStudentLookup" data-ng-bind="semester.name"></h3>
```
davidlimcheng: @nkutub Lines 14+15 have some pretty gnarly logic in there, would it be possible encapsulate this logic in the Controller instead?  Then we can give it a readable name like "showSemesterLink" or something like that.

We can do this in a future merge too if you're pressed for time - I'll leave that up to your discretion.

lyttam: +1.  Lots of logic in this template, and any way it could be grouped into meaningful business terms would be helpful.

nkutub: @davidlimcheng after discussing with you seems like it will be more consistent with entire card to keep the flags visible

davidlimcheng: @nkutub @lyttam Yeah - Nader and I discussed and due to the immense amount of sharing for these flags on this page, it'd ironically make it less readable if we encapsulate just a piece of the pie.  

One thing we can do in a future PR is get rid of all the isStudentAdvisingLookup flags, since the Semesters card does not appear on that page anymore.  Created [SISRP-32292](https://jira.berkeley.edu/browse/SISRP-32292) for that purpose.

davidlimcheng:
good stuff, @nkutub

PR #6584, opened 2017-04-17
SISRP-30166 - Hides popover alert icon when zero alerts

lyttam:
https://jira.berkeley.edu/browse/SISRP-30166

We were setting `hasAlerts` to true even when the number of holds was zero.


davidlimcheng:
Great use of the conditional, @lyttam

PR #6586, opened 2017-04-17
SISRP-31200 - Revision asset files in development build mode

lyttam:
https://jira.berkeley.edu/browse/SISRP-31200

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -374,15 +374,11 @@
    * Add hashes to the files and update the includes (production)
    */
   gulp.task('revall', function() {
-    if (!isProduction) {
-      return;
-    }
-
     var path = require('path');
     var RevAll = require('gulp-rev-all');
     var revAllAssets = new RevAll({
-      // Since we only run this in production mode, add some extra logging
-      debug: true,
+      // Add some extra logging in production mode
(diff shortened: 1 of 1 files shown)


davidlimcheng:
@lyttam, if I'm understanding this correctly, with this change, we'll be caching our static assets in all environments, including local development environments.  Production will have extra logging, but Development env's will behave similarly.  I'd imagine this might clash with Browser-sync's "real-time" reloading.  Am I understanding this correctly?

lyttam:
@davidlimcheng this change will prevent the browser from hanging on to the old version of an asset after you've made changes to it.  Actually, we were already doing that in dev, qa, and prod, but this will extend it to local development as well.  

The browser already caches the assets - it _tries_ to be smart about it, but sometimes it's not, so the rev-all plugin is a cache-busting implementation aimed at strategically circumventing the caching.

dev, qa, and prod already had that extra logging, so I wanted to preserve that without adding it to the local build process because it's too verbose for that.

It doesn't seem to interfere with browser-sync, but if anyone experiences issues with that we can revisit this.

davidlimcheng:
Ahh, gotcha, thanks for the clarification @lyttam.  This will be a welcome change.

PR #6590, opened 2017-04-18
CLC-6625 Include OEC Freshman/Sophomore seminar support in DB migrate

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6625

This PR should fix the Bamboo build.

### DIFF
--- a/spec/models/edo_oracle/oec_spec.rb
+++ b/spec/models/edo_oracle/oec_spec.rb
@@ -18,9 +18,9 @@ describe EdoOracle::Oec do
         ]
       end
       it { should include(
-        "(sec.\"displayName\" LIKE 'CATALAN %')",
-        "(sec.\"displayName\" LIKE 'PORTUG %')",
-        "(sec.\"displayName\" LIKE 'SPANISH %')"
+        "(sec.\"displayName\" LIKE 'CATALAN %'",
+        "(sec.\"displayName\" LIKE 'PORTUG %'",
+        "(sec.\"displayName\" LIKE 'SPANISH %'"
       ) }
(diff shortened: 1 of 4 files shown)


spec/models/edo_oracle/oec_spec.rb
```diff
@@ -19,8 +19,8 @@
       end
       it { should include(
         "(sec.\"displayName\" LIKE 'CATALAN %')",
```
davidlimcheng: @raydavis, it looks like you're expecting the `)` here but not on the following two lines.  Should one case be different than the others?

raydavis: Thanks for asking, @davidlimcheng . I made the minimal changes needed to fix the test, but it would have been less puzzling if I'd either added a comment or changed the CATALAN line to match (even though it doesn't need the change to work). My only excuse is that the INTEGBI check farther down shows the same mysterious difference.

As background, these tests are running against two different DBs. In RAILS_ENV=test, the in-memory sqlite3 DB has nothing but the CourseCode rows added by the spec. But in RAILS_ENV=testext, the Postrgres DB may have records which result in clauses like ```(sec."displayName" LIKE 'PORTUG %' and sec."displayName" != 'PORTUG 24')```

davidlimcheng: Ahh, very tricky.  Thanks for the clarification, @raydavis

raydavis:
Amended, @davidlimcheng . I added a more minor OEC update, as well.

davidlimcheng:
Looks good, @raydavis!

PR #6591, opened 2017-04-18
[QA] SISRP-31567 - Removes JS logging

lyttam:

Removes logging introduced in #6573 for debugging purposes.  Also returns jshint config to its original state.  This should be merged only into the qa branch, as we want to leave the logging in place in master for now.


davidlimcheng:
Logs begone!

PR #6596, opened 2017-04-19
SISRP-28717 - 8.1: Update SOR to CS for CalCentral

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-28717

* Cleans up some minor bugs when switching the `allow_legacy_fallback` flag to `false`.
* Updates the Rspec tests to not expect a `semesters[:additionalCredits]` node when only pulling from CS.
* Double-checked with Student Records regarding the `semesters[:additionalCredits]` node - we are pulling EAP/AP/UC EXT data from there according to #3483 - all of which are covered by other avenues in CalCentral (can expand more on this if needed)
* Would like @nkutub to give this a review before merging
* Once this is merged, we'll be able to flip the switch in all clusters to only pull from EDODB.
* QA PR to come

### DIFF
--- a/app/models/my_academics/academics_module.rb
+++ b/app/models/my_academics/academics_module.rb
@@ -22,6 +22,10 @@ module MyAcademics
         termCode: term_cd,
         termYear: term_yr,
         timeBucket: time_bucket(term_yr, term_cd),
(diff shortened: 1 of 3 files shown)


nkutub:
good catch @davidlimcheng

PR #6609, opened 2017-04-24
SISRP-32402 - Semesters - Withdrawal Status

nkutub:
https://jira.berkeley.edu/browse/SISRP-32402

* refactor the merge to incorporate as part of the semester's sort.

### DIFF
--- a/app/models/my_academics/semesters.rb
+++ b/app/models/my_academics/semesters.rb
@@ -30,7 +30,8 @@ module MyAcademics
     end
 
     def semester_feed(enrollment_terms, transcript_terms, withdrawal_data)
-      (enrollment_terms.keys | transcript_terms.keys).sort.reverse.map do |term_key|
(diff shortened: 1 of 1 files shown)


redconfetti:
@nkutub - This doesn't have the "SISRP-32402" in the commit message.

app/models/my_academics/semesters.rb
```diff
@@ -40,39 +41,44 @@ def semester_feed(enrollment_terms, transcript_terms, withdrawal_data)
...
+          merge_withdrawals(semester, withdrawal_data)
         end
-        semester unless semester[:classes].empty?
+        semester unless semester[:classes].empty? && !semester[:hasWithdrawalData]
       end
     end
 
     def merge_withdrawals (semester, withdrawal_data)
       withdrawal_data.each do |row|
+        withdrawal_status = map_withdrawal_status(row)
```
lyttam: @nkutub I liked this better when it was inside the `if` block - that way it only gets executed if it's needed.

nkutub: @lyttam , good catch..

PR #6621, opened 2017-04-26
SISRP-32149 - Refactors academic roles / adds new summer visitor behavior

lyttam:
*Note:* don't merge this PR unless given the greenlight in the comments.  Still in discussion over whether we can postpone this fix until we have time to cure the disease rather than just treat the symptoms.

https://jira.berkeley.edu/browse/SISRP-32149

In a prior release we designed a special experience for summer visitors (students with a summer visitor academic plan) where they can't see some things that normal students can see, and they can see special things that normal students cannot see.  For example, we hide or show certain cards and links throughout the site.

A new case was brought to light where a student can be in a summer visitor plan and another plan AT THE SAME TIME.  This adds an additional layer of complexity, as these students need to see the special stuff that summer visitors can see, but they also need to see everything a normal student can see.

I created a new role for this and called it `summerVisitorPlus`.  I welcome any suggestions for a better way to handle this mess.


db/schema.rb
```diff
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
```
redconfetti: @lyttam - How did this schema update make it into your PR? I assume you perhaps ran a migration, and these changes were put in place... but should we include this? It's only rearranging the column ordering needlessly.

lyttam: @redconfetti it's also changing the number on line 14.  Should that be part of a commit, or not?  Should I add to .gitignore?

app/models/hub_edos/my_academic_status.rb
```diff
@@ -57,5 +58,21 @@ def collect_roles(status)
...
+      active_plan_codes.compact
+    end
+
+    def extract_active_plans(status)
+      status.try(:[], 'studentPlans').collect do |plan|
+        if active? plan
+          plan.try(:[], 'academicPlan').try(:[], 'plan').try(:[], 'code')
+        end
+      end
+    end
```
redconfetti: @lyttam - Is this also a new requirement, that we disregard plans that are not active? I thought that the iHub Student Academic Status API, which is used by MyAcademics::CollegeAndLevel, only returns active plans.

db/schema.rb
lyttam: Looking at the history of this file, it does need to be modified for every new migration.

redconfetti: @lyttam - I think this is a question for the entire team, not just me.
@davidlimcheng @nkutub What do you think?

raydavis: FWIW, I eventually found it less noisy to hit ``/api/my/campuslinks/refresh`` at release time instead of having to do so many migrations and schema updates. My Campus contents are expected to be fairly volatile.

For best of both worlds (automatic update on release but no migration noise), the ``script/migrate.sh`` script could run a Rake task to run delete_links and load_links. We never got around to that "because priorities."

redconfetti: @raydavis - I don't think we're using migrations to introduce campus links here. I think that for some reason `rake db:migrate` sometimes causes modifications to be made to db/schema.rb that are just a re-arrangement of column ordering. I've typically used `git checkout db/schema.rb` when this happens, so that it's not part of a PR.

I'm not sure if we should check these changes in, just to avoid other people having the same result, or if we should avoid them because they're just useless noise.

app/models/hub_edos/my_academic_status.rb
lyttam: @redconfetti no, we were doing it this way before (only considering active plans in determining the roles).  If the academic status API is only returning active plans, then I wasn't aware of that.

db/schema.rb
lyttam: I like @raydavis' suggestion - that way we don't have to do a migration at all, right?

@redconfetti I'm using a migration here to refresh the campus links.

redconfetti:
FYI, we had a discussion in this JIRA concerning this update in general being a layer of complication to Berkeley::AcademicRoles, which are used for:

1. Globally for UID to determine a students including in certain careers or plans (FPF, LAW, UCBX, HaasMba, HaasEwMba, etc). See /api/my/academics > [CollegeAndLevel](https://github.com/ets-berkeley-edu/calcentral/blob/c0c4d250/app/models/my_academics/college_and_level.rb#L26) > [Roles](https://github.com/ets-berkeley-edu/calcentral/blob/c0c4d250/app/models/hub_edos/my_academic_status.rb#L40)
1. [Advising card](https://confluence.ets.berkeley.edu/confluence/display/MYB/Advising+Widget)
1. [Class Enrollment card](https://confluence.ets.berkeley.edu/confluence/display/MYB/Class+Enrollment+Widget)

I'm not sure if it's wise for us to implement this code change as it greatly complicates the logic, and we're not sure if it's worth the complication for the population that it serves.

lyttam:
Based on discussion we just had in the design meeting, I'm going to close this PR and potentially submit a new one (if Bernie decides we need these changes for 8.1)

PR #6626, opened 2017-04-27
NOJIRA - Bumping version number to next release

davidlimcheng:


versions/previous_release_db_schema.txt
```diff
@@ -1 +1 @@
-20170307233607
+20170418172849
```
redconfetti: Confirmed that [20170418172849](https://github.com/ets-berkeley-edu/calcentral/blob/62025926/db/migrate/20170418172849_add_oec_course_codes.rb) is the latest DB migration.

PR #6627, opened 2017-04-28
SISRP-31724 - Hard Code Links to Petition to Change Courseload (Add/Drop)

redconfetti:
https://jira.berkeley.edu/browse/SISRP-31724

* Removes link from Student Resources card (My Dashboard for Students)
* Adds link to Class Enrollment card (My Academics for Students) for display only under the 'Class Adjustment' section when viewing a summer term


lyttam:
@redconfetti looks pretty straightforward.  Where's the logic that causes the link to display on Class Enrollment only when viewing a summer term?

src/assets/templates/widgets/enrollment/adjust.html
```diff
@@ -13,3 +13,10 @@
     data-ng-include="'widgets/enrollment/adjust_waitlisted.html'">
   </div>
 </div>
+
+<div class="cc-enrollment-card-section-content" data-ng-if="enrollmentInstruction.termIsSummer">
```
redconfetti: @lyttam - The ngIf ensures that the link only shows when `termIsSummer` is present for the enrollment Instruction. Luckily that flag was already in place, so I didn't have to add it to the feed. 🍾

lyttam: @redconfetti Cool, looks good then.

PR #6628, opened 2017-04-28
SISRP-32562 - CalnetCrosswalk::ByCsId being called with a UID, returning Error

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-32562

* Lots of stuff in the JIRA, but basically CalNet changed their 404 response body and we started logging errors because of it.
* This conforms to the new 404 response body.


lyttam:
@davidlimcheng excellent, thanks for cleaning up the error and for the nice explanation.

PR #6630, opened 2017-05-01
SISRP-32379 Map legacy CCNs to CS section IDs for Course Captures

raydavis:
https://jira.berkeley.edu/browse/SISRP-32379

The PR attempts to minimize programmatic joins between DBs rather than attempting full data coverage.

It also removes some obsolete EDO DB queries.

### DIFF
--- a/app/models/berkeley/legacy_terms.rb
+++ b/app/models/berkeley/legacy_terms.rb
@@ -1,0 +1,69 @@
+module Berkeley
+  class LegacyTerms
+    include ClassLogger
+    extend Cache::Cacheable
+
+    # When pre-Fall-2016 course section data was migrated to the CS DB, the original section Course Control Numbers
+    # were not included. Since external SIS-integrated resources frequently relied on CCNs as a unique section ID,
+    # this utility class provides a programmatic mapping.
+
+    def self.legacy_ccns_to_section_ids(cs_term_id, legacy_ccns)
+      mapping = {}
+      remaining = []
+      legacy_ccns.each do |legacy_ccn|
+        legacy_ccn = normalized(legacy_ccn)
+        ckey = cs_section_id_key(cs_term_id, legacy_ccn)
+        # We may have cached a nil value; i.e., there is no match and no need to query again.
+        if in_cache? ckey
+          mapping[legacy_ccn] = fetch_from_cache ckey
+        else
+          remaining << legacy_ccn
+        end
+      end
+      if remaining.present?
+        term_yr, term_cd = Berkeley::TermCodes.from_edo_id(cs_term_id).values
+        legacy_sections = CampusOracle::Queries.get_sections_from_ccns(term_yr, term_cd, remaining)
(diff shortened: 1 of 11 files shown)


raydavis:
I'd welcome reviews from @johncrossman and @nkutub, if they're available.

johncrossman:
Reviewing now, @raydavis

app/models/berkeley/legacy_terms.rb
```diff
@@ -0,0 +1,69 @@
...
+    # When pre-Fall-2016 course section data was migrated to the CS DB, the original section Course Control Numbers
+    # were not included. Since external SIS-integrated resources frequently relied on CCNs as a unique section ID,
+    # this utility class provides a programmatic mapping.
+
+    def self.legacy_ccns_to_section_ids(cs_term_id, legacy_ccns)
+      mapping = {}
+      remaining = []
+      legacy_ccns.each do |legacy_ccn|
+        legacy_ccn = normalized(legacy_ccn)
+        ckey = cs_section_id_key(cs_term_id, legacy_ccn)
```
johncrossman: @raydavis Seems that `legacy_ccn` is normalized twice above, the second time is in `cs_section_id_key()`. I see no harm tho'.

app/models/berkeley/legacy_terms.rb
```diff
@@ -0,0 +1,69 @@
...
+      write_cache(normalized(legacy_ccn), legacy_ccn_key(cs_term_id, section_id)) if cs_term_id.present?
+    end
+
+    def self.normalized(legacy_ccn)
+      if legacy_ccn.present?
+        # Normalize to deal with legacy zero-padding in some legacy systems.
+        legacy_ccn.to_i.to_s
+      else
+        legacy_ccn
+      end
```
johncrossman: @raydavis A blank legacy_ccn would cause problems in `legacy_ccns_to_section_ids` above. I assume such blankness is not expected.

app/models/berkeley/legacy_terms.rb
raydavis: Yes, that does look odd, @johncrossman -- I should have commented, I guess. It needs to be done in both places because legacy_ccn is used below in a context where it needs to be normalized, and I also expect cs_section_id_key() to be called from outside the ``legacy_ccns_to_section_ids`` method.

app/models/berkeley/legacy_terms.rb
raydavis: It's not expected in ``legacy_ccns_to_section_ids``, true.

johncrossman:
The perilous bridge to legacy data has been nicely fortified... For now!!  Thank you, @raydavis

redconfetti:
@raydavis - So the Legacy Oracle views will continue to be available with static data that we'll use to transform CCNs to Section IDs, until the CCNs are migrated?

raydavis:
Yes, @redconfetti , that's Karen K.'s preferred plan.

PR #6631, opened 2017-05-02
SISRP-31567 - Removes sorting from termGpa parsing

lyttam:


davidlimcheng:
One green Travis

PR #6637, opened 2017-05-03
SISRP-31567 - Student Overview student success card is blank when page loads first time

davidlimcheng:

* The actual fix for the big kahuna
* Moves the "is active career?" logic to the back-end
* Had to move some code around from `CollegeAndLevel` to `AcademicsModule` now that it was being shared.

### DIFF
--- a/app/models/student_success/term_gpa.rb
+++ b/app/models/student_success/term_gpa.rb
@@ -1,31 +1,51 @@
 module StudentSuccess
   class TermGpa
+    include MyAcademics::AcademicsModule
 
     def initialize(opts={})
       @student_uid_param = opts[:user_id]
+      @active_careers = get_active_careers
     end
 
     def merge(data={})
-      data[:termGpa] = term_gpa
+      data[:termGpa] = get_term_gpas
     end
 
-    def term_gpa
+    def get_term_gpas
       response = CampusSolutions::StudentTermGpa.new(user_id: @student_uid_param).get
       parse_term_gpa response
     end
 
     def parse_term_gpa(response)
-      current_term = Berkeley::Terms.fetch.current[:campus_solutions_id]
-      term_gpa = response.try(:[], :feed).try(:[], :ucAaTermData).try(:[], :ucAaTermGpa)
-      if term_gpa.present?
-        term_gpa.each do |term_obj|
-          if term_obj[:termId].to_i >= current_term.to_i || term_obj[:termGpaUnits].to_i == 0
-            term_obj[:termGpa] = nil
-          end
-          term_obj[:termName] = Berkeley::TermCodes.normalized_english term_obj[:termName]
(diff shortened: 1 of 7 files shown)


lyttam:
@davidlimcheng Very nice:  ` [].tap do |careers|`

app/models/student_success/term_gpa.rb
```diff
@@ -1,32 +1,52 @@
...
-      current_term = Berkeley::Terms.fetch.current[:campus_solutions_id]
-      term_gpa = response.try(:[], :feed).try(:[], :ucAaTermData).try(:[], :ucAaTermGpa)
-      if term_gpa.present?
-        term_gpa.each do |term_obj|
-          if term_obj[:termId].to_i >= current_term.to_i || term_obj[:termGpaUnits].to_i == 0
-            term_obj[:termGpa] = nil
-          end
-          term_obj[:termName] = Berkeley::TermCodes.normalized_english term_obj[:termName]
+      if (term_gpas = response.try(:[], :feed).try(:[], :ucAaTermData).try(:[], :ucAaTermGpa))
+        term_gpas.delete_if {|term| invalid_term_gpa? term}
```
lyttam: @davidlimcheng This is much easier to read. I like having the logic broken out into functions with descriptive names.

lyttam:
@davidlimcheng Good refactor.  I'll go ahead and merge it, but we should keep an eye out for this type of problem where we have two chunks of code, each wrapped in a promise, where the second one relies on a piece of data that the first one initializes (in this case, the active careers on the scope object) - because the way we're doing things makes it appear that the second promise waits for the first to resolve, but that's not the case.  Maybe the solution is to avoid the dependencies, or to structure things differently altogether?

davidlimcheng:
Thanks @lyttam!  Definitely agree that we need to be cognizant of this type of problem.  As @redconfetti pointed out though, it might not be such a bad thing.  When there are no dependencies between AJAX calls, it allows for asynchronous loading of multiple cads on a page, when we don't structure things in a rigidly synchronous way.  It's tough, because avoiding front-end dependencies usually means making inefficient calls in the back-end for data that will eventually be available in the front-end through a separate process (as is the case here).

I think the solution is definitely a case-by-case kind of thing.  What didn't help in this case was the tangled mess that is the userOverview page, which I do think we need to do something about structure-wise.  Hopefully, when the rate of change on that page slows down, it'll be stable enough for us to optimize with better practices.

redconfetti:
@lyttam - I noted to @davidlimcheng in person that I see a pattern where we use $scope.$on to setup a watcher on a scope variable that performs the necessary action when a change is detected. These could be used to continue processing once a dependency becomes available. Big picture I agree that merging data into one data feed for a card is a good idea.

It's easier for us to test using Rspec

PR #6639, opened 2017-05-08
SISRP-32440 - Update CSS to Include flow-root alternative to clearfix

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-32440

* Lots of good stuff in the JIRA
* The beginning of the end for the clearfix hack

### DIFF
--- a/src/assets/stylesheets/_useful.scss
+++ b/src/assets/stylesheets/_useful.scss
@@ -35,6 +35,14 @@
 .cc-clearfix-container::after {
   clear: both;
 }
+@supports (display: flow-root) {
+  .cc-clearfix {
+    display: flow-root;
+    &::after {
+      content: none;
+    }
+  }
+}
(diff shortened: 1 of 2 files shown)


lyttam:
@davidlimcheng very excited about this improvement!  I'm glad you heard about it and implemented it.  One question though:  won't `clear: both` still be applied to any element that has the `.cc-clearfix` class?  So we won't know whether `display: flow-root` is working.  Or, does adding `content: none` on the pseudo-element take care of that?

davidlimcheng:
Thanks @lyttam!  Correct, the `clear` property needs an element to be applied to, so if we remove the psuedo-elements, the `display` and `clear` properties will have nothing to attach to, and will be essentially removed.

lyttam:
@davidlimcheng Cool!  Merged.

PR #6641, opened 2017-05-09
SISRP-32544 - Delegate UX: remove page links, bConnected icons, profile menu item

lyttam:
https://jira.berkeley.edu/browse/SISRP-32544

Only users who are delegates and nothing else should have the special delegate experience.  If they are a delegate and also an advisor, for example, they still need to see stuff that is relevant to advisors.

Also did some refactoring of api/my/status feed.

### DIFF
--- a/app/models/user/api.rb
+++ b/app/models/user/api.rb
@@ -94,49 +94,98 @@ module User
     end
 
     def is_delegate_user?
+      return false if is_delegate_user_emulating_student?
       authentication_state.directly_authenticated? && !@delegate_students.nil? && @delegate_students.any?
     end
 
-    def has_academics_tab?(roles, has_instructor_history, has_student_history)
+    def is_delegate_user_emulating_student?
+      authentication_state.authenticated_as_delegate?
+    end
+
+    def super_user?
+      authentication_state.policy.can_administrate? && !is_delegate_user_emulating_student?
+    end
+
+    def viewer?(current_user_policy)
+      return false if is_delegate_user_emulating_student?
+      current_user_policy.can_view_as?
+    end
+
+    def can_act_on_finances?
+      return authentication_state.directly_authenticated? unless is_delegate_user_emulating_student?
+      !!authentication_state.delegated_privileges[:financial]
+    end
+
+    def can_view_grades?(can_view_academics)
+      return !!authentication_state.delegated_privileges[:viewGrades] if is_delegate_user_emulating_student?
+      can_view_academics || !!@user_attributes[:roles][:advisor]
+    end
+
+    def has_dashboard_tab?
+      return false if is_delegate_user_emulating_student?
+      @user_attributes[:roles].values.any? || super_user?
+    end
+
+    def has_academics_tab?(has_instructor_history, has_student_history)
+      if is_delegate_user_emulating_student?
+        return !!authentication_state.delegated_privileges[:viewEnrollments] || !!authentication_state.delegated_privileges[:viewGrades]
+      end
+      roles = @user_attributes[:roles]
       roles[:student] || roles[:faculty] || roles[:applicant] || has_instructor_history || has_student_history
     end
 
-    def has_financials_tab?(roles, has_student_history)
+    def has_financials_tab?(has_student_history)
+      if is_delegate_user_emulating_student?
+        return !!authentication_state.delegated_privileges[:financial]
+      end
+      roles = @user_attributes[:roles]
       roles[:student] || roles[:exStudent] || roles[:applicant] || has_student_history
     end
 
-    def has_toolbox_tab?(policy, roles)
-      return false unless authentication_state.directly_authenticated? && authentication_state.user_auth.active?
-      policy.can_administrate? || authentication_state.real_user_auth.is_viewer? || is_delegate_user? || !!roles[:advisor]
+    def has_campus_tab?
+      @user_attributes[:roles].values.any? || super_user?
     end
 
-    def filter_user_api_for_delegator(feed)
-      view_as_privileges = authentication_state.delegated_privileges
-      feed[:delegateViewAsPrivileges] = view_as_privileges
-      # Delegate users get a pared-down UX.
-      feed[:hasDashboardTab] = false
-      feed[:showSisProfileUI] = false
-      # Delegate users do not have access to preferred name and similar sensitive data.
-      feed[:firstName] = feed[:givenFirstName]
-      feed[:fullName] = feed[:givenFullName]
-      feed[:preferredName] = feed[:givenFullName]
-      feed.delete :firstLoginAt
-      # Extraordinary privileges are set to false.
-      feed[:isDelegateUser] = false
-      feed[:isViewer] = false
-      feed[:isSuperuser] = false
-      # Filter based on delegation rights chosen by the student.
-      feed[:canViewGrades] = false unless view_as_privileges[:viewGrades]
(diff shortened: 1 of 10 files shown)


app/models/user/api.rb
```diff
@@ -94,49 +94,107 @@ def record_first_login
...
+    end
+
+    def can_view_grades?(can_view_academics)
+      return !!authentication_state.delegated_privileges[:viewGrades] if is_delegate_user_emulating_student?
+      can_view_academics || !!@user_attributes[:roles][:advisor]
+    end
+
+    def has_dashboard_tab?
+      return false if is_delegate_user_emulating_student?
+      @user_attributes[:roles].values.any? || super_user?
```
redconfetti: @lyttam - I almost want to suggest putting `@user_attributes[:roles].values.any?` in a `has_legitimate_role?` method.

redconfetti:
@raydavis - You may want to review this as well, given that it touches the the Status API settings that may affect LTI tools.

app/models/user/api.rb
```diff
@@ -145,50 +203,50 @@ def get_feed_internal
...
-        isLegacyStudent: @user_attributes[:isLegacyStudent],
-        isDelegateUser: is_delegate_user?,
-        showSisProfileUI: @user_attributes[:sisProfileVisible],
-        isDirectlyAuthenticated: directly_authenticated,
-        actingAsUid: acting_as_uid,
-        advisorActingAsUid: !directly_authenticated && authentication_state.original_advisor_user_id,
-        delegateActingAsUid: !directly_authenticated && authentication_state.original_delegate_user_id,
-        canSeeCSLinks: directly_authenticated || authentication_state.classic_viewing_as?,
-        canActOnFinances: directly_authenticated
+        uid: @uid
```
redconfetti: @lyttam - Thanks for the alphabetical order!

raydavis:
Thanks, @redconfetti , I gave it a quick look & will test it on the servers when it lands.

app/models/user/api.rb
lyttam: @redconfetti yeah, I was thinking about putting it in a method but I couldn't come up with a good name for it. Basically, the check is for determining if they have any role *besides* delegate, since the delegate flag exists outside of `roles`.  It's pretty fragile but naming it would help.  What do you think about `has_campus_role`?

redconfetti: @lyttam - Yeah, `has_campus_role?` sounds good!

davidlimcheng: +1

app/models/user/api.rb
```diff
@@ -94,49 +94,98 @@ def record_first_login
...
+    def is_delegate_user_emulating_student?
+      authentication_state.authenticated_as_delegate?
+    end
+
+    def super_user?
+      authentication_state.policy.can_administrate? && !is_delegate_user_emulating_student?
+    end
+
+    def viewer?(current_user_policy)
+      return false if is_delegate_user_emulating_student?
```
davidlimcheng: @lyttam, so are we defining a delegate emulating a student as not a viewer?  I can see this becoming confusing with current naming conventions - specifically the `AllowDelegateViewAs` mixin.  Perhaps we should change the name of that mixin?  I'm not sure if it's a necessity or not though, certainly not a blocker for merge.  Just a question!

app/models/user/api.rb
```diff
@@ -94,49 +94,98 @@ def record_first_login
...
-      feed[:isDelegateUser] = false
-      feed[:isViewer] = false
-      feed[:isSuperuser] = false
-      # Filter based on delegation rights chosen by the student.
-      feed[:canViewGrades] = false unless view_as_privileges[:viewGrades]
-      feed[:hasFinancialsTab] = false unless view_as_privileges[:financial]
-      feed[:hasAcademicsTab] = false unless view_as_privileges[:viewEnrollments] || view_as_privileges[:viewGrades]
-      feed[:canActOnFinances] = !!view_as_privileges[:financial]
-      feed
+      policy.can_administrate? || authentication_state.real_user_auth.is_viewer? || is_delegate_user? || !!@user_attributes[:roles][:advisor]
```
davidlimcheng: @lyttam This is more of a functional question, but is it assumed that all advisors should have view-as access, and thus, a toolbox tab?  Or is there a reason for them having a toolbox tab other than having view-as access?

app/models/user/api.rb
```diff
@@ -145,50 +194,49 @@ def get_feed_internal
...
+        isSuperuser: super_user?,
+        isViewer: viewer?(current_user_policy),
+        lastName: names[:last_name],
         officialBmailAddress: @user_attributes[:officialBmailAddress],
         primaryEmailAddress: @user_attributes[:primaryEmailAddress],
-        preferredName: self.preferred_name,
+        preferredName: names[:preferred_name],
         roles: roles,
-        uid: @uid,
+        showSisProfileUI: show_sis_profile_ui?,
```
davidlimcheng: @lyttam I see that this is pretty much mapping to the `profile` feature flag in `User::AggregatedAttributes`.  Again, this is not the right place for it, but I wonder if we can obsolete this flag.  This feed is definitely an overwhelming amount of information already.  Let me know if you agree, and I'll create a JIRA for this work.

davidlimcheng:
Really good stuff, @lyttam!  Couple of questions for you, but overall looks awesome.

app/models/user/api.rb
lyttam: @davidlimcheng sure, that would be great.  I think we can get rid of `@sis_profile_visible` in `AggregatedAttributes` and just use `is_cs_profile_feature_enabled`.

app/models/user/api.rb
lyttam: @davidlimcheng  Yes, it's weird but that's what the original logic did (`feed[:isViewer] = false`) and I didn't want to mess with it too much.  I'm guessing it's because we have two types of viewing - View As vs. a delegate or advisor emulating a student.  I'd support using different language to avoid confusion.  Maybe 'View As' for one and 'Emulate' for the other.  I opened a JIRA:  https://jira.berkeley.edu/browse/SISRP-33091

app/models/user/api.rb
lyttam: @davidlimcheng I don't know, but it would be good to find out the answer.  I left it in because it's how the code was working before.

app/models/user/api.rb
redconfetti: @davidlimcheng @lyttam - Yeah. I like the idea of 'View As' meaning total viewer ability, where-as the others are 'emulating'... and thus have some limitations in certain areas... like certain links disabled, or certain data not shown.

Long term I hope we can move towards building certain pages for these special users, such as the Student Overview for Advisors, so that we can remove the 'View As' session mode from Advisors and Delegates.

app/models/user/api.rb
davidlimcheng: Sounds good, created [SISRP-33106](https://jira.berkeley.edu/browse/SISRP-33106) to do so in the future.

app/models/user/api.rb
lyttam: @davidlimcheng @redconfetti I implemented the suggestion.  It will show up in the latest commit (I did this to avoid merge conflicts).

lyttam:
Ready to merge once Travis is appeased.

davidlimcheng:
Great stuff, @lyttam

PR #6642, opened 2017-05-09
SISRP-32561 - Investigate 'GCMT' CS IDs in Prod Error Logs

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-32561

* Cuts down on # of erroneous calls by not making API calls to `MyCommitteesController#member_photo` or `CalnetCrosswalk::ByCsId` if the EMPL ID given by CS starts with `GCMT`, which is a sign that the committee member is not Berkeley faculty.

Would appreciate it if @lyttam can take a look at this one, along w/ any other reviewers.

### DIFF
--- a/app/controllers/my_committees_controller.rb
+++ b/app/controllers/my_committees_controller.rb
@@ -19,9 +19,9 @@ class MyCommitteesController < ApplicationController
   end
 
   def member_photo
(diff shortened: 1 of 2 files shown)


lyttam:
@davidlimcheng a little late to the party, but looks good and thanks for adding the `nil` check to `member_photo`!

PR #6644, opened 2017-05-11
SISRP-32542 - FA Cost of Attendance Card not rendering for Summer Only students

redconfetti:
https://jira.berkeley.edu/browse/SISRP-32542


nkutub:
LGTM @redconfetti

PR #6651, opened 2017-05-16
SISRP-33192 - Adds graceful handling of empty Link API responses

lyttam:
https://jira.berkeley.edu/browse/SISRP-33187

This will prevent our proxies that fetch links from blowing up when the link API response is empty or otherwise bad.

### DIFF
--- a/app/models/campus_solutions/link.rb
+++ b/app/models/campus_solutions/link.rb
@@ -33,65 +33,76 @@ module CampusSolutions
     def get_url(url_id, placeholders = {})
       # get cached response
       link_resources = get
+      status = link_resources[:status]
 
       if url_id
         # Need this due to side-effect of the lift process in normalize_response.
         link_id = url_id.downcase.camelize(:lower).to_sym
-        entry = link_resources[:feed][:ucLinkResources][:links][link_id]
-
-        if entry.nil?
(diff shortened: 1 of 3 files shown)


lyttam:
@nkutub @redconfetti @davidlimcheng Should we log an ERROR when Link API response is bad, or possibly just a WARN?

davidlimcheng:
@lyttam If a link is "turned off" on the CS side, without any code changes on the CalCentral side, would that be considered a "bad" response?  For that particular scenario, I think WARN would be better.

I think ERROR is apt if the response status is not 200.  It seems like the Link API response is fluid, even in between releases, so logging an ERROR might be a bit misleading.

lyttam:
@davidlimcheng When a link is disabled in CS, we want to treat that as a normal response.  This PR is for dealing with a different situation where we get a totally empty response from the Link API (just the `<UC_LINK_RESOURCES />` node and nothing else).  In that case we still get a 200, so I guess it's up to us to decide how to deal with the response.  So, sounds like you think ERROR is an appropriate level.

redconfetti:
@lyttam - I do think that an ERROR logged is best, however with graceful failure (links disappearing from user interface), is the approach we should take.

PR #6657, opened 2017-05-19
SISRP-20513 - Haas Students - Enrollment Card

redconfetti:
https://jira.berkeley.edu/browse/SISRP-20513
https://jira.berkeley.edu/browse/SISRP-28580


src/assets/javascripts/angular/controllers/widgets/enrollment/enrollmentCardController.js
```diff
@@ -116,6 +160,7 @@ angular.module('calcentral.controllers').controller('EnrollmentCardController',
    * Determines the sections displayed for card by instruction type
    */
   var setSections = function(enrollmentInstruction) {
+    enrollmentInstruction.notificationTemplate = 'notifications';
```
davidlimcheng: @redconfetti - is `notificationTemplate` an already initialized property of `enrollmentInstruction`?  Or is it a new property that is meant to be initialized here?

src/assets/templates/widgets/enrollment/adjust.html
```diff
@@ -5,6 +5,16 @@
   data-title="section.title"
 ></div>
 
+<div data-ng-if="isInstructionType(enrollmentInstruction, ['haasFullTimeMba','haasEveningWeekendMba'])">
```
davidlimcheng: Very nice!

davidlimcheng:
Great stuff @redconfetti, I just had a question for you but other than that I couldn't find much of anything!

src/assets/javascripts/angular/controllers/widgets/enrollment/enrollmentCardController.js
redconfetti: It is not already initialized. It is at this point that it's initializing it with a default value, and modifying it if the card type is other than the default (i.e. if it's a haas card).

redconfetti: @davidlimcheng - It is not already initialized. It is at this point that it's initializing it with a default value, and modifying it if the card type is other than the default (i.e. if it's a haas card).

davidlimcheng: @redconfetti In that case, can we use the Lodash [set](https://lodash.com/docs/4.17.4#set) method (or any other method that seems applicable) to initialize this?  Similar to the reason why we use Ruby's `#try` method, I think that Javascript will throw an error if it sees that `notificationTemplate` isn't already an initialized property of `enrollmentInstruction`.

src/assets/templates/widgets/enrollment/notifications.html
```diff
@@ -29,4 +29,8 @@
       </li>
     </ul>
   </div>
+
+  <div data-ng-if="!enrollmentInstruction.enrollmentPeriod[0]">
```
lyttam: @redconfetti Consider using `_.first` here - it won't blow up if `enrollmentPeriod` is empty.  https://lodash.com/docs/4.17.4#head

src/assets/templates/widgets/enrollment/adjust.html
lyttam: +1

src/assets/templates/widgets/enrollment/notifications.html
redconfetti: @lyttam - I just tried this but I didn't get the results I expected. I don't think that AngularJS is able to evaluate the string we pass to ngIf with LoDash included.

src/assets/javascripts/angular/controllers/widgets/enrollment/enrollmentCardController.js
redconfetti: @davidlimcheng - The 'enrollmentInstruction' variable is an object, not an array. I don't think it will throw an error if I try to establish a 'notificationTemplate' value.

davidlimcheng: @redconfetti You're absolutely [right](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_Accessors#Dot_notation).

```
Dot notation

get = object.property;
object.property = set;
```
This is very good to know!  Thanks for the lesson.

davidlimcheng:
Good stuff, @redconfetti

PR #6660, opened 2017-05-19
SISRP-27378 - Refactors AdvisingResources and adds career param to URL

lyttam:
https://jira.berkeley.edu/browse/SISRP-27378

- Removes dependency on the UC_ADVISING_RESOURCES API since we're not using any of those links
- Moves the remaining calls to Link API from a model into a module
- Creates two separate functions, one for links on the advisor's dashboard and one for the Student Overview links, so that we no longer have to filter one big list to pick out the appropriate links
- Adds the ACAD_CAREER parameter to the student-specific URL for Service Indicators


app/controllers/concerns/advising_resources.rb
```diff
@@ -0,0 +1,72 @@
+module AdvisingResources
```
nkutub: @lyttam, this is not being cached like the old AdvisingResources was cached. will this create a performance issue?

lyttam: @nkutub Thanks for asking.  By testing the caching behavior directly, I verified that links are getting cached on the `CampusSolutions::Link` level - for the generic links, this benefits everyone.

For the student-specific links, it's still caching them with the EMPLID parameter attached to the URL - and each time an advisor does a student lookup, those links get replaced with the new student's links.  Not ideal whatsoever.  I think we should move the logic that attaches the EMPLID to the URL so that it happens after the link is cached, so that student-specific links are still cached in their generic form (and we pull from cache for each student, then tack on the EMPLID).  I could do this in a separate PR.

nkutub: ok I will approve with the idea that you are doing another JIRA to fix caching issue.

PR #6661, opened 2017-05-19
CLC-6662 CLC-6666 OEC Freshman & Sophomore Seminars take 2

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6662
https://jira.ets.berkeley.edu/jira/browse/CLC-6666


raydavis:
@paulkerschen , I guess you're the ideal victim so far as code review goes -- sorry about the length!

app/models/oec/department_mappings.rb
```diff
@@ -0,0 +1,68 @@
...
+    private
+
+    def all_mappings
+      self.class.fetch_from_cache @term_code do
+        mappings = Oec::CourseCode.all
+        mappings.concat create_virtural_department_mappings @term_code
+      end
+    end
+
+    def create_virtural_department_mappings(term_code)
```
paulkerschen: @raydavis Turpographic error here and 4 lines up.

raydavis: Gor darnit.

raydavis:
@paulkerschen , I'm starting to have a bad conscience for assuming that the seminars won't ever be cross-listed. What do you think? Doesn't seem like it would take much longer to move those CourseCode methods to in-memory versions.

paulkerschen:
@raydavis, it looks like seminar cross-listing is not recommended but contemplated:

> Not recommended. If a seminar is cross-listed, please work closely with the other department scheduler and FSS staff to insure consistency in scheduling and enrollment settings.

http://fss.berkeley.edu/facstaff/Scheduler_Chart.html

If it looks doable to you while you're in there, I suppose we might have cause to thank you later.

paulkerschen:
Let it be. Thanks @raydavis.

raydavis:
Looks like a wealth of testext breakages in Bamboo thanks to the local DB setup moving from in-memory to real-PostgreSQL. The easiest solution is probably to restrict them to RAILS_ENV=test -- I'll amend for that.

PR #6662, opened 2017-05-19
SISRP-32471 - On SPRING enrollment card choose classes link navigates only for summer schedule

redconfetti:
https://jira.berkeley.edu/browse/SISRP-32471

### DIFF
--- a/src/assets/templates/widgets/enrollment/decide.html
+++ b/src/assets/templates/widgets/enrollment/decide.html
@@ -7,7 +7,7 @@
 ></div>
 
 <div class="cc-enrollment-card-section-content" data-ng-if="section.show">
-  <div data-ng-if="api.user.profile.actAsOptions.canSeeCSLinks && !isInstructionType(enrollmentInstruction, ['law','haasEveningWeekendMba','haasFullTimeMba'])">
+  <div data-ng-if="api.user.profile.actAsOptions.canSeeCSLinks && !enrollmentInstruction.isGradeBaseAvailable && !isInstructionType(enrollmentInstruction, ['law','haasEveningWeekendMba','haasFullTimeMba'])">
     <div data-ng-if="enrollmentInstruction.isClassScheduleAvailable">
       <a data-cc-campus-solutions-link-directive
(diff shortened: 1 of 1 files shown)


lyttam:
@redconfetti looks good... I don't know what "grade base" means though.  You could wrap the logic in a boolean that has a more descriptive name if you wish, so that someone less familiar with enrollment can still understand what it's doing.

redconfetti:
@lyttam - Good note. I don't know what it means either, but it would help if we did and converted it to a value that makes more sense. I'll check with functionals and follow up.

redconfetti:
@lyttam - I think I see what it means in [SISRP-15504](https://jira.berkeley.edu/browse/SISRP-15504#comment-338444). The 'Add', 'Drop', 'Swap', and 'Options' links are supposed to disappear, and the 'Change Grading Option' link displays ([same link that was previously 'Options'](https://github.com/ets-berkeley-edu/calcentral/blob/77197e49/src/assets/templates/widgets/enrollment/adjust_enrolled.html#L50,L74)) once `isGradeBaseAvailable` becomes true. I think this link historically was going to be named 'Grade Base Update', but students didn't know what that meant.

redconfetti:
@lyttam - I really don't know what to do at this point. I'd like to clarify, but we would need to discuss further with functionals. I've started to document some details in [Confluence](https://confluence.ets.berkeley.edu/confluence/display/MYB/Class+Enrollment+Widget#ClassEnrollmentWidget-GradeBase)

lyttam:
@redconfetti Thanks for looking into it.  So, 'grade base available' means they have the ability to change grading options.  You could rename the boolean to something like 'canChangeGradingOptions', which would obscure the business logic behind it but I think that's appropriate for CalCentral (we shouldn't need to worry about what a 'grade base' is if we're not surfacing that language to students).

I don't want to hold up your PR any longer - let me know if you're going to make a change, and if not I'll just merge it.  It's up to you whether you think it's worth it to change the wording.

davidlimcheng:
@redconfetti @lyttam - Seems like it's good and Nader is about to cut the QA branch, so I'm going to go ahead and merge this.

PR #6667, opened 2017-05-23
CLC-6666 Do not run stub-crazed OEC tests against live Postgres DB

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6666

Fixes the 13 testext failures introduced by my earlier PR.


johncrossman:
Fixing 13 testext failures is no small feat.  Thanks, @raydavis

PR #6669, opened 2017-05-24
CLC-6666 Fix OEC cross-listings bug caught by testext

raydavis:
I missed this boneheaded refactoring error because my higher-level tests all stubbed the broken method. Thanks for the catch, Bamboo!


raydavis:
@nkutub , any objections to my making a QA PR for this after it's merged?

PR #6672, opened 2017-05-25
SISRP-27378 - moves url param replacement outside of cached model

lyttam:
https://jira.berkeley.edu/browse/SISRP-27378

This moves the URL parameter replacement up one layer from the caching, so that only the generic form of each link will be cached (and not the form that has particular URL params).

The way it was working before, a link with EMPLID param would be cached with "EMPLID=12345", and when that link was requested again for a new EMPLID, the item in the cache would be replaced by the new one.

### DIFF
--- a/app/models/campus_solutions/link.rb
+++ b/app/models/campus_solutions/link.rb
@@ -30,7 +30,7 @@ module CampusSolutions
       container
     end
 
-    def get_url(url_id, placeholders = {})
+    def get_url(url_id)
       # get cached response
       link_resources = get
       status = link_resources[:status]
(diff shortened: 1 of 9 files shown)


app/models/campus_solutions/link.rb
```diff
@@ -53,25 +53,15 @@ def get_url(url_id, placeholders = {})
...
 
-      placeholders.try(:each) do |k, v|
-        if v.nil?
-          logger.debug "Could not set placeholder for #{k} on link id #{url_id}"
-          link = nil
-          break
-        else
-          link[:url] = link[:url].gsub("{#{k}}", v)
-        end
-      end
```
redconfetti: @lyttam - I see you moved this logic to LinkFetcher#replace_url_params. Is that supposed to be the new interface for getting a link that includes parameters that need to be replaced with the Student ID, career code, etc?

The CampusSolutions::Link model only performs cacheing of the link template table, and the ability to query for the template to a specific link?

redconfetti: Anne verbally confirmed for me that yes, the CampusSolutions::Link model is now like a globally cached proxy that provides the raw data. The LinkFetcher class is more like the service/utility class that is used to obtain links, or even obtain the link template and replace the parameters at will as needed.

davidlimcheng:
Looks great, @lyttam

PR #6673, opened 2017-05-25
SISRP-31074 - Transfer Units in Student Success Card show TC adjust for continuing students

redconfetti:
https://jira.berkeley.edu/browse/SISRP-31074

### DIFF
--- a/src/assets/javascripts/angular/controllers/pages/academicSummaryController.js
+++ b/src/assets/javascripts/angular/controllers/pages/academicSummaryController.js
@@ -32,7 +32,8 @@ angular.module('calcentral.controllers').controller('AcademicSummaryController',
   var parseGpaUnits = function() {
     // Testing units are lumped in with Transfer Units on the academic summary
     if ($scope.gpaUnits && !$scope.gpaUnits.errored) {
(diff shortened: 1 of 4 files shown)


davidlimcheng:
Good stuff, and thanks for digging deep to find a workaround, @redconfetti.  Two questions:

1. I see in the JIRA that the Academic Summary is also showing an incorrect value.  Were you going to put the same workaround into that card as well?
2. I also see in the JIRA that this is meant to be a temporary workaround until the API starts to return the correct value.  If that is the case, do you mind creating a JIRA for this future work so that it doesn't fall through the cracks?

redconfetti:
@davidlimcheng

1. This update does address the Academic Summary card. That is why it modifies `academicSummaryController.js`, using the shared `academicsService.totalTransferUnits` function to safely convert either testing units or transfer unit values to `0` if the value is null.
2. There is already JIRA for future work to undo this work-around in [SISRP-33519](https://jira.berkeley.edu/browse/SISRP-33519)

davidlimcheng:
Ah!  I should have realized that @redconfetti - looks mergarific to me.

PR #6678, opened 2017-05-26
SISRP-31677 - Updates logic to allow link visibility for Advisors viewing as Students

redconfetti:
https://jira.berkeley.edu/browse/SISRP-31677

The CalCentral back-end already ensures that Students are able to get the data from the My Academics feed, and that Advisors are able to obtain the Advisor Student Academics feed (for Student Overview page).

Because the back-end ensures that the user is gaining access ONLY to the link they should see, this updates the logic to only show the link that is available, with Advisor link presence suppressing the Student link from displaying. This closes the gap of an Advisor viewing-as a student, but not seeing the Advisor link.

### DIFF
--- a/src/assets/templates/widgets/transfer_credit.html
+++ b/src/assets/templates/widgets/transfer_credit.html
@@ -74,7 +74,7 @@
             </td>
           </tr>
         </tbody>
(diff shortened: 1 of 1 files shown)


davidlimcheng:
LGTM @redconfetti

PR #6681, opened 2017-05-31
SISRP-33567 - Add Classic Home Thumbnail asset to CalCentral media library

redconfetti:
https://jira.berkeley.edu/browse/SISRP-33567

@nkutub @lyttam @davidlimcheng - If we continue to receive requests to host certain images or large file assets, we should use [Github Large File storage (LFS)](https://git-lfs.github.com/). Do you think we should perhaps begin with this case? This will keep the size of our repository down.

~~This file is a 52KB file.~~

Update: Chara further compressed the file down to 11 KB. I think that the Github LFS solution would affect our build/deployment processing, so that would need to be something we do in the future if needed.

We're being asked to host it for to meet the requirements of a new PeopleSoft design in [SISRP-32460](https://jira.berkeley.edu/browse/SISRP-32460)


redconfetti:
@nkutub vetoed this

PR #6692, opened 2017-06-05
CLC-6687 Do not verify Crosswalk endpoint if disabled

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6687

Crosswalk is now disabled on Junction, but our pre-release API verification script didn't get the news.


johncrossman:
I haven't seen this script in a while... Thx, @raydavis

PR #6693, opened 2017-06-05
SISRP-33364 - Status and Holds Remove inaccurate "not registered" message

nkutub:
https://jira.berkeley.edu/browse/SISRP-33364

* removed the misleading message with no registration


davidlimcheng:
Looks good @nkutub waiting on Travis

PR #6695, opened 2017-06-06
SISRP-33381 - Profile Card Default to Open

redconfetti:
https://jira.berkeley.edu/browse/SISRP-33381


src/assets/javascripts/angular/controllers/widgets/finaid/FinaidProfileController.js
```diff
@@ -20,6 +20,10 @@ angular.module('calcentral.controllers').controller('FinaidProfileController', f
         angular.extend($scope.finaidProfile, _.get(response, 'data.feed.status'));
         $scope.finaidProfileInfo.errored = _.get(response, 'data.errored');
         $scope.finaidProfileInfo.isLoading = false;
+        var firstCategory = _.get($scope, 'finaidProfile.categories[0]');
```
davidlimcheng: @redconfetti, can the same thing be accomplished without creating a pointer variable?  E.g. 
```
if ($scope.finaidProfile.categories[0]) {
  _.set(finaidProfile.categories[0], 'show', true)
}
```

redconfetti: @davidlimcheng - Yes that works. Thanks!

PR #6696, opened 2017-06-07
SISRP-33671 - Removes unused appointments calendar code

lyttam:
Part 1 of this effort:  looks like we got most of the way to implementing an appointments calendar for students, but never wound up using it.  This cleans up all the dead code.

https://jira.berkeley.edu/browse/SISRP-33671


redconfetti:
@lyttam - Looks good.

PR #6697, opened 2017-06-08
SISRP-32578 - Instructor My Academics > Enrollment, sort not working for DIS column

redconfetti:
https://jira.berkeley.edu/browse/SISRP-32578

* Fixes column sorting to the section number columns
* Adds column sorting to dates and locations on Section Overview

### DIFF
--- a/src/assets/templates/academics_classinfo.html
+++ b/src/assets/templates/academics_classinfo.html
@@ -247,8 +247,8 @@
               <th data-cc-sortable-column-directive="is_primary" data-column-heading="P/S"></th>
               <th data-cc-sortable-column-directive="section_label" data-column-heading="Section"></th>
               <th data-cc-sortable-column-directive="ccn" data-column-heading="Class Number"></th>
-              <th class="show-for-large-up">Day and Time</th>
-              <th class="show-for-large-up">Location</th>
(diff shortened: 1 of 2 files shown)


src/assets/templates/academics_classinfo.html
```diff
@@ -247,8 +247,8 @@ <h2 class="cc-left">Section Overview</h2>
               <th data-cc-sortable-column-directive="is_primary" data-column-heading="P/S"></th>
               <th data-cc-sortable-column-directive="section_label" data-column-heading="Section"></th>
               <th data-cc-sortable-column-directive="ccn" data-column-heading="Class Number"></th>
-              <th class="show-for-large-up">Day and Time</th>
-              <th class="show-for-large-up">Location</th>
+              <th data-cc-sortable-column-directive="dates.join(' ')" data-column-heading="Day and Time" class="show-for-large-up"></th>
+              <th data-cc-sortable-column-directive="locations.join(' ')" data-column-heading="Location" class="show-for-large-up"></th>
```
lyttam: @redconfetti It's trivial, but I don't think there's any reason to join with a space here.  You could use an empty string.

redconfetti: @lyttam - I've updated it.

PR #6702, opened 2017-06-15
SISRP-33544 - Revert v2 CS Link Hack

redconfetti:
https://jira.berkeley.edu/browse/SISRP-33544

Reverts changes applied in #6677 and #6699


redconfetti:
@nkutub - I merged this myself because other developers weren't present and the PS dev team is wishing to continue testing without the CC Link hack.

PR #6703, opened 2017-06-15
SISRP-33646 - 9.2: Replace Grad CNP Messaging

davidlimcheng:
Parent JIRA:  [SISRP-33646 - 9.2: Replace Grad CNP Messaging](https://jira.berkeley.edu/browse/SISRP-33646)
Subtask JIRAs, corresponds to commits:
[SISRP-33746 - Refactor Current Registration Status Logic to Back-End](https://jira.berkeley.edu/browse/SISRP-33746)
[SISRP-33747 - Replace Graduate Students' CNP Messaging](https://jira.berkeley.edu/browse/SISRP-33747)
[SISRP-34118 - Rspec Unit Tests for Refactored/Renewed CNP Messaging](https://jira.berkeley.edu/browse/SISRP-34118)

Giving this feature a much-needed facelift:  Took all of the front-end logic and put it in the back-end, obsoleting some factories/models that we were using.  Tried to separate it up into commits, but they still ended up pretty big.  Thanks in advanced for review!


redconfetti:
[In Progress]

I want to try to summarize some of the changes here, to make the interpretation of the PR easier.

* MyAcademics::Registrations logic was moved to MyRegistrations::Statuses
* HubEdos::StudentAttributes data was integrated into the new MyRegistrations::Statuses to reduce front-end merging complexity

davidlimcheng:
@redconfetti That's a bingo!

redconfetti:
image

app/models/my_registrations/statuses.rb
```diff
@@ -0,0 +1,162 @@
...
+        term_values[:positiveIndicators] = [].tap do |term_indicators|
+          # Each indicator has a "fromTerm" and a "toTerm", but UC Berkeley usage of the positive service indicator is
+          # term-specific, so these should always be the same.
+          pos_indicators.delete_if do |indicator|
+            if term_id.try(:to_i) == indicator.try(:[], 'fromTerm').try(:[], 'id').try(:to_i)
+              term_indicators << indicator
+              # Delete the indicator from our array so we don't have to process it again
+              true
+            end
+          end
```
redconfetti: @davidlimcheng - I like this pattern you introduce here where you use #tap on an empty array to return the same array that was modified in the code block. I would have initialized an empty array, and append to it based on the logic.

I would move the logic that returns the indicators that match the term_id into another method, thus simplifying the method. Something like `positive_term_indicators(term_id, positive_indicators)`

app/models/my_registrations/registrations_module.rb
```diff
@@ -0,0 +1,263 @@
...
+      else
+        summary = 'Not Enrolled in Classes'
+      end
+      summary
+    end
+
+    def set_regstatus_explanation_grad(term, summary)
+      case summary
+        when 'Fees Unpaid'
+          return regstatus_messages[:feesUnpaidGrad]
```
redconfetti: @davidlimcheng - If I tried to include this module in any other model, it would break. See my other comment concerning the #regstatus_messages method.

app/models/my_registrations/statuses.rb
```diff
@@ -0,0 +1,162 @@
...
+      {
+        notOfficiallyRegistered: find_message_by_number(100),
+        cnpNotificationUndergrad: find_message_by_number(101),
+        feesUnpaidGrad: find_message_by_number(102),
+        cnpWarningUndergrad: find_message_by_number(103),
+        cnpWarningGrad: find_message_by_number(104),
+        notEnrolledUndergrad: find_message_by_number(105),
+        notEnrolledGrad: find_message_by_number(106)
+      }
+    end
```
redconfetti: Methods in your module rely on this #regstatus_messages. I think you should move `#find_message_by_number` to `RegistrationsModule`, and move the logic for `#registration_messages` to CampusSolutions#EnrollmentVerificationMessages itself like so:
```
module CampusSolutions
  class EnrollmentVerificationMessages < GlobalCachedProxy

    def self.registration_messages
      reg_messages = CampusSolutions::EnrollmentVerificationMessages.new().get
      reg_messages.try(:[], :feed).try(:[], :root).try(:[], :getMessageCatDefn) || []
    end
  end
end
```

You could still make a method in the module that uses the `@reg_messages ||= CampusSolutions::EnrollmentVerificationMessages.registration_messages` to load and re-use the same hash, to keep it performant.

davidlimcheng: Very interesting, @redconfetti - I do agree that the logic to bypass `:feed` and `:root` should live in `EnrollmentVerificationMessages`, but I don't think a new `self` method should be created for that.  Reason being, that's the only relevant data in the API response, so it should just be baked into the initial `EnrollmentVerificationMessages#get` method, via a `process_response` method or something similar.  

I don't think this should be included in this PR though, since I'll also have to modify any other models or front-end code that touches upon `EnrollmentVerificationMessages`.

app/models/my_registrations/registrations_module.rb
davidlimcheng: @redconfetti, Would it break?  Or would it only break if this particular method is called?  In either case, moving this into its own `MyRegistrations` module should safeguard it from any misguided usages, but I do agree that it is bad practice to refer to a variable that doesn't exist within its own scope.  Your suggestions in your other comment should fix this, I'll create a JIRA for this work and schedule it to happen for the near-future.

app/models/my_registrations/registrations_module.rb
```diff
@@ -0,0 +1,229 @@
...
+    def set_regstatus_flags(term_registrations)
+      term_registrations.each do |term_id, term_value|
+        past_end_of_instruction = get_term_flag(term_value, :pastEndOfInstruction)
+        term_value[:showRegStatus] = term_includes_indicator?(term_value, '+S09') && !past_end_of_instruction
+      end
+    end
+
+    def set_summer_flags(term_registrations)
+      term_registrations.each do |term_id, term_value|
+        term_value[:isSummer] = term_value.try(:[], 'term').try(:[], 'name').include?('Summer')
```
lyttam: @davidlimcheng would it be better to use the `term_id` to determine if it's summer?  What if the term name changes, or contains 'summer' or 'SUMMER'?

app/models/my_registrations/statuses.rb
davidlimcheng: @redconfetti Makes sense!  Although I am unsure about the behavior of using `#delete_if` and passing the array in question around multiple methods.  I'll give it a shot and see if it works.

app/models/my_registrations/registrations_module.rb
davidlimcheng: @lyttam You're right, this part of the code is very brittle.  I also just saw that `Berkeley::Term#edo_id_is_summer?` exists, and is perfect to use in this scenario.  Will amend.

app/models/my_registrations/registrations_module.rb
```diff
@@ -0,0 +1,229 @@
...
+          explanation: set_cnp_explanation(has_r99, has_rop, past_financial_disbursement, undergrad),
+          popoverSummary: set_cnp_popover_summary(has_r99, has_rop, past_financial_disbursement)
+        }
+      end
+    end
+
+    def set_cnp_summary(has_r99, has_rop, past_financial_disbursement)
+      if has_r99
+        return 'You Will Not Be Canceled for Non-Payment'
+      elsif !has_r99 && has_rop
```
lyttam: @davidlimcheng technically, you don't need to include `!has_r99` after every `elsif` because it is already ruled out after the first `if`.  But maybe you want it here for readability.

app/models/my_registrations/registrations_module.rb
```diff
@@ -0,0 +1,229 @@
...
+      elsif !has_r99 && !has_rop && !past_financial_disbursement
+        return 'Cancel for Non-Payment Notification'
+      elsif !has_r99 && !has_rop && past_financial_disbursement
+        return 'Cancel for Non-Payment Warning'
+      end
+    end
+
+    def set_cnp_popover_summary(has_r99, has_rop, past_financial_disbursement)
+      if has_r99
+        return '<strong>Exception: </strong>Your enrollment is not subject to cancellation this semester.'
```
lyttam: @davidlimcheng I'm actually not a fan of putting view logic in the backend... or HTML tags.  I think this logic should live in the template.

app/models/my_registrations/registrations_module.rb
```diff
@@ -0,0 +1,229 @@
...
+
+    def set_cnp_explanation(has_r99, has_rop, past_financial_disbursement, undergrad)
+      if has_r99
+        'You have an exception from Cancellation for Non-Payment (CNP) for this term.  You will not be dropped from your classes for this term.
+         You remain financially responsible for all charges on your Student Account.  Please monitor your communications and tasks in CalCentral for updates.'
+      elsif !has_r99 && has_rop
+        'The deadline to pay for this term has been extended.  To maintain enrollment in your current class schedule and avoid an administrative withdrawal, please pay
+         at least 20% of your tuition and fees by August 30th.  Note that if you are dropped for non-payment you will be subject to the pro-rated fee schedule.
+         <br><br>
+         To learn more about the consequences of withdrawal please visit <a href="http://registrar.berkeley.edu/registration/cancellation-withdrawal/refunds-after-withdrawl">refunds after
```
lyttam: @davidlimcheng I definitely would rather see hard-coded links in a template than here. But it's your call.

app/models/my_registrations/registrations_module.rb
davidlimcheng: @lyttam Yeah, it was more for me to keep my sanity.  Do you think I should remove it?

davidlimcheng: I'll remove it - some of the other conditionals here could use some clean-up as well.

app/models/my_registrations/registrations_module.rb
lyttam: +1

app/models/my_registrations/registrations_module.rb
```diff
@@ -0,0 +1,263 @@
...
+      end
+    end
+
+    def show_cnp?(term)
+      summer = term.try(:[], :isSummer)
+      regstatus = term.try(:[], :regStatus).try(:[], :summary)
+      undergrad = term.try(:[], 'academicCareer').try(:[], 'code') == 'UGRD'
+      past_classes_start = get_term_flag(term, :pastClassesStart)
+
+      # Only consider showing CNP status for undergraduate non-summer terms in which the student is not already Officially Registered
```
lyttam: @davidlimcheng More of a personal preference, but I think this method would be easier to read if the comments were method-level, like so:
`# Only consider showing CNP status for undergraduate non-summer terms in which the student is not already Officially Registered.
# If a student is Not Enrolled and does not have CNP protection through R99 or ROP, do not show CNP warning as there are no classes to be dropped from.
# If a student is not Officially Registered but is protected from CNP via R99, show protected status regardless of where we are in the term.  Otherwise, show CNP status until CNP action is taken (start of classes)
def show_cnp?(term)
       summer = term.try(:[], :isSummer)
       regstatus = term.try(:[], :regStatus).try(:[], :summary)
       undergrad = term.try(:[], 'academicCareer').try(:[], 'code') == 'UGRD'
       past_classes_start = get_term_flag(term, :pastClassesStart)
...
end`

src/assets/javascripts/angular/controllers/widgets/academicsStatusHoldsBlocksController.js
```diff
@@ -6,31 +6,26 @@ var _ = require('lodash');
...
+  $scope.$watchGroup(['regStatus.registrations.length', 'api.user.profile.features.csHolds'], function(newValues) {
     var enabledSections = [];
 
-    if (newValues[0] || newValues[1]) {
+    if (newValues[0]) {
       enabledSections.push('Status');
     }
 
-    if (newValues[2]) {
+    if (newValues[1]) {
```
lyttam: @davidlimcheng Can `newValues` be empty, or contain only one element, or null... anything that would cause indexing to bomb?

src/assets/javascripts/angular/controllers/widgets/academicsStatusHoldsBlocksController.js
```diff
@@ -145,16 +86,14 @@ angular.module('calcentral.controllers').controller('AcademicsStatusHoldsBlocksC
     angular.merge($scope.residency, residency);
   };
 
-  $scope.showCNP = function(registration) {
-    return statusHoldsService.showCNP(registration);
-  };
+  $scope.cnpStatusIcon = statusHoldsService.cnpStatusIcon;
+  $scope.regStatusIcon = statusHoldsService.regStatusIcon;
```
lyttam: @davidlimcheng these two lines should be inside a function definition that gets called somewhere.  That way we are intentionally executing them, not letting them execute as this file is parsed.  The way it's written, the JS parser will go line by line reading the file and execute these lines immediately when it reads them (can lead to unexpected bugs down the line).

src/assets/javascripts/angular/controllers/widgets/statusController.js
```diff
@@ -172,6 +153,9 @@ angular.module('calcentral.controllers').controller('StatusController', function
     $scope.statusLoading = '';
   };
 
+  $scope.cnpStatusIcon = statusHoldsService.cnpStatusIcon;
+  $scope.regStatusIcon = statusHoldsService.regStatusIcon;
```
lyttam: @davidlimcheng as with my previous comment, I'd like to see these lines wrapped in a function definition.

lyttam:
@davidlimcheng awesome job sorting out all these exceedingly complicated rules and translating them into readable code!

src/assets/javascripts/angular/controllers/widgets/academicsStatusHoldsBlocksController.js
davidlimcheng: @lyttam ah, great catch.  So, would the following be a safer way to assign these functions?
```
$scope.cnpStatusIcon = function() { return statusHoldsService.cnpStatusIcon };
```

src/assets/javascripts/angular/controllers/widgets/academicsStatusHoldsBlocksController.js
davidlimcheng: @lyttam Since newValues is a callback of the array of watchExpressions, we should be safe here.  The expressions in the $watchGroup array should always be there - `regStatus.registrations.length` will always be an `int` or possibly `undefined`, and `api.user.profile.features.csHolds` will always be `true` or `false`.

app/models/my_registrations/registrations_module.rb
davidlimcheng: That does seem much cleaner - will amend.

src/assets/javascripts/angular/controllers/widgets/academicsStatusHoldsBlocksController.js
lyttam: @davidlimcheng I was thinking either of the following would be better.  The way it's written, those lines are going to get executed immediately when the parser hits them.  It's better if we have control over exactly when they are executed, either by placing them at the bottom of the file...

```
...
  var loadStatusInformation = function() {
    getCalResidency()
    .then(parseCalResidency)
    .then(getSlrDeeplink)
    .then(getRegistrations)
    .finally(function() {
      $scope.residency.isLoading = false;
      $scope.regStatus.isLoading = false;
    });
  };

  loadStatusInformation();
  $scope.cnpStatusIcon = statusHoldsService.cnpStatusIcon;
  $scope.regStatusIcon = statusHoldsService.regStatusIcon;
});
```

... or by wrapping them in a function, and calling the function at the moment we want them to be executed.

```
...
  var assignStatusIconGenerators = function() {
    $scope.cnpStatusIcon = statusHoldsService.cnpStatusIcon;
    $scope.regStatusIcon = statusHoldsService.regStatusIcon;
  };

  var loadStatusInformation = function() {
    getCalResidency()
    .then(parseCalResidency)
    .then(getSlrDeeplink)
    .then(getRegistrations)
    .finally(function() {
      $scope.residency.isLoading = false;
      $scope.regStatus.isLoading = false;
    });
  };

  loadStatusInformation();
  assignStatusIconGenerators();
});
```

app/models/my_registrations/statuses.rb
davidlimcheng: Created https://jira.berkeley.edu/browse/SISRP-34147 for this work.

davidlimcheng:
@redconfetti @lyttam - PR amended with recommendations.  Thanks for reviewing!

PR #6704, opened 2017-06-16
SISRP-33580 - Fixes bug in delegate-filtered semesters feed

lyttam:
https://jira.berkeley.edu/browse/SISRP-33580

For each semester in the feed, if the feed was being filtered for a delegate user then we were removing the semester `slug`.  We would then rely on the existence or non-existence of that key in the UI to determine whether to suppress information that a delegate shouldn't see.

Unfortunately this caused `merge_withdrawals` to blow up because it relies on the slug being present.  This fix sets a `filteredForDelegate` flag instead of removing the `slug`.

### DIFF
--- a/app/models/my_academics/semesters.rb
+++ b/app/models/my_academics/semesters.rb
@@ -33,7 +33,7 @@ module MyAcademics
       withdrawal_terms = withdrawal_data.map {|row| Berkeley::TermCodes.edo_id_to_code(row['term_id'])}
       (enrollment_terms.keys | transcript_terms.keys | withdrawal_terms).sort.reverse.map do |term_key|
         semester = semester_info term_key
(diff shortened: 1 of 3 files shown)


redconfetti:
@lyttam - I get the impression that this is allowing something to surface that would have been deleted under certain circumstances. Is it safe that this data is being sent to the front-end, even though the delegate doesn't see it due to the ngIf conditional statements? Or might it expose something that delegates shouldn't see, if they were to look at the JSON data?

lyttam:
@redconfetti Yes, we were deleting the semester `slug` from the delegate-filtered feed; now we're not.  It seems fine to me though since it's not sensitive data.  I believe the only reason we were deleting it was so that front-end logic could rely on its existence or non-existence.

PR #6707, opened 2017-06-19
SISRP-34100 - Links on Committees Card do not appear as links

nkutub:
https://jira.berkeley.edu/browse/SISRP-34100


lyttam:
@nkutub nasty copy-paste error, thanks for fixing it.

PR #6708, opened 2017-06-19
SISRP-31635 - Limits the late add/drop link to students with certain acad plans

lyttam:
https://jira.berkeley.edu/browse/SISRP-31635

We were showing the late add/drop link on the Summer term view of the Enrollment card to any and all students with a summer term.  We will now be limiting it to only the following types of students with a summer term:

- students with a summer visitor plan
- students with a coursework-only plan
- students with a GRAD career


davidlimcheng:
LGTM, @lyttam

PR #6709, opened 2017-06-20
SISRP-34193 - For UCBX student there shouldn't be CNP message

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-34193

* Takes care of the JIRA, and makes the process a bit more efficient as well by removing CNP message processing if we are not showing the registration status.

### DIFF
--- a/app/models/my_registrations/registrations_module.rb
+++ b/app/models/my_registrations/registrations_module.rb
@@ -51,8 +51,7 @@ module MyRegistrations
 
     def set_regstatus_flags(term_registrations)
       term_registrations.each do |term_id, term_value|
-        past_end_of_instruction = get_term_flag(term_value, :pastEndOfInstruction)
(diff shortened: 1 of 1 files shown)


app/models/my_registrations/registrations_module.rb
```diff
@@ -203,18 +206,25 @@ def set_cnp_explanation(r99, rop, past_financial_disbursement)
       end
     end
 
+    def show_regstatus?(term)
+      past_end_of_instruction = get_term_flag(term, :pastEndOfInstruction)
+      term_career = get_term_career(term)
+      term_includes_indicator?(term, '+S09') && !past_end_of_instruction && (term_career != 'UCBX')
+    end
```
redconfetti: @davidlimcheng - This is how I like my methods. Small little byte sized chunks that I can chew on.

PR #6710, opened 2017-06-20
SISRP-31823 - UX: Academic Summary - Non-Adjusted and Transfer Credit Total Switched

redconfetti:
https://jira.berkeley.edu/browse/SISRP-31823

The API that this template depends on is reflecting the wrong values for two nodes. The values should be swapped in the API, but we do not have time for that in GL 9.2

After this update the Academic Summary card will reflect the correct values. The [Transfer Credit card template](https://github.com/ets-berkeley-edu/calcentral/blob/e2ca0a6a/src/assets/templates/widgets/transfer_credit.html#L27,L28) already reflects this swap. 

It is noted in [SISRP-34230](https://jira.berkeley.edu/browse/SISRP-34230) that both of these templates should be updated once the API is fixed to reflect the proper values in the right nodes in [SISRP-34229](https://jira.berkeley.edu/browse/SISRP-34229).


davidlimcheng:
LGTM @redconfetti.  Thanks for creating JIRAs for the future work so that it doesn't fall through the crack.  I can imagine one of us being very confused in the future for that work.

PR #6711, opened 2017-06-20
SISRP-34223 - For Grad/LAW student not enrollment then CNP message should be Red Colored Text 'Not Enrolled in Classes'

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-34223

* Luckily, there was an existing bug that made this implementation very easy to do!

### DIFF
--- a/app/models/my_registrations/registrations_module.rb
+++ b/app/models/my_registrations/registrations_module.rb
@@ -90,7 +90,7 @@ module MyRegistrations
           end
         end
       else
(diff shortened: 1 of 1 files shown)


redconfetti:
@davidlimcheng - We could possible move to Ruby symbols instead of Strings if all the summaries begin with capital letters
```
:not_enrolled.to_s.gsub("_", ' ').titleize
```

PR #6714, opened 2017-06-23
CLC-6694 OEC: Exclude non-participating courses from room share checks

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6694

I'm impressed (in a queasy way) that the bug was legal Ruby syntax...


paulkerschen:
This bug was the subtlest of all God's creatures. Thanks @raydavis.

PR #6727, opened 2017-06-27
SISRP-31635 - adds 'grad' to enrollment roles

lyttam:
https://jira.berkeley.edu/browse/SISRP-31635

Students with the 'grad' role see the default version of the Enrollment card... but we still need to flag their enrollment as 'grad' so that we can show the Late Add/Drop link on the card (only shows for grad, summerVisitor, and courseworkOnly roles).


davidlimcheng:
LGTM

PR #6730, opened 2017-06-28
SISRP-33364 - Status and Holds Remove inaccurate 'not registered' message

nkutub:
* fix bug introduced by PR #6703 
* Status title was not being added under $watch

### DIFF
--- a/src/assets/javascripts/angular/controllers/pages/userOverviewController.js
+++ b/src/assets/javascripts/angular/controllers/pages/userOverviewController.js
@@ -63,7 +63,7 @@ angular.module('calcentral.controllers').controller('UserOverviewController', fu
     isLoading: true
   };
 
-  $scope.$watchGroup(['regStatus.registrations.length', 'api.user.profile.features.csHolds'], function(newValues) {
+  $scope.$watchGroup(['regStatus.registrations', 'api.user.profile.features.csHolds'], function(newValues) {
     var enabledSections = [];
 
     if (newValues[0]) {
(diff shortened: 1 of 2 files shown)


davidlimcheng:
@nkutub: Correct me if I'm wrong, but this fix means that instead of looking for a boolean from the length of the array, we are checking for the existence of an array, which will always be there (since it is declared in the controller itself):
```
  $scope.regStatus = {
    registrations: [],
    isLoading: true
  };
```
This will always resolve to `true`, and we will show the section no matter what.  As said in the JIRA, it's possible that we could not have any registrations AND not have any residency status, in which case we would want the title to only be "Holds" and not "Status and Holds."

So, how about instead of looking at the registrations array, we look at the existence of registrations OR residency status; something like this:
```
  $scope.$watchGroup([('regStatus.registrations.length' || 'residency.official.description'), 'api.user.profile.features.csHolds'], function(newValues) {
    var enabledSections = [];
```
Does that make sense?

redconfetti:
@davidlimcheng - I don't think we're looking for a boolean. The [$watchGroup method](https://docs.angularjs.org/api/ng/type/$rootScope.Scope#$watchGroup) watches for changes to the object in `regStatus.registrations`, which is an array.

I'm not sure how recursive the method is, but I interpreted this update as ensuring that the callback is executed not only when the number of items in the array changes, but also if one of those objects is modified.

davidlimcheng:
Ahh thanks @redconfetti.  In that case, I think we'd still want to `$watch` the residency status too, right? It's still entirely possible that a student's `regStatus.registrations` array will be blank, but they will have a residency status.  

What complicates this is that the condition to show the residency status is based on the function `$scope.showResidency`, which lives in the `academicsController`.  I think we need to bring that logic to the `academicsStatusAndHoldsController`, but definitely not in this PR.  With this realization, I think this is fine, and OK to merge, but can we create a JIRA for the work of bringing the Residency logic to the `academicStatusHoldsController`?

davidlimcheng:
Created https://jira.berkeley.edu/browse/SISRP-34493 for future work, going to go ahead and merge.

PR #6747, opened 2017-07-03
SISRP-34598 - DIS sort still not working in calcentral rosters

redconfetti:
https://jira.berkeley.edu/browse/SISRP-34598

Fixes sorting issue in Rosters page, List view. Similar to #6697


lyttam:
Looks good.

PR #6749, opened 2017-07-06
CLC-6707 Set home page of new sites to our preferred default

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6707

Aimed at a Junction-only emergency release to be deployed before July 15, when Instructure's change takes effect.


johncrossman:
Looks right to me, @raydavis  Another Canvas-workaround to add to the collection.

PR #6750, opened 2017-07-10
SISRP-26133 Upgrade JRuby / Torquebox / OpenSSL

lyttam:
This was all the work of @raydavis; I just cherry-picked his commit (https://github.com/ets-berkeley-edu/calcentral/commit/dbb265f1032f84b488c17db445001c3e8d1b01a9)

Ray's notes:
Torquebox 3.2.0 requires JDBC driver JARs to be directly in the JBoss load path.
Classpath conflicts occur if the drivers are only in the JRuby-wide lib directory.


Gemfile
```diff
@@ -63,7 +66,8 @@ gem 'jruby-activemq', '~> 5.13.0', git: 'https://github.com/ets-berkeley-edu/jru
 # To support SSL TLSv1.2.
 # jruby-openssl versions 0.9.8 through 0.9.16 trigger runaway memory consumption in CalCentral.
 # Track progress at https://github.com/jruby/jruby-openssl/issues/86 and SISRP-18781.
-gem 'jruby-openssl', '0.9.7'
+#gem 'jruby-openssl', '0.9.7'
```
davidlimcheng: @lyttam Renegade comment here

Gemfile
```diff
@@ -110,7 +114,6 @@ gem 'closure-compiler', '~> 1.1.11'
 # Oracle adapter
 # Purposely excluding this for test environments since folks have to install ojdbc6
 group :development, :testext, :production do
-  gem 'activerecord-oracle_enhanced-adapter', '~> 1.5.6'
```
davidlimcheng: @lyttam Is there a reason why we no longer need this gem?

raydavis:

Gemfile
raydavis: We never used it.

script/build-torquebox.sh
```diff
@@ -14,8 +14,8 @@ source .rvmrc
 export RAILS_ENV=${RAILS_ENV:-production}
 export LOGGER_STDOUT=only
 # Temporary workaround for a JRuby 1.7.4 + Java 1.7 + JIT/invokedynamic bug : CLC-2732
-export JRUBY_OPTS="-Xcompile.invokedynamic=false -Xcext.enabled=true -J-Xmx900m -J-Djruby.compile.mode=OFF"
-# export JRUBY_OPTS="-Xcext.enabled=true -J-Xmx900m"
```
davidlimcheng: @lyttam A quick Google search didn't really give me any answers on the state of C Extensions on Jruby 9000 - is there any particular reason we're removing this setting from our options?

Gemfile
lyttam: @davidlimcheng good eye; will remove. And that reminds me; we can switch to 0.9.18 now.  I'll amend the PR.

script/build-torquebox.sh
raydavis: https://github.com/jruby/jruby/wiki/DifferencesBetweenMriAndJruby

app/models/calnet_crosswalk/proxy.rb
```diff
@@ -3,7 +3,6 @@ class Proxy < BaseProxy
 
     include ClassLogger
     include Proxies::Mockable
-    include Proxies::Tls12
```
davidlimcheng: Does this mean that we can should be able to remove all explicit TLS v1.2 calls in our code?  Deleting stuff is always fun - should I create a follow-up JIRA to remove this mixin + its references?

Gemfile
```diff
@@ -63,7 +66,7 @@ gem 'jruby-activemq', '~> 5.13.0', git: 'https://github.com/ets-berkeley-edu/jru
 # To support SSL TLSv1.2.
 # jruby-openssl versions 0.9.8 through 0.9.16 trigger runaway memory consumption in CalCentral.
 # Track progress at https://github.com/jruby/jruby-openssl/issues/86 and SISRP-18781.
-gem 'jruby-openssl', '0.9.7'
+gem 'jruby-openssl', '0.9.20'
```
lyttam: @davidlimcheng Removed comment and changed jruby-openssl to latest version.

lyttam:
One green Travis!

app/models/calnet_crosswalk/proxy.rb
lyttam: @davidlimcheng good question... @raydavis, was this removed because we're no longer using  `RssProxy`?  If that's the case, I could remove it in this PR.

raydavis: Yes, between this change & RssProxy's removal, I don't see any more Tls12 usage.

raydavis:
@lyttam Travis started failing with your last amendment. It looks like Gemfile.lock didn't get updated to match Gemfile?

app/models/calnet_crosswalk/proxy.rb
lyttam: Digging deeper, it looks like {{RssProxy}} is used by {{EtsBlog::ReleaseNotes}}, which is part of the {{ServiceAlerts}} merged feed. @raydavis, can we get rid of {{EtsBlog::ReleaseNotes}}?  

If so, it looks like slightly more work than I'd like to add to this PR, so I'll create a separate Jira.

raydavis: Someone on your team (don't remember whether it was @nkutub , @davidlimcheng, or @redconfetti ) chatted with me last week about yanking out the EtsBlog code, so I assumed the removal was either already done or that you knew about it.

lyttam:
@raydavis ah, thanks.  I amended again with the updated Gemfile.lock.

davidlimcheng:
Great stuff @lyttam and @raydavis !  Let's give this baby a whirl.

app/models/calnet_crosswalk/proxy.rb
redconfetti: @raydavis - I recall that you made an update that made it so that CalCentral is authoritative for both release notes and service alerts, instead of ONLY service alerts, with [EtsBlog still coded as a backup for release notes](https://github.com/ets-berkeley-edu/calcentral/blob/master/app/models/service_alerts/merged.rb).

We need to make the call to either update this code to rely on a hosted SIS RSS feed, or if we should just continue to post a notice with a link to the actual release notes.

I've created a JIRA for this in [SISRP-34730](https://jira.berkeley.edu/browse/SISRP-34730)

PR #6752, opened 2017-07-12
SISRP-34230 - Revert Non-Adjusted and Total Transfer Credit Unit count after API Swap

redconfetti:
https://jira.berkeley.edu/browse/SISRP-34230

We ended up not reverting the swap, because the swap was actually correct. We instead ensured that the keys used are more correct, specifically the 'unitsNonAdjusted' being renamed as 'unitsAdjusted'... because that's actually what it represents.

### DIFF
--- a/app/models/my_academics/transfer_credit.rb
+++ b/app/models/my_academics/transfer_credit.rb
@@ -11,9 +11,9 @@ module MyAcademics
       response = CampusSolutions::TransferCredit.new(user_id: @uid).get
       response = response.try(:[], :feed).try(:[], :root).try(:[], :ucTransferCredits).try(:[], :transferCredit)
       if (credit = response.try(:[], :ucTransferCrseSch))
-        adjusted_units = credit[:unitsAdjust].to_f
-        nonadjusted_units = credit[:unitsNonAdjusted].to_f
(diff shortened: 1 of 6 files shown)


app/models/my_academics/transfer_credit.rb
```diff
@@ -11,8 +11,8 @@ def transfer_credit
       response = CampusSolutions::TransferCredit.new(user_id: @uid).get
       response = response.try(:[], :feed).try(:[], :root).try(:[], :ucTransferCredits).try(:[], :transferCredit)
       if (credit = response.try(:[], :ucTransferCrseSch))
-        adjusted_units = credit[:unitsAdjust].to_f
-        nonadjusted_units = credit[:unitsNonAdjusted].to_f
+        adjusted_units = credit[:tcAdjust].to_f
+        nonadjusted_units = credit[:unitsAdjusted].to_f
```
lyttam: @redconfetti should `nonadjusted_units` be changed to `adjusted_units`, since it represents units adjusted?  And should the variable now called `adjusted_units` be changed to something else?

redconfetti: @lyttam - You're right. This area is still unclear and confusing.

PR #6754, opened 2017-07-12
SISRP-32020 - 9.3 - Semesters CampusOracle Clean-up

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-32020

* Removes the "transcript" mechanism for classes with grades (more info in JIRA)
* Removes CampusOracle "Transcripts" query, as well as "Additional Credits" query.  This data will be obtained using EDODB, Hub, and Transfer Credit API.
* Updates (hopefully) all functionality dependent on grading data in the Semesters feed.


redconfetti:
@davidlimcheng - Do you add your PRs to the comments in your JIRAs? I didn't notice this one.

src/assets/javascripts/angular/controllers/widgets/academicsFinalGradesController.js
```diff
@@ -2,25 +2,33 @@
...
+    _.forEach(selectedCourses, function(course) {
+      var sections = _.get(course, 'sections');
+
+      _.forEach(sections, function(section) {
+        if (section.is_primary_section && section.grading.grade) {
+          hasGrades = true;
+          // Lodash uses 'return false' to break from a loop.
+          return false;
+        }
+      });
```
redconfetti: @davidlimcheng - It might be possible to replace this with the [Lodash find()](https://lodash.com/docs/4.17.4#find) method, which returns an object if any item in the collection causes the callback function to return true.

```
  var hasGrades = function() {
    var selectedCourses = $scope.selectedCourses;

    var courseWithGrades = _.find(selectedCourses, function(course) {
      var sections = _.get(course, 'sections');
      var firstSectionWithGrades = _.find(sections, function(section) {
        return section.is_primary_section && section.grading.grad;
      });

      // If we found a section with grades, return course.
      if (firstSectionWithGrades !== undefined) {
        return true;
      }
    });
    return (courseWithGrades !== undefined);
  };
```

Just a suggestion, not a requirement.

src/assets/javascripts/angular/controllers/widgets/academicsFinalGradesController.js
```diff
@@ -2,25 +2,33 @@
...
-        break;
+      // If we found a grade, break from the loop.
+      if (hasGrades) {
+        return false;
       }
-    }
-    return response;
+    });
+    return hasGrades;
   };
```
redconfetti: @davidlimcheng - It might be possible to replace this with the [Lodash find()](https://lodash.com/docs/4.17.4#find) method, which returns an object if any item in the collection causes the callback function to return true.

```
  var hasGrades = function() {
    var selectedCourses = $scope.selectedCourses;

    var courseWithGrades = _.find(selectedCourses, function(course) {
      var sections = _.get(course, 'sections');
      var firstSectionWithGrades = _.find(sections, function(section) {
        return section.is_primary_section && section.grading.grad;
      });
      return (firstSectionWithGrades !== undefined);
    });
    return (courseWithGrades !== undefined);
  };
```

Just a suggestion, not a requirement.

src/assets/javascripts/angular/controllers/widgets/academicsGpaController.js
```diff
@@ -81,22 +82,27 @@ angular.module('calcentral.controllers').controller('AcademicsGpaController', fu
...
 
   var accumulateUnits = function(courses, accumulator) {
-    angular.forEach(courses, function(course) {
-      var gradingSource = course.transcript || course.estimatedTranscript;
-      angular.forEach(gradingSource, function(gradingData) {
-        if (gradingData.units) {
+    var gradingSource;
+    _.forEach(courses, function(course) {
+      _.forEach(course.sections, function(section) {
+        if (section.is_primary_section && section.grading.grade) {
```
redconfetti: @davidlimcheng - This is the third place in the code where it loops through courses, and loops through the sections within the course. I wonder if it would be possible to create a shared function in a service that does this, and simply runs a callback function on each section.

src/assets/javascripts/angular/controllers/widgets/academicsGpaController.js
```diff
@@ -81,22 +82,27 @@ angular.module('calcentral.controllers').controller('AcademicsGpaController', fu
...
+          gradingSource = section.grading;
+        } else if (section.is_primary_section && !section.grading.grade) {
+          gradingSource = section.estimatedGrading;
+        }
+        if (gradingSource.units) {
           var grade;
-          if (gradingData.grade) {
-            grade = findWeight(gradingData.grade);
+          if (_.isNaN(gradingSource.grade)) {
```
lyttam: @davidlimcheng did you mean `!_.isNan`?

lyttam:
@davidlimcheng lots of good work here, and I'm glad to see a bunch of code get deleted!

src/assets/javascripts/angular/controllers/widgets/academicsFinalGradesController.js
davidlimcheng: @redconfetti I actually considered using `_.find`, but having to deal with an object instead of a boolean, and checking for `undefined` made me a bit uneasy.  But good suggestion, I'm sure it'd work as well either way.

src/assets/javascripts/angular/controllers/widgets/academicsGpaController.js
davidlimcheng: Great suggestion, @redconfetti.  After looking into it though, I found that of the three occurrences:  
1 uses some logic in between the nested loops
1 uses a references to a variable outside of the loops' scope
1 does not use a reference to a variable outside of the loops' scope.

So, I would be using the service in only two places, and the service would have to be able to handle an optional input parameter.  IF we were using ES6, this wouldn't be a problem, since we'd be able to use the Ruby-esque pattern:
```function myFunc(a, b = 0) {
   // function body
}```

But since we are still on ES5, we'd have to do something like this:
```function myFunc(a,b) {
  b = b || 0;

  // b will be set either to b or to 0.
}```
(Taken from [a Stack Overflow thread](https://stackoverflow.com/questions/12797118/how-can-i-declare-optional-function-parameters-in-javascript)

This is totally a shameless plug for us to move to ES6.

src/assets/javascripts/angular/controllers/widgets/academicsGpaController.js
davidlimcheng: @lyttam Actually no, we want `gradingSource.grade` to be a number.  So, if it's NaN, we use `findWeight()` to convert it to a number.

davidlimcheng:
Thanks @redconfetti and @lyttam !  I think I've responded to all of your guys's comments.  Jason - I usually do comment the PR link on the JIRA, but sometimes I forget until I resolve the ticket.

PR #6755, opened 2017-07-13
SISRP-34776 Fix Bamboo deployment support for JRuby 2

raydavis:
https://jira.berkeley.edu/browse/SISRP-34776

I tested this script fix on Junction's test server, which is now running the latest Master branch.


lyttam:
Thanks for the fix!

PR #6756, opened 2017-07-13
SISRP-33547 - Disable Google Analytics tracking in Local, Dev, and QA environments

redconfetti:
https://jira.berkeley.edu/browse/SISRP-33547

### DIFF
--- a/src/assets/javascripts/angular/services/analyticsService.js
+++ b/src/assets/javascripts/angular/services/analyticsService.js
@@ -11,7 +11,9 @@ angular.module('calcentral.services').service('analyticsService', function(calce
    * More info on https://developers.google.com/analytics/devguides/collection/analyticsjs/events
    */
   var sendEvent = function(category, action, label) {
-    window.ga('send', 'event', category, action, label);
+    if (isProduction()) {
+      window.ga('send', 'event', category, action, label);
+    }
   };
 
   /**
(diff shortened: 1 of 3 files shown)


src/assets/javascripts/angular/services/analyticsService.js
```diff
@@ -11,6 +11,9 @@ angular.module('calcentral.services').service('analyticsService', function(calce
    * More info on https://developers.google.com/analytics/devguides/collection/analyticsjs/events
    */
   var sendEvent = function(category, action, label) {
+    if (!isProduction()) {
+      return;
+    }
     window.ga('send', 'event', category, action, label);
```
davidlimcheng: @redconfetti Suggestion, not a requirement, but how about wrapping the `send event` function inside of the conditional instead?

```
if (isProduction()) {
  window.ga('send', 'event', category, action, label);
}
```
This way, we wouldn't need an explicit `return` statement

src/assets/javascripts/angular/services/analyticsService.js
```diff
@@ -39,7 +41,9 @@ angular.module('calcentral.services').service('analyticsService', function(calce
    * e.g. /, /dashboard
    */
   var trackPageview = function() {
-    window.ga('send', 'pageview', $location.path());
+    if (!isProduction()) {
```
davidlimcheng: @redconfett, I think we want `isProduction` here?

davidlimcheng:
Looks good, @redconfetti

PR #6757, opened 2017-07-20
SISRP-26133 - Fixes rspec-rails dependency issue

lyttam:

When running `bundle install` after the jruby upgrade, we were seeing this message:

```
Downloading rspec-rails-3.1.0 revealed dependencies not in the API or the
lockfile (rspec-support (~> 3.1.0)).
Either installing with `--full-index` or running `bundle update rspec-rails`
should fix the problem.
```

The "problem" was that our Gemfile and Gemfile.lock were both specifying rspec-support ~> 3.1.2 (which allows anything 3.1.2 and above, short of 3.2).

I changed it to allow rspec-support-3.1.0 and above, short of 3.2 (see https://rubygems.org/gems/rspec-rails/versions/3.1.0).  This should allow us to run a clean `bundle install` without doing any extra steps.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -121,7 +121,7 @@ group :development, :test , :testext do
   gem 'rspec-core', '~> 3.1.7'
   gem 'rspec-rails', '~> 3.1.0'
   gem 'rspec-mocks', '~> 3.1.3'
-  gem 'rspec-support', '~> 3.1.2'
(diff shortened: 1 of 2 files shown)


lyttam:
@raydavis I would like this PR to have your blessing before getting merged.

raydavis:
Go in peace & upgrade no more.

PR #6758, opened 2017-07-20
SISRP-31448 - 9.3 - Node Packages Updates

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-31448

* Updates most packages to latest, usable version, and changes a bit of code due to method deprecation (`angular.merge`).
* `npm-shrinkwrap.json` - similar to `Gemfile.lock`.  Lots of good info in the JIRA.
* Sheer size of `npm-shrinkwrap.json` due to the number of child dependencies we rely on.  Here is a small snippet of what's in `npm-shrinkwrap.json` in lieu of a diff:
```
{
  "name": "calcentral",
  "version": "51.0.0",
  "dependencies": {
    "add-stream": {
      "version": "1.0.0",
      "from": "add-stream@>=1.0.0 <1.1.0",
      "resolved": "https://registry.npmjs.org/add-stream/-/add-stream-1.0.0.tgz"
    },
    "angular": {
      "version": "1.6.5",
      "from": "angular@>=1.6.5 <1.7.0",
      "resolved": "https://registry.npmjs.org/angular/-/angular-1.6.5.tgz"
    },
```
`"version"` locks in the version we are using, and `"from"` is sourced from the original instructions given in `package.json`.  Since we are using ~ (e.g. `"angular": "~1.6.5"`), we are specifying to install **at least** the version specified, **or the latest patched version.**  More info in the JIRA.

I'll be typing up a Confluence page as well, for instructions on how to keep on top of updated packages each sprint.


redconfetti:
@davidlimcheng noted in the JIRA:
> Angular@1.6.5 deprecates the angular.merge function, so I'll be replacing all of them with Lodash's _.merge function

redconfetti:
@lyttam @nkutub 

Now that this is merged, you should delete your `node_modules` folder, then do a refresh NPM install.

```
rm -rf node_modules
npm install
```

PR #6760, opened 2017-07-21
CLC-6714 Prefer "L & S" to "L&S" as an OEC department name

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6714

Whitespace: can't live with it, can't live without it.


paulkerschen:
It's not short for Lettersandscience, after all.

PR #6763, opened 2017-07-24
My Campus > Academic Planning - Remove DARS Link

nkutub:
https://jira.berkeley.edu/browse/SISRP-24196
* just delete a link.


nkutub:
need to remove gem.lock

PR #6770, opened 2017-07-26
Revert "SISRP-31448 - Update font-awesome in npm-shrinkwrap.json"

davidlimcheng:
Reverts ets-berkeley-edu/calcentral#6769

* 1 of 3 reversions.  3rd reversion being #6758 - but hopefully we won't need to do that one.  
* Needs to be reverted due to the bug outlined in https://jira.berkeley.edu/browse/SISRP-31448, that @raydavis and @pfarestveit unearthed.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -179,9 +179,9 @@
               "resolved": "https://registry.npmjs.org/is-binary-path/-/is-binary-path-1.0.1.tgz",
               "dependencies": {
                 "binary-extensions": {
-                  "version": "1.9.0",
+                  "version": "1.8.0",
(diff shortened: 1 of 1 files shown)


raydavis:
@davidlimcheng , now that we know the bug is triggered by `gulp build --env production` failure, I just tried out your PR branch. It failed just like master, so I think this PR won't do the job.

(I sanity-checked by checking out a version one or two commits before your big update, and confirming that `gulp build --env production` succeeded.)

davidlimcheng:
@raydavis Definitely, I'll be reverting the three PRs all the way back to #6758, and sticking these upgrades into the next sprint.

raydavis:
@nkutub , why did you merge this PR?

PR #6771, opened 2017-07-26
SISRP-35076 - Update Roster feed to display greatest value for Terms in Attendance

redconfetti:
https://jira.berkeley.edu/browse/SISRP-35076

### DIFF
--- a/app/models/edo_oracle/queries.rb
+++ b/app/models/edo_oracle/queries.rb
@@ -333,15 +333,19 @@ module EdoOracle
           TRIM(enroll."GRADING_BASIS_CODE") AS grading_basis,
           plan."ACADPLAN_DESCR" AS major,
           plan."STATUSINPLAN_STATUS_CODE",
-          stdgroup."STDNT_GROUP" AS terms_in_attendance_group
+          stdgroup."HIGHEST_STDNT_GROUP" AS terms_in_attendance_group
           #{email_col}
         FROM SISEDO.ENROLLMENTV00_VW enroll
         LEFT OUTER JOIN
           SISEDO.STUDENT_PLAN_CC_V00_VW plan ON enroll."STUDENT_ID" = plan."STUDENT_ID" AND
           plan."ACADPLAN_TYPE_CODE" IN ('CRT', 'HS', 'MAJ', 'SP', 'SS')
         LEFT OUTER JOIN
-          SISEDO.STUDENT_GROUPV00_VW stdgroup ON enroll."STUDENT_ID" = stdgroup."STUDENT_ID" AND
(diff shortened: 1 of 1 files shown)


app/models/edo_oracle/queries.rb
```diff
@@ -333,15 +333,19 @@ def self.get_rosters(ccns, term_id)
...
           #{email_col}
         FROM SISEDO.ENROLLMENTV00_VW enroll
         LEFT OUTER JOIN
           SISEDO.STUDENT_PLAN_CC_V00_VW plan ON enroll."STUDENT_ID" = plan."STUDENT_ID" AND
           plan."ACADPLAN_TYPE_CODE" IN ('CRT', 'HS', 'MAJ', 'SP', 'SS')
         LEFT OUTER JOIN
-          SISEDO.STUDENT_GROUPV00_VW stdgroup ON enroll."STUDENT_ID" = stdgroup."STUDENT_ID" AND
-          stdgroup."STDNT_GROUP" IN ('R1TA', 'R2TA', 'R3TA', 'R4TA', 'R5TA', 'R6TA', 'R7TA', 'R8TA')
+          (
+            SELECT s."STUDENT_ID", Max(s."STDNT_GROUP") AS "HIGHEST_STDNT_GROUP" FROM SISEDO.STUDENT_GROUPV00_VW s
```
davidlimcheng: @redconfetti Is that a renegade `s` hanging out at the end of the line?

redconfetti: @davidlimcheng - Nope. The 's' is an alias for the table name. That is why each column referenced begins with 's' such as `s."STUDENT_ID"` and `s."STDNT_GROUP"`. 

Because this is a self contained sub-query, I decided to use just 's' instead of 'stdgroup'. You can see on line 347 I actually alias the results of this sub-query as 'stdgroup' so that in the outer select statement `stdgroup."HIGHEST_STDNT_GROUP"` makes sense. It's including the highest student group value in the row for each result.

redconfetti: If you think it's advisable, I could change `s` to `stdgroup2` or `stdgroup_inner`, just to make it clearer yet distinguished

lyttam:
Looks good to me; will merge unless you are going to change that alias.

app/models/edo_oracle/queries.rb
davidlimcheng: AH I see.  My weak SQL just thought that there needed to be something before the alias.  Nah I think you're good, this LGTM.

redconfetti:
@davidlimcheng - The alias seems self evident to me. Please proceed with the merge.

PR #6773, opened 2017-07-27
Revert "SISRP-31448 - 9.3 - Node Packages Updates"

davidlimcheng:
Reverts ets-berkeley-edu/calcentral#6769

3rd of 3 reversions.
Needs to be reverted due to the bug outlined in https://jira.berkeley.edu/browse/SISRP-31448, that Ray & PaulF unearthed.


davidlimcheng:
One successful Travis

PR #6776, opened 2017-07-28
CLC-6719 Fix Travis-CI & Bamboo scripts

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6719

- Bamboo scripts will now fail if the front-end build is unsuccessful.
- Front-end build and test steps are more explicit.
- Work around Travis-CI's recent loss of Node.js installs.


johncrossman:
Thanks, @raydavis  I do like the DevOps stuff

PR #6777, opened 2017-07-31
SISRP-32435 - Update SOR to CS- Pre-Spring 2010 Data is Missing

nkutub:
https://jira.berkeley.edu/browse/SISRP-32435
* temporarily pulling all terms from local JSON file 
* bug fix to terms logic 
* removed IN terms where clause for EDODB queries 
* all controlled by hub_terms_enabled feature flag.


raydavis:
I'm reviewing this now, @nkutub

nkutub:
thanks. @raydavis

app/models/berkeley/terms.rb
```diff
@@ -50,6 +51,7 @@ class Terms
     attr_reader :campus
 
     def self.fetch(options = {})
+      oldest =
```
raydavis: Typo?

nkutub: yup, @raydavis , fixed

app/models/berkeley/term_codes.rb
```diff
@@ -34,8 +34,9 @@ def to_edo_id(term_yr, term_cd)
 
     def from_edo_id(edo_term_id)
```
raydavis: This is getting pretty opaque. Would you mind adding a comment describing how year "19xx" and "20xx" terms are actually mapped to a CS term ID?

app/models/berkeley/terms.rb
```diff
@@ -111,6 +113,21 @@ def init
       self
     end
 
+    def load_terms_from_file
+      cs_terms = []
+      filename = "/public/json/terms.json"
```
raydavis: This forces us into a code change and emergency release whenever terms data changes. If the logic is expected to hang around for more than one release, do you think it would be possible to configure the terms file path in settings, so that Sys Admins could handle it with a restart?

app/models/edo_oracle/queries.rb
```diff
@@ -52,6 +52,7 @@ class Queries < Connection
     #   - 'cred_cd' and 'pnp_flag' replaced by 'grading_basis'
     def self.get_enrolled_sections(person_id, terms = nil)
       # The push_pred hint below alerts Oracle to use indexes on SISEDO.API_COURSEV00_VW, aka crs.
+      in_term_where_clause = "enr.\"TERM_ID\" IN (#{terms_query_list terms}) AND " if Settings.features.hub_term_api
```
raydavis: Because the Hub Terms API isn't directly used in this logic, it took a while for me to understand that this line is trying to reduce the performance hit of queries which explicitly allow for terms stretching back to 1969. A comment might be a good idea.

raydavis:
All done with my pass, @nkutub !

app/models/berkeley/terms.rb
nkutub: @raydavis, does it make sense to add another level of configuration? Either way, we would need a PR and same level of effort to add a new file and re-point to it. It's just as easy to change the content of the file with PR and do a restart.

raydavis: Sorry, @nkutub , I was unclear. If you, for example, add a ``settings.yml`` line like so:

```
  # Allows override of terms data with a JSON file.
  term_definitions_json_file:
```

then your ``load_terms_from_file`` code would use that rather than the default. That JSON file could then be edited in place on the clusters, if needed.

redconfetti: @nkutub - I do think it would be good to develop a configurable fallback for emergencies.

nkutub:
@raydavis , All fixed

app/models/berkeley/terms.rb
```diff
@@ -111,6 +113,21 @@ def init
       self
     end
 
+    def load_terms_from_file
+      cs_terms = []
+      filename = Settings.terms.term_definitions_json_file
+      begin
+        file = File.open("#{Rails.root}#{filename}")
```
raydavis: @nkutub , on a deployed cluster, Rails.root files are only changeable by making a code change, building a new KNOB binary, and doing a release. I would recommend only configuring a non-null ``term_definitions_json_file`` when we want to _override_ the default. Or, in sample code form:

```
filename = Settings.terms.term_definitions_json_file ||
  "#{Rails.root}/public/json/terms.json"
```

Does that make sense?

raydavis:
The documentation makes it much clearer for us senile old folk -- thank you, @nkutub !

I think I was still unclear about how optional JSON overrides might work, though.

PR #6779, opened 2017-08-01
SISRP-32435 - Update SOR to CS- Pre-Spring 2010 Data is Missing

nkutub:
* revert upgrade to rake done in #6777


davidlimcheng:
Thx @nkutub

PR #6796, opened 2017-08-08
[QA] SISRP-34230 - Fix logic for hiding adjusted credits when same as total

redconfetti:
https://jira.berkeley.edu/browse/SISRP-34230


davidlimcheng:
@redconfetti Can you link the master PR?

PR #6799, opened 2017-08-09
SISRP-26133 - Reverts scripts back to pre-jruby-upgrade state

lyttam:

Fixes a JDBC dependency issue:  https://jira.berkeley.edu/browse/SISRP-35332

### DIFF
--- a/script/install-jars.rb
+++ b/script/install-jars.rb
@@ -1,0 +1,24 @@
+#!/usr/bin/env ruby
+
+# Extremely primitive JRuby dependency handler for JARs we can't redistribute publicly:
+# 1. Check for the JAR libraries we expect to be in place.
+# 2. If they're not installed yet, try to download from MAVEN_REPO to the JRuby's main "lib" directory.
+# The MAVEN_REPO environment variable must point to a private repository.
+# This confluence page has docs on keeping MAVEN_REPO updated: https://confluence.media.berkeley.edu/confluence/x/jYVcAg
+
+abort "ERROR: No MAVEN_REPO is defined" unless ENV['MAVEN_REPO']
+
+def knows_class?(classname)
+  begin
+    eval("Java::#{classname}.class")
+    true
+  rescue NameError
+    false
+  end
+end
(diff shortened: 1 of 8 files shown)


script/install-jars.rb
```diff
@@ -0,0 +1,24 @@
...
+  rescue NameError
+    false
+  end
+end
+
+puts "Installing Oracle jar from #{ENV['MAVEN_REPO']}"
+puts "  to #{ENV['MY_RUBY_HOME']}/lib"
+`wget "#{ENV['MAVEN_REPO']}/com/oracle/ojdbc6/11.2.0.3/ojdbc6-11.2.0.3.jar" -P "#{ENV['MY_RUBY_HOME']}/lib"`
+puts "Copying Oracle jar into ./lib"
+`cp -f "#{ENV['MY_RUBY_HOME']}/lib/ojdbc6-11.2.0.3.jar" "#{File.expand_path(File.dirname(__FILE__))}/../lib/"`
```
redconfetti: I see that this Ruby version of the install-jars script was [present before we upgraded JRuby](https://github.com/ets-berkeley-edu/calcentral/commit/2d0b20ba).

raydavis: OK, here's the full horrible story:

1. That MAVEN_REPO URL is long dead, but because the shell script aborts if it's undefined, it hung around the Bamboo build environment.
2. The "wget" line naturally fails, since it's looking at a dead URL...
3. ... but the shell script doesn't _check_ for failure, and so it marches on down to the next line.
4. Hey, what do you know? The .rvm JRuby lib directory had a JAR file named that, and so the copy worked.

SUGGESTION: Replace this by the {{install-jars.sh}} script which replaced it for the JRuby 2 upgrade.

lyttam: Thanks @raydavis.  We will abandon this script and use install-jars.sh with ojdbc7.jar

davidlimcheng:
Closing, we have decided to keep the current version of the script.

PR #6802, opened 2017-08-10
SISRP-35206 - Enrollment Verification page accessible to users without academic history

redconfetti:
https://jira.berkeley.edu/browse/SISRP-35206


davidlimcheng:
LGTM

PR #6803, opened 2017-08-14
CLC-6459 Map Administrative Proxy role to Lead TA in bCourses

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6459


paulkerschen:
Looking good to me, @raydavis; just waiting for Travis to do its thing.

PR #6808, opened 2017-08-28
[QA] SISRP-35757 - Upgrades JRuby to 9.1.5.0 (take 3)

lyttam:
https://jira.berkeley.edu/browse/SISRP-35757


Gemfile
```diff
@@ -119,7 +121,7 @@ group :development, :test , :testext do
   gem 'rspec-core', '~> 3.1.7'
   gem 'rspec-rails', '~> 3.1.0'
   gem 'rspec-mocks', '~> 3.1.3'
-  gem 'rspec-support', '~> 3.1.2'
+  gem 'rspec-support', '~> 3.1.0'
```
davidlimcheng: @lyttam Why do we have to downgrade this gem?

davidlimcheng:
Thanks @lyttam.  We're holding off on merging this until we get the green light from the Ops team, right?

lyttam:
@davidlimcheng it can be merged anytime, but we won't deploy until after Ops does their work.

Gemfile
lyttam: @davidlimcheng It was causing conflicts before, though that may have only been the case under the newer bundler version.

davidlimcheng:
Thanks @lyttam.  Since the testext build will fail if we merge before upgrading the "Docker" server, I'll hold off on merging this until we get our Ops team back.

davidlimcheng:
Looks like we're ready to merge!

Note:  @lyttam has added the configuration change to the JVM_OPTS that raises the threshold for OldGenHeap size that kicks off a full GC cycle.

PR #6809, opened 2017-08-31
SISRP-35815 - GSI's not able to view waitlist details in Cal Central 

redconfetti:
https://jira.berkeley.edu/browse/SISRP-35815

* Updates logic to rely on 'enroll_status' code, rather than the 'waitlist_position' value, as it appears that the waitlist position is not always present for waitlisted rosters
* Also includes update to docs to specify specific Bundler version when setting up workstation

### DIFF
--- a/src/assets/javascripts/angular/services/rosterService.js
+++ b/src/assets/javascripts/angular/services/rosterService.js
@@ -90,10 +90,10 @@ angular.module('calcentral.services').factory('rosterService', function($filter)
   var doesStudentMatchEnrollStatus = function(student, enrollStatus) {
     switch (enrollStatus) {
       case 'enrolled': {
-        return (!_.get(student, 'waitlist_position'));
+        return (_.get(student, 'enroll_status') === 'E');
       }
       case 'waitlisted': {
-        return (_.get(student, 'waitlist_position'));
+        return (_.get(student, 'enroll_status') === 'W');
       }
       default: {
         return true;


README.md
```diff
@@ -114,7 +114,7 @@ http://www.oracle.com/technetwork/java/javase/downloads/index.html
 1. Download the appropriate gems with [Bundler](http://gembundler.com/rails3.html):
 
     ```bash
-    gem install bundler
+    gem install bundler --version '1.11.2'
```
davidlimcheng: @redconfetti, it looks like we're using `1.15.1` in the `master` build directory:
```
[bamboo@docker MYB-CLCMVPQA-JOB1]$ bundle version
Bundler version 1.15.1
```
Maybe it'd be better to leave this vague in our documentation?
That being said, I know that our docs are in need of a major re-haul, especially if we upgrade JRuby in this next release.  Ultimately I think it's your call, but I'm still more in favor of leaving the versioning vague, at least until our stack has been stabilized after the upgrade.

redconfetti: Well, I think that when you install Bundler, it uses the latest version (1.15.4) by default. Our current gemset installs properly without errors under version 1.11.2, but I got errors when I tried to use the default.

I can omit this, but I thought it would be a very helpful suggestion. It's not a big deal. I think I'll just remove it then.

davidlimcheng:
Thanks @redconfetti, LGTM

PR #6812, opened 2017-09-05
SISRP-34657 - Marks faculty committee as active based on service date

lyttam:
https://jira.berkeley.edu/browse/SISRP-34657

Previously, a committee was presented to faculty as 'Completed' if the student had finished the necessary milestones.  With this change we will instead show a faculty committee as 'Completed' if the faculty member's service on the committee has ended.

### DIFF
--- a/app/models/my_committees/committees_module.rb
+++ b/app/models/my_committees/committees_module.rb
@@ -1,5 +1,6 @@
 module MyCommittees::CommitteesModule
   extend self
+  include ClassLogger
 
   COMMITTEE_TYPES = {
     QE: {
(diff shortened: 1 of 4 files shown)


davidlimcheng:
Travis errored out in the bundler step, I'm going to restart it to see if that was just a hiccup.

lyttam:
`bundle install` is failing. Somehow we're back to using bundler v1.15.4, which won't work with our Gemfile.lock.

lyttam:
Actually, it's not the bundler version that changed - it's the RubyGems version.

Successful build:
```$ bundle --version
Bundler version 1.15.4
$ gem --version
2.4.5
```$ bundle --version
Bundler version 1.15.4
$ gem --version
2.6.13```

**Comment by raydavis** [comment]
@lyttam , yep. If I'm reading the docs right, adding this to `.travis.yml` might fix our sad little PRs:

```
before_install:
  - gem update --system 2.4.5
```

lyttam:
@raydavis Yeah, I tried running `gem update --system 2.4.5` locally before running `bundle install` with bundler v1.15.2 and it worked.

I supposed we'll have to get to the root of the problem eventually though.  I ran `bundle update` with `--verbose` and it looks like RubyGems can't find or parse the .gemspec file in jruby-activemq.

lyttam:
I pulled in Ray's fix to .travis.yml and the build succeeded.  This is ready to merge.  I will keep looking for a long-term solution to the problem.

PR #6813, opened 2017-09-05
CLC-6739 OEC should include explicitly selected non-participating departments

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6739

These changes all affect the include-only-these-explicit-departments code path in the OEC UX.

The include-all-participating-departments logic remains buried in `Oec::Tasks initialize`:
```
      @departments_filter = if opts[:dept_codes]
                             {dept_code: opts[:dept_codes].split}
                           else
                             {include_in_oec: true}
                           end
```

and `Oec::SisImportTask run_internal`:
```
      Oec::DepartmentMappings.new(term_code: @term_code).by_dept_code(@departments_filter).each do |dept_code, course_codes|
```

where the `course_codes` values eventually make their way to `EdoOracle::Oec.depts_clause` calls.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,9 @@ matrix:
 env:
   - JRUBY_OPTS="--dev -J-Xmx900m" DISPLAY=:99.0 LOGGER_LEVEL=WARN TRAVIS_NODE_VERSION="4"
 
+before_install:
+  - gem update --system 2.4.5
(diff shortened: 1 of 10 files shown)


paulkerschen:
If it suits Travis it suits me. Thanks @raydavis.

PR #6816, opened 2017-09-08
SISRP-34658 - Surfaces member service dates on Student Committees card

lyttam:
https://jira.berkeley.edu/browse/SISRP-34658

Students will now be able to see the start and end date of each faculty member's service on that student's committees.

### DIFF
--- a/app/models/my_committees/faculty_committees.rb
+++ b/app/models/my_committees/faculty_committees.rb
@@ -17,7 +17,6 @@ module MyCommittees
       }
       feed = CampusSolutions::FacultyCommittees.new(user_id: @uid).get.try(:[], :feed)
       if feed && (cs_committees = feed[:ucSrFacultyCommittee].try(:[], :facultyCommittees))
-        # Service range will be parsed from nested member data with match on emplid
         @emplid = feed[:ucSrFacultyCommittee].try(:[], :emplid)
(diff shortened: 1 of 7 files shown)


app/models/my_committees/faculty_committees.rb
```diff
@@ -72,28 +76,23 @@ def sort_committees_by_svc(cs_committees)
...
 
-    def to_display(date)
-      if date === '2999-01-01'
-        'Present'
-      else
-        format_date date
+    def parse_cs_faculty_committee_svc(cs_committee)
+      current_user_member = cs_committee.try(:[], :committeeMembers).try(:find) do |member|
+        member.try(:[], :memberEmplid) == @emplid
       end
```
redconfetti: I assume that the `committeeMembers` element in the array is also an Array. If it's not found using `#try`, this do block is going to fail. Do we want this to return some alternative data if no members are present?

lyttam: @redconfetti If no members are present, the block shouldn't fail.  If `committeeMembers` is `[]` or `nil`, it will just result in `current_user_member` being `nil`, which is what we'd expect if there are no members on the committee.

redconfetti: You're right @lyttam 
```
nil.try(:find) do |member|
  member.try(:[], :memberEmplid) == @emplid
end
=> nil
```

For some reason I expected an error trying to pass a block to `nil`.

PR #6820, opened 2017-09-14
SISRP-34910 - Surfaces APR link to UCLS students on Ugrd Degree Progress

lyttam:
https://jira.berkeley.edu/browse/SISRP-34910

Advisors already have the link to run the APR.  We are adding a student-specific version of that link that will only show to students who are active in the UCLS program.


nkutub:
LGTM @lyttam

PR #6823, opened 2017-09-18
SISRP-35467 CalCentral displaying letter grading option for all P/NP classes, correct in BCS

nkutub:
https://jira.berkeley.edu/browse/SISRP-35467
* simple change for grading option to grading basis


redconfetti:
@nkutub - Looks good. Glad that TravisCI is passing now too.

PR #6830, opened 2017-09-20
SISRP-34542 - Refactors academic roles to handle multiple roles

lyttam:
https://jira.berkeley.edu/browse/SISRP-34542

This is a big refactor of academic roles, and how they are used by the enrollment card to determine which version of the card to display.  It will enable more flexibility in how these roles are assigned, so that we can assign multiple roles to a single student based on their career, program, plan, or any other CPP stack data.

### DIFF
--- a/app/models/my_academics/class_enrollments.rb
+++ b/app/models/my_academics/class_enrollments.rb
@@ -6,6 +6,8 @@ module MyAcademics
     include CampusSolutions::EnrollmentCardFeatureFlagged
     include LinkFetcher
 
+    ENROLLMENT_DECK_KEYS = ['fpf', 'haasFullTimeMba', 'haasEveningWeekendMba', 'haasExecMba', 'summerVisitor', 'law', 'concurrent']
+
     def get_feed_internal
       return {} unless is_feature_enabled && user_is_student?
       HashConverter.camelize({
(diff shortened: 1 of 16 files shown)


spec/models/hub_edos/my_academic_status_spec.rb
```diff
@@ -12,23 +12,47 @@
     shared_examples 'a translator that maps academic status to roles' do
       it 'translates roles' do
         roles = subject[:feed]['student']['roles']
+        puts roles.pretty_inspect
```
redconfetti: @lyttam - You left a puts

lyttam: @redconfetti thanks, will remove

app/models/my_academics/class_enrollments.rb
```diff
@@ -25,17 +27,30 @@ def grouped_student_plan_roles
...
         }
       }
-      active_plans.to_a.each do |plan|
-        role_code = plan[:enrollmentRole]
-        career_code = plan[:career][:code]
-        role_key = [role_code, career_code]
-        grouped_roles[:data][role_key] = { role: role_code, career_code: career_code, academic_plans: [] } if grouped_roles[:data][role_key].blank?
-        grouped_roles[:data][role_key][:academic_plans] << plan
-        grouped_roles[:metadata][:includes_fpf] = true if role_code == 'fpf'
+      academic_status.try(:[], :feed).try(:[], 'student').try(:[], 'academicStatuses').each do |status|
```
redconfetti: @lyttam - We can probably replace this #academic_status call and narrowing down by using [HubEdos::MyAcademicsStatus.get_statuses(@uid)](https://github.com/ets-berkeley-edu/calcentral/blob/master/app/models/hub_edos/my_academic_status.rb#L7)

redconfetti:
@lyttam - Looks good to me other than the recommended call to #get_statuses. If you're not interested in making that update, I'll merge.

app/models/my_academics/class_enrollments.rb
lyttam: @redconfetti true, but we also need the response from `HubEdos::MyAcademicStatus.new(@uid)` in `user_has_holds?` (line 147).  If I change the `academic_status` function to call `HubEdos::MyAcademicsStatus.get_statuses(@uid)` I won't be able to take advantage of the locally cached data in `@academic_status`.

It's kind of like six/one, half dozen/other.

lyttam:
@redconfetti I'll leave it as-is now, and for future we should consider caching the Academic Status data at a higher level and making it more available to other feeds.  That can be part of the caching review/refactor that we'll be doing.

app/models/my_academics/class_enrollments.rb
redconfetti: @lyttam - Aah okay. Well then maybe you can use [#parse_academic_statuses](https://github.com/ets-berkeley-edu/calcentral/blob/master/app/models/hub_edos/my_academic_status.rb#L25) instead, and perhaps even create a #parse_holds class method that can be used by any other models that consume this feed in the future.

PR #6831, opened 2017-09-20
SISRP-34493 - Change  Logic in Status and Holds Controller (+userOverview)

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-34493

* Finally makes `My Academics` + `User Overview` follow the same logic on whether or not we show Residency information.
* Removes crazy spinner behavior, and unifies all proxy calls within the `Status and Holds` card to one spinner.
* Most importantly, removes unnecessary `$watch` usage.  [Here](https://www.airpair.com/angularjs/posts/angularjs-performance-large-applications) are some thoughts on why `$watch` should be avoided:
> scope.$watch() has now been discussed on several occasions. In general, scope.$watch is indicative of bad architecture. There are very few cases when some combination of services and reference bindings can not achieve the same results with lower overhead.

### DIFF
--- a/src/assets/javascripts/angular/controllers/pages/userOverviewController.js
+++ b/src/assets/javascripts/angular/controllers/pages/userOverviewController.js
@@ -6,7 +6,7 @@ var _ = require('lodash');
 /**
  * Preview of user profile prior to viewing-as
  */
-angular.module('calcentral.controllers').controller('UserOverviewController', function(academicsService, adminService, advisingFactory, apiService, enrollmentVerificationFactory, linkService, statusHoldsService, $route, $routeParams, $scope) {
+angular.module('calcentral.controllers').controller('UserOverviewController', function(academicsService, academicStatusFactory, adminService, advisingFactory, apiService, enrollmentVerificationFactory, linkService, statusHoldsService, $route, $routeParams, $scope) {
   linkService.addCurrentRouteSettings($scope);
 
   $scope.expectedGradTermName = academicsService.expectedGradTermName;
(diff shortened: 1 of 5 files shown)


src/assets/javascripts/angular/controllers/pages/userOverviewController.js
```diff
@@ -167,6 +155,15 @@ angular.module('calcentral.controllers').controller('UserOverviewController', fu
     });
   };
 
+  var loadAcademicRoles = function() {
+    academicStatusFactory.getAcademicRoles({
+      uid: $routeParams.uid
+    }).then(function(response) {
+      $scope.academicStatus.roles = _.get(response, 'roles');
```
lyttam: @davidlimcheng Is `$scope.academicStatus.roles` used anywhere by this controller or the view(s) that it controls?  Doesn't appear that way - if it's not directly used, you don't need to be setting it in scope.  Maybe just pass it directly to `academicsService.showResidency()`.  

```
}).then(function(response) {
  $scope.showResidency = $scope.targetUser.roles.student && academicsService.showResidency(_.get(response, 'roles'));
```

lyttam:
+100 for making everything more logical and less annoying.

src/assets/javascripts/angular/controllers/pages/userOverviewController.js
davidlimcheng: Great suggestion!  Will amend.

davidlimcheng:
@lyttam Amended, thanks for the review!

PR #6832, opened 2017-09-20
SISRP-36097 - Removes milestone attempts and status from Student Committees card

lyttam:
https://jira.berkeley.edu/browse/SISRP-36097

Removes some data from the student-facing card, but leaves it as-is on the faculty card.  This required some moving around of code, since we want to have `CommitteesModule` contain only functionality that is shared between the faculty and student versions.


davidlimcheng:
LGTM

PR #6834, opened 2017-09-22
SISRP-35805 - P/NP Bug Fix - Calculate P/NP Total Correctly

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-35805

* Obtains cumulative P/NP units from EDODB, which holds values that take into account repeat classes, which are more accurate than the values from the Hub (which are not technically wrong).
* Had to get rid of a "total unit" sanity check for this, because the cumulative units will no longer align with the EDODB units.  According to Design team, this is an expected non-issue.

### DIFF
--- a/spec/models/my_academics/gpa_units_spec.rb
+++ b/spec/models/my_academics/gpa_units_spec.rb
@@ -45,7 +45,7 @@ describe 'MyAcademics::GpaUnits' do
     context 'empty status feed' do
       before { status_proxy.set_response(status: 200, body: '{}') }
       it 'reports empty' do
-        expect(feed[:gpaUnits][:empty]).to eq true
+        expect(feed[:gpaUnits][:hub_empty]).to eq true
       end
     end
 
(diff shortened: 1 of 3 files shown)


spec/models/my_academics/gpa_units_spec.rb
```diff
@@ -45,7 +45,7 @@
     context 'empty status feed' do
       before { status_proxy.set_response(status: 200, body: '{}') }
       it 'reports empty' do
-        expect(feed[:gpaUnits][:empty]).to eq true
+        expect(feed[:gpaUnits][:hub_empty]).to eq true
       end
```
redconfetti: @davidlimcheng - Why is there not a spec scenario for the :edo_empty value? Where is the front-end code that will be consuming this new :hub_empty or :edo_empty response?

davidlimcheng: @redconfetti, there was actually no existing front-end code that handled the `:empty` response.  I had not included a spec scenario for the `:edo_empty` value for this reason - and, due to time constraints as today is the last day of the sprint, I made the decision to move forward without adding additional functionality that we didn't have before.

PR #6836, opened 2017-09-28
SISRP-35822 - Document / Develop formal Tsung and VisualVM Approach

redconfetti:
https://jira.berkeley.edu/browse/SISRP-35822

This is just a script for developers to use to run Torquebox locally. It has no affect on any functionality in production.

### DIFF
--- a/script/local/start-torquebox.sh
+++ b/script/local/start-torquebox.sh
@@ -1,0 +1,32 @@
+#! /bin/bash
+
+# Script to start CalCentral Torquebox in a local / development environment.
+# WARNING: You must also set your local.yml configuration to include:
+#
+# cache:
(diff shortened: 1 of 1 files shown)


davidlimcheng:
Nice, @redconfetti.  This is very minor but what do you think about renaming the script to `start-local-torquebox.sh` to really hammer home the point that this is for local use only?

redconfetti:
@davidlimcheng - I'm sorry, but that seems like overkill to me. It's in a 'local' folder, which already makes typing the command to run it longer. If it was in `script`, I'd make it 'local-dev-start-torquebox.sh`.

davidlimcheng:
@redconfetti Fair enough!

PR #6837, opened 2017-09-30
SISRP-35360 - Update SISEDO queries to return term specific course title

redconfetti:
https://jira.berkeley.edu/browse/SISRP-35360

### DIFF
--- a/app/models/edo_oracle/queries.rb
+++ b/app/models/edo_oracle/queries.rb
@@ -3,7 +3,6 @@ module EdoOracle
     include ActiveRecordHelper
     include ClassLogger
 
-
     ABSENTIA_CODE = 'OGPFABSENT'.freeze
     FILING_FEE_CODE = 'BGNNFILING'.freeze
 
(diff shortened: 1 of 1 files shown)


app/models/edo_oracle/queries.rb
```diff
@@ -279,15 +302,19 @@ def self.terms
       SQL
     end
 
+    # TODO: Update this and dependencies to require term
     def self.get_cross_listed_course_title(course_code)
       result = safe_query <<-SQL
-        SELECT DISTINCT
```
lyttam: Good call on removing DISTINCT, as it's not needed here and is a fairly expensive SQL operation.

app/models/edo_oracle/queries.rb
```diff
@@ -49,6 +47,27 @@ class Queries < Connection
...
+          FROM
+            SISEDO.API_COURSEV01_MVW crs2, SISEDO.EXTENDED_TERM_MVW term2
+          WHERE
+            crs2."cms-version-independent-id" = crs."cms-version-independent-id" AND
+            crs2."displayName" = crs."displayName" AND
+            #{enroll_acad_career_filter}
+            term2.STRM = sec."term-id" AND
+            CAST(crs2."fromDate" AS DATE) <= term2.TERM_BEGIN_DT AND
+            CAST(crs2."toDate" AS DATE) >= term2.TERM_END_DT)
+            OR ( CAST(crs."updatedDate" AS DATE) = TO_DATE('1901-01-01', 'YYYY-MM-DD') )
```
lyttam: @redconfetti What's the significance of `'1901-01-01'`?

redconfetti: @lyttam - The new query approach to resolve this bug was provided by Marc Pechaitis. The course data is provided by the Course Management System (CMS), external to Campus Solutions.

Courses that have never had their name replaced don't have a fromDate or toDate (nor updatedDate). To ensure that we're getting a relevant record we're trying to match on a record that is within the term dates, or the one that has never been updated.

PR #6838, opened 2017-10-02
SISRP-34910 - Updates text and adds static links

lyttam:
https://jira.berkeley.edu/browse/SISRP-34910

### DIFF
--- a/src/assets/templates/widgets/academics/degree_progress_undergrad.html
+++ b/src/assets/templates/widgets/academics/degree_progress_undergrad.html
@@ -54,11 +54,28 @@
     </div>
     <div data-ng-if="!api.user.profile.roles.advisor" class="cc-degree-progress-footer">
       <span data-ng-if="isLettersAndScience && degreeProgress.undergraduate.links.academicProgressReport.url">
-        Run your
-        <a data-cc-campus-solutions-link-directive="degreeProgress.undergraduate.links.academicProgressReport"
-           data-cc-campus-solutions-link-directive-cc-page-name="currentPage.name"
-           data-cc-campus-solutions-link-directive-cc-page-url="currentPage.url"
-        ></a> to view your College of Letters and Science requirements. Contact your major advisor with questions about your major requirements.
+        <p>Run your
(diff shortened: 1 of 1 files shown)


src/assets/templates/widgets/academics/degree_progress_undergrad.html
```diff
@@ -54,11 +54,15 @@
...
-           data-cc-campus-solutions-link-directive-cc-page-url="currentPage.url"
-        ></a> to view your College of Letters and Science requirements. Contact your major advisor with questions about your major requirements.
+        <p>Run your
+          <a data-cc-campus-solutions-link-directive="degreeProgress.undergraduate.links.academicProgressReport"
+             data-cc-campus-solutions-link-directive-cc-page-name="currentPage.name"
+             data-cc-campus-solutions-link-directive-cc-page-url="currentPage.url"
+          ></a> to view your College of Letters and Science requirements. Major requirements may not be updated.
+        </p>
+        <p>
+          Please contact your <a href="https://ls.berkeley.edu/advising/about-undergraduate-advising-services" target="_blank" title="How to contact your L&S college advisor">college advisor</a> with college requirement questions or your <a href="https://ls.berkeley.edu/sites/default/files/umalist.pdf" target="_blank" title="List of major advisors">major advisor</a> with major requirement questions.
```
davidlimcheng: @lyttam Instead of using `target="_blank"`, can we use [outboundLinkDirective.js](https://github.com/ets-berkeley-edu/calcentral/blob/master/src/assets/javascripts/angular/directives/outboundLinkDirective.js) instead?  In addition to keeping consistent with other links on CalCentral, I believe it adds a message for accessibility as well.

davidlimcheng: Also, I don't think this is a big deal, but it spurred my curiosity. According to [this StackOverflow post](https://stackoverflow.com/questions/3705591/do-i-encode-ampersands-in-a-href), `&` should be changed to `&amp;` even within the `title` attribute, but, it's your call.

lyttam:
@davidlimcheng Thanks, I didn't know about that directive!  I made the changes you suggested.

davidlimcheng:
Thanks @lyttam!

PR #6839, opened 2017-10-02
SISRP-36337 - Academic Summary Page is showing a blank row for one of the term under Enrollment

nkutub:
https://jira.berkeley.edu/browse/SISRP-36337

### DIFF
--- a/src/assets/templates/widgets/academic_summary/academic_summary_semesters.html
+++ b/src/assets/templates/widgets/academic_summary/academic_summary_semesters.html
@@ -1,11 +1,9 @@
 <div class="cc-academic-summary-border-wrapper">
-  <div class="row">
-    <div class="small-6 columns">
-      <h4 class="cc-academic-summary-enrollment-header">
-        <a data-ng-if="semester.hasEnrollmentData && semester.slug" data-ng-bind="semester.name" data-ng-href="/academics/semester/{{semester.slug}}"></a>
-        <span data-ng-if="!semester.hasEnrollmentData || !semester.slug" data-ng-bind="semester.name"></span>
-      </h4>
-    </div>
+  <div class="row cc-academic-summary-content cc-clearfix" >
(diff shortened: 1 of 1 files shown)


src/assets/templates/widgets/academic_summary/academic_summary_semesters.html
```diff
@@ -1,10 +1,12 @@
 <div class="cc-academic-summary-border-wrapper">
-  <div class="row">
-    <div class="small-6 columns">
-      <h4 class="cc-academic-summary-enrollment-header">
-        <a data-ng-if="semester.hasEnrollmentData && semester.slug" data-ng-bind="semester.name" data-ng-href="/academics/semester/{{semester.slug}}"></a>
-        <span data-ng-if="!semester.hasEnrollmentData || !semester.slug" data-ng-bind="semester.name"></span>
-      </h4>
+  <div class="row cc-academic-summary-content cc-clearfix" >
+    <h3 data-ng-if="!semester.hasWithdrawalData && semester.hasEnrollmentData && !semester.filteredForDelegate && !isAdvisingStudentLookup"><a class="cc-left" data-ng-href="/academics/semester/{{semester.slug}}" data-ng-bind="semester.name"></a></h3>
```
davidlimcheng: @nkutub We don't need `filteredForDelegate` check here (that check is done on the `academics_semesters.html` card) and we don't need `isAdvisingStudentLookup` here, since this will never show up on that page.

The `h3` tags also need to be changed to `h4' tags to preserve the same formatting that was here before.

Lastly, I don't know if we want to even show the "You are currently not enrolled" message on this page, since students will be using this as an unofficial transcript. There's not much discussion in the JIRA, can you validate that's the behavior we want?

nkutub:
@davidlimcheng , you are correct on the logic, changes made.

lyttam:
@nkutub @davidlimcheng Ready to merge? I can't see David's comment or what changed.

PR #6840, opened 2017-10-02
SISRP-22589 - Remove Berkeley::SummerSubTerm and dependencies

redconfetti:
https://jira.berkeley.edu/browse/SISRP-22589

I got caught up trying to fix the migrations that referenced the model itself that I was removing (Berkeley::SummerSubTerm). After a discussion with @lyttam and @nkutub we concluded that Rails migrations don't need to be fixed across the board. If anything we should perhaps flatten them to setup only the schema we currently have to clean them up.

I left the code changes I made to fix the migrations that go back to when this table was made, but not further back than that.


.rspec
```diff
@@ -1,2 +1,3 @@
 --color
 --require spec_helper
+
```
davidlimcheng: @redconfetti Erroneous newline here?

davidlimcheng:
Whew, LGTM @redconfetti.  Just a question about a new line.

.rspec
redconfetti: @davidlimcheng - No. See [ThoughtBot - No Newline at End of File](https://robots.thoughtbot.com/no-newline-at-end-of-file) or [Stack Overflow - Why is it recommended to have empty line in the end of file](https://stackoverflow.com/questions/2287967/why-is-it-recommended-to-have-empty-line-in-the-end-of-file).

Our [.editorconfig](https://github.com/ets-berkeley-edu/calcentral/blob/066cba9a/.editorconfig#L7) specifies this behavior, which the plugin I use in Sublime Text enforces.

davidlimcheng: Thanks @redconfetti! Good read.

PR #6852, opened 2017-10-06
SISRP-36120 - Class enrollment card not loading for law student

redconfetti:
https://jira.berkeley.edu/browse/SISRP-36120

Sort method cannot sort on nil, so forced code to convert to string just in case.


lyttam:
@redconfetti good call.

PR #6854, opened 2017-10-09
CLC-6751 Tighten regexp for OEC semester migration

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6751


paulkerschen:
That's our alphabet.

PR #6859, opened 2017-10-11
NOJIRA - Removes intermittently failing unit test

lyttam:
This test breaks our Travis build occasionally.  I don't think it provides enough value to justify putting any more effort into fixing, so I'm removing it.


redconfetti:
I agree @lyttam

PR #6860, opened 2017-10-11
SISRP-36566 - Enrollment Card and Student Status Not Displaying In Card

redconfetti:
https://jira.berkeley.edu/browse/SISRP-36566

This is a proxy that provides parsed data. I think we want to make this fault tolerant, so that it fails gracefully, especially due to other cards relying on the data it provides.

### DIFF
--- a/app/models/hub_edos/my_academic_status.rb
+++ b/app/models/hub_edos/my_academic_status.rb
@@ -65,27 +65,35 @@ module HubEdos
     #  or as some combination of the two in custom logic.
     def assign_roles(status)
       assign_career_based_role(status.try(:[], 'studentCareer'))
-      status.try(:[], 'studentPlans').each do |plan|
-        if MyAcademics::AcademicsModule.active? plan
-          assign_plan_based_role plan
-          assign_program_based_role plan.try(:[], 'academicPlan').try(:[], 'academicProgram')
+      if plans = status.try(:[], 'studentPlans')
+        plans.each do |plan|
+          if MyAcademics::AcademicsModule.active? plan
+            assign_plan_based_role plan
+            assign_program_based_role plan.try(:[], 'academicPlan').try(:[], 'academicProgram')
+          end
         end
       end
     end
 
     def assign_career_based_role(studentCareer)
-      career_code = studentCareer.try(:[], 'academicCareer').try(:[], 'code')
-      studentCareer[:role] = get_academic_career_role_code(career_code)
+      if studentCareer
+        career_code = studentCareer.try(:[], 'academicCareer').try(:[], 'code')
+        studentCareer[:role] = get_academic_career_role_code(career_code)
+      end
     end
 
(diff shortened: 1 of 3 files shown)


app/models/hub_edos/my_academic_status.rb
```diff
@@ -65,39 +65,47 @@ def roles(statuses)
...
         end
       end
     end
 
     def assign_career_based_role(studentCareer)
-      career_code = studentCareer.try(:[], 'academicCareer').try(:[], 'code')
-      studentCareer[:role] = get_academic_career_role_code(career_code)
+      if career_code = studentCareer.try(:[], 'academicCareer').try(:[], 'code')
+        studentCareer[:role] = get_academic_career_role_code(career_code)
+      end
```
lyttam: @redconfetti Your goal is to handle the case where `studentCareer` is nil, I believe.  However, if `studentCareer` is present but studentCareer.try(:[], 'academicCareer').try(:[], 'code') is nil, line 80 will get bypassed.  

It makes more sense to me in that case to execute line 80 so that `studentCareer[:role]` gets set to `nil`.

These 3 functions could instead look like this:
```
def assign_career_based_role(studentCareer)
  if studentCareer
    career_code = studentCareer.try(:[], 'academicCareer').try(:[], 'code')
    studentCareer[:role] = get_academic_career_role_code(career_code)
  end
end
```

Or better yet, do the nil check somewhere upstream in the code.

redconfetti: @lyttam - If the `studentCareer` is present, but the `academicCareer` is nil, or the `code` is nil, then I do want line 80 to be bypassed. There is no role to merge in if that object isn't present. The function that processes it isn't setup to process a `nil` value.

lyttam: @redconfetti 
>Also if studentCareer is nil then the []= operator won't work on it because it's not a hash.

Right, so wrapping it with `if studentCareer` would prevent []= from getting called if `studentCareer` is nil.

>There is no role to merge in if that object isn't present. The function that processes it isn't setup to process a nil value.

I don't have a strong opinion, only a slight preference for setting `studentCareer[:role] = nil` if `studentCareer` is present but `studentCareer.try(:[], 'academicCareer').try(:[], 'code')` evaluates to nil. I think it's mostly a style preference - using nil to represent a missing value in this context feels more semantic.  If downstream code can't handle it then I would change the downstream code.

I'm fine with merging it if you'd rather leave it as-is.

redconfetti: @lyttam - I see what you're saying now. I was thinking that wouldn't work because the `get_academic_career_role_code` method fails if you pass it a nil value. I'll just make that method (and others) fault tolerant, so that it returns `nil` if it's passed `nil`.

I agree actually, that it would be better if the hash is present that it ends up having a `:role` key that has a nil, rather than no key at all.

PR #6861, opened 2017-10-11
CLC-6660 Do not include CalCentral-only data in Canvas roster export

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6660

Didn't try to get fancy with this one...


spec/models/rosters/common_spec.rb
```diff
@@ -79,59 +79,6 @@
   end
   subject { Rosters::Common.new(teacher_login_id, course_id: course_id) }
 
-  context 'when serving roster feed based content' do
-    before { allow_any_instance_of(Rosters::Common).to receive(:get_feed_internal).and_return(fake_feed) }
```
paulkerschen: @raydavis With this example gone, looks like the exhaustively defined `fake_feed` above can also go, and maybe some of its dependencies too.

spec/models/rosters/campus_spec.rb
```diff
@@ -407,6 +408,41 @@ def attributes
           expect(feed[:students][1][:majors][0]).to eq 'Break Science BA'
           expect(feed[:students][1][:terms_in_attendance]).to eq '6'
         end
+
+        it "returns rosters csv" do
```
paulkerschen: Double quotes without interpolation? I smell a python!

spec/models/rosters/common_spec.rb
raydavis: To paraphrase Pope Cerebus, "One less line to maintain is one less line to maintain."

PR #6862, opened 2017-10-12
SISRP-36573 - Fixes issue with missing font-awesome assets

lyttam:
https://jira.berkeley.edu/browse/SISRP-36573

"Fixes" by forcing our builds to use gulp-clean-css@3.4.0.  It's not clear to me what the breaking change to the underlying clean-css library was, but this seems like an acceptable workaround.


raydavis:
You may not think it's pretty, but I'll take it! Thanks for grinding through that list of updates.

Just waiting on Travis...

PR #6864, opened 2017-10-16
SISRP-36681 - Remove HotPlate

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-36681

Pretty straightforward, since nothing is using this code, it's pretty easy to remove.


davidlimcheng:
Missed some functionality and specs, will tweak the commit.  Thanks Travis!

davidlimcheng:
Got the thumbs up from Travis, PR should be ready for review.

config/torquebox.rb
```diff
@@ -13,24 +13,8 @@
   # CALCENTRAL-ONLY
   service JmsWorker
 
-  # warm up active users caches once a day
-  # CALCENTRAL-ONLY
-  job HotPlate do
-    cron '0 1 8 * * ? *'
```
davidlimcheng: Pretty cool, I didn't know Torquebox supported scheduled jobs:
http://torquebox.org/documentation/2.1.2/scheduled-jobs.html

redconfetti: Yeah. A pretty nice feature. Too bad we don't want to rely on this behemoth for that cool functionality.

redconfetti:
@davidlimcheng - Is User::CacheWarmer used by LiveUpdates still currently, so that's why you're leaving that in? It sounds like a HotPlate-only model.

Also, should we perhaps also remove User::Visit? I think that tracks when users log into the system, just so that the HotPlate feature can query the users that have recently visited, so it can "warm them up".

davidlimcheng:
@redconfetti Correct, User::CacheWarmer is still being used by LivedUpdates.  I haven't done research to see if it's being used beyond that, but if not, I'll rip it out in my next PR.

As for if we should remove `User::Visit`, that's a good question.  It is still being used in `UserApiController`, but I'm not really sure why.  I do see that we are logging some information about the user from `User::Visit`:
```
[2017-10-17 10:27:12] [DEBUG] [CalCentral] User::Visit using connection_id: 135798, connected = true
[2017-10-17 10:27:12] [DEBUG] [CalCentral] User::Visit current connection pool (#<ActiveRecord::ConnectionAdapters::ConnectionSpecification:0x52e7ba85>-postgresql_connection) count: 1; thread 135796
[2017-10-17 10:27:12] [DEBUG] [CalCentral] User::Visit current connection pool (#<ActiveRecord::ConnectionAdapters::ConnectionSpecification:0x26e0eaa2>-jdbc_connection) count: 1; thread 135796
[2017-10-17 10:27:12] [DEBUG] [CalCentral] User::Visit current connection pool (#<ActiveRecord::ConnectionAdapters::ConnectionSpecification:0x36fee6ba>-jdbc_connection) count: 1; thread 135796
```
And for that reason, I'd err more to the side of leaving it in for now, and revisiting it once we complete our log inventory/review.

redconfetti:
Yeah @davidlimcheng, that PostgreSQL based ActiveRecord model was added so that we could track when users log in, just so HotPlate could query users that have recently logged in and warm them up. See #1025

If there is no other code using that model, we should create a migration to drop that table and remove the model.

davidlimcheng:
@redconfetti Totally valid, and I think we should have a conversation with the sys admins about the usefulness of those particular records.  There are LiveUpdate dependencies for `User::Visit`, though, which I can deal with in my LiveUpdates removal PR, so would you be OK with merging this for now while we continue the conversation with the sys admins?

davidlimcheng:
Created [SISRP-36700](https://jira.berkeley.edu/browse/SISRP-36700) as a separate task to potentially remove `User::Visit`.

PR #6865, opened 2017-10-17
SISRP-35360 - Update SISEDO queries to return term specific course title

redconfetti:
https://jira.berkeley.edu/browse/SISRP-35360

Update 2 (See #6837 for update 1)

### DIFF
--- a/app/models/edo_oracle/queries.rb
+++ b/app/models/edo_oracle/queries.rb
@@ -48,21 +48,22 @@ module EdoOracle
     SQL
 
     def self.where_course_term_updated_date(with_career_filter = true)
-      enroll_acad_career_filter = with_career_filter ? 'term2.ACAD_CAREER = enr."ACAD_CAREER" AND' : ''
+      enroll_acad_career_filter = with_career_filter ? 'AND term2.ACAD_CAREER = enr."ACAD_CAREER"' : ''
       sql_clause = <<-SQL
-        AND (crs."updatedDate" = (
-          SELECT
-            MAX(crs2."updatedDate")
-          FROM
-            SISEDO.API_COURSEV01_MVW crs2, SISEDO.EXTENDED_TERM_MVW term2
-          WHERE
-            crs2."cms-version-independent-id" = crs."cms-version-independent-id" AND
-            crs2."displayName" = crs."displayName" AND
-            #{enroll_acad_career_filter}
-            term2.STRM = sec."term-id" AND
-            CAST(crs2."fromDate" AS DATE) <= term2.TERM_BEGIN_DT AND
-            CAST(crs2."toDate" AS DATE) >= term2.TERM_END_DT)
-            OR ( CAST(crs."updatedDate" AS DATE) = TO_DATE('1901-01-01', 'YYYY-MM-DD') )
+        AND crs."updatedDate" = (
+          SELECT MAX(crs2."updatedDate")
+          FROM SISEDO.API_COURSEV01_MVW crs2, SISEDO.EXTENDED_TERM_MVW term2
+          WHERE crs2."cms-version-independent-id" = crs."cms-version-independent-id"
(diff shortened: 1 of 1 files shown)


marcpechaitis:
@redconfetti per our Slack conversation - I recommend in this case that we add 'DISTINCT' to this SQL

redconfetti:
@davidlimcheng - This is ready for review.

app/models/edo_oracle/queries.rb
```diff
@@ -50,19 +50,24 @@ class Queries < Connection
...
+              crs2."displayName" = crs."displayName" AND
+              #{enroll_acad_career_filter}
+              term2.STRM = sec."term-id" AND (
+                (
+                  CAST(crs2."fromDate" AS DATE) <= term2.TERM_BEGIN_DT AND
+                  CAST(crs2."toDate" AS DATE) >= term2.TERM_END_DT
+                )
+                OR CAST(crs."updatedDate" AS DATE) = TO_DATE('1901-01-01', 'YYYY-MM-DD')
+              )
+          )
```
lyttam: @redconfetti Something doesn't look right about the alignment of these parentheses. Or is it just the way Github displays it?

lyttam:
@redconfetti Looks good; the only non-cosmetic change was adding DISTINCT to those three queries, right?

app/models/edo_oracle/queries.rb
redconfetti: Here are the parenthesis I'm trying to make clear:
* Lines 53 and 71 (not highlighted in green)
* Lines 54 and 70
* Lines 63 and 69
* Lines 64 and 67

You do bring up the question of whether or not the 54 - 71 grouping is really necessary. I'm just trying to implement the query the way @marcpechaitis provided it. What do you think Marc?

marcpechaitis: Here is how SQL Developer formats that section of the SQL statement (including a line before and after to help put in context:
```sql
AND ENR."STDNT_ENRL_STATUS_CODE" != 'D'
AND crs."updatedDate"             =
  (SELECT MAX(crs2."updatedDate")
  FROM SISEDO.API_COURSEV01_MVW crs2,
    SISEDO.EXTENDED_TERM_MVW term2
  WHERE crs2."cms-version-independent-id" = crs."cms-version-independent-id"
  AND crs2."displayName"                  = crs."displayName"
  AND term2.ACAD_CAREER                   = enr."ACAD_CAREER"
  AND TERM2.STRM                          = SEC."term-id"
  AND (( CAST(CRS2."fromDate" AS DATE)   <= TERM2.TERM_BEGIN_DT
  AND CAST(CRS2."toDate" AS      DATE)   >= TERM2.TERM_END_DT)
  OR CAST(crs."updatedDate" AS DATE)      = TO_DATE('1901-01-01', 'YYYY-MM-DD'))
  )
ORDER BY TERM_ID DESC,
```

to answer your question @redconfetti, yes line parens on lines 54 and 70 are necessary for the "updatedDate" subquery.

redconfetti:
@lyttam @marcpechaitis - Is the SQL clearer now with parenthesis applied?

marcpechaitis:
@redconfetti, would it be better to tighten up the standalone parens on lines 61 and 64, so they are together with the "fromDate" && "toDate" lines?  more readable?
cc: @lyttam

redconfetti:
@marcpechaitis - I think that those two lines being indented inside of the parenthesis makes it clearer what is happening. When the parenthesis are placed together with the other statements, they seem to easily get mistaken for being part of the `CAST` statements.

Even your "pretty" version has that going on where it's cognitively difficult to reason what's going on.

davidlimcheng:
Seeing as how this needs to be tested tomorrow and judging from the finality of the comments, this seems like it's ready to go.  Good stuff, @redconfetti

PR #6867, opened 2017-10-18
widgetService: Fix nodeName check in `toggleShow`

mc10:
https://jira.berkeley.edu/browse/SISRP-36753

This prevents the corresponding "Class Enrollment" section from collapsing when a link inside a section is clicked.

### DIFF
--- a/src/assets/javascripts/angular/services/widgetService.js
+++ b/src/assets/javascripts/angular/services/widgetService.js
@@ -7,9 +7,9 @@ angular.module('calcentral.services').service('widgetService', function(analytic
    * Toggle whether an item for a widget should be shown or not
    */
   var toggleShow = function(event, items, item, widget) {
-    var tagName = (event && event.toElement && event.toElement.tagName);
-    // Ignore toggling on Anchor events
-    if (['A', 'INPUT', 'TEXTAREA'].indexOf(tagName) !== -1) {
+    var ignoreTogglingTags = ['A', 'INPUT', 'TEXTAREA'];
+    var nodeName = (event && event.target && event.target.nodeName);
+    if (nodeName && ignoreTogglingTags.indexOf(nodeName) !== -1) {
       return;
     }
     // Toggle the current item


davidlimcheng:
Thanks @mc10! One thing I noticed is that this _should_ be checked in [`widgetService.js`](https://github.com/ets-berkeley-edu/calcentral/blob/master/src/assets/javascripts/angular/services/widgetService.js):
```
  var toggleShow = function(event, items, item, widget) {
    var tagName = (event && event.toElement && event.toElement.tagName);
    // Ignore toggling on Anchor events
    if (['A', 'INPUT', 'TEXTAREA'].indexOf(tagName) !== -1) {
      return;
    }
```
We are using this service for this particular section of the card, so if it is a bug with the service, it might be better to fix it at that level, which will propagate the fix application-wide (as opposed to just this particular card).

If you can find the bug there, that would be a great way to put in a more permanent fix.  I'll take a look as well, when time permits.

mc10:
It seems to me that the bug is using `event.toElement` (an IE-only attribute) instead of `event.target`. After some digging, it seems that #1299 was the original PR that introduced these lines of code, but there seems to be no justification in using `toElement`.

src/assets/javascripts/angular/services/widgetService.js
```diff
@@ -7,9 +7,9 @@ angular.module('calcentral.services').service('widgetService', function(analytic
    * Toggle whether an item for a widget should be shown or not
    */
   var toggleShow = function(event, items, item, widget) {
-    var tagName = (event && event.toElement && event.toElement.tagName);
-    // Ignore toggling on Anchor events
-    if (['A', 'INPUT', 'TEXTAREA'].indexOf(tagName) !== -1) {
+    var IGNORE_TOGGLING_TAGS = ['A', 'INPUT', 'TEXTAREA'];
```
davidlimcheng: @mc10, do you mind changing this to camelcase?  We're still on ES5, so we can't use `const` yet, and this array is technically still variable.  We hope to upgrade our ES soon, though.

davidlimcheng:
Looks great, @mc10!  Left one request, then should be good to merge after.  Thanks so much for the bug fix!

redconfetti:
Great catch @mc10 on that use of `toElement` that should be `target`.

redconfetti:
@mc10, we want to make sure we test this fix, so we need to describe a test scenario. Can you explain what you noticed, and which browser you were using when this occurred?

src/assets/javascripts/angular/services/widgetService.js
mc10: Sounds good to me.

mc10:
@redconfetti The place I've noticed this issue is the "Class Enrollment" section on the "My Academics" page; if you click on any of the links within the collapsible sections, the section still collapses, despite the link opening (in a new tab).

davidlimcheng:
Looks great @mc10, thanks again!

PR #6873, opened 2017-10-23
SISRP-36611 - Improves caching of MyAcademics feed

lyttam:
https://jira.berkeley.edu/browse/SISRP-36611

This is an attempt to improve performance of the `AcademicStatus` feed and its downstream consumers, while tidying up the dependencies in the code and increasing DRYness.  The architecture now consists of one proxy model and two layers of domain models, for reasons explained below.

- `HubEdos::AcademicStatus` is the iHub proxy that passes through the data from the `v1/students/{EMPLID}/academic-status` endpoint without modifying it.

- `MyAcademics::MyAcademicStatus` is the second layer - it parses through the feed and attaches a `role` to each career, program, and plan object in the feed.  This must be done ahead of time so that the 

- `MyAcademics::CollegeAndLevel` and `MyAcademics::MyAcademicRoles` are the third layer.  They each provide a feed for direct consumption by the front-end.  

- I moved some of the common functionality into a 'concern' (basically, a mixin).  I did some research and formed some opinions on how best to use concerns.  I'm happy to discuss more on that in person if anyone is curious.
 
- I also made all of the models under `MyAcademics` inherit their initialize methods from `UserSpecificModel` rather than get them from the mixin from `AcademicsModule` because it makes more sense.  Eventually I intend to split up `AcademicsModule` into logical parts and give the parts more descriptive names.

- I made the `MyAcademics::MyAcademicRoles` data available to the front-end via the existing endpoint `api/my/status`.  I updated all the front-end code that depends on that data to look for it in `apiService.user.profile.academicRoles`


davidlimcheng:
Really good stuff, @lyttam. Reviewing now - in the JIRA, it looks like you were planning to create a new subdirectory/class called `Concerns` that would be used for pure Ruby mixins.  But, it looks like you opted to keep the `Berkeley` class for `Berkeley::AcademicStatus`.  Why did you decide to keep `Berkeley` instead of moving it to `Concerns`, and do you still plan on creating that new class?

app/models/berkeley/academic_status.rb
```diff
@@ -0,0 +1,37 @@
...
+          end
+        end
+      end.uniq
+    end
+
+    def newest_career(statuses)
+      newest_career_status = statuses.try(:sort) do |this_status, that_status|
+        this_from_date = this_status['studentCareer'].try(:[], 'fromDate').to_s
+        that_from_date = that_status['studentCareer'].try(:[], 'fromDate').to_s
+        this_from_date <=> that_from_date
```
davidlimcheng: I see that we're currently implementing this sort, and that we're only using it for the Advisor view, but is it possible we'll run into any problems here for continuing students?  For example, a scenario where a student is finishing up his last undergraduate term, and has been accepted to graduate school.  Would CalCentral "shift" to a GRAD experience, when we still need the UGRD experience?  

Would it make sense to add an `active?` check here, effectively making it `newest_active_career` instead of `newest_career`?

app/models/degree_progress/my_undergrad_requirements.rb
```diff
@@ -30,11 +30,8 @@ def add_links(response)
     end
 
     def should_see_links?
-      academic_status.try(:[], :feed).try(:[], 'student').try(:[], 'roles').try(:[], 'lettersAndScience')
-    end
-
-    def academic_status
-      @academic_status ||= HubEdos::MyAcademicStatus.new(@uid).get_feed
+      roles = MyAcademics::MyAcademicRoles.new(@uid).get_feed
+      !!roles['lettersAndScience']
```
davidlimcheng: Will this raise a NoMethodError, despite the double-bang?  It seems fine via Rails Console, but were you able to confirm that an error won't be raised if `'lettersAndScience'` does not exist?

spec/models/campus_solutions/my_enrollment_term_spec.rb
```diff
@@ -23,7 +23,6 @@
       before { allow_any_instance_of(described_class).to receive(:get_feed).and_return(api_response) }
       it 'returns enrollment term object' do
         result = described_class.get_term(user_id, term_id)
-        puts "result: #{result.inspect}"
```
davidlimcheng: Nice catch!

lyttam:
@davidlimcheng I think some changes got lost in a merge... I swear I had a /concerns directory.  I will make another push.

app/models/degree_progress/my_undergrad_requirements.rb
lyttam: @davidlimcheng `roles` will always have the key `'lettersAndScience'` because of how that object is constructed - see `AcademicRoles#role_defaults`.  There are unit tests in place that will break if someone unwittingly changes that behavior.

lyttam:
@davidlimcheng I moved the three mixins into a /concerns directory.  Planning to do some more moving/refactoring/renaming of other mixins in the future.

app/models/berkeley/academic_status.rb
redconfetti: @davidlimcheng - The career code logic is used only for the [Service Indicators link](https://github.com/lyttam/calcentral/blob/e27352d6/app/controllers/concerns/advising_resources.rb#L42) used by Advisors. I wouldn't suggest changing the logic for that link, because perhaps the inactive yet immediately new career is more relevant for advisors.

Also AFAIK I think the Advisor is able to change to the other career easily from the CS page they land on.

src/assets/javascripts/angular/controllers/pages/enrollmentVerificationController.js
```diff
@@ -64,18 +57,24 @@ angular.module('calcentral.controllers').controller('EnrollmentVerificationContr
     }
   };
 
+  var parseRoles = function() {
+    if (_.isPlainObject(apiService.user.profile.academicRoles)) {
```
davidlimcheng: @lyttam Can you talk me through the usage of [_.isPlainObject](https://lodash.com/docs/4.17.4#isPlainObject) here?  I'm having a hard time understanding why that check is being used.

src/assets/javascripts/angular/controllers/pages/myFinancesController.js
```diff
@@ -1,35 +1,21 @@
...
-  };
-
-  var loadAcademicRoles = function() {
-    return academicStatusFactory.getAcademicRoles().then(
-      function(parsedAcademicRoles) {
-        $scope.academicStatus.roles = _.get(parsedAcademicRoles, 'roles');
-      }
-    );
+    apiService.util.redirectToHome();
+    return false;
```
davidlimcheng: Why do we need a `return false` statement here?

davidlimcheng:
Really good stuff, @lyttam.  Wholeheartedly agree with the refactor of academic roles, especially considering its omnipresent usage throughout our code.  Just a couple of questions to fill gaps in my understanding.

lyttam:
Please hold off on merging - I found some things I missed and will be making another push.

spec/controllers/user_controller_spec.rb
```diff
@@ -16,60 +16,66 @@
   context 'when not logged in' do
     let(:user_id) { nil }
     it 'should not have a logged-in status' do
-      get :mystatus
+      get :my_status
```
redconfetti: @lyttam - i_appreciate_the_proper_snake_case_here

image
)

spec/models/campus_solutions/my_enrollment_term_spec.rb
redconfetti: It [was me](https://github.com/ets-berkeley-edu/calcentral/blame/a420ca35/spec/models/campus_solutions/my_enrollment_term_spec.rb#L26) 😊 , and @lyttam merged it ;)

spec/controllers/advising_student_controller_spec.rb
```diff
@@ -46,133 +46,162 @@
...
+      context 'when viewing a student' do
+        let(:student) { true }
 
-      context 'confidential' do
-        let(:confidential) { true }
-        it_behaves_like 'an endpoint refusing a request'
+        it_behaves_like 'an endpoint receiving a valid request'
+        it 'should provide a filtered academics feed' do
+          feed = JSON.parse(body = subject.body)
+          puts feed.pretty_inspect
```
redconfetti: @lyttam - rogue puts

spec/models/my_academics/my_academic_roles_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+      it 'returns an empty array' do
+        expect(subject).to eq []
+      end
+    end
+  end
+
+  describe '#extract_roles' do
+    subject { described_class_instance.extract_roles(status) }
+
+    context 'when status is nil' do
```
redconfetti: @lyttam - Why does this only test when it's nil, but doesn't test the common path of the results of it's parsing?

redconfetti: Oh nevermind. I see that the primary feed itself is tested for that functionality, because it's used by the feed.

src/assets/javascripts/angular/controllers/pages/enrollmentVerificationController.js
lyttam: @davidlimcheng That was in the existing code, and I was trying not to mess with it too much besides just pulling it out into its own function (`parseRoles).

It seems to be doing a somewhat unnecessary check to make sure the academic roles are an object and not, say, an array or something.  Since we're confident it will always be an object (because we create it that way in the model), I could probably safely remove this check.  What do you think?

src/assets/javascripts/angular/controllers/pages/myFinancesController.js
lyttam: @davidlimcheng `return false` does nothing when calling this function from another JS function.  But when binding it to a click event, like we're doing in delegate_access_denied.html, it will do two things (in jQuery at least; I assume it also works in Angular's jQuery lite):

event.stopPropagation()
event.preventDefault()

See https://css-tricks.com/lodge/learn-jquery/08-preventdefault-stoppropagation-return-false/ for details.

src/assets/javascripts/angular/controllers/pages/enrollmentVerificationController.js
davidlimcheng: @lyttam, yeah, I think it's a small but good check to remove.  We see in [userService.js](https://github.com/ets-berkeley-edu/calcentral/blob/master/src/assets/javascripts/angular/services/userService.js#L7) that it's initiated as an object, with data only ever being merged, so like you said, I don't think we need to worry about it ever being something else.

src/assets/javascripts/angular/controllers/pages/myFinancesController.js
davidlimcheng: Very cool, thanks for the lesson!

src/assets/javascripts/angular/controllers/pages/enrollmentVerificationController.js
lyttam: @davidlimcheng Cool, I removed the `_.isPlainObject` check.

lyttam:
Additional work is finished and I responded to and/or implemented all the suggestions.  Ready to merge.

davidlimcheng:
LGTM!

PR #6875, opened 2017-10-25
SISRP-36868 - Update CampusSolutions::Link to include only proper camelcase key for isCsLink

redconfetti:
https://jira.berkeley.edu/browse/SISRP-36868

I originally thought that we needed to move from using HashConverter.camelize to [HashConverter.downcase_and_camelize](https://github.com/ets-berkeley-edu/calcentral/blob/eb795453/lib/hash_converter.rb#L15), but it turns out that this isn't a good idea, because the downcase will destroy existing camelcase keys. Both are used where appropriate for now.

We don't use this 'IS_CS_LINK' key, nor does the ccCampusSolutionsLinkDirective recognize :ISCsLink as a key. So this is just a simple maintenance task to remove the unused key.

```
>> hash = {'IS_CS_LINK' => 'test'}
=> {"IS_CS_LINK"=>"test"}
>> HashConverter.downcase_and_camelize(hash)
=> {:isCsLink=>"test"}
>> HashConverter.camelize(hash)
=> {:iSCsLink=>"test"}

>> hash = {:isCsLink => 'test'}
=> {:isCsLink=>"test"}
>> HashConverter.downcase_and_camelize(hash)
=> {:iscslink=>"test"}
```

### DIFF
--- a/app/models/campus_solutions/link.rb
+++ b/app/models/campus_solutions/link.rb
@@ -89,7 +89,6 @@ module CampusSolutions
         link[:title] = link.delete(:hoverOverText) || ''
         if !link[:showNewWindow]
           link[:isCsLink] = true
-          link['IS_CS_LINK'] = true
         end
       end
       link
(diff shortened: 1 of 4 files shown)


davidlimcheng:
```
>> HashConverter.camelize(hash)
=> {:iSCsLink=>"test"}
```
Interesting, I wonder if that's a Rails bug that's been patched.  Anyway, LGTM @redconfetti

PR #6877, opened 2017-10-27
CLC-6759 Work around inefficient Google Sheets wrapper

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6759


paulkerschen:
I wish I could say that wrapper was working hard so we don't have to.

PR #6880, opened 2017-11-06
SISRP-36611 - Fixes spinner issues

lyttam:
https://jira.berkeley.edu/browse/SISRP-36611

In my refactoring of the academic-status feed, I inadvertently broke the loading spinners on a couple of the cards.

### DIFF
--- a/src/assets/javascripts/angular/controllers/widgets/enrollment/enrollmentCardController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/enrollment/enrollmentCardController.js
@@ -8,7 +8,9 @@ var _ = require('lodash');
  * Main controller for the enrollment card on the My Academics and Student Overview pages
  */
 angular.module('calcentral.controllers').controller('EnrollmentCardController', function(apiService, enrollmentFactory, linkService, $route, $scope) {
-  $scope.isLoading = true;
+  $scope.enrollment = {
+    isLoading: true
+  };
   $scope.accessibilityAnnounce = apiService.util.accessibilityAnnounce;
   linkService.addCurrentRouteSettings($scope);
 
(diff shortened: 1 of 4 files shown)


src/assets/javascripts/angular/controllers/widgets/enrollment/enrollmentCardController.js
```diff
@@ -227,7 +229,7 @@ angular.module('calcentral.controllers').controller('EnrollmentCardController',
   };
 
   var loadEnrollmentInstructionDecks = function() {
-    return enrollmentFactory.getEnrollmentInstructionDecks();
+    return enrollmentFactory.getEnrollmentInstructionDecks().then(parseEnrollmentInstructionDecks);
```
davidlimcheng: @lyttam - instead of having two `isLoading = false;` lines on 267 & 275, would it be more efficient to add a [`.finally()`](https://stackoverflow.com/questions/37075972/angular-is-finally-called-at-the-end-of-every-path-in-a-chained-promise-dec) clause to this line (or 279), and setting `isLoading = false;` there?  That way, the flag will always be set to `false` eventually, even if there was an error somewhere along the lines of grabbing the data and parsing it.

lyttam: @davidlimcheng yeah, I thought about that too.  But in order to use a `.finally()` you need to create a Promise object and eventually resolve it.  Within the `if` block, `loadEnrollmentInstructionDecks()` returns that Promise, but there's no Promise in the `else` block and I thought it would be awkward to create one just for this purpose.

davidlimcheng: Ah!  I see.  In that case, shouldn't we either:

1. Remove the existing `.finally()` block, since we're removing `$scope.isLoading` or
2. Replace line 272 with `$scope.enrollment.isLoading = false` and remove line 267

Personally I think #2 would be cleaner, but either way, I believe we have to remove the now-defunct `$scope.isLoading`, unless there is a parent controller somewhere that's using it that I'm not seeing.

lyttam: @davidlimcheng ahh, I see what you're pointing out!  That's a mistake, I'll fix it and do another push

davidlimcheng:
LGTM! Thanks @lyttam

PR #6886, opened 2017-11-13
SISRP-37100 - CalCentral Pulling in Inactive Addresses

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-37100

### DIFF
--- a/app/models/hub_edos/contacts.rb
+++ b/app/models/hub_edos/contacts.rb
@@ -15,5 +15,29 @@ module HubEdos
       %w(identifiers affiliations names addresses phones emails urls emergencyContacts confidential)
     end
 
+    def build_feed(response)
+      remove_invalid_addresses super
+    end
+
+    def remove_invalid_addresses(response)
+      filtered_addresses = response.try(:[], 'student').try(:[], 'addresses').try(:delete_if) do |address|
+        address_is_invalid? address
+      end
+      response['student']['addresses'] = filtered_addresses unless filtered_addresses.nil?
+      response
+    end
+
+    def address_is_invalid?(address)
+      if (to_date = address.try(:[], 'toDate'))
+        to_date = Date.parse(to_date)
+        from_date = Date.parse(address.try(:[], 'fromDate'))
+        current_date = Settings.terms.fake_now ? Date.parse(Settings.terms.fake_now.to_s) : Date.today
+        if (current_date < from_date) || (current_date > to_date)
+          return true
+        end
+      end
+      false
+    end
+
   end
 end
--- a/fixtures/json/hub_contacts.json
+++ b/fixtures/json/hub_contacts.json
@@ -77,6 +77,7 @@
                 "formattedAddress": "1234 56TH ST\nAPT 789, BOX 101112\nBERKELEY, California 94708",
                 "disclose": false,
                 "fromDate": "2015-01-12",
+                "toDate": "2142-01-01",
                 "type": {
                   "code": "DIPL",
                   "description": "Diploma"
@@ -106,6 +107,7 @@
                 "countryName": "United States",
                 "formattedAddress": "2111 BANCROFT WAY  #550\nBERKELEY, California 94720",
                 "fromDate": "2013-10-31",
+                "toDate": "2013-11-01",
                 "disclose": false,
                 "type": {
                   "code": "HOME",
--- a/spec/models/hub_edos/contacts_spec.rb
+++ b/spec/models/hub_edos/contacts_spec.rb
@@ -1,6 +1,9 @@
 describe HubEdos::Contacts do
 
   context 'mock proxy' do
+    before(:each) do
+      allow(Settings.terms).to receive(:fake_now).and_return('2017-11-07 00:00:00')
+    end
     let(:proxy) { HubEdos::Contacts.new(fake: true, user_id: '61889') }
     subject { proxy.get }
 
@@ -15,6 +18,10 @@ describe HubEdos::Contacts do
       expect(subject[:feed]['student']['addresses'][0]['country']).to eq 'USA'
       expect(subject[:feed]['student']['addresses'][0]['formattedAddress']).to eq "1234 56TH ST\nAPT 789, BOX 101112\nBERKELEY, California 94708"
       expect(subject[:feed]['student']['phones'].length).to eq 1
+    end
+
+    it 'properly removes inactive addresses' do
+      expect(subject[:feed]['student']['addresses'].length).to eq 2
     end
   end
 


app/models/hub_edos/contacts.rb
```diff
@@ -15,5 +15,31 @@ def whitelist_fields
...
+        end
+      end
+      response
+    end
+
+    def address_is_visible?(address)
+      if (to_date = address.try(:[], 'toDate'))
+        to_date = Date.parse(to_date)
+        from_date = Date.parse(address.try(:[], 'fromDate'))
+        current_date = Settings.terms.fake_now ? Date.parse(Settings.terms.fake_now.to_s) : Date.today
```
lyttam: @davidlimcheng this could be written more succinctly as `Settings.terms.fake_now || DateTime.now` (and we use `DateTime.now` rather than `Date.today` elsewhere in the code).

lyttam:
@davidlimcheng I see you're adding a visibility flag to each address, but where does it get used?

Why not just remove any addresses that aren't current before returning the response?

app/models/hub_edos/student.rb
```diff
@@ -61,6 +61,10 @@ def wrapper_keys
       %w(apiResponse response any students)
     end
 
+    def parse_response(response)
+      super
+    end
```
redconfetti: @davidlimcheng - Why is this being defined here, when it's not doing anything different than the class it inherits from ([HubEdos::Proxy#parse_response](https://github.com/davidlimcheng/calcentral/blob/fba419af/app/models/hub_edos/proxy.rb#L120))?

app/models/hub_edos/contacts.rb
davidlimcheng: @lyttam I was actually worried to use `DateTime` - since CS is sending us a simple date string (e.g. `2017-11-08`), I figured it would be safest to use `Date` class to parse.  In addition, since there are so many ways us developers/testers fill out `Settings.terms.fake_now`, such as `2017-11-08` or `2017-11-08 00:00:00 -0700` etc., I found out through local testing that this can potentially already be parsed as a `DateTime` object already, which will cause `Date.parse` to error out, so I had to append a `to_s` method to whatever is brought from `Settings` - which is why I couldn't simply put `Settings.terms.fake_now || Date.today`

Which is a very long-winded way of saying - the only reason why I object is because CS will always send us a simple date string, with no time.  Which is actually kind of nice, since that means we get to keep it simple.  What do you think?

app/models/hub_edos/student.rb
davidlimcheng: @redconfetti Because I just needed it as a passthrough, since `HubEdos::Contacts` is a child of `HubEdos::Student`.  If I wanted to use the built-in `parse_response` method in `Contacts`, I needed a way to access the existing `parse_response` method in `HubEdos::Proxy`:
```
    def parse_response(response)
      safe_json response.body.force_encoding('UTF-8')
    end
```
I also could have just copy and pasted this, but redefining `parse_response` under `HubEdos::Student` opens up the method to any child of `Student`.

That being said, I could be understanding this incorrectly.  [Here is what I based my understanding of `super` on.](https://stackoverflow.com/questions/4557304/calling-supers-super-method)

davidlimcheng:
@lyttam Massive brainfart on my end.  I definitely should have removed the addresses being flagged as `visible: false`, or included some front-end changes to filter out the non-visible addresses.  I'll go with your recommendation and simply remove the addresses, amendment coming.  Nice catch!

app/models/hub_edos/student.rb
davidlimcheng: That being said, this does make me realize an error in my code - I'm calling `super` but not assigning anything to it, so as of now it's an erroneous call.  Will amend.

redconfetti: @davidlimcheng - Is this necessary? I did double inheritance with classes in [this REPL](https://repl.it/@redconfetti/Second-Generation-Inheritance-and-Super) and it worked.

davidlimcheng: @redconfetti It does work!  Thanks for clarifying that.  Next time I'll test it myself instead of relying on SO ;)

davidlimcheng:
@lyttam @redconfetti Should be good for another review now.

app/models/hub_edos/contacts.rb
```diff
@@ -15,5 +15,31 @@ def whitelist_fields
       %w(identifiers affiliations names addresses phones emails urls emergencyContacts confidential)
     end
 
+    def parse_response(response)
+      remove_invalid_addresses(super)
```
lyttam: @davidlimcheng one last thing... shouldn't we be passing `response` to this function, not `super`?

davidlimcheng: @lyttam From http://rubylearning.com/satishtalim/ruby_overriding_methods.html:

"When you invoke super with no arguments Ruby sends a message to the parent of the current object, asking it to invoke a method of the same name as the method invoking super. It automatically forwards the arguments that were passed to the method from which it's called."

I'm invoking `super` from the Grandparent `HubEdos::Proxy` class, which forces UTF-8 encoding on the response, then jsonify's it:

```
    def parse_response(response)
      safe_json response.body.force_encoding('UTF-8')
    end
```

And then removing the invalid addresses from that JSON output.  I believe passing `response` would instead bypass this step.

lyttam: @davidlimcheng fine, but you have an unused parameter (`response`) and something seems weird about overriding `parse_response`. 

Other `HubEdos` proxies (e.g. `WorkExperience`) override `build_feed` instead, and call `super(response)` to invoke `HubEdos::Proxy#build_feed` - which calls `parse_response`.  I know `parse_response` is technically a public method but it feels like it should be private.

So I would suggest this instead:
```
  def build_feed(response)
      remove_invalid_addresses super(response) 
   end
```

That way you also take advantage of some common response handling functionality that's built into `HubEdos::Proxy#build_feed`.

davidlimcheng: @lyttam Works for me!  I wish there was a way to document these very specific "best practices" cases.  I was actually looking for an existing case where we parse the `Student` response, but didn't see `WorkExperience`.  

I would argue though, that in my current `parse_response` implementation, the `response` parameter is not unused.  To quote that article again:

"It automatically forwards the arguments that were passed to the method from which it's called."

So I think `super == super(response)`.  Only when `super()` is invoked does it not use the parameters passed into the method from which it was called.  So, I think a more concise implementation would be
```
  def build_feed(response)
      remove_invalid_addresses super
   end
```
In any case, amendment coming to override `build_feed` instead of `parse_response`.

davidlimcheng:
@lyttam Amended, should be good once we fix whatever is ailing Travis.

PR #6889, opened 2017-11-17
SISRP-30118 - Updates Degrees Conferred with newly available data

lyttam:
https://jira.berkeley.edu/browse/SISRP-30118

New degrees conferred data is available in the academic-status API. We'll now display the names of majors and certificates correctly, and surface the college name, designated emphasis, and any minors. Also adds an icon for each degree.

### DIFF
--- a/app/models/my_academics/college_and_level.rb
+++ b/app/models/my_academics/college_and_level.rb
@@ -112,9 +112,16 @@ module MyAcademics
 
       filtered_statuses.each do |status|
         Array.wrap(status.try(:[], 'studentPlans')).each do |plan|
-          flattened_plan = flatten_plan(plan)
-          plan_set[:plans] << flattened_plan
+          flattened_plan = flatten_plan(plan.try(:[], 'academicPlan'))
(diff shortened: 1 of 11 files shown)


app/models/my_academics/college_and_level.rb
```diff
@@ -123,9 +130,26 @@ def parse_hub_plans(statuses)
 
     def parse_hub_degrees(response)
       if (degrees = response.try(:[], :feed).try(:[], 'student').try(:[], 'degrees'))
-        awarded_degrees = degrees.select do |degree|
-          status = degree.try(:[], 'status').try(:[], 'code')
-          status === 'Awarded'
+        awarded_degrees = []
+        degrees.each do |degree|
+          if degree.try(:[], 'status').try(:[], 'code') === 'Awarded'
```
davidlimcheng: @lyttam Did you mean to use `===` (subsumption operator) instead of `==` here, as well as on line 150?

lyttam: @davidlimcheng No, good catch!

PR #6890, opened 2017-11-17
SISRP-29924 - Faculty enrollment drop request does not differentiate section #s

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-29924


lyttam:
Looks good.

PR #6891, opened 2017-11-17
SISRP-37237 - Removes extraneous hard-coded 'https' from URLs

lyttam:
https://jira.berkeley.edu/browse/SISRP-37237


davidlimcheng:
LGTM, nice reformatting.

PR #6905, opened 2017-12-06
SISRP-37144 - Surface timezone offset and render with provided offset

redconfetti:
https://jira.berkeley.edu/browse/SISRP-37144


nkutub:
LGTM @redconfetti

PR #6907, opened 2017-12-06
SISRP-37144 - Adds timezone offset for schedule of classes period. Updates times displayed in Class Enrollment sections

redconfetti:
https://jira.berkeley.edu/browse/SISRP-37144

### DIFF
--- a/app/models/my_academics/class_enrollments.rb
+++ b/app/models/my_academics/class_enrollments.rb
@@ -141,20 +141,33 @@ module MyAcademics
         instructions[term_id] = CampusSolutions::MyEnrollmentTerm.get_term(@uid, term_id)
         instructions[term_id][:concurrentApplyDeadline] = get_concurrent_apply_deadline(term_id)
         instructions[term_id][:termIsSummer] = Berkeley::TermCodes.edo_id_is_summer?(term_id)
-        apply_enrollment_period_timezones(instructions[term_id])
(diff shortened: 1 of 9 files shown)


app/models/my_academics/class_enrollments.rb
```diff
@@ -141,22 +141,35 @@ def get_enrollment_term_instructions
...
       end
       instructions
     end
 
-    def apply_enrollment_period_timezones(instruction)
+    def apply_period_timezones(instruction)
+      if schedule_of_classes_period = instruction.try(:[], :scheduleOfClassesPeriod)
+        if offset = get_timezone_offset(schedule_of_classes_period.try(:[], :date))
+          instruction[:scheduleOfClassesPeriod][:date][:offset] = offset
+        end
```
lyttam: @redconfetti I'd prefer it if the `offset` key was always present, rather than having it included conditionally. 

```
instruction[:scheduleOfClassesPeriod][:date][:offset] = get_timezone_offset(schedule_of_classes_period.try(:[], :date))
```

This way it's easier to deal with on the front end, and avoids the potential for bugs of a certain kind.

redconfetti: @lyttam - Good call. I've fixed it.

lyttam:
Looks good

PR #6913, opened 2017-12-12
CLC-6777 Separate bCourses email deletion from user inactivation

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6777

This should eliminate a sometimes painful timing conflict between the two stages of our occasional user account inactivation.


johncrossman:
Looks good. Thx, @raydavis

PR #6917, opened 2017-12-18
SISRP-37370 - Fixes errors in academics feed

lyttam:
https://jira.berkeley.edu/browse/SISRP-37370

These two errors have been occurring in instances where a response from CS is missing data and we aren't handling it gracefully.

### DIFF
--- a/app/models/my_academics/grading.rb
+++ b/app/models/my_academics/grading.rb
@@ -248,7 +248,7 @@ module MyAcademics
 
     def parse_cs_grading_status(cs_grading_status, is_law, is_summer)
       if unexpected_cs_status?(cs_grading_status, is_law)
-        logger.warn "Unexpected CS Final Grading Status Received (Final: #{cs_grading_status[:finalStatus]}#{', Midpoint: ' + cs_grading_status[:midpointStatus] if cs_grading_status.key?(:midpointStatus)}) for Class #{self.class.name} feed, uid = #{@uid}"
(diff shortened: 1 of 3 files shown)


app/models/my_academics/grading.rb
```diff
@@ -248,7 +248,7 @@ def find_status_in_rosters(rosters, is_law, is_summer)
 
     def parse_cs_grading_status(cs_grading_status, is_law, is_summer)
       if unexpected_cs_status?(cs_grading_status, is_law)
-        logger.warn "Unexpected CS Final Grading Status Received (Final: #{cs_grading_status[:finalStatus]}#{', Midpoint: ' + cs_grading_status[:midpointStatus] if cs_grading_status.key?(:midpointStatus)}) for Class #{self.class.name} feed, uid = #{@uid}"
+        logger.warn "Unexpected CS Final Grading Status Received (Final: #{cs_grading_status.try(:[], :finalStatus)}#{', Midpoint: ' + cs_grading_status.try(:[], :midpointStatus) if cs_grading_status.try(:key?).try(:[], :midpointStatus)}) for Class #{self.class.name} feed, uid = #{@uid}"
```
redconfetti: @lyttam - I'm confused about this part of the command: `if cs_grading_status.try(:key?).try(:[], :midpointStatus)`. I don't think that a call to #key? returns a hash that would have `:midpointStatus` present. I think you meant to use `cs_grading_status.try(:[], :midpointStatus)`?

lyttam: @redconfetti good catch, `key?` is unnecessary here. I will correct it.

lyttam:
Fixed issue pointed out by @redconfetti. Enhanced unit tests in relation to that fix.

PR #6918, opened 2017-12-19
Revert "SISRP-35757 - Upgrades codebase to use JRuby 9.1.5.0"

lyttam:
https://jira.berkeley.edu/browse/SISRP-35757

Reverts ets-berkeley-edu/calcentral#6916 because it broke Junction deployments.


raydavis:
Thank you, @lyttam . I see that automatic deployment was already off for CC-SIS-QA, so it doesn't look like this will interfere with today's load test.

PR #6921, opened 2017-12-21
SISRP-37724 - New Admit UX Dev: Refactor existing SIR logic from front-end to back-end

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-37724

Pretty hefty PR, but that's deceptive:
* Moves all the front-end SIR logic to the back-end
* * Because of that, we can consolidate from exposing two endpoints to one
* * Back-end now serves the front-end a nice package that contains all of the data
* Moves a lot of the SIR-related models to the SIR module/directory
* Ensured back-end design was the same so no changes to front-end necessary


src/assets/javascripts/angular/factories/sirFactory.js
```diff
@@ -6,36 +6,30 @@ var angular = require('angular');
...
  */
 angular.module('calcentral.factories').factory('sirFactory', function(apiService, $http) {
-  var urlChecklist = '/api/campus_solutions/checklist';
-  // var urlChecklist = '/dummy/json/cs_checklist_ugrd.json';
   var urlDeposit = '/api/campus_solutions/deposit';
   var urlHigherOne = '/api/campus_solutions/higher_one_url';
-  var urlSirConfig = '/api/campus_solutions/sir_config';
-  // var urlSirConfig = '/dummy/json/cs_sir_config.json';
   var urlSirResponse = '/api/campus_solutions/sir_response';
+  var urlSirStatuses = '/api/my/sir_statuses';
```
davidlimcheng: I also want to fold in `urlDeposit` into the `sir_statuses` endpoint, but that will probably come next as that requires front-end changes.

app/models/campus_solutions/checklist.rb
```diff
@@ -1,7 +1,7 @@
 module CampusSolutions
   class Checklist < Proxy
 
-    include SirFeatureFlagged
+    include Sir::SirFeatureFlagged
```
redconfetti: @davidlimcheng - When we were going through the process of setting up new features in CalCentral for certain releases, we implemented many various feature flags, so that we could turn off that functionality.

I don't think we truly intended to have the possibility of completely turning off a crucial feature in CalCentral like the SIR card long term. I can only think we wanted the ability to turn it off inbetween releases if we needed to, with some other work-around put in place in an emergency.

I could be wrong. We should ask if certain feature flags should be removed, because it doesn't make sure that we're supporting a feature (i.e. feature flags) that would never be used.

app/models/campus_solutions/sir/sir_statuses.rb
```diff
@@ -0,0 +1,97 @@
...
+  module Sir
+    class SirStatuses < UserSpecificModel
+
+      include Cache::CachedFeed
+      include Cache::UserCacheExpiry
+      include SirFeatureFlagged
+
+      HEADER_DATA = {
+        DEFAULT: {
+          background: 'cc-widget-sir-background-berkeley'
```
lyttam: @davidlimcheng In general we should avoid letting UI stuff like CSS class names leak into our back end code.  It's a bad architecture smell and creates unnecessary dependencies between front end and back end.  However, I know you're just moving around code that already existed, so no need to change it now.

redconfetti: I was concerned about this myself also. I wasn't sure how one would apply the image URL from a feed to the background image of a DIV, but I just noticed the the [ngStyle directive](https://docs.angularjs.org/api/ng/directive/ngStyle) would make that possible.

lyttam:
@davidlimcheng nice work! One suggestion (you don't have to implement it now):  since you're moving SIR-related stuff into a module called Sir, you could remove the now-redundant "Sir" from the name of each class within the Sir module.  So, instead of `Sir::SirResponse` and `Sir::SirConfig`, we would have `Sir::Response` and `Sir::Config`.

app/models/campus_solutions/checklist.rb
davidlimcheng: @redconfetti Totally agreed, and the thought came to mind as I was developing this.  Since this is more of a "carry over existing functionality" PR though, I'll add that to my next set of JIRAs, which are "add new functionality"

davidlimcheng: Documented in https://jira.berkeley.edu/browse/SISRP-37567

app/models/campus_solutions/sir/sir_statuses.rb
davidlimcheng: Yeah it feels wrong to me too.  I'll look into `ngStyle` when implementing the new code, thanks @redconfetti

PR #6926, opened 2018-01-05
 SISRP-37567 - New Admit UX Dev: My Dashboard - SIR card Updates

davidlimcheng:
Dev JIRA:  https://jira.berkeley.edu/browse/SISRP-37567
Design JIRA: https://jira.berkeley.edu/browse/SISRP-36594

Pretty hefty change, there were a surprising number of features that had to be added for undergraduates, while grad/law user experiences stayed the same.

* Started rebranding from "Sir" to "New Admit", but it got out of hand pretty quickly due to the number of front-end references to "Sir", so I'll do that in a separate nice-to-have PR.
* Second commit removes the SIR feature flag and all references to it


src/assets/templates/widgets/sir/sir_item_completed.html
```diff
@@ -1,10 +1,33 @@
 <div class="cc-widget cc-widget-sir" data-ng-controller="SirItemController">
-  <div data-ng-include="'widgets/sir/sir_item_header.html'" data-ng-if="item.studentResponse.option.progAction !== 'WAPP'"></div>
-  <div class="cc-widget-title" data-ng-if="item.studentResponse.option.progAction === 'WAPP'">
-    <h2>Enrollment Declined</h2>
+  <div data-ng-include="'widgets/sir/sir_item_header.html'" data-ng-if="item.studentResponse.option.progAction !== 'WAPP' || item.isUndergraduate"></div>
+  <div data-ng-if="!item.isUndergraduate">
```
redconfetti: @davidlimcheng - What stops Graduate and Law students from seeing this card? Might they have a completed checklist item that causes this to show?

app/models/campus_solutions/sir/sir_feature_flagged.rb
```diff
@@ -1,9 +0,0 @@
-module CampusSolutions
-  module Sir
-    module SirFeatureFlagged
```
redconfetti: image

Bye feature flag

app/models/campus_solutions/sir/sir_statuses.rb
```diff
@@ -89,8 +128,49 @@ def add_header_info(sir_checklist_items)
           header_info = header_cd.nil? ? HEADER_DATA.try(:[], :DEFAULT) : HEADER_DATA.try(:[], header_cd)
```
redconfetti: @davidlimcheng - Isn't this supposed to reference `:GENERIC` instead of `:DEFAULT`? Where else does it reference the HEADER_DATA and not reference `:GENERIC`?

src/assets/templates/widgets/sir/sir_item_completed.html
davidlimcheng: @redconfetti - Great question - this is a bit odd, and will take a bit to explain:

In a nutshell, graduate/law students only see the "SIR Completed" state if they *decline* their offer.  This is due to the front-end controller filtering out any "Completed" checklist SIR items when the user hits the dashboard, as seen in [SirController.js](https://github.com/ets-berkeley-edu/calcentral/blob/master/src/assets/javascripts/angular/controllers/widgets/sir/sirController.js#L36). In other words, we only want to see "Completed" checklist items *when they happen*.  If a checklist status is switched to "Completed" and a user logs in after it's done, they won't see anything.

This is why the only content here before my PR "Enrollment Declined" text.  Now, we are opening up this state to undergraduates that have *accepted* their offer, which will also be in a "Completed" state.  Undergraduates, by design, are not given the option to decline, so they don't have any content related to declining their offer here.

Clear as mud?

app/models/campus_solutions/sir/sir_statuses.rb
davidlimcheng: Great catch @redconfetti, amendment coming.

spec/models/campus_solutions/sir/sir_statuses_spec.rb
```diff
@@ -188,6 +286,52 @@
...
+        end
+      end
+
+      context 'as a transfer student' do
+        before do
+          EdoOracle::Queries.stub(:get_new_admit_status) { new_admit_attributes_transfer }
+        end
+        subject { (proxy.new(uid).get_feed)[:sirStatuses]}
+        it 'adds the correct links to the status' do
+          pp subject
```
lyttam: @davidlimcheng I know this PR is already merged, but is `pp subject` a typo? My Travis build failed after I rebased and pulled in this changeset.

PR #6931, opened 2018-01-09
SISRP-37580 - Updates CampusSolutions::Sir::SirStatuses to include new admit status roles

davidlimcheng:
Dependency for both @redconfetti  and mine's work under [SISRP-37480](https://jira.berkeley.edu/browse/SISRP-37480).  I pulled this branch from @redconfetti's repo, so it'll be under his JIRA.

JIRA: https://jira.berkeley.edu/browse/SISRP-37580

### DIFF
--- a/app/models/campus_solutions/sir/sir_statuses.rb
+++ b/app/models/campus_solutions/sir/sir_statuses.rb
@@ -2,6 +2,7 @@ module CampusSolutions
   module Sir
     class SirStatuses < UserSpecificModel
 
+      include Berkeley::TermCodes
       include Cache::CachedFeed
       include Cache::UserCacheExpiry
       include Cache::RelatedCacheKeyTracker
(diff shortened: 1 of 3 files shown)


app/models/campus_solutions/sir/sir_statuses.rb
```diff
@@ -2,6 +2,7 @@ module CampusSolutions
   module Sir
     class SirStatuses < UserSpecificModel
 
+      include Berkeley::TermCodes
```
redconfetti: @davidlimcheng - Do we have to include this module? Why not simply call `Berkeley::TermCodes.codes`, as that returns the term codes you're looking for. It makes it less ambiguous which parts of this model make use of those methods in that module.

davidlimcheng: @redconfetti We're also using the `from_edo_id` method, so I would have to call it by its full name twice, which seems a bit anti-DRY to me.

PR #6934, opened 2018-01-11
SISRP-37573 - New Admit UX Dev: New Admit Resources card

davidlimcheng:
Dev JIRA: https://jira.berkeley.edu/browse/SISRP-37573
Design JIRA: https://jira.berkeley.edu/browse/SISRP-36596

* 2 commits, one for back-end and one for the front-end
* Speaking with the designer, she mentioned that the format of the New Admit Resources card, which is a collection of expandable lists, will most likely be re-used in the future.  This is why I chose to create an angular directive (`cc-expandable-link-list`) for it, even though I'm only using it twice (couldn't use it more because of one-off weirdness, like inline links in the link description)

### DIFF
--- a/spec/models/campus_solutions/new_admit_resources_spec.rb
+++ b/spec/models/campus_solutions/new_admit_resources_spec.rb
@@ -210,7 +210,6 @@ describe CampusSolutions::NewAdmitResources do
         end
         subject { proxy.new(uid).get_feed }
         it 'returns a simple response indicating expiration of the card' do
-          pp subject
           expect(subject[:visible]).to eq false
         end
       end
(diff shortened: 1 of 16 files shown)


spec/models/campus_solutions/new_admit_resources_spec.rb
```diff
@@ -0,0 +1,236 @@
...
+    end
+  end
+
+  describe 'attempting to view new admit resources' do
+
+    context 'as an incomplete new admit' do
+      context 'as a pre-matriculated non-first-year pathway freshman with an admissions evaluator' do
+        before do
+          sir_statuses_incomplete_base_response[:sirStatuses][0].merge!(new_admit_attributes_freshman)
+          CampusSolutions::Sir::SirStatuses.stub_chain(:new, :get_feed).and_return sir_statuses_incomplete_base_response
```
redconfetti: @davidlimcheng - Oh wow. I didn't know about the ability to stub a chain. I've always made #new return a double that responds to another call. This is must easier.

PR #6937, opened 2018-01-12
[QA] SISRP-38047 - Prevents user from updating any name other than the preferred name

lyttam:
https://jira.berkeley.edu/browse/SISRP-38047

Fixes a security exploit that was discovered by a student.

This fix is going straight to the qa branch for emergency release on 1/21.  I will follow up with a master PR.


davidlimcheng:
Looks good @lyttam

PR #6939, opened 2018-01-16
Display assigned advisor program in parentheses only if truthy

sethlu:
So that it will not show an empty pair of parentheses at cases


lyttam:
Thanks for the PR, @sethlu! Your change looks good; we just need to determine which release it will be included in before we merge it.

https://jira.berkeley.edu/browse/SISRP-38096 (for tracking purposes)

PR #6941, opened 2018-01-17
SISRP-37580 - New Admit UX Dev: Financial Aid Housing card

redconfetti:
https://jira.berkeley.edu/browse/SISRP-37580

### DIFF
--- a/app/models/campus_solutions/financial_aid_housing.rb
+++ b/app/models/campus_solutions/financial_aid_housing.rb
@@ -1,0 +1,56 @@
+module CampusSolutions
+  module FinancialAidHousing
+
+    INSTRUCTIONAL_MESSAGE_NUMBERS = {
+      generic: '117',
+      fall_pathways: '116',
+      spring_pathways: '118'
+    }
+    def self.append_housing(uid, feed)
+      if feed.try(:[], :feed).try(:[], :housing)
+        message_type = nil
+        if link_title_present = feed[:feed][:housing].try(:[], :link).try(:has_key?, :title)
+          feed[:feed][:housing][:link][:name] = 'Update Housing'
+          feed[:feed][:housing][:link][:title] = 'Update Housing'
+        end
+
+        sir_status_feed = CampusSolutions::Sir::SirStatuses.new(uid).get_feed
+        sir_statuses = sir_status_feed.try(:[], :sirStatuses)
+        if sir_statuses.present? && ugrd_status = sir_statuses.find {|status| status[:isUndergraduate]}
+
+          message_type = :generic
+          new_admit_attributes = ugrd_status.try(:[], :newAdmitAttributes)
+
+          first_year_pathway = new_admit_attributes.try(:[], :roles).try(:[], :firstYearPathway)
+          admit_term = new_admit_attributes.try(:[], :admitTerm)
+          if first_year_pathway
+
+            if admit_term.try(:[], :type) == 'Fall'
+              message_type = :fall_pathways
+              feed[:feed][:housing][:title] = 'Housing or Pathway'
+              if link_title_present
+                feed[:feed][:housing][:link][:title] = 'Update Housing or First-Year Pathway'
+                feed[:feed][:housing][:link][:name] = 'Update Housing / Pathway'
+              end
+            end
+            if admit_term.try(:[], :type) == 'Spring'
+              if link_title_present
+                pathways_link = LinkFetcher.fetch_link('UC_ADMT_FYPATH_FA_SPG')
+                pathways_message = CampusSolutions::MessageCatalog.get_message_catalog_definition('26500', INSTRUCTIONAL_MESSAGE_NUMBERS[:spring_pathways])
+                feed[:feed][:housing][:pathways_link] = pathways_link
+                feed[:feed][:housing][:pathways_message] = pathways_message
+              end
+            end
+          end
+        else
+          feed[:feed][:housing].delete(:link)
+        end
+        if message_type
+          instruction = CampusSolutions::MessageCatalog.get_message_catalog_definition('26500', INSTRUCTIONAL_MESSAGE_NUMBERS[message_type])
+          feed[:feed][:housing][:instruction] = instruction
+        end
+      end
+      feed
+    end
+  end
+end
--- a/app/models/campus_solutions/message_catalog.rb
+++ b/app/models/campus_solutions/message_catalog.rb
@@ -1,5 +1,17 @@
 module CampusSolutions
   class MessageCatalog < GlobalCachedProxy
+    include ClassLogger
+
+    def self.get_message_catalog_definition(message_set_nbr, message_nbr)
+      instance = self.new({message_set_nbr: message_set_nbr, message_nbr: message_nbr})
+      response = instance.get
+      if response.try(:[], :statusCode) == 200
+        return response.try(:[], :feed).try(:[], :root).try(:[], :getMessageCatDefn)
+      else
+        logger.warn "Failed to obtain message catalog definition: message_set_nbr: #{message_set_nbr}; message_nbr: #{message_nbr}"
+      end
+      nil
+    end
 
     def initialize(options = {})
       super options
--- a/app/models/campus_solutions/my_financial_aid_data.rb
+++ b/app/models/campus_solutions/my_financial_aid_data.rb
@@ -13,8 +13,9 @@ module CampusSolutions
     def get_feed_internal
       if is_feature_enabled && (self.aid_year ||= CampusSolutions::MyAidYears.new(@uid).default_aid_year)
         logger.debug "User #{@uid}; aid year #{aid_year}"
-        feed = CampusSolutions::FinancialAidData.new(user_id: @uid, aid_year: aid_year).get
-        CampusSolutions::FinancialAidDataHousingSegregator.segregate(feed)
+        finaid_feed = CampusSolutions::FinancialAidData.new(user_id: @uid, aid_year: aid_year).get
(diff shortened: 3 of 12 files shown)


app/models/campus_solutions/message_catalog.rb
```diff
@@ -1,6 +1,15 @@
 module CampusSolutions
   class MessageCatalog < GlobalCachedProxy
 
+    def self.get_message_catalog_definition(message_set_nbr, message_nbr)
+      instance = self.new({message_set_nbr: message_set_nbr, message_nbr: message_nbr})
+      response = instance.get
+      if response.try(:[], :statusCode) == 200
+        return response.try(:[], :feed).try(:[], :root).try(:[], :getMessageCatDefn)
+      end
```
davidlimcheng: @redconfetti What do you think about adding some logging here if the response returns with `statusCode != 200`? It might be useful in the context of troubleshooting a problematic message catalog entry.

app/models/campus_solutions/financial_aid_housing.rb
```diff
@@ -0,0 +1,56 @@
+module CampusSolutions
+  module FinancialAidHousing
```
lyttam: @redconfetti Why "Financial Aid Housing"?  Is there another kind of housing?

app/models/campus_solutions/financial_aid_housing.rb
```diff
@@ -0,0 +1,56 @@
...
+      generic: '117',
+      fall_pathways: '116',
+      spring_pathways: '118'
+    }
+    def self.append_housing(uid, feed)
+      if feed.try(:[], :feed).try(:[], :housing)
+        message_type = nil
+        if link_title_present = feed[:feed][:housing].try(:[], :link).try(:has_key?, :title)
+          feed[:feed][:housing][:link][:name] = 'Update Housing'
+          feed[:feed][:housing][:link][:title] = 'Update Housing'
```
lyttam: @redconfetti  I would rather see this translation done on the front end (lines 13-14).

spec/models/campus_solutions/financial_aid_housing_spec.rb
```diff
@@ -0,0 +1,186 @@
...
+        expect(subject).to have_key(:feed)
+        expect(subject[:feed]).to_not have_key(:housing)
+      end
+    end
+
+    context 'when housing data present within feed' do
+      let(:feed) { feed_with_housing }
+      context 'when sir statuses not present (non-ugrd)' do
+        let(:sir_statuses) { nil }
+        it 'exludes generic housing instructional message' do
```
davidlimcheng: @redconfetti typo here

app/models/campus_solutions/my_financial_aid_data.rb
```diff
@@ -14,7 +14,8 @@ def get_feed_internal
       if is_feature_enabled && (self.aid_year ||= CampusSolutions::MyAidYears.new(@uid).default_aid_year)
         logger.debug "User #{@uid}; aid year #{aid_year}"
         feed = CampusSolutions::FinancialAidData.new(user_id: @uid, aid_year: aid_year).get
-        CampusSolutions::FinancialAidDataHousingSegregator.segregate(feed)
+        feed = CampusSolutions::FinancialAidDataHousingSegregator.segregate(feed)
+        feed = CampusSolutions::FinancialAidHousing.append_housing(@uid, feed)
```
lyttam: @redconfetti Reusing variables makes code harder to read and is generally frowned upon.  Could use a different variable each time to make this code more descriptive (e.g. `feed`, `segregated_feed`, `housing_feed`) and perhaps encapsulate lines 16-18 in a method that returns the final feed.

Also, it looks like this method never returns the feed.  Am I missing something?

spec/models/campus_solutions/financial_aid_housing_spec.rb
```diff
@@ -0,0 +1,186 @@
+describe CampusSolutions::FinancialAidHousing do
+  subject { CampusSolutions::FinancialAidHousing.append_housing(uid, feed) }
+  let(:uid) { random_id }
+  let(:raw_feed) { {:feed => {}} }
+  let(:feed_with_housing) do
+    raw_feed[:feed][:housing] = {
```
lyttam: @redconfetti this line actually modifies `raw_feed`, which is unexpected behavior when defining `feed_with_housing`. This could cause mysterious bugs in the future.  I would prefer

```
let(:feed_with_housing) do
  feed: {
    housing: {
```
    ...etc

app/models/campus_solutions/message_catalog.rb
redconfetti: Sure. I'll add an else clause with a debug logger statement.

app/models/campus_solutions/financial_aid_housing.rb
redconfetti: @lyttam = It's a card that shows up on the 'Financial Aid and Scholarships' page. I don't know if housing information might show up elsewhere. I believe that the housing data is only related to financial data.

Also currently the model is meant to transform the data we receive from the Financial Aid Data API.

src/assets/templates/widgets/finaid_housing.html
```diff
@@ -0,0 +1,47 @@
...
+      data-ng-if="housing.link.url && !housing.errored"
+      class="cc-right cc-widget-title-link"
+      data-cc-campus-solutions-link-directive="housing.link"
+    ></a>
+  </div>
+
+  <div class="cc-widget-text" data-cc-spinner-directive="housing.isLoading">
+    <div data-ng-if="housing.errored" data-ng-include="'widgets/finaid_errored.html'"></div>
+    <div data-ng-if="!housing.errored">
+      <div data-ng-if="housing.values.length > 0">
```
lyttam: @redconfetti you might want to change this to the following in case `housing.values` does not have a property called `length`:
```<div data-ng-if="housing.values.length">```

app/models/campus_solutions/my_financial_aid_data.rb
redconfetti: @lyttam - You're right, introducing other variables doesn't necessarily increase memory usage... so I'll do that for the sake of clarity.

The use of `return` isn't necessary with Ruby. The return of the last line in the method becomes the return of the method.  That is either going to be the contents of `feed`, or the `{}` should no aid year be present (or feature flag enabled).

app/models/campus_solutions/financial_aid_housing.rb
lyttam: @redconfetti It's your call. It's nice to be able to rspec test something, but on the other hand it creates an extra dependency when the back end has knowledge of some text (or html, or css class names) that is used only by the UI.  I prefer to keep a clean separation between UI and back end concerns.

src/assets/templates/widgets/finaid_housing.html
redconfetti: @lyttam - That doesn't cover if it is an empty array, which is the case where I do not want it to display. 

Expressions passed to AngularJS directives are forgiving, meaning that they evaluate to undefined or null
> *[Forgiving](https://docs.angularjs.org/guide/expression#angularjs-expressions-vs-javascript-expressions)*: In JavaScript, trying to evaluate undefined properties generates `ReferenceError` or `TypeError`. In AngularJS, expression evaluation is forgiving to `undefined` and `null`.

This does bring up the point that my message about no housing data might not have a `length` method. I found an approach that does work - see [jsfiddle](https://jsfiddle.net/redconfetti/voke5jgq/).

lyttam: @redconfetti In JavaScript, 0 is falsey - so if `housing.values` is not an array, `housing.values.length` evaluates as false (however, if it's a string it will have a length property.  So the approach should depend on what the possible types are)

app/models/campus_solutions/financial_aid_housing.rb
redconfetti: @lyttam - I see where you're coming from. The reason the code does this is that the link itself is being shared, with only a single Link API entry shared for both links. The design calls for the same link to be here, but with two variations of wording. You'll see that on line 32-33 it has a variant.

So really this link technically exists in 3 variations (with the same url):
1. The default link location (likely in New Admit Resources card on Dashboard)
2. The 'Update Housing' link in this card (default)
3. The 'Update Housing / Pathway' link shown only to Fall Pathway eligible students

Honestly, it seems like we should probably have configured 3 different links in the Link API, for the 3 variations. I tried to ask Chara for clarification on this, but this whole thing has been so rushed due to limited time in this sprint.

lyttam: @redconfetti I see - so we're overriding the name and title that are configured for the one link?  I agree that 3 different links should have been configured.

spec/models/campus_solutions/financial_aid_housing_spec.rb
redconfetti: @lyttam - Good point. That could cause much wasted time debugging.

src/assets/templates/widgets/finaid_housing.html
redconfetti: @lyttam - True. I have an alternative approach I'm about to submit.

app/models/campus_solutions/financial_aid_housing.rb
redconfetti: @lyttam - Correction. There are only 2 variations of this link, and it doesn't come from the Link API... it comes from the Finaid Data CS API. I'll make sure that moving this link to the CS Link API is included in the follow-up refactor tasks we do in the next/future sprint.

lyttam:
Looks good, thanks for the changes.

PR #6942, opened 2018-01-17
SISRP-37574 - New Admit UX Dev: Financial Aid and Scholarships Summary card updates

davidlimcheng:
Dev JIRA: https://jira.berkeley.edu/browse/SISRP-37574
Design JIRA: https://jira.berkeley.edu/browse/SISRP-36597

* Feature flags can be removed in a couple of cycles, when the changes have stabilized in prod
* Adds some minor features to the Finaid Summary card, adding links and incomplete tasks count

### DIFF
--- a/src/assets/templates/widgets/finaid_summary_netcost.html
+++ b/src/assets/templates/widgets/finaid_summary_netcost.html
@@ -3,7 +3,20 @@
   class="cc-widget-finaid-summary-netcost-summary"
 >
   <div data-ng-repeat="item in finaidSummaryData.netCost.items">
-    <div data-cc-finaid-summary-item-directive data-item="item"></div>
+    <div data-ng-if="item.title === 'Estimated Cost of Attendance' && !isMainFinaid && finaidUrl" class="cc-clearfix-container">
+      <div class="cc-left">
+        <span data-ng-bind="item.title"></span>
+        <span data-ng-bind="itemSubTitle" class="cc-text-light"></span>
+      </div>
+      <a
+        class="cc-right"
+        data-cc-amount-directive="item.amount"
+        data-cc-amount-directive-color="false"
+        data-ng-href="{{finaidUrl}}"
+        title="View Estimated Cost of Attendance"
+      ></a>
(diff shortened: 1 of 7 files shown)


src/assets/templates/widgets/finaid_summary_netcost.html
```diff
@@ -3,7 +3,20 @@
   class="cc-widget-finaid-summary-netcost-summary"
 >
   <div data-ng-repeat="item in finaidSummaryData.netCost.items">
-    <div data-cc-finaid-summary-item-directive data-item="item"></div>
+    <div data-ng-if="item.title === 'Estimated Cost of Attendance' && !isMainFinaid && finaidUrl" class="cc-clearfix-container">
```
lyttam: @davidlimcheng It's risky to do string comparison like this, although sometimes we have no choice... are you sure the text 'Estimated Cost of Attendance' will never change?  For example, if someone changed it to 'Est. Cost of Attendance', that would break this code.

Does `item` have another property like an id, code, or short description that might be more reliable?

src/assets/templates/widgets/finaid_summary_netcost.html
```diff
@@ -3,7 +3,20 @@
...
+      </div>
+      <a
+        class="cc-right"
+        data-cc-amount-directive="item.amount"
+        data-cc-amount-directive-color="false"
+        data-ng-href="{{finaidUrl}}"
+        title="View Estimated Cost of Attendance"
+      ></a>
+    </div>
+    <div data-ng-if="item.title !== 'Estimated Cost of Attendance' || isMainFinaid || !finaidUrl" data-cc-finaid-summary-item-directive data-item="item"></div>
```
redconfetti: @davidlimcheng - I don't like that we have to take this approach, but I like that it fails gracefully if someone was to change the title from 'Estimated Cost of Attendance' to something else. This seems unlikely though.

If we can in the future, we should push for PS Devs to add keys to data points like this so that applying special behavior like this is possible.

PR #6944, opened 2018-01-19
SISRP-38173 - Housing update - First year pathways link missing banner parameters

redconfetti:
https://jira.berkeley.edu/browse/SISRP-38173

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ matrix:
     - { rvm: jruby-1.7.19 }
 
 env:
-  - JRUBY_OPTS="--dev -J-Xmx900m" DISPLAY=:99.0 LOGGER_LEVEL=WARN TRAVIS_NODE_VERSION="4"
+  - JRUBY_OPTS="--dev -J-Xmx900m" DISPLAY=:99.0 LOGGER_LEVEL=WARN TRAVIS_NODE_VERSION="7"
 
 before_install:
   - gem update --system 2.4.5
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "build-production": "gulp build --env production"
   },
   "engines": {
-    "node": ">=4.4.3"
+    "node": ">=7.8.0"
   },
   "license": "ECL-2.0",
   "dependencies": {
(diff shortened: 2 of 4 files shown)


redconfetti:
It appears that the sc5 styleguide module is causing an issue with TravisCI.
```bash
> gulp build --env production
/home/travis/build/ets-berkeley-edu/calcentral/node_modules/sc5-styleguide/node_modules/vinyl-fs/node_modules/gulp-sourcemaps/node_modules/source-map/lib/source-map-consumer.js:72
  let consumer = null;
  ^^^
SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:373:25)
    at Object.Module._extensions..js (module.js:416:10)
    at Module.load (module.js:343:32)
    at Function.Module._load (module.js:300:12)
    at Module.require (module.js:353:17)
    at require (internal/module.js:12:17)
    at Object.<anonymous> (/home/travis/build/ets-berkeley-edu/calcentral/node_modules/sc5-styleguide/node_modules/vinyl-fs/node_modules/gulp-sourcemaps/node_modules/source-map/source-map.js:7:29)
    at Module._compile (module.js:409:26)
    at Object.Module._extensions..js (module.js:416:10)
```

redconfetti:
@lyttam @davidlimcheng @nkutub - Let's discuss the NodeJS version we're using. I couldn't get TravisCI to pass until I updated the configuration as shown in the last commit. We obviously need to decide to update our version of NodeJS together, and across our entire platform (Bamboo/Docker server, etc).

lyttam:
@redconfetti was there a reason you decided to use Node 7.8?  I have not read the changelogs, but if we're moving to Node 7 we might consider 7.10.1.

nkutub:
@redconfetti ,   I think we should be fine. We don't use Node in prod, it's just a tool for CI. I'm good moving forward. we can always roll back if other issues come up.

redconfetti:
@lyttam - Lets go to 7.10.1 then. I'll make a follow up PR, and a JIRA to track the upgrade for Node. Most important is Travis and Bamboo, because those are our build environments where NodeJS is used to compile our assets.

PR #6949, opened 2018-01-22
SISRP-38231 - EFC amount need a Dollar amount$) and decimal value on the Profile card on FA page

nkutub:
https://jira.berkeley.edu/browse/SISRP-38231

### DIFF
--- a/src/assets/templates/widgets/finaid_profile.html
+++ b/src/assets/templates/widgets/finaid_profile.html
@@ -41,7 +41,7 @@
                   <span class="cc-visuallyhidden" data-ng-bind="category.title"></span>
                 </a>
               </div>
-              <div class="cc-right" data-ng-bind="item.value" data-ng-if="item.value"></div>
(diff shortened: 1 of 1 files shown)


src/assets/templates/widgets/finaid_profile.html
```diff
@@ -41,7 +41,7 @@ <h3 class="cc-page-finaid-header cc-left" data-ng-bind="category.title"></h3>
                   <span class="cc-visuallyhidden" data-ng-bind="category.title"></span>
                 </a>
               </div>
-              <div class="cc-right" data-ng-bind="item.value" data-ng-if="item.value"></div>
+              <div class="cc-right" data-cc-amount-directive="item.value" data-cc-amount-directive-color="false" data-ng-if="item.value"></div>
```
redconfetti: @nkutub - We don't want a [space before the dollar sign](https://github.com/ets-berkeley-edu/calcentral/blob/ff2e8adf/src/assets/javascripts/angular/directives/amountDirective.js#L11)?

nkutub: @redconfetti we do and the default is true.

PR #6959, opened 2018-01-23
SISRP-38224 - On New Admit Resource Card > Your Financial Aid & Scholarships Awards - Link under Financial Aid & Scholarships section should take you directly to financial page

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-38224

* Link to latest-aid-year financial aid details page if aid years exist, otherwise, link to the "My Finances" page.

### DIFF
--- a/app/models/campus_solutions/new_admit_resources.rb
+++ b/app/models/campus_solutions/new_admit_resources.rb
@@ -69,7 +69,7 @@ module CampusSolutions
 
     def parse_fin_aid_links(fin_aid_links)
       fin_aid_links[:finAidAwards] = {
-         url: '/finances',
+         url: calcentral_financial_aid_link,
          linkDescription: 'View your estimated cost of attendance and financial aid awards.',
          showNewWindow: false,
          name: 'Your Financial Aid & Scholarships Awards',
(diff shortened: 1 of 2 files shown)


app/models/campus_solutions/new_admit_resources.rb
```diff
@@ -103,6 +103,23 @@ def parse_first_year_pathways_links(pathways_links)
...
+    end
+
+    def get_latest_aid_year
+      aid_years = []
+      aid_years_feed = CampusSolutions::MyAidYears.new(@uid).get_feed
+      aid_years_feed.try(:[], :feed).try(:[], :finaidSummary).try(:[], :finaidYears).try(:each) do |aid_year_obj|
+        aid_years.push(aid_year_obj.try(:[], :id).try(:to_i))
+      end
+      aid_years.max
+    end
```
redconfetti: @davidlimcheng - This isn't a blocker, but I'm curious why we don't put logic like this into CampusSolutions::MyAidYears as some sort of class method that other models could share? You could even present one as a method that returns all aid years, and another one that uses that one and simply returns the max.

davidlimcheng: Good question.  The [MyAidYears model](https://github.com/ets-berkeley-edu/calcentral/blob/master/app/models/campus_solutions/my_aid_years.rb) has a similar method to get the **default** aid year which is used across CalCentral, but nowhere else is the **latest** aid year being used.  If there were multiple places that were using it, I think it'd totally make sense to add an additional method called `max_aid_year` to the MyAidYears model - but until then, I figured it'd make more sense to leave that work here.

lyttam: @davidlimcheng @redconfetti I know I'm repeating myself, but I'm against making bloated models with tons of methods on them.  I'd rather stick a method like that in a concern, where it could potentially be shared by more than one model.

PR #6965, opened 2018-01-24
[JQA] CLC-6803 Work around new bug in Canvas Course User API

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6803

QA-first, since Junction-Dev is a bit behind the current master branch.


app/models/canvas/course_user.rb
```diff
@@ -16,7 +17,25 @@ def initialize(options = {})
...
+        feed = response[:body]
+        if feed['enrollments']
+          # CLC-6803 : An Instructure bug may have injected enrollments from other site memberships into this response.
+          # If so, weed them out.
+          fixed_enrollments = []
+          feed['enrollments'].each do |enr|
+            if enr['course_id'] == @course_id
+              fixed_enrollments << enr
+            end
+          end
```
paulkerschen: @raydavis FWIW I would tend to do this kind of thing with a one-liner:

`fixed_enrollments = feed['enrollments'].select { |enr| enr['course_id'] == @course_id }`

Is this filtering worth a test, or are we hoping it will only be a short-term guest in the codebase?

raydavis: Agreed on the one-liner, @paulkerschen -- I didn't wait to recover my Ruby legs. I'll amend.

My main concern is how the code tests against the Beta Canvas server, but I'll see if an automated test is easy to drop in.

raydavis:
PR amended -- thanks for the review, @paulkerschen !

PR #6975, opened 2018-01-29
SISRP-36883 - Check for nil params being passed to EDODB queries

lyttam:
https://jira.berkeley.edu/browse/SISRP-36883

### DIFF
--- a/spec/models/edo_oracle/queries_spec.rb
+++ b/spec/models/edo_oracle/queries_spec.rb
@@ -1,5 +1,5 @@
-describe EdoOracle::Queries, :ignore => true do
-  # These tests go against the DB, which could change at any time, and thus is not included in Travis.
+describe EdoOracle::Queries do
+
   # Stubbing terms not available in TestExt env
   let(:summer_2016_db_term) do
     {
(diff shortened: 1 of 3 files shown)


spec/models/edo_oracle/queries_spec.rb
```diff
@@ -47,195 +47,196 @@
...
-    let(:term) { Berkeley::Terms.fetch.campus['spring-2010'] }
-    let(:uid) { '30' }
-    it 'fetches expected data' do
-      results = EdoOracle::Queries.get_instructing_sections(uid, [term])
-      expect(results.count).to eq 17
-      expected_keys = %w(course_title course_title_short dept_name catalog_id primary section_num instruction_format primary_associated_section_id catalog_root catalog_prefix catalog_suffix enroll_limit waitlist_limit)
-      results.each do |result|
-        expect(result['term_id']).to eq '2102'
-        expect(result).to have_keys(expected_keys)
+  context 'when hitting the database', :ignore => true do
```
davidlimcheng: @lyttam, did you mean to leave the `:ignore` tag here?

lyttam: @davidlimcheng yes - let me know if you agree. I wanted to prevent the testext tests from running on bamboo, but leave them here for future reference.

davidlimcheng: @lyttam I don't think I feel very strongly either way, but it seems like the testext tests have been more problematic than helpful, so I'm inclined to agree.

PR #6976, opened 2018-01-30
NOJIRA - Fixes academic records error and updates SISEDO view checker

lyttam:
1.  Fixes the following error by replacing `nil` with an empty Hash.
```[2018-01-30 10:23:08] [FATAL] 
NoMethodError (undefined method `delete' for nil:NilClass):
  app/models/my_academics/academic_records.rb:19:in `get_official_transcript_request_data'
  app/models/my_academics/academic_records.rb:8:in `get_feed_internal'
```

2.  Adds some new views to the SISEDO view testing script.

### DIFF
--- a/app/models/edo_oracle/view_checker.rb
+++ b/app/models/edo_oracle/view_checker.rb
@@ -134,6 +134,14 @@ class EdoOracle::ViewChecker
       ]
     },
     {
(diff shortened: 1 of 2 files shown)


davidlimcheng:
Nice!  Regression testing should catch anything, so I don't think we need to create a JIRA for this.

PR #6977, opened 2018-02-01
SISRP-37719 - Creates new holds feed and endpoint, removes disused CS holds code

lyttam:
https://jira.berkeley.edu/browse/SISRP-37719

This is part 1 of an attempt clean up the MyAcademicStatus feed (see #1 in the Jira description).  The entire academic status feed was being requested by the front end just to find out if the user has holds.  I created a separate feed containing the holds data by itself.

I also found some dead code that wasn't being referenced anywhere in the front end - only by the Selenium scripts.  I removed all that code and the corresponding feature flag, but left the scripts intact.  They will probably fail if we run them, but as documentation I think they are better than nothing.

I welcome opinions on whether I should have done something different with the {{cs_holds}} feature flag or the Selenium scripts.


app/models/my_academics/my_holds.rb
```diff
@@ -0,0 +1,35 @@
...
+        errored: academic_status.try(:[], :errored),
+        statusCode: academic_status.try(:[], :statusCode)
+      }
+    end
+
+    def parse_holds(academic_status)
+      hub_response = academic_status.try(:[], :feed).try(:[], 'student').try(:[], 'holds') || []
+      hub_response.map do |hub_hold|
+        {
+          'reason' => hub_hold.try(:[], 'reason'),
```
davidlimcheng: @lyttam Just curious - is there a reason why you went with strings here instead of symbols for the keys?

spec/ui_selenium/pages/api_cs_holds_page.rb
```diff
@@ -3,6 +3,11 @@ class ApiCSHoldsPage
   include PageObject
   include ClassLogger
 
+  def initialize(root, visit=false)
+    logger.warn 'DEPRECATION WARNING: ApiCSHoldsPage is deprecated. The CampusSolutions::Holds feed is not in use and the route \'/api/campus_solutions/holds\' has been deleted.'
```
davidlimcheng: Nice!

src/assets/javascripts/angular/factories/holdsFactory.js
```diff
@@ -1,22 +1,36 @@
...
-  var urlHolds = '/api/campus_solutions/holds';
-  // var urlHolds = '/dummy/json/holds_empty.json';
+  var urlHolds = '/api/my/holds';
+  // var urlHolds = '/dummy/json/holds_errored.json';
   // var urlHolds = '/dummy/json/holds_present.json';
   var urlAdvisingStudentHolds = '/api/advising/holds/';
   // var urlAdvisingStudentHolds = '/dummy/json/holds_present.json';
 
-  var getHolds = function(options) {
+  var parseHolds = function(response) {
```
davidlimcheng: @lyttam I don't think I've seen this pattern of parsing feeds in the Factory layer in our code base before.  I'm not totally against it, but I also don't think we should have too much logic in the Factory layer controlling what the feed looks like before it gets to the Controller.  Are there any particular advantages to having this here instead of the Controller?

lyttam: @davidlimcheng Good point. There are some factories that do this kind of parsing (activityFactory, enrollmentFactory, campusLinksFactory, etc) but I like the idea of keeping factories simple.  The only time I see an advantage is if some extra fancy parsing is needed by multiple controllers.  Then it might make sense to put that work in the factory so that it isn't repeated in each controller.

app/models/my_academics/my_holds.rb
lyttam: @davidlimcheng No, symbols would be better. Will make the change.

lyttam:
@davidlimcheng Thanks for the feedback. I made the changes you suggested.

davidlimcheng:
Thx @lyttam, looks good!

PR #6978, opened 2018-02-02
SISRP-38358 - Allows Housing card to display to advisors and delegates

lyttam:
https://jira.berkeley.edu/browse/SISRP-38358

I address two separate bugs here - thus the two commits.

1.  The housing card showed an infinite spinner
2.  The `CampusSolutions::FinancialAidHousing` data wasn't being added to the `MyFinancialAidFilteredForAdvisor` or `MyFinancialAidFilteredForDelegate` feeds

### DIFF
--- a/app/models/campus_solutions/my_financial_aid_data.rb
+++ b/app/models/campus_solutions/my_financial_aid_data.rb
@@ -3,7 +3,6 @@ module CampusSolutions
 
     include ClassLogger
     include Cache::CachedFeed
-    include Cache::JsonifiedFeed
     include Cache::UserCacheExpiry
     include Cache::RelatedCacheKeyTracker
     include CampusSolutions::FinaidFeatureFlagged
--- a/app/models/campus_solutions/my_financial_aid_filtered_for_advisor.rb
+++ b/app/models/campus_solutions/my_financial_aid_filtered_for_advisor.rb
@@ -15,11 +15,9 @@ module CampusSolutions
     end
 
     def get_feed_internal
-      if is_feature_enabled && (self.aid_year ||= CampusSolutions::MyAidYears.new(@uid).default_aid_year)
-        apply_filter CampusSolutions::FinancialAidData.new(user_id: @uid, aid_year: aid_year).get
-      else
-        {}
-      end
+      model = CampusSolutions::MyFinancialAidData.new(@uid)
+      model.aid_year = aid_year
+      apply_filter model.get_feed
     end
 
     def apply_filter(feed)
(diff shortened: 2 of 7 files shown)


app/models/campus_solutions/my_financial_aid_data.rb
```diff
@@ -3,7 +3,6 @@ class MyFinancialAidData < UserSpecificModel
 
     include ClassLogger
     include Cache::CachedFeed
-    include Cache::JsonifiedFeed
```
lyttam: This data gets converted to JSON later on, at the controller level.

app/models/campus_solutions/my_financial_aid_filtered_for_advisor.rb
```diff
@@ -15,11 +15,9 @@ def get_feed_as_json(force_cache_write=false)
     end
 
     def get_feed_internal
-      if is_feature_enabled && (self.aid_year ||= CampusSolutions::MyAidYears.new(@uid).default_aid_year)
-        apply_filter CampusSolutions::FinancialAidData.new(user_id: @uid, aid_year: aid_year).get
-      else
-        {}
-      end
+      model = CampusSolutions::MyFinancialAidData.new(@uid)
+      model.aid_year = aid_year
```
davidlimcheng: @lyttam Previously, the call to `CampusSolution::MyAidYears` was explicit - where is it being called after this change?  I see it being called in `CampusSolutions::MyFinancialAidData`, but I'm having a hard time finding where it is defined in the current scope.

lyttam: @davidlimcheng  It's done on line 13 in `CampusSolutions::MyFinancialAidData`. It's still explicit; it's just being done upstream.

davidlimcheng: Ah, that still leaves me a bit confused on this line.  Why do we need to reassign the `aid_year` variable in `model` here?

lyttam: @davidlimcheng By assigning it, we are passing the value of the `aid_year` parameter to `MyFinancialAidData`.  If we didn't do that, `self.aid_year` on line 13 would always be nil and we would set it to the `default_aid_year` provided by `CampusSolutions::MyAidYears`.

davidlimcheng: I was confused by the lack of `aid_year` definition, but I now see that it is coming from the controller level as part of the request params.  Makes sense now - thanks!

PR #6979, opened 2018-02-06
SISRP-38229 - Adds validation to address posting proxy

lyttam:
https://jira.berkeley.edu/browse/SISRP-38229

New `valid?` method can be overridden to do specific validation in any class inheriting from `PostingProxy`.  I imagine we'll only need to do validation on POST requests.  In this case, we validate the address type against a whitelist because CS apparently does not do this type of validation, nor does the API.

### DIFF
--- a/src/assets/javascripts/angular/controllers/widgets/profile/profileAddressController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/profile/profileAddressController.js
@@ -167,7 +167,8 @@ angular.module('calcentral.controllers').controller('ProfileAddressController', 
 
     apiService.profile
       .save($scope, profileFactory.postAddress, merge)
-      .then(saveCompleted);
+      .then(saveCompleted)
+      .catch(saveCompleted);
   };
 
   $scope.showAdd = function() {
--- a/app/controllers/campus_solutions/address_controller.rb
+++ b/app/controllers/campus_solutions/address_controller.rb
@@ -1,5 +1,6 @@
 module CampusSolutions
   class AddressController < CampusSolutionsController
+    rescue_from Errors::ClientError, with: :handle_client_error
 
     before_filter :exclude_acting_as_users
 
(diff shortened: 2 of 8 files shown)


src/assets/javascripts/angular/controllers/widgets/profile/profileAddressController.js
```diff
@@ -167,7 +167,8 @@ angular.module('calcentral.controllers').controller('ProfileAddressController',
 
     apiService.profile
       .save($scope, profileFactory.postAddress, merge)
-      .then(saveCompleted);
+      .then(saveCompleted)
+      .catch(saveCompleted);
```
davidlimcheng: @lyttam I see that `saveCompleted` eventually calls `actionCompleted`, which handles the error:

```
  var actionCompleted = function($scope, response, callback) {
    if (response.data.errored) {
      $scope.errorMessage = _.get(response, 'data.feed.errmsgtext') || 'An error occurred while saving your data.';
    } else {
      $scope.closeEditor();
      callback({
        refresh: true
      });
    }
  };
```
This is not a blocker for merging, but for the future, what are your thoughts on creating separate functions - one to handle a resolved promise and one to handle a rejected one?  I don't see anything wrong with the current flow, but the error handling is implicit, and requires some follow-up to validate that the error is actually being dealt with.  Creating a separate function with an appropriate name could explicitly show the error-handling at this layer, instead of having to go to `profileService.js` to validate.

lyttam: @davidlimcheng yeah, I would support that idea. The reason why this code is weird is because if there is an error in CS, we get a 200 response back, which causes the `.then()` to get executed - so `saveCompleted` must include the error handling code.

The validation I added will send a 400 response if it fails, which causes the `.catch()` to get exectued. I want the same error handling to happen in that case.

We could move the error handling logic to its own function which could be called in two places - inside `actionCompleted` and inside the `.catch()`.  That will affect more than just  this one controller, but I'm probably going to have to update most of these controllers that post profile information anyway, so maybe I'll do it in a future PR.

davidlimcheng: Makes sense!

PR #6981, opened 2018-02-06
CLC-6775 Remove empty object from OEC controller scope

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6775

Why was it ever in there? Why did it work?


raydavis:
Those are good questions, @paulkerschen . Let's hope we never find an answer.

PR #6983, opened 2018-02-07
SISRP-38178 - Fixes race condition on Advising card

lyttam:
https://jira.berkeley.edu/browse/SISRP-38178

This will go in an ER.  I will submit a QA PR once I verify in cc-dev that the fix worked.


nkutub:
LGTM @lyttam

PR #6984, opened 2018-02-07
SISRP-38178 - Provides missing academic roles to advisor-facing Advising card

lyttam:
https://jira.berkeley.edu/browse/SISRP-38178

My previous PR #6983 didn't go far enough, so this is a follow-up.  The Advising card was loading before the page-level controller (`userOverviewController`) was finished fetching the student's academic roles.  I added an explicit call to get the roles when the card is being viewed by the student's advisor.

### DIFF
--- a/src/assets/javascripts/angular/controllers/widgets/myAdvisingController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/myAdvisingController.js
@@ -6,7 +6,7 @@ var _ = require('lodash');
 /**
  * My Advising controller
  */
-angular.module('calcentral.controllers').controller('MyAdvisingController', function(myAdvisingFactory, apiService, $route, $routeParams, $scope) {
+angular.module('calcentral.controllers').controller('MyAdvisingController', function(myAdvisingFactory, advisingFactory, apiService, $route, $routeParams, $scope, $q) {
   $scope.myAdvising = {
     isLoading: true
   };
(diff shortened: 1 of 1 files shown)


src/assets/javascripts/angular/controllers/widgets/myAdvisingController.js
```diff
@@ -29,16 +29,33 @@ angular.module('calcentral.controllers').controller('MyAdvisingController', func
     $scope.showAppointmentLinks = apiService.user.profile.features.csAdvisingLinks && _.get(response, 'data.feed.links');
   };
 
-  var loadStudentAcademicRoles = function() {
-    var isAdvisingStudentLookup = $route.current.isAdvisingStudentLookup;
-    var academicRoles = isAdvisingStudentLookup ? $scope.targetUser.academicRoles : apiService.user.profile.academicRoles;
+  var getStudentAcademicRoles = function() {
+    var params;
```
davidlimcheng: @lyttam Since `params` and `deferred` never leave the scope of their respective `if`/`else` statements, can we remove their declarations here?

lyttam: @davidlimcheng Actually, they are here for a reason - to avoid the variable hoisting problem.  It looks bad but I think it's a best practice.  https://stackoverflow.com/questions/3725546/variable-hoisting

davidlimcheng: I had no clue that was a thing - thanks for the share!

PR #6986, opened 2018-02-08
SISRP-37758 - Grading Links at Term Change - Dev for CLC

redconfetti:
https://jira.berkeley.edu/browse/SISRP-37758

* Updates the semester shown by default inside of the [Semester Classes card](https://confluence.ets.berkeley.edu/confluence/display/MYB/Semester+Classes+Widget) so that it prefers the semester that has grading in progress first, current semester second, the next upcoming semester third, and whatever semester is found that doesn't meet any of that criteria as a fallback.

### DIFF
--- a/src/assets/javascripts/angular/services/academicsService.js
+++ b/src/assets/javascripts/angular/services/academicsService.js
@@ -6,25 +6,26 @@ var angular = require('angular');
 
 angular.module('calcentral.services').service('academicsService', function() {
 
-  // Selects the semester of most pressing interest. Choose the current semester if available.
+  // Selects the semester of most pressing interest.
+  // Choose the semester with grading in progress
+  // Otherwise choose the current semester, if available.
   // Otherwise choose the next semester in the future, if available.
-  // Otherwise,choose the most recent semester.
+  // Otherwise choose the most recent semester.
   var chooseDefaultSemester = function(semesters) {
-    var oldestFutureSemester = false;
-    for (var i = 0; i < semesters.length; i++) {
-      if (semesters[i].timeBucket === 'current') {
-        return semesters[i];
-      } else if (semesters[i].timeBucket === 'future') {
-        oldestFutureSemester = semesters[i];
-      } else {
-        if (oldestFutureSemester) {
-          return oldestFutureSemester;
-        } else {
-          return semesters[i];
-        }
-      }
-    }
-    return oldestFutureSemester;
+    var groupedSemesters = _.groupBy(semesters, 'timeBucket');
+    var gradingInProgress = _.find(semesters, ['gradingInProgress', true]);
+
+    // Create array with terms objects in preferred order
+    // (grading in progress, current, most recent future, most recent past)
+    var sortedSemesters = [
+      gradingInProgress,
+      _.head(groupedSemesters.current),
+      _.last(groupedSemesters.future),
+      _.head(groupedSemesters.past)
+    ];
+
+    // return first present in list
+    return _.head(_.compact(sortedSemesters));
   };
 
   var countSectionItem = function(selectedCourse, sectionItem) {


src/assets/javascripts/angular/services/academicsService.js
```diff
@@ -6,18 +6,25 @@ var angular = require('angular');
...
+      } else if (semesters[i].timeBucket === 'current') {
+        currentSemester = semesters[i];
       } else if (semesters[i].timeBucket === 'future') {
         oldestFutureSemester = semesters[i];
       } else {
-        if (oldestFutureSemester) {
+        if (currentSemester) {
+          return currentSemester;
+        } else if (oldestFutureSemester) {
           return oldestFutureSemester;
```
redconfetti: Do you think this would be good to reword as `upcomingSemester`?

src/assets/javascripts/angular/services/academicsService.js
```diff
@@ -6,18 +6,25 @@ var angular = require('angular');
 
 angular.module('calcentral.services').service('academicsService', function() {
 
-  // Selects the semester of most pressing interest. Choose the current semester if available.
+  // Selects the semester of most pressing interest.
+  // Choose the semester with grading in progress
+  // Otherwise choose the current semester, if available.
   // Otherwise choose the next semester in the future, if available.
-  // Otherwise,choose the most recent semester.
+  // Otherwise choose the most recent semester.
```
lyttam: @redconfetti Are the semesters guaranteed to be in order?  Will the first one always be the one where grading is in progress, if it exists?  Your algorithm will return the first semester it finds that meets any of the criteria (grading in progress, current, future) - which might not be the one you want, if they aren't in the right order.

Also, I think you can effectively shorten this implementation and accomplish the same thing:
```
var chooseDefaultSemester = function(semesters) {
  for (var i = 0; i < semesters.length; i++) {
    if (semesters[i].gradingInProgress || semesters[i].timeBucket === 'current' || semesters[i].timeBucket === 'future') {
      return semesters[i];
    } 
  }
}
```
Again, this only works if the order of semesters is guaranteed.

src/assets/javascripts/angular/services/academicsService.js
davidlimcheng: How about `nextSemester`, to keep with the Terms API verbiage?

redconfetti: Actually, I see why it uses that term, because the state of it changes as it iterates over the semesters.

src/assets/javascripts/angular/services/academicsService.js
redconfetti: @lyttam

The `semesters` originate from the [teachingSemesters](https://github.com/ets-berkeley-edu/calcentral/blob/fc990a89d/src/assets/javascripts/angular/controllers/pages/academicsController.js#L204) node of the /api/my/academics JSON response. Those originate from [EdoOracle::UserCourses::All.get_all_campus_courses](https://github.com/ets-berkeley-edu/calcentral/blob/fc990a89d/app/models/my_academics/teaching.rb#L6,L11). The instructing sections obtained for grading are provided by [EdoOracle::UserCourses::Base#merge_instructing](https://github.com/ets-berkeley-edu/calcentral/blob/fc990a89d/app/models/edo_oracle/user_courses/all.rb#L12). This obtains data from [EdoOracle::Queries.get_instructing_sections](https://github.com/ets-berkeley-edu/calcentral/blob/fc990a89d/app/models/edo_oracle/user_courses/base.rb#L36). It turns out that [#get_instructing_sections](https://github.com/ets-berkeley-edu/calcentral/blob/fc990a89d/app/models/edo_oracle/queries.rb#L110,L133) does order by term ID descending.

The issue with your logic however is that it's wanting the last of the 'future' semesters if one in grading or a current one isn't present. And it still needs to fallback on the first one it encounters from the remaining category 'past'... which because of the ordering happens to be the most recent past term.

It seems the way it's coded now is the simplest way to implement this logic without having to iterate over the array more than once, and stopping execution as soon as a relevant one is found.

lyttam: @redconfetti thanks for checking the ordering.

IMO, the code as it is now looks overly complicated.  This isn't an issue of performance, but of code being too verbose.  If the goal can be accomplished in fewer `if` statements, I would prefer that.

redconfetti: @lyttam - I've updated the code with the new algorithm we discussed.

src/assets/javascripts/angular/services/academicsService.js
```diff
@@ -6,25 +6,26 @@ var angular = require('angular');
...
-    return oldestFutureSemester;
+    var groupedSemesters = _.groupBy(semesters, 'timeBucket');
+    var gradingInProgress = _.find(semesters, ['gradingInProgress', true]);
+
+    // Create array with terms objects in preferred order
+    // (grading in progress, current, most recent future, most recent past)
+    var sortedSemesters = [
+      gradingInProgress,
+      _.head(groupedSemesters.current),
+      _.head(_.reverse(groupedSemesters.future)),
```
davidlimcheng: According to the Lodash documentation:

"The order of grouped values is determined by the order they occur in collection. "

If we want to get the "most recent future semester", would we need the `reverse` here?

redconfetti: @davidlimcheng - The semesters are provided by the API in descending order for the Term ID. This means from the future to the past.

After isolating the 'future' semesters, I want the one from the bottom of the list. That is the one that falls right AFTER the 'current' semester.

I just noticed that it would be simpler if I just use [_.last()](https://lodash.com/docs/4.17.5#last). I'll update that now.

davidlimcheng:
Nice!

PR #6987, opened 2018-02-09
SISRP-38230 - Adds validation to email address update

lyttam:
https://jira.berkeley.edu/browse/SISRP-38230

Also includes some refactoring of the JavaScript response handlers at @davidlimcheng's request

### DIFF
--- a/app/models/campus_solutions/user_api_updating_model.rb
+++ b/app/models/campus_solutions/user_api_updating_model.rb
@@ -2,7 +2,7 @@ module CampusSolutions
   module UserApiUpdatingModel
     def passthrough(model_name, params)
       proxy = model_name.new({user_id: @uid, params: params})
-      result = proxy.get
+      result = proxy.post
       UserApiExpiry.expire @uid
       result
     end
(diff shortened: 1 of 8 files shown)


app/models/campus_solutions/user_api_updating_model.rb
```diff
@@ -2,7 +2,7 @@ module CampusSolutions
   module UserApiUpdatingModel
     def passthrough(model_name, params)
       proxy = model_name.new({user_id: @uid, params: params})
-      result = proxy.get
+      result = proxy.post
```
redconfetti: @lyttam - How did we switch from using HTTP GET to HTTP POST? We should use POST or PUT for updating obviously, but does the Campus Solutions API support either HTTP method?

spec/controllers/campus_solutions/email_controller_spec.rb
```diff
@@ -17,7 +17,7 @@
         post :post,
              {
                bogus_field: 'abc',
-               type: 'CAMP',
+               type: 'OTHR',
```
redconfetti: This makes me wonder if we originally planned on making the campus email updatable, but then that was scratched, but we kept it mocked up as such anyway in our spec.

PR #6992, opened 2018-02-13
SISRP-38682 - Remove GPA Estimator from CalCentral and place message in CLC to students

nkutub:
https://jira.berkeley.edu/browse/SISRP-38682

### DIFF
--- a/src/assets/templates/academics_gpa.html
+++ b/src/assets/templates/academics_gpa.html
@@ -2,7 +2,17 @@
   <div class="cc-widget-title">
     <h2>GPA Calculator</h2>
   </div>
+  <!-- temp show ths message until issues with GPA calculator are fixed SISRP-38000 -->
   <div class="cc-widget-padding">
+    <p class="cc-left">
+      <span class="cc-widget-padding">
+        <i class="fa fa-wrench fa-2x"></i>
+        <div>The GPA Calculator is currently under construction. We apologize for any inconvenience.</div>
+      </span>
+    </p>
+  </div>
+  <!-- temp hide ths card until issues with GPA calculator are fixed SISRP-38000 -->
+  <div class="cc-widget-padding" data-ng-if="false">
     <div>
       <h4>Estimate your grades for <span data-ng-bind="selectedSemester.name"></span></h4>
       <div class="cc-table" data-ng-if="selectedCourses.length">


src/assets/templates/academics_gpa.html
```diff
@@ -2,7 +2,17 @@
   <div class="cc-widget-title">
     <h2>GPA Calculator</h2>
   </div>
+  <!-- temp show ths message until issues with GPA calculater are fixed SISRP-38000 -->
   <div class="cc-widget-padding">
+    <p class="cc-left">
+    <span class="cc-widget-padding">
```
davidlimcheng: @nkutub  This `span` should be indented

src/assets/templates/academics_gpa.html
```diff
@@ -2,7 +2,17 @@
...
+      <i class="fa fa-wrench fa-2x"></i>
+      <div>The GPA card is currently under construction. We apologize for any inconvenience.</div>
+    </span>
+    </p>
+  </div>
+  <!-- temp hide ths card until issues with GPA calculater are fixed SISRP-38000 -->
+  <div class="cc-widget-padding" data-ng-if="false">
```
davidlimcheng: @nkutub - Instead of directly inserting a boolean into the front-end here, what do you think about creating a feature flag instead that flips between the two "versions" of the card?  I know that the chances of flipping the feature back on before a fix is extremely low, but the extra work to pull from a feature flag instead will give us more flexibility in case anything happens.

nkutub: @davidlimcheng, I thought of that, but all the extra work will most likely not be used as feature flag, there are major issues with card and will need code to fix before we show it. The JIRA was already assigned to next sprint to fix and is high priority.

src/assets/templates/academics_gpa.html
nkutub: @davidlimcheng done!

lyttam:
@nkutub Suggestion: change the text of the message to say "GPA Calculator" rather than "GPA card".  It's more correct, and "card" is an abstraction we use in our code - not something our users know about.

`The GPA card is currently under construction. We apologize for any inconvenience.`

nkutub:
@lyttam I'm just following the design https://jira.berkeley.edu/browse/SISRP-38733.

lyttam:
@nkutub I know; I'm suggesting we deviate from the design.

davidlimcheng:
@nkutub @lyttam The design actually says "GPA Calculator"

image

nkutub:
oh crap!, I thought It was opposite. I thought I had Calculator. My bad. I will update.

nkutub:
@davidlimcheng , card changed to calculator.

davidlimcheng:
Looks good, @nkutub

PR #6993, opened 2018-02-13
SISRP-38585 - FA Aid Year Specific Page Redirect

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-38585

* `redirectToPage()` falls back to `redirectToHome()` if for some reason the page isn't available to the user.
* I had to do two redirects here because `finaidController` and `finaidSummaryController` both have code that can lead to a redirect.  Unfortunately, I had to leave them both in because there is very little overlap in the cases in which they are activated.

### DIFF
--- a/src/assets/javascripts/angular/controllers/pages/finaidController.js
+++ b/src/assets/javascripts/angular/controllers/pages/finaidController.js
@@ -22,10 +22,6 @@ angular.module('calcentral.controllers').controller('FinaidController', function
     blank: 'blank'
   };
 
-  /**
-   * Set the current finaid year
-   * If we don't receive a finaid year through the route params, use the default aid year
-   */
   var setFinaidYear = function(data, finaidYearId) {
     if (finaidYearId) {
       $scope.finaidYear = finaidService.findFinaidYear(data, finaidYearId);
@@ -35,27 +31,19 @@ angular.module('calcentral.controllers').controller('FinaidController', function
     }
   };
 
-  /**
-   * See whether the finaid year exist, otherwise, send them to the 404 page
-   */
   var combinationExists = function(data, finaidYearId) {
     var combination = finaidService.combinationExists(data, finaidYearId);
-
-    // If no correct finaid year comes back, make sure to send them to the 404 page.
     if (!combination) {
-      apiService.util.redirect('404');
+      apiService.user.redirectToPage('finances');
       return false;
     }
   };
 
-  /**
-   * Get the finaid summary information
-   */
   var getFinaidSummary = function(options) {
     return finaidFactory.getSummary(options).then(
       function successCallback(response) {
-        setFinaidYear(response.data.feed, $routeParams.finaidYearId);
         combinationExists(response.data.feed, _.get($scope, 'finaidYear.id'));
+        setFinaidYear(response.data.feed, $routeParams.finaidYearId);
         $scope.finaidSummary = response.data.feed.finaidSummary;
         $scope.finaid.isLoading = false;
       }
--- a/src/assets/javascripts/angular/services/finaidService.js
+++ b/src/assets/javascripts/angular/services/finaidService.js
@@ -3,7 +3,7 @@
 var _ = require('lodash');
 var angular = require('angular');
 
-angular.module('calcentral.services').service('finaidService', function($rootScope) {
+angular.module('calcentral.services').service('finaidService', function($rootScope, userService) {
   var options = {
     finaidYear: false
   };
@@ -38,7 +38,11 @@ angular.module('calcentral.services').service('finaidService', function($rootSco
   var setDefaultFinaidYear = function(data, finaidYearId) {
     if (data && data.finaidSummary && data.finaidSummary.finaidYears) {
       if (finaidYearId) {
-        setFinaidYear(findFinaidYear(data, finaidYearId));
+        if (combinationExists(data, finaidYearId)) {
+          setFinaidYear(findFinaidYear(data, finaidYearId));
+        } else {
+          userService.redirectToPage('finances');
+        }
       } else {
         // If no aid year has been selected before, select the default one
         var finaidYear = findDefaultFinaidYear(data.finaidSummary.finaidYears);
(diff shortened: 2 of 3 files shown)


src/assets/javascripts/angular/controllers/pages/finaidController.js
```diff
@@ -36,14 +36,12 @@ angular.module('calcentral.controllers').controller('FinaidController', function
   };
 
   /**
-   * See whether the finaid year exist, otherwise, send them to the 404 page
+   * See whether the finaid year exist, otherwise, send them to the finances page
```
lyttam: @davidlimcheng I think this comment is unnecessary; the code explains itself.

src/assets/javascripts/angular/services/finaidService.js
```diff
@@ -38,7 +38,11 @@ angular.module('calcentral.services').service('finaidService', function($rootSco
   var setDefaultFinaidYear = function(data, finaidYearId) {
     if (data && data.finaidSummary && data.finaidSummary.finaidYears) {
       if (finaidYearId) {
-        setFinaidYear(findFinaidYear(data, finaidYearId));
+        if (combinationExists(data, finaidYearId)) {
+          setFinaidYear(findFinaidYear(data, finaidYearId));
+        } else {
+          userService.redirectToPage('finances');
+        }
```
lyttam: @davidlimcheng This looks like the same logic that you added to FinaidController. Why is it repeated here?

src/assets/javascripts/angular/controllers/pages/finaidController.js
```diff
@@ -54,8 +52,8 @@ angular.module('calcentral.controllers').controller('FinaidController', function
   var getFinaidSummary = function(options) {
     return finaidFactory.getSummary(options).then(
       function successCallback(response) {
-        setFinaidYear(response.data.feed, $routeParams.finaidYearId);
         combinationExists(response.data.feed, _.get($scope, 'finaidYear.id'));
+        setFinaidYear(response.data.feed, $routeParams.finaidYearId);
```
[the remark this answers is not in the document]
davidlimcheng: @nkutub Sure - `combinationExists()` is a quick check to see if the aid year is acceptable - if not, it will redirect to the `/finances` path.  If we do redirect, then everything done in `setFinaidYear()` is unnecessary, so I put the `combinationExists()` check above it for those cases.

src/assets/javascripts/angular/services/finaidService.js
davidlimcheng: @lyttam Yes - it's pretty funky stuff that's going on.

[finaidSummaryController.js:43](https://github.com/ets-berkeley-edu/calcentral/blob/master/src/assets/javascripts/angular/controllers/widgets/finaid/finaidSummaryController.js#L43) is code that gets run from [finaid_summary.html](https://github.com/ets-berkeley-edu/calcentral/blob/master/src/assets/templates/widgets/finaid_summary.html), which is a child card of [finaid.html](https://github.com/ets-berkeley-edu/calcentral/blob/master/src/assets/templates/finaid.html), which has two controllers defined for it - one on the route level in [routeConfiguration.js:86](https://github.com/ets-berkeley-edu/calcentral/blob/master/src/assets/javascripts/angular/configuration/routeConfiguration.js#L86) and one on the page level in [finaid.html:1](https://github.com/ets-berkeley-edu/calcentral/blob/master/src/assets/javascripts/angular/configuration/routeConfiguration.js#L86).  Both of these controllers also have redirect logic.

So in short, there are multiple controllers with `redirect` logic that is kicked off due to different actions.  For this particular PR, the logic that I want to kick off a `redirect` to the `/finances` page is:

* User goes to path `/finances/finaid/{aid_year}` where `aid_year` is a nonexistent aid year for that student

This requires a change in `FinaidController` (as you mentioned), which checks (at the page level) the validity of the aid year.  
The change in `finaidService.js` (this code) is required because code is also kicked off in [`finaidSummaryController.js:43`](https://github.com/ets-berkeley-edu/calcentral/blob/master/src/assets/javascripts/angular/configuration/routeConfiguration.js#L86) when a student tries to access a non-existent aid year via route parameters that ultimately leads to a redirect.  Before this change, `findFinaidYear()` would equal to `nil` when given an invalid aid year, which this function would lead to a `$broadcast` that would kick off a ton more invalid actions, and CalCentral would essentially try to show a bunch of bad information from a bad aid year.  This change stops the process immediately when the aid year is bad, and redirects to `/finances` instead.

The architecture of what's going on in the front-end can definitely be cleaned up. There are essentially two "is aid year valid?" checks when given an aid year via route param - one at the page level and one at the card level.  I think it definitely requires a more pre-planned, extensive cleanup.

src/assets/javascripts/angular/controllers/pages/finaidController.js
davidlimcheng: Agreed - will amend.

davidlimcheng: @lyttam Removed comment, also removed other comments in this controller that I thought were  unnecessary.

src/assets/javascripts/angular/services/finaidService.js
lyttam: @davidlimcheng I see, thanks for the explanation. Perhaps it would make sense to put the logic in the service, and have the two controllers call it - but that refactoring might be out of scope for your task here.

davidlimcheng: @lyttam Yeah, I think it needs a refactor, but since the code is shared between so many front-end controllers, it will require a more extensive effort.

src/assets/javascripts/angular/controllers/pages/finaidController.js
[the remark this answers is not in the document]
nkutub: @davidlimcheng, this turned out to be a bug, you need to change line 45 from  _.get($scope, 'finaidYear.id') to $routeParams.finaidYearId since the scope is not set yet, it gets set in setFinaidYear.  I will fix in SISRP-38686 since it directly related.

PR #6997, opened 2018-02-19
SISRP-38686 - FinAId Title IV Response Page in CalCentral should NOT be Aid Year Specific

nkutub:
https://jira.berkeley.edu/browse/SISRP-38686

### DIFF
--- a/src/assets/javascripts/angular/configuration/routeConfiguration.js
+++ b/src/assets/javascripts/angular/configuration/routeConfiguration.js
@@ -106,7 +106,7 @@ angular.module('calcentral.config').config(function($routeProvider, calcentralCo
       controller: 'MyFinancesController',
       pageName: 'Financial Aid and Scholarships'
     }).
-    when('/finances/finaid/:finaidYearId?/t4', {
+    when('/finances/finaid/t4/auth', {
       templateUrl: 'finaid_t4.html',
       controller: 'MyFinancesController',
       pageName: 'Financial Aid and Scholarships'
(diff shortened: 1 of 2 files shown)


src/assets/javascripts/angular/controllers/pages/finaidController.js
```diff
@@ -22,32 +21,14 @@ angular.module('calcentral.controllers').controller('FinaidController', function
...
-      apiService.user.redirectToPage('finances');
-      return false;
-    }
-  };
-
   var getFinaidSummary = function(options) {
     return finaidFactory.getSummary(options).then(
       function successCallback(response) {
-        combinationExists(response.data.feed, _.get($scope, 'finaidYear.id'));
-        setFinaidYear(response.data.feed, $routeParams.finaidYearId);
```
davidlimcheng: Sorry @nkutub, I added a comment but I dunno if it was visible to you.  Won't be using the "Start a review" function anymore.

Why are these functions being removed?  These set the default finaid year in `finaidService`, which a lot of other cards depend on.  `combinationExists` also has important redirect logic for invalid aid years.

nkutub: There is no more need for identifying the year for this card anymore. title IV is no longer fin year specific. Also the other finaid controllers still call to these functions on other cards so there are no longer need here. @davidlimcheng

nkutub:
@davidlimcheng , I left the controller file the same per your analysis. At this point, it's the entry point for all FA activity including the title IV auth card. Since the title IV card will not pass a year parameter the controller will set the default year just in case it is referenced.

davidlimcheng:
Thanks @nkutub !

PR #6999, opened 2018-02-22
SISRP-38227 - Adds validation to Terms & Conditions

lyttam:
https://jira.berkeley.edu/browse/SISRP-38227

Prevents a user from accepting the terms and conditions for an aid year that isn't available.  For example, if they can only see data for aid years 2017 and 2018, they shouldn't be able to accept for 2019.

Also fixes a bug in FinaidController.


davidlimcheng:
LGTM

PR #7000, opened 2018-02-22
SISRP-38360 - Add link to Summer Aid Estimator in CalCentral

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-38360

* In addition to adding a link to Summer Aid Estimator, this also lays the groundwork to move the work done in [financesLinksController.js](https://github.com/davidlimcheng/calcentral/blob/master/src/assets/javascripts/angular/controllers/widgets/financesLinksController.js) to the back-end, specifically to `CampusSolutions::FinancialResourcesGeneral` and `CampusSolutions::FinancialResourcesParameterized`.
* Why two models?  `FinancialResourcesGeneral` was made for all general all-purpose links that aren't user-specific and don't need any special parameters.  Since this is only serving one link at the moment, which is already cached in `Links`, I did not put any caching mechanisms on this model.
* `FinancialResourcesParameterized` was made for all links that may be user-specific and need parameters passed in.  At the moment it's only the Summer Aid Estimator, but in the future it will also serve the EFT Enrollment data that is currently being served via `MyEftEnrollment`.
* I also reordered the Campus Solutions endpoints because they were annoying out-of-order.
* Also refactored the front-end controller to be a bit more performant.  It needs a bigger refactor, which we don't have time for at the moment.

### DIFF
--- a/app/models/campus_solutions/financial_resources_parameterized.rb
+++ b/app/models/campus_solutions/financial_resources_parameterized.rb
@@ -1,0 +1,23 @@
+module CampusSolutions
+  class FinancialResourcesParameterized < UserSpecificModel
+    include Cache::CachedFeed
+    include Cache::UserCacheExpiry
+
+    def initialize(options = {})
+      super options
+      @aid_year = options.try(:[], :aid_year)
+    end
+
+    def get_feed_internal
+      get_summer_estimator_link
+    end
+
+    def get_summer_estimator_link
+      return {} if @aid_year.nil?
+      link_config = { cs_link_key: 'UC_FA_SUMMR_EST_FLU', cs_link_params: { :AID_YEAR => @aid_year } }
+      link = LinkFetcher.fetch_link(link_config[:cs_link_key], link_config[:cs_link_params])
+      { summerEstimator: link }
+    end
+
+  end
+end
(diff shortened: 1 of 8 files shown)


app/models/campus_solutions/financial_resources_parameterized.rb
```diff
@@ -0,0 +1,23 @@
+module CampusSolutions
+  class FinancialResourcesParameterized < UserSpecificModel
```
lyttam: @davidlimcheng The UC_FA_SUMMR_EST_FLU link doesn't appear to be user specific (I don't see EMPLID being passed as a param) so in theory couldn't you just have one model that returns both links?

davidlimcheng: @lyttam What's user-specific about the link is the student's default aid year.  The default aid year can vary between students, and it's possible that some students don't have an aid year at all.  

In addition, we want to show the emergencyLoan link regardless, whether the student has an aid-year or not.  So if we have one model returning both links, we'd have to asynchronously make two calls: 

1. Get general EmergencyLoan link
2. Get aid years

Once #2 returns, we'll have to check if the student has a default aid year, and if they do, call the model again, except this time with an aid year as a parameter.  So, the model will have to return back both the emergencyLoan link, *and* the summerEstimator link.  With just two links that's not so bad, but if/when we move all the `campuslinks.json` links to the CS Link API, it'd be quite clunky to have to call the same model twice.

lyttam: @davidlimcheng Ohh, I see. Thanks for explaining!

PR #7001, opened 2018-02-23
SISRP-38681 - Show Committee Card in Adviser Dashboard - UX Development

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-38681

* Brings Committees to Advisors
* Reorders some stuff that wasn't alphabetical
* User Overview, like so many other things, is in dire need of a refactor

### DIFF
--- a/app/controllers/advising_student_controller.rb
+++ b/app/controllers/advising_student_controller.rb
@@ -9,38 +9,28 @@ class AdvisingStudentController < ApplicationController
   rescue_from Errors::ClientError, with: :handle_client_error
   rescue_from Pundit::NotAuthorizedError, with: :user_not_authorized
 
-  def profile
-    student_uid = student_uid_param
-    render json: {
-      academicRoles: MyAcademics::MyAcademicRoles.new(student_uid).get_feed,
-      attributes: User::AggregatedAttributes.new(student_uid).get_feed,
-      contacts: HubEdos::Contacts.new(user_id: student_uid, include_fields: %w(names addresses phones emails)).get,
-      residency: MyAcademics::Residency.new(student_uid).get_feed
-    }
-  end
-
   def academics
     render json: MyAcademics::FilteredForAdvisor.from_session('user_id' => student_uid_param).get_feed_as_json
+  end
+
+  def academics_cache_expiry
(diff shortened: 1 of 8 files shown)


lyttam:
@davidlimcheng Thanks for alphabetizing stuff and refactoring the front-end committees code.

One thing I notice is it looks like the studentOverviewController is looking for both student and faculty committees. Was there any discussion of whether an advisor should see the faculty committees card? That card is meant to be shown to a user who is a member of a committee.

I suspect advisors should never see faculty committee data. If that's the case, the controller should get an advisor-only version of the feed that only contains the data from `MyCommittees::StudentCommittees`. We don't want to expose the faculty data to the front-end.

davidlimcheng:
Ah, great catch @lyttam !  That makes total sense, I was in a hurry and didn't realize `Merged` had both types of data.  Seems very obvious in hindsight.  I'll amend this PR.

davidlimcheng:
@lyttam While working on this, I saw that `MyCommittees::Merged` is cached, while `MyCommittees::StudentCommittees` is not.  So, if we switch to `StudentCommittees`, we'll have to make the decision whether to cache the `StudentCommittees` data by itself or not.

According to Google Analytics over the past month, `/user/overview/:uid` got 106,222 views (1.57% of total), while `/academics` got 2,470,145 (36.44% of total).  It might actually be worth it to pull from `Merged` as opposed to `StudentCommittees` to leverage the already-existing caching.

What do you think about removing all the Faculty data parsing in the front-end, but leaving the endpoint to pull from `Merged`?

lyttam:
@davidlimcheng  The endpoint can pull from `Merged` but then it needs to remove the Faculty data before passing to the front-end.  Otherwise, a user can see that data in their browser console.

Or, you could cache the `Merged` feed and the `StudentCommittees` feed separately, and decide in the controller which one to call. {{MyAcademicsController}} does this.

davidlimcheng:
@lyttam Good point.  Amended with:

1. `facultyCommittees` is parsed out in `AdvisingStudentController`
2. Renamed the URL and controller method from `committees` to `student_committees`.

Ready for re-review.

lyttam:
Thanks!

PR #7002, opened 2018-02-23
SISRP-38829 - Adds Co-Curricular links to Student Resources

lyttam:
https://jira.berkeley.edu/browse/SISRP-38688


davidlimcheng:
@lyttam  Judging from the code, I think you linked to the correct JIRA but the PR title and branch name is incorrect?

lyttam:
@davidlimcheng No, it's correct - perhaps I should have titled my commit differently. I called it "Co-Curricular" because that's the heading of the new section where these links are displayed.

PR #7003, opened 2018-02-26
SISRP-33773 - Update CalCentral to use new IS_END_OF_DROP_ADD_TIME_PERIOD boolean

redconfetti:
https://jira.berkeley.edu/browse/SISRP-33773

We asked for the name of a boolean to be changed so that it's reflecting whether or not the time period to add or drop enrollments has ended. This is an update to correspond with that change.


lyttam:
@redconfetti Thanks for asking for this change. Much easier to understand now.

PR #7004, opened 2018-02-26
SISRP-38829 - Adds APR link to My Academics > Grad Degree Progress

lyttam:
https://jira.berkeley.edu/browse/SISRP-38587

Three versions of the link to run the APR were created specifically for Grad, Law, and Haas students.  Each of those three links can be enabled/disabled independently of the others.

- Adds two new Haas plan-based roles. 
-- Only these 2 Haas roles should see the Haas link - other Haas roles will not see a link.
-- I did not update the other places in the code that use Haas roles to take these new roles into account.
- Adds 4 new plan-based roles for Law students

### DIFF
--- a/app/models/degree_progress/my_graduate_milestones.rb
+++ b/app/models/degree_progress/my_graduate_milestones.rb
@@ -8,16 +8,48 @@ module DegreeProgress
     include MilestonesModule
     include LinkFetcher
 
+    APR_LINK_ID_HAAS = 'UC_CX_APR_RPT_GRD_STDNT_HAAS'
+    APR_LINK_ID_LAW = 'UC_CX_APR_RPT_GRD_STDNT_LAW'
+    APR_LINK_ID_GRAD = 'UC_CX_APR_RPT_GRD_STDNT'
+
     def get_feed_internal
       return {} unless is_feature_enabled? && target_audience?
       response = CampusSolutions::DegreeProgress::GraduateMilestones.new(user_id: @uid).get
-      response[:feed] = HashConverter.camelize({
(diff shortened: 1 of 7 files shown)


app/models/degree_progress/my_graduate_milestones.rb
```diff
@@ -8,17 +8,55 @@ class MyGraduateMilestones < UserSpecificModel
...
+
+    def law_student?(roles)
+      !!roles['doctorScienceLaw'] || !!roles['jurisSocialPolicyMasters'] || !!roles['jurisSocialPolicyPhC'] || !!roles['jurisSocialPolicyPhD']
+    end
+
+    def non_haas_grad_student?(roles)
+      !!roles['grad'] && (!roles['haasBusinessAdminMasters'] && !roles['haasBusinessAdminPhD'] && !roles['haasFullTimeMba'] && !roles['haasEveningWeekendMba'] && !roles['haasExecMba'] && !roles['haasMastersFinEng'] && !roles['haasMbaPublicHealth'] && !roles['haasMbaJurisDoctor'])
+    end
+
+    def get_link(id)
```
davidlimcheng: @lyttam  - I don't see this method being used anywhere in the model.  Is it being used somewhere else that I'm not seeing?

lyttam: @davidlimcheng good catch, I need to remove that.

lyttam:
@davidlimcheng I removed that unused method.

davidlimcheng:
Thanks @lyttam, LGTM

PR #7006, opened 2018-02-28
SISRP-38366 - New admits admission offer acceptance - improve validation

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-38366

* Adds validation to the SIR response POST
* Also creates `ChecklistDataUpdatingModel` because `PersonDataUpdatingModel` really had nothing to do with SIR
* Changes `valid?` from a class method to an instance method.  I believe that it needed to be an instance method so that classes outside of the `CampusSolutions` module (in this PR, `SirStatuses` is under `CampusSolutions::Sir::SirStatuses`) could access inherited instance variables
* For that reason, changes `valid?` to an instance method on all the other models it was already defined in

### DIFF
--- a/app/models/campus_solutions/sir/my_sir_response.rb
+++ b/app/models/campus_solutions/sir/my_sir_response.rb
@@ -2,10 +2,9 @@ module CampusSolutions
   module Sir
     class MySirResponse < UserSpecificModel
 
-      include CampusSolutions::PersonDataUpdatingModel
+      include CampusSolutions::ChecklistDataUpdatingModel
 
       def update(params = {})
-        CampusSolutions::ChecklistDataExpiry.expire @uid
         passthrough(CampusSolutions::Sir::SirResponse, params)
       end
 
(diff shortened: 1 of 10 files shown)


app/models/campus_solutions/sir/my_sir_response.rb
```diff
@@ -2,10 +2,9 @@ module CampusSolutions
   module Sir
     class MySirResponse < UserSpecificModel
 
-      include CampusSolutions::PersonDataUpdatingModel
+      include CampusSolutions::ChecklistDataUpdatingModel
 
       def update(params = {})
-        CampusSolutions::ChecklistDataExpiry.expire @uid
```
davidlimcheng: I got rid of this because we clear the cache *after* the POST is made already, in `ChecklistDataUpdatingModel`

lyttam:
@davidlimcheng Looks great! Nice work

PR #7007, opened 2018-02-28
 SISRP-38270 - UX Develop: Tasks Card (New Admit UX)

redconfetti:
https://jira.berkeley.edu/browse/SISRP-38270

Note: I tried to squash this into a single commit, but every time it ends up as 2 commits, one owned by @lyttam for some unknown reason.


app/models/my_tasks/merged.rb
```diff
@@ -10,7 +10,6 @@ def initialize(uid, options={})
       super(uid, options)
       #To avoid issues with tz, use DateTime instead of Date (http://www.elabs.se/blog/36-working-with-time-zones-in-ruby-on-rails)
       @starting_date = Time.zone.today.in_time_zone.to_datetime
-      @now_time = Time.zone.now
     end
```
redconfetti: This was simply not used in the model, so I removed it.

src/assets/templates/widgets/tasks.html
```diff
@@ -18,101 +18,169 @@ <h2 class="cc-left">Tasks</h2>
...
+
+        <div data-ng-show="incompleteTaskTotal() !== 0">
+          <div class="cc-widget-text cc-widget-tasks-no-padding-bottom">
+            <p>Click "View" to see tasks, due dates, and instructions.</p>
           </div>
 
-          <div class="cc-clearfix-container cc-widget-tasks-bottom-margin">
-            <h3 class="cc-widget-tasks-section-label">Today <span class="cc-widget-tasks-section-label-count cc-widget-tasks-section-label-count-green" data-ng-bind="dueTodayTasks.length"></span> <span class="cc-right">Due</span></h3>
-            <ul class="cc-widget-tasks-list cc-widget-list-color cc-widget-list-color-green">
+          <!-- Over Due Section -->
```
lyttam: @redconfetti personally I think HTML comments are unnecessary and usually not helpful (it's obvious from the markup that this is the Overdue section). Up to you whether to leave them in.

lyttam:
One suggestion; looks good otherwise.

src/assets/templates/widgets/tasks.html
redconfetti: @lyttam - Oh I agree on this. I was doing that to help myself as I worked on it. I do thinks the comments in dashboard_task_loop.html are helpful though in pointing out what the sections of HTML represent.

redconfetti: @lyttam - Actually, I think the spacing between sections is enough even in dashboard_task_loop.html, so I'm removing those too.

src/assets/templates/dashboard_task_loop.html
```diff
@@ -47,22 +46,49 @@
...
+        <span data-ng-if="!task.cs.itemStatusCode" class="cc-widget-tasks-title-sub cc-uppercase">Completed</span>
+        <span data-ng-if="task.cs.itemStatusCode === 'C'" class="cc-widget-tasks-title-sub cc-uppercase">Completed</span>
+        <span data-ng-if="task.cs.itemStatusCode === 'W'" class="cc-widget-tasks-title-sub cc-uppercase">Waived</span>
+
+        <!-- Completed Dates -->
+        <div class="cc-widget-tasks-item-completed-date"
+          data-ng-bind="task.completedDate.epoch * 1000 | dateInYearFilter:'MMM d':'MMM d, yyyy'">
+        </div>
+      </div>
+
```
redconfetti: @lyttam - Actually, I think the spacing between sections is enough even in dashboard_task_loop.html, so I'm removing those too. The ngIf attributes like `task.status !== 'completed'` and `task.status === 'completed'` make it clear. I'll just leave space between the "statuses" and "dates" sections of code.

app/models/my_tasks/sis_tasks.rb
```diff
@@ -109,5 +112,38 @@ def format_checklist(result)
...
+    def display_category(admin_func_code, checklist_item_code)
+      return 'residency' if checklist_item_code[0,2] == 'RR'
+      case admin_func_code
+        when 'ADMA' # Admissions Application
+          'newStudent'
+        when 'ADMP' # Admissions Program
+          'admission'
+        when 'FINA' # Financial Aid
+          'finaid'
+        when 'GEN', 'SPRG', 'STRM' # General, Student Program, Student Term
```
davidlimcheng: @redconfetti  Since the default is already 'student', can we remove this `when` block?

redconfetti:
@lyttam - I've removed all the HTML comments

src/assets/javascripts/angular/controllers/widgets/tasksController.js
```diff
@@ -8,21 +8,102 @@ var _ = require('lodash');
...
-    $scope.counts.current = isScheduled ? $scope.counts.scheduled : $scope.counts.unscheduled;
-    $scope.counts.opposite = isScheduled ? $scope.counts.unscheduled : $scope.counts.scheduled;
+  var getTasks = function(options) {
+    return tasksFactory.getTasks(options).then(function(data) {
+      var tasks = _.get(data, 'tasks');
+      if (tasks) {
+        $scope.tasks = tasks;
+        linkService.addCurrentPagePropertiesToResources($scope.tasks, $scope.currentPage.name, $scope.currentPage.url);
+        $scope.updateTaskLists();
+        $scope.isLoading = false;
```
davidlimcheng: @redconfetti - I think `isLoading` should be set to `false` regardless if a student has tasks or not.  How about instead of putting it in this `if` block, appending a `.finally()` to the promise to make `$scope.isLoading = false` ?

app/models/my_tasks/sis_tasks.rb
redconfetti: @davidlimcheng - Good catch. I want to say that the design only called for these codes, and I decided to make it default to 'student' for any other unknown task types... but that will still remain in the last design, which is like our spec doc. So yeah, I'll make that change now to make for more efficient code

src/assets/javascripts/angular/controllers/widgets/tasksController.js
```diff
@@ -154,23 +203,90 @@ angular.module('calcentral.controllers').controller('TasksController', function(
     );
   };
 
-  var filterOverdue = function(task) {
-    return (task.status !== 'completed' && task.bucket === 'Overdue');
+  $scope.incompleteTaskTotal = function() {
+    if ($scope.lists && $scope.lists.incomplete && $scope.lists.furtherActionNeeded && $scope.lists.overdue) {
```
davidlimcheng: @redconfetti If we are returning the summation of all of these task statuses, should these be `||` conditionals instead of `&&`?  If we have `incomplete` tasks but no `overdue` tasks, we still want to return a count, right?

davidlimcheng:
@redconfetti Good stuff!  Left some comments, otherwise looks great.  Nice job!

src/assets/javascripts/angular/controllers/widgets/tasksController.js
redconfetti: @davidlimcheng - Each of those should be empty arrays, and thus the length should return `0` for each if there are none. 

I'm only making this function safely check for each because `updateTaskLists()` may have not yet initialized those as empty arrays, which would result in an error.

If there are no tasks, and `incompleteTaskTotal()` returns `0`, it displays the message that there are no tasks present under this tab.

src/assets/javascripts/angular/controllers/widgets/tasksController.js
redconfetti: @davidlimcheng - great catch

PR #7008, opened 2018-02-28
SISRP-37998 - Creates 'under review' status for new admit degree progress

lyttam:
https://jira.berkeley.edu/browse/SISRP-37998

Establishes a grace period for new admits during which their incomplete requirements appear as "Under Review".  This accounts for the time it takes Transfer Credit to process their incoming credits (which may satisfy some of their requirements).

The grace period is different for each term:
- Fall grace period ends 60 days into the term
- Spring grace period ends 30 days into the term
- Summer grace period ends 60 days after the term ends


lyttam:
Probable unit test failure. Investigating.

PR #7011, opened 2018-03-01
SISRP-37998 - Surfaces transfer credit review date for new admits

lyttam:
https://jira.berkeley.edu/browse/SISRP-37998

This PR is a follow-up to #7008 

Only the new admits that see the special "Under Review" status should see the help icon and the information explaining the status (which includes the transfer credit review date).

### DIFF
--- a/src/assets/javascripts/angular/controllers/widgets/academics/undergraduateDegreeProgressController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/academics/undergraduateDegreeProgressController.js
@@ -10,13 +10,20 @@ angular.module('calcentral.controllers').controller('UndergraduateDegreeProgress
     }
   };
 
+  var showTip = function() {
+    $scope.degreeProgress.undergraduate.tipVisible = true;
+    return false;
+  };
+
   var loadDegreeProgress = function() {
     return degreeProgressFactory.getUndergraduateRequirements().then(
       function(response) {
         $scope.degreeProgress.undergraduate.progresses = _.get(response, 'data.feed.degreeProgress.progresses');
+        $scope.degreeProgress.undergraduate.transferCreditReviewDeadline = _.get(response, 'data.feed.degreeProgress.transferCreditReviewDeadline');
         $scope.degreeProgress.undergraduate.links = _.get(response, 'data.feed.links');
         $scope.degreeProgress.undergraduate.errored = _.get(response, 'data.errored');
         $scope.degreeProgress.undergraduate.showCard = apiService.user.profile.features.csDegreeProgressUgrdStudent && apiService.user.profile.roles.undergrad;
+        $scope.degreeProgress.undergraduate.showTip = showTip;
       }
     );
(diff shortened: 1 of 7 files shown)


src/assets/stylesheets/_icons.scss
```diff
@@ -218,10 +218,10 @@
 // Icon Color List Bullet Grey
 //
 // markup:
-// <i class="fa fa-exclamation-triangle cc-icon-list-bullet-grey"></i>
+// <i class="fa fa-exclamation-triangle cc-color-air-force-blue"></i>
 //
 // Styleguide 3.3.5
-.cc-icon-list-bullet-grey {
+.cc-icon-air-force-blue {
```
lyttam: Just renaming this to be more general.

src/assets/javascripts/angular/controllers/widgets/academics/undergraduateDegreeProgressController.js
```diff
@@ -10,13 +10,20 @@ angular.module('calcentral.controllers').controller('UndergraduateDegreeProgress
     }
   };
 
+  var showTip = function() {
+    $scope.degreeProgress.undergraduate.tipVisible = true;
+    return false;
```
davidlimcheng: @lyttam  - You mentioned that this is done this way to prevent the `<a>` node from acting like an actual link.

Another way we can do this that's a bit less hacky is instead of using `<a>`, we can make it a `<button>` instead.  There is existing CSS to style a button to looks like an anchor link to match designs:

_buttons.scss
```
// Button Link
//
// Used to style a button as a link. Recommended for Javascript actions
// that do not result in a new page loading.
//
// markup:
// <button class="cc-button-link">Show more</button>
//
// Styleguide 2.9
.cc-button-link {
  padding: 0;
  &, &:hover, &:focus {
    background-color: transparent;
    border: 0;
    color: $primary-color;
    font-size: 13px;
    font-weight: normal;
  }
  &:hover i, &:focus i {
    text-decoration: underline;
  }
  &:disabled {
    background-color: inherit;
    color: $cc-color-nobel;
    pointer-events: none;
  }
}
```
What do you think?

lyttam: @davidlimcheng Good question. I opted to make it a link because it seems more accessible (according to https://webaim.org/discussion/mail_thread?thread=5041).

davidlimcheng: Good stuff!  I did a quick search, it seems like there are a lot of conflicting opinions about this. 

The main difference I saw was that if we use `<a>`, then the "enter" key is used to activate it.  If we use `<button>`, then the "space" key is used to activate.

I know that we're using both in multiple places, so it would be good if we changed them all to the same thing in a future PR.  I'll open a JIRA for this.

For this PR though, it LGTM.

PR #7027, opened 2018-03-08
SISRP-39180 / SISRP-39181 / SISRP-39229 - Resolves categorization and count involving overdue tasks. Fixes duplicated FA task statuses

redconfetti:
https://jira.berkeley.edu/browse/SISRP-39180
https://jira.berkeley.edu/browse/SISRP-39181
https://jira.berkeley.edu/browse/SISRP-39229

* Fixes issue with counts
* Fixes display of messaging when no tasks are present, including bConnected messaging
* Refactored to make code simpler using LoDash [_.clone](https://lodash.com/docs/4.17.5#clone) and [_.remove](https://lodash.com/docs/4.17.5#remove)
* Fixed issue with CS display status showing twice for Financial Aid Tasks

```
var objects = [1, 5, 1, 2];
var shallow = _.clone(objects);
_.remove(shallow, function(i){ return i === 1});
console.log(objects);
// => [1, 5, 1, 2]
console.log(shallow);
// => [5, 2]
```

### DIFF
--- a/src/assets/javascripts/angular/controllers/widgets/finaid/finaidCommunicationsController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/finaid/finaidCommunicationsController.js
@@ -7,7 +7,6 @@ var _ = require('lodash');
  * Finaid Communications controller
  */
 angular.module('calcentral.controllers').controller('FinaidCommunicationsController', function($q, $scope, activityFactory, finaidFactory, finaidService, linkService, tasksFactory, tasksService) {
-  $scope.tasksService = tasksService;
 
   $scope.communicationsInfo = {
     aidYear: '',
@@ -41,6 +40,13 @@ angular.module('calcentral.controllers').controller('FinaidCommunicationsControl
       }
       $scope.activityInfo.isLoading = false;
     });
+  };
+
+  // Initializes methods required by dashboard_task_loop.html
+  var initializeServiceMethods = function() {
+    $scope.isCompletedTask = tasksService.isCompletedTask;
+    $scope.isCsBeingProcessedTask = tasksService.isCsBeingProcessedTask;
+    $scope.isOverdueTask = tasksService.isOverdueTask;
   };
 
   /**
@@ -93,5 +99,6 @@ angular.module('calcentral.controllers').controller('FinaidCommunicationsControl
     );
   };
 
+  initializeServiceMethods();
   $scope.$on('calcentral.custom.api.finaid.finaidYear', loadCommunications);
 });
--- a/src/assets/javascripts/angular/controllers/widgets/finaid/finaidCommunicationsController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/finaid/finaidCommunicationsController.js
@@ -6,7 +6,9 @@ var _ = require('lodash');
 /**
  * Finaid Communications controller
  */
-angular.module('calcentral.controllers').controller('FinaidCommunicationsController', function($q, $scope, activityFactory, finaidFactory, finaidService, linkService, tasksFactory) {
+angular.module('calcentral.controllers').controller('FinaidCommunicationsController', function($q, $scope, activityFactory, finaidFactory, finaidService, linkService, tasksFactory, tasksService) {
+  $scope.tasksService = tasksService;
+
   $scope.communicationsInfo = {
     aidYear: '',
     isLoading: true,
--- a/src/assets/javascripts/angular/controllers/widgets/tasksController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/tasksController.js
@@ -7,8 +7,6 @@ var _ = require('lodash');
  * Tasks controller
  */
 angular.module('calcentral.controllers').controller('TasksController', function(apiService, linkService, tasksFactory, tasksService, $http, $interval, $filter, $scope) {
-  $scope.tasksService = tasksService;
-
   // Initial mode for Tasks view
   $scope.currentTaskMode = 'incomplete';
   $scope.taskModes = ['incomplete', 'completed'];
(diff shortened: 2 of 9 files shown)


src/assets/javascripts/angular/controllers/widgets/finaid/finaidCommunicationsController.js
```diff
@@ -6,7 +6,9 @@ var _ = require('lodash');
 /**
  * Finaid Communications controller
  */
-angular.module('calcentral.controllers').controller('FinaidCommunicationsController', function($q, $scope, activityFactory, finaidFactory, finaidService, linkService, tasksFactory) {
+angular.module('calcentral.controllers').controller('FinaidCommunicationsController', function($q, $scope, activityFactory, finaidFactory, finaidService, linkService, tasksFactory, tasksService) {
+  $scope.tasksService = tasksService;
```
davidlimcheng: @redconfetti Is it necessary to attach the `tasksService` object to the scope?  Can we just refer to it as `tasksService.method` instead of `$cope.tasksService.method`?

lyttam:
@redconfetti I'm curious why you're storing so much in `$scope` inside that controller.  What's the benefit? It seems like an anti-pattern to me.

src/assets/javascripts/angular/controllers/widgets/finaid/finaidCommunicationsController.js
redconfetti: @davidlimcheng - I'm adding it to the scope so that it can be used within `dashboard_task_loop.html`. I'm doing it in both controllers because that template is used under both of those controllers.

davidlimcheng: @redconfetti Ah, I see.  In that case, can we just attach the three used functions to the scope, as opposed to the entirety of the `tasksService` object?

redconfetti:
@lyttam - I'm not sure what you mean. What am I storing in the $scope that isn't needed?

src/assets/javascripts/angular/controllers/widgets/finaid/finaidCommunicationsController.js
redconfetti: @davidlimcheng - So how do we share those methods with both the FinaidCommunicationsController and also the TasksController at the same time?

These methods are not only being used by the TasksController and child controllers such as taskAdderController and taskEditorController. That's why I put them in the service. They're all related to checking the status of the tasks.

I could understand if you want me to remove the methods from the service that aren't used by the dashboard_task_loop.html. But it seemed to make more sense to move them all over together to lighten up the TaskController code, and make them accessible to the template should they be needed for use in ngIf / ngShow / ngHide expressions.

davidlimcheng: @redconfetti , what I'm suggesting is a function that initializes only the `tasksService` functions you need and attaches them to the scope, not the only object.  For example, something like:
``` 
var initializeServiceMethods = function() {
  $scope.taskMethodOne = tasksService.taskMethodOne;
  $scope.taskMethodTwo = tasksService.taskMethodTwo;
}
```
This way, we won't be unnecessarily bringing over functions from `tasksService` that aren't being used.

redconfetti:
@davidlimcheng - I've revised the code as discussed

src/assets/javascripts/angular/controllers/widgets/finaid/finaidCommunicationsController.js
```diff
@@ -6,7 +6,15 @@ var _ = require('lodash');
 /**
  * Finaid Communications controller
  */
-angular.module('calcentral.controllers').controller('FinaidCommunicationsController', function($q, $scope, activityFactory, finaidFactory, finaidService, linkService, tasksFactory) {
+angular.module('calcentral.controllers').controller('FinaidCommunicationsController', function($q, $scope, activityFactory, finaidFactory, finaidService, linkService, tasksFactory, tasksService) {
+  // Initializes methods required by dashboard_task_loop.html
+  var initializeServiceMethods = function() {
```
davidlimcheng: Sorry to be nitpicky, @redconfetti, but for both controllers, can we put this further down, after all of the `$scope` initializations?  Maybe for this one it can go above `getMyFinaidActivity()`.

src/assets/javascripts/angular/controllers/widgets/finaid/finaidCommunicationsController.js
```diff
@@ -6,7 +6,15 @@ var _ = require('lodash');
...
+    $scope.isCompletedTask = tasksService.isCompletedTask;
+    $scope.isCsBeingProcessedTask = tasksService.isCsBeingProcessedTask;
+    $scope.isOverdueTask = tasksService.isOverdueTask;
+  };
+  initializeServiceMethods();
```
davidlimcheng: Similarly, can we move this all the way down, @redconfetti?  If it's something that's going to run when the controller instantiates, I think it should be at or close to the very bottom, similar to the pattern of other controllers.

redconfetti:
@davidlimcheng - I've moved the methods as requested. They're alphabetically in order with the other non-$scope functions.

davidlimcheng:
Thanks for the changes @redconfetti, LGTM!

src/assets/javascripts/angular/controllers/widgets/tasksController.js
```diff
@@ -231,28 +240,23 @@ angular.module('calcentral.controllers').controller('TasksController', function(
...
-      overdueTasks = overdueAndIncompleteTasks[0] || [];
-      remainingIncompleteTasks = overdueAndIncompleteTasks[1] || [];
+    if (incompleteTasks.length > 0) {
+      overdueTasks = _.remove(incompleteTasks, $scope.isOverdueTask);
     }
 
     $scope.lists = {
-      completed: $scope.tasks.filter(isCompletedTask).sort(sortByCompletedDateReverse),
-      incomplete: remainingIncompleteTasks,
+      completed: $scope.tasks.filter($scope.isCompletedTask).sort(sortByCompletedDateReverse),
```
lyttam: @redconfetti I would suggest using `tasksService.isCompletedTask` instead of `$scope.isCompletedTask`.  Maybe I'm superstitious, but I've several instances where `$scope` isn't fully initialized when it's used by an Angular controller, resulting in a race condition and a really tough to crack bug.

lyttam:
@redconfetti Yes, `updateTaskLists` is one example.  If a function is getting called in the template, then it belongs in `$scope` - otherwise, it should go in a service.

redconfetti:
@lyttam - I've replaced the references with `taskService` to avoid the possible race condition. Are you still expecting updateTaskLists to be converted to a service method? This could greatly increase development time required for this. I don't see this as a problematic pattern, given that the taskAdderController and taskEditorController are like "children" components (controller/templates) of the taskController.

lyttam:
@redconfetti Ultimately it's up to you, but code that needs to be shared between controllers belongs in a service. Here's a basic explanation of why: https://stackoverflow.com/a/15390245

I'm not confident that relying on this kind of controller inheritance is a good idea, given the mysterious race condition bugs we've seen before. It's cleaner and more predictable to move controller behavior into a service - and that's how we do it elsewhere in the app.  If we do merge your code as-is, I recommend refactoring it in the next release.

redconfetti:
@lyttam - I anticipated many complications, but I took a shot at moving the `updateTaskLists` method and all it's dependencies to the tasksService.

lyttam:
@redconfetti Looks good, thanks for the changes.

PR #7049, opened 2018-03-19
SISRP-39460 - Tasks with Received Status are not displaying in CalCentral

redconfetti:
https://jira.berkeley.edu/browse/SISRP-39460

See QA PR #7048


nkutub:
@redconfetti , Merging test failing for other reasons.

PR #7054, opened 2018-03-19
SISRP-39298 - Task card Further Action Needed section icon, due date and title should be red

redconfetti:
https://jira.berkeley.edu/browse/SISRP-39298


davidlimcheng:
LGTM @redconfetti

PR #7062, opened 2018-04-03
SISRP-39506 - Add grading basis to Semesters card - CLC Dev

redconfetti:
https://jira.berkeley.edu/browse/SISRP-39506


davidlimcheng:
Very nice @redconfetti

PR #7067, opened 2018-04-05
SISRP-38429 - Suppress P/NP units for Law students

lyttam:
https://jira.berkeley.edu/browse/SISRP-38429

P/NP units are not relevant for Law students, so we'll hide them on the Academic Summary.

### DIFF
--- a/app/models/my_academics/gpa_units.rb
+++ b/app/models/my_academics/gpa_units.rb
@@ -33,16 +33,8 @@ module MyAcademics
       else
         result[:hub_empty] = true
       end
-      if (edo_response)
-        if (totalUnitsTakenNotForGpa = parse_edo_total_units_taken_not_for_gpa edo_response) && totalUnitsTakenNotForGpa.present?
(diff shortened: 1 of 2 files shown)


app/models/my_academics/gpa_units.rb
```diff
@@ -33,16 +33,8 @@ def hub_gpa_units
...
       end
-      if (edo_response)
-        if (totalUnitsTakenNotForGpa = parse_edo_total_units_taken_not_for_gpa edo_response) && totalUnitsTakenNotForGpa.present?
-          result[:totalUnitsTakenNotForGpa] = totalUnitsTakenNotForGpa
-        end
-        if (totalUnitsPassedNotForGpa = parse_edo_total_units_passed_not_for_gpa edo_response) && totalUnitsPassedNotForGpa.present?
-          result[:totalUnitsPassedNotForGpa] = totalUnitsPassedNotForGpa
-        end
-      else
-        result[:edo_empty] = true
```
lyttam: This was not being used, so I removed it.

PR #7073, opened 2018-04-06
SISRP-39790 - Summer Estimator - CalCentral Dev

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-39790

* The requirement was not just "undergrads", but "new admit undergrads", which are not assigned the "undergrad" role until they are matriculated
* So, we have to see if the user has any undergraduate SIR offers
* Also moves all of the conditional links within an `if` block, so we don't needlessly make the proxy call

### DIFF
--- a/src/assets/javascripts/angular/controllers/widgets/financesLinksController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/financesLinksController.js
@@ -6,7 +6,7 @@ var _ = require('lodash');
 /**
  * Footer controller
  */
-angular.module('calcentral.controllers').controller('FinancesLinksController', function(apiService, campusLinksFactory, finaidService, financesLinksFactory, userService, $scope, $q) {
+angular.module('calcentral.controllers').controller('FinancesLinksController', function(apiService, campusLinksFactory, finaidService, financesLinksFactory, sirFactory, userService, $scope, $q) {
   $scope.isLoading = true;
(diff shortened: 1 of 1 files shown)


src/assets/javascripts/angular/controllers/widgets/financesLinksController.js
```diff
@@ -102,11 +102,21 @@ angular.module('calcentral.controllers').controller('FinancesLinksController', f
...
+  var parseSirStatuses = function(response) {
+    var hasUndergraduateRole = userService.profile.roles.undergrad;
+    var sirStatuses = _.get(response, 'data.sirStatuses');
+    var hasUndergraduateSir = !!_.find(sirStatuses, {
+      isUndergraduate: true
+    });
+    $scope.canViewSummerEstimatorLink = hasUndergraduateRole || hasUndergraduateSir;
+  };
+
+  var setSummerEstimatorVisibility = sirFactory.getSirStatuses().then(parseSirStatuses);
```
redconfetti: @davidlimcheng - Could we move this line, which makes the first actual method call, down above line 150?

davidlimcheng: @redconfetti That makes a ton of sense.  Nice catch, will amend.

redconfetti:
@davidlimcheng LGTM

PR #7077, opened 2018-04-11
SISRP-38415 - Establish academic roles for JSP/JSD and JD/LMM groups

redconfetti:
https://jira.berkeley.edu/browse/SISRP-38415


lyttam:
@redconfetti why did you add the styleguide exclusions?

redconfetti:
@lyttam - The SC5 styleguide is generated and added to /public/styleguide when we run the `gulp styleguide` task locally. The styleguide is [ignored by Git](https://github.com/ets-berkeley-edu/calcentral/blob/bbc34be8/.gitignore#L41), so not checked into our repo.

It's presence was causing errors when I'd run `scripts/front-end-tests.sh`. I added those lines to avoid getting the errors locally in the future with [JSCS](http://jscs.info/overview) and [JsHint](http://jshint.com/).

PR #7081, opened 2018-04-16
SISRP-39179 - Redefine Affiliations Logic

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-39179

Updates Affiliations logic with the following steps:
- Removes all dependencies/usages of CampusOracle::UserAttributes
- Adds logic to get "expiredAccount" affiliation data from LDAP
- Changes CS-based `ADMT_UX` affiliation from `applicant` to `releasedAdmit`, and changes `releasedAdmit` role to be the gatekeeper to the SIR card
- Removes CampusOracle::UserAttributes
- Updates `held_applicant` logic in `User::AuthenticationValidator` to filter out undergraduates who shouldn't be in CalCentral before their SIR has been released
- Creates a `base_roles` object to list all possible roles for each user
- Puts LDAP Student Affiliations on a feature flag

It's a pretty large PR unfortunately, I suggest working through each commit one-by-one to make more sense of it.

Lastly, I'd like to request a PR review from @raydavis, @johncrossman, or @paulkerschen if possible - to ensure that I'm not stepping on the toes of anything Junction is dependent on.  (Just the first commit should be good!  Should've said that earlier.)


app/models/user/authentication_validator.rb
```diff
@@ -43,25 +43,37 @@ def cached_held_applicant?
...
-        calnet_attributes['affiliations'] != 'STUDENT-TYPE-NOT-REGISTERED'
       cs_feed = HubEdos::Affiliations.new(user_id: @auth_uid).get
-      if cs_feed[:feed] && (student = cs_feed[:feed]['student']) && student['affiliations']
+      if cs_feed.try(:[], :feed) && (student = cs_feed[:feed].try(:[], 'student')) && student.try(:[], 'affiliations')
         cs_feed = HashConverter.symbolize student
-        applicant_in_process?(cs_feed[:affiliations]) && roles_from_cs_affiliations(cs_feed[:affiliations]).blank?
+        held = unreleased_applicant?(cs_feed[:affiliations])
+        has_ldap_affiliations = held ? has_ldap_affiliations? : nil
+        is_graduate = held && (!has_ldap_affiliations.nil? && !has_ldap_affiliations) ? is_graduate_applicant? : nil
+        held && (!has_ldap_affiliations.nil? && !has_ldap_affiliations) && (!is_graduate.nil? && !is_graduate)
```
davidlimcheng: Not proud of this!  I wrote it this way to avoid any unnecessary proxy calls - i.e., we shouldn't be calling `#has_ldap_affiliations` if `held == false`.

If anybody can suggest a cleaner pattern, I'd gladly take it.

redconfetti:
@davidlimcheng - What happened with the Education Abroad Program?

app/models/hub_edos/user_attributes.rb
```diff
@@ -132,7 +132,7 @@ def extract_roles(edo, result)
       # CS Affiliations are expected to exist for any working CS ID.
       if (affiliations = edo[:affiliations])
         result[:roles] = roles_from_cs_affiliations(affiliations)
-        if result[:roles].slice(:student, :exStudent, :applicant).has_value?(true)
+        if result[:roles].slice(:student, :exStudent, :applicant, :releasedAdmit).has_value?(true)
```
redconfetti: @davidlimcheng - I know that Released Admits are the ones we want to give access to in CalCentral, and Applicants are not.

>  if a student has "APPLICANT" and absolutely nothing else, it means they probably don't belong in CalCentral, so do not let them log in

If this is true, should we be excluding applicants?

redconfetti: Oh nevermind, they can be both an `:applicant` and also a `releasedAdmit` at the same time.

app/models/berkeley/user_roles.rb
```diff
@@ -78,7 +107,10 @@ def roles_from_cs_affiliations(cs_affiliations)
         # TODO: We still need to cover staff, guests, concurrent-enrollment students and registration status.
         case active_affiliation[:type][:code]
           when 'ADMT_UX'
-            result[:applicant] = true
+            # A 'releasedAdmit' is a user whose Statement of Intent to Register has been released to them
+            result[:releasedAdmit] = true
+          when 'APPLICANT'
+            result[:applicant] = true unless active_affiliation[:detail] == 'Applied'
```
redconfetti: @davidlimcheng - I'm curious what this exception is about.

raydavis:
Thank you for the ping, @davidlimcheng -- I'm diving into the PR's implications now & will let you know ASAP.

app/models/user/authentication_validator.rb
redconfetti: @davidlimcheng - The conditional to avoid calling the proxy when not needed makes sense. 

I'm curious about this pattern: `(!has_ldap_affiliations.nil? && !has_ldap_affiliations)`. If it's nil, then isn't it also not present already. You also specifically set it to be `nil` if the user is not held.

Also I wonder if you can simplify that IF statement.

```ruby
student = cs_feed.try(:[], :feed).try(:[], 'student')
if student.try(:[], 'affiliations')
  # ...
else
  false
end
```

I'm also a little uncomfortable with you using `cs_feed` twice like that. Within the IF block perhaps you should assign the student data to a different var like `cs_student = HashConverter.symbolize student`.

raydavis:
@pfarestveit will have the final say when this code reaches Junction clusters. But so far as I can see, you've made it through the shared-code labyrinth, @davidlimcheng -- thank you for the effort.

I'll create an ETS ticket to enable {{features.ldap_student_affiliations}} on our systems.

app/models/user/authentication_validator.rb
redconfetti: @davidlimcheng - We discussed this and decided that this is the simplest way to implement the staged logic, without the use of a [state machine](https://github.com/pluginaweek/state_machine). You could however check for the presence of a true value, excluding nil and false, by using `present?`.

davidlimcheng: @redconfetti Agree about your other comments though - will amend

app/models/berkeley/user_roles.rb
davidlimcheng: @redconfetti - The CS "APPLICANT" role comes in three flavors, defined in `detail`:

1. APPLICANT - Applied
2. APPLICANT - Admitted
3. APPLICANT - SIR Completed

"APPLICANT - Applied" means that the student has applied to UCB, and no decision has been made yet.  "APPLICANT - Admitted" means that the student has applied to UCB, and they received a SIR offer.  And 3 is self-explanatory.

Since "APPLICANT - Applied" is so upstream, and students shouldn't be in CalCentral during this part of the process, I decided to ignore that population until they have a reason to be in CalCentral.

davidlimcheng:
Great, thanks for the review @raydavis!

davidlimcheng:
@redconfetti - The Education Abroad Program was put on Hold when we switched over to CS, and has not been updated since.  More info at [SISRP-22698](https://jira.berkeley.edu/browse/SISRP-22698).  Since we're no longer getting the data to show Education Abroad status, I removed the parts dependent on it.  I imagine once SISRP-22698, there will be updated designs and data sources to implement.

davidlimcheng:
@redconfetti Amended!

app/models/user/aggregated_attributes.rb
```diff
@@ -35,16 +34,15 @@ def get_feed_internal
...
 
     private
 
     def get_campus_roles
-      ldap_roles = (@ldap_attributes && @ldap_attributes[:roles]) || {}
-      oracle_roles = (@oracle_attributes && @oracle_attributes[:roles]) || {}
-      campus_roles = oracle_roles.merge ldap_roles
+      base_roles = Berkeley::UserRoles.base_roles
+      campus_roles = (@ldap_attributes && @ldap_attributes[:roles]) || {}
+      campus_roles = base_roles.merge campus_roles
```
lyttam: @davidlimcheng I believe it's best practice to avoid variable reassignment. I suggest changing this to 

```
      ldap_roles = (@ldap_attributes && @ldap_attributes[:roles]) || {}
      campus_roles = base_roles.merge ldap_roles
```

PR #7080, opened 2018-04-16
CLC-6849 Feed course and enrollment snapshots to the data loch

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6849

PR #7082, opened 2018-04-17
CLC-6849 Make data-loch-snapshot.sh executable

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-6849

Follow-up to #7080 with an extra permissions bit.


johncrossman:
We're ready for the executioner...

PR #7087, opened 2018-04-24
SISRP-38419 - Profile/Academic Summary - hide GPA from Law students - CalCentral Dev

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-38419

* Introduces multiple-career parsing for `MyAcademics::GpaUnits`
* Eventually we need the ability to parse multiple careers for all applicable features - but we'll take things one step at a time.
* @lyttam since you'll also be working on a lot of the law stuff that requires multiple-career parsing, can you give this one a review?

### DIFF
--- a/app/models/my_academics/gpa_units.rb
+++ b/app/models/my_academics/gpa_units.rb
@@ -10,23 +10,24 @@ module MyAcademics
 
     def gpa_units
       feed = MyAcademics::MyAcademicStatus.new(@uid).get_feed
-      academic_status = parse_academic_status feed
+      # TODO: Eventually we want to use #parse_academic_statuses to pull all careers and parse associated values for each career
+      academic_statuses = parse_academic_statuses feed
+      academic_status_first = academic_statuses.try(:first)
 
       result = {
         :errored => feed.try(:[], :errored),
-        :cumulativeGpa => parse_cumulative_gpa(academic_status),
+        :gpa => parse_cumulative_gpa(academic_statuses),
         :totalUnitsTakenNotForGpa => parse_total_units_not_for_gpa(pnp_units, 'pnp_taken'),
         :totalUnitsPassedNotForGpa => parse_total_units_not_for_gpa(pnp_units, 'pnp_passed'),
         :totalTransferAndTestingUnits => units_transferred
       }
-      result.merge!(parse_total_units academic_status)
-      result.merge!(parse_total_units_for_gpa academic_status)
+      result.merge!(parse_total_units academic_status_first)
+      result.merge!(parse_total_units_for_gpa academic_status_first)
       result
     end
 
-    def parse_academic_status(feed)
-      #TODO: Consult with SR concerning GPA displayed when multiple academic careers present
-      academic_statuses(feed).try(:first)
+    def parse_academic_statuses(feed)
+      academic_statuses(feed)
     end
 
     def pnp_units
@@ -51,8 +52,18 @@ module MyAcademics
       lookup_campus_solutions_id
     end
 
-    def parse_cumulative_gpa(status)
-      status.try(:[], 'cumulativeGPA').try(:[], 'average').try(:to_s)
+    def parse_cumulative_gpa(statuses)
+      gpa = []
+      statuses.try(:each) do |status|
+        role = status.try(:[], 'studentCareer').try(:[], :role)
+        gpa.push(
+          {
+            role: role,
+            roleDescr: role == 'concurrent' ? 'UCB Extension' : status.try(:[], 'studentCareer').try(:[], 'academicCareer').try(:[], 'formalDescription'),
+            cumulativeGpa: status.try(:[], 'cumulativeGPA').try(:[], 'average').try(:to_s)
+          })
+      end
+      gpa
     end
 
     # Ignores unimportant unit types given back by the hub, including 'unitsOther' (holds total units that exceed limits for other categories, e.g. transfer units)
--- a/src/assets/javascripts/angular/controllers/pages/academicSummaryController.js
+++ b/src/assets/javascripts/angular/controllers/pages/academicSummaryController.js
@@ -14,6 +14,22 @@ angular.module('calcentral.controllers').controller('AcademicSummaryController',
     apiService.util.printPage();
   };
 
+  var mergeTermHonors = function(semester) {
+    var honors = $scope.collegeAndLevel.awardHonors[semester.termId] || [];
+    semester.honors = _.keyBy(honors, function(honor) {
+      return honor.code;
+    });
+  };
+
+  var parseTransferCredit = function() {
+    $scope.showTransferCredit = showTransferCredit();
+    if (!showTransferCredit) {
+      return;
+    } else {
+      $scope.showPointsColumn = showTransferCreditPoints(_.get($scope, 'transferCredit.ucTransferCrseSch.items'));
+    }
+  };
+
   var showSemesters = function() {
     return !!(_.get($scope, 'semesters.length') && apiService.user.profile.hasStudentHistory);
   };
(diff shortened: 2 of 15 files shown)


src/assets/javascripts/angular/services/academicsService.js
```diff
@@ -28,6 +28,29 @@ angular.module('calcentral.services').service('academicsService', function() {
     return _.head(_.compact(sortedSemesters));
   };
 
+  var containsLawClass = function(selectedTeachingSemester) {
```
davidlimcheng: A lot of these are existing functions that I just moved to alphabetical order.

src/assets/javascripts/angular/services/academicsService.js
```diff
@@ -212,6 +264,47 @@ angular.module('calcentral.services').service('academicsService', function() {
     }
   };
 
+  /**
+   * Converts each value given in gpaUnits to a Number type to be processed regularly.  `parseFloat` returns NaN if input value does not contain at least one digit.
+   * GPAs are displayed with 4 significant digits.
+   * Also sets any law career-based GPAs to 'N/A' due to law classes being P/NP
+   */
+  var parseGpaUnits = function(gpaUnits) {
```
davidlimcheng: `parseGpaUnits` and `showGpa` are new functions.

src/assets/javascripts/angular/services/academicsService.js
```diff
@@ -212,6 +264,47 @@ angular.module('calcentral.services').service('academicsService', function() {
...
+      }
+    });
+    gpaUnits.totalUnits = parseFloat(gpaUnits.totalUnits);
+    return gpaUnits;
+  };
+
+  var showGpa = function(gpaArray) {
+    return _.some(gpaArray, function(gpa) {
+      var career = _.get(gpa, 'careerCode');
+      if (career && career !== 'law') {
```
lyttam: @davidlimcheng are we guaranteed that the string value of `career` will be lowercase? If not, could convert it before doing the `!==` comparison.

src/assets/javascripts/angular/controllers/pages/academicSummaryController.js
```diff
@@ -29,44 +45,37 @@ angular.module('calcentral.controllers').controller('AcademicSummaryController',
...
+    _.each($scope.semesters, mergeTermHonors);
+  };
+
   // Similar to academicsController, we wait until user profile is fully loaded before hitting academics data
   $scope.$on('calcentral.api.user.isAuthenticated', function(event, isAuthenticated) {
     if (isAuthenticated) {
       $scope.canViewAcademics = apiService.user.profile.hasAcademicsTab;
       academicsFactory.getAcademics()
       .then(parseAcademics)
+      .then(parseGpa)
```
lyttam: @davidlimcheng slight preference for calling `parseGpa` inside `parseAcademics`, instead of chaining `parseGpa` to `parseAcademics` here. Then you don't need `parseAcademics` to return a promise, and besides, the GPA is academic data too.

app/models/my_academics/gpa_units.rb
```diff
@@ -10,25 +10,30 @@ def merge(data)
 
     def gpa_units
       feed = MyAcademics::MyAcademicStatus.new(@uid).get_feed
-      academic_status = parse_academic_status feed
+      academic_status_first = parse_academic_status feed
+      academic_statuses = parse_academic_statuses feed
```
redconfetti: I wonder if this would be more efficient to parse the multiple academic statuses, and then call `academic_statuses.try(:first)` to create `academic_status_first`.

app/models/my_academics/gpa_units.rb
```diff
@@ -10,25 +10,30 @@ def merge(data)
...
-      result.merge!(parse_total_units academic_status)
-      result.merge!(parse_total_units_for_gpa academic_status)
+      result.merge!(parse_total_units academic_status_first)
+      result.merge!(parse_total_units_for_gpa academic_status_first)
       result
     end
 
     def parse_academic_status(feed)
-      #TODO: Consult with SR concerning GPA displayed when multiple academic careers present
+      # TODO: Eventually we want to use #parse_acaedemic_statuses to pull all careers and parse associated values for each career
```
redconfetti: @davidlimcheng - Typo - `#parse_acaedemic_statuses`

app/models/my_academics/gpa_units.rb
```diff
@@ -51,8 +56,18 @@ def get_campus_solutions_id
       lookup_campus_solutions_id
     end
 
-    def parse_cumulative_gpa(status)
-      status.try(:[], 'cumulativeGPA').try(:[], 'average').try(:to_s)
+    def parse_cumulative_gpa(statuses)
+      gpa = []
+      statuses.try(:each) do |status|
+        career_code = status.try(:[], 'studentCareer').try(:[], :role)
```
redconfetti: @davidlimcheng - Is there a reason why we're calling the `role` as `career_code` instead of actually using the actual [career code](https://github.com/davidlimcheng/calcentral/blob/9e555f6ba/fixtures/json/hub_academic_status.json#L103) from the academic status? The roles are a CalCentral construct, and are more likely to change than career codes. 
```javascript
"studentCareer": {
    "academicCareer": {
    "code": "UGRD",
    "description": "Undergraduate"
    },
    "fromDate": "2016-01-12"
}
```

src/assets/javascripts/angular/services/academicsService.js
redconfetti: Very @christianvuerings

src/assets/javascripts/angular/services/academicsService.js
```diff
@@ -28,6 +28,29 @@ angular.module('calcentral.services').service('academicsService', function() {
...
+      dept: 'LAW'
+    });
+  };
+
+  var containsMidpointClass = function(selectedTeachingSemester) {
+    var classes = _.get(selectedTeachingSemester, 'classes');
+    var isSummer = isSummerSemester(selectedTeachingSemester);
+    if (!isSummer && classes && classes.length) {
+      return !_.every(classes, function(klass) {
+        if (_.get(klass, 'dept') === 'LAW') {
```
christianvuerings: @davidlimcheng instead of

```js
if (_.get(klass, 'dept') === 'LAW') {
  return true;
} else {
  return false;
}
```

you can write this as
```js
return _.get(klass, 'dept') === 'LAW'
```

app/models/my_academics/gpa_units.rb
davidlimcheng: @redconfetti Absolutely, good catch!  Will amend.

app/models/my_academics/gpa_units.rb
davidlimcheng: @redconfetti Yeah, I was thinking to use the CalCentral-defined `role` here to ensure that we recognize the career coming in.  If there is a career coming in we don't recognize, we'd have a lot more problems, but it also helps to keep verbiage consistent across the app.  I do agree that calling it `careerCode` is a bit misleading, will change it to `role`.

src/assets/javascripts/angular/controllers/pages/academicSummaryController.js
davidlimcheng: @lyttam Absolutely, nice catch.  I chained it to take advantage of `$scope.showGpa` - that is, if we don't need to show GPA, then we don't need to parse it - but `parseGpa` also parses the units as well.  Will de-promisify, and change from `parseGpa` to `parseGpaUnits`.

src/assets/javascripts/angular/services/academicsService.js
davidlimcheng: Nice!

redconfetti:
@davidlimcheng - LGTM

PR #7089, opened 2018-04-26
[JQA] CLC-6843 Base new Project Sites on template

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6843

This will be tested on J-QA before J-Dev.

### DIFF
--- a/app/models/canvas_lti/project_provision.rb
+++ b/app/models/canvas_lti/project_provision.rb
@@ -20,13 +20,26 @@ module CanvasLti
     def create_project(project_name)
       project_account_id = Settings.canvas_proxy.projects_account_id
       term_id = Settings.canvas_proxy.projects_term_id
+      template_id = Settings.canvas_proxy.projects_template_id
       worker = Canvas::Course.new(user_id: @uid)
       response = worker.create(project_account_id, project_name, project_name, term_id, unique_sis_project_id)
       if (course_details = response[:body])
         site_id = course_details['id']
-        Canvas::CourseSettings.new(course_id: site_id).fix_default_view course_details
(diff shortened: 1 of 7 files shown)


app/models/canvas_lti/project_provision.rb
```diff
@@ -20,13 +20,27 @@ def unique_sis_project_id
...
       project_account_id = Settings.canvas_proxy.projects_account_id
       term_id = Settings.canvas_proxy.projects_term_id
+      template_id = Settings.canvas_proxy.projects_template_id
       worker = Canvas::Course.new(user_id: @uid)
       response = worker.create(project_account_id, project_name, project_name, term_id, unique_sis_project_id)
       if (course_details = response[:body])
         site_id = course_details['id']
-        Canvas::CourseSettings.new(course_id: site_id).fix_default_view course_details
+        response = Canvas::CourseCopyImport.new(canvas_course_id: site_id).import_course_template(template_id)
+        logger.info("New Project site #{site_id} import template response: #{response}")
```
johncrossman: @raydavis I don't think logging response object is a prob but I've seen `response.inspect` used to get more concise (and safer?) summary. If memory serves, with Google api we were getting key/secret in the logs.

johncrossman:
Looks good, @raydavis  Waiting on you-know-who.

app/models/canvas_lti/project_provision.rb
raydavis: You're right that I'm not following the usual format -- I've been away from this code too long, I guess. I'll amend.

raydavis:
Looks like our shared codebase / Travis is still having those sticky test issues. I think they'll clear up once we re-align with the SIS master and JRuby 2.

johncrossman:
These failures are unrelated to the change. Merging.

PR #7094, opened 2018-04-27
SISRP-40115 - Resolve gulp-concat NodeJS package issue causing builds to fail

redconfetti:
https://jira.berkeley.edu/browse/SISRP-40115


davidlimcheng:
Nice @redconfetti , can we keep the ticket open as a reminder to remove this once https://github.com/gulp-community/gulp-concat/issues/142 is resolved?

PR #7095, opened 2018-04-27
SISRP-39361 - Update links on My Campus - Wellness Section

nkutub:
https://jira.berkeley.edu/browse/SISRP-39361

### DIFF
--- a/public/json/campuslinks.json
+++ b/public/json/campuslinks.json
@@ -1345,6 +1345,25 @@
       },
       "categories": [
         {
+          "topcategory": "Support Services",
+          "subcategory": "Students"
+        }
+      ]
+    },
+    {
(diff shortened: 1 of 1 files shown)


public/json/campuslinks.json
```diff
@@ -3485,9 +3485,9 @@
       ]
     },
     {
-      "name": "UC Berkeley Wellness Letter",
+      "name": "Well e-News",
```
davidlimcheng: @nkutub I think the JIRA says this should be called "Be Well e-News"

nkutub: @davidlimcheng nice catch!

nkutub:
@davidlimcheng , all fixed

PR #7099, opened 2018-05-02
NOJIRA - Suppress database initialization output. Resolve deprecated serve_static_assets setting.

redconfetti:
Reduces the output to STDOUT during initialization of the application, thus cutting down on unnecessary output during `rails console` or `rspec` test runs.

Relies on use of [silence_stream](https://rails.lighthouseapp.com/projects/8994/tickets/4946-kernelsilence_stream-is-not-thread-safe), which is reported to not be thread safe. We don't run this code in a multi-threaded manner (initialization) so I think we're fine. I've included the note for when we upgrade to Rails 5. We may need to replace it with our own version then.


redconfetti:
Taken from [ApiDock](https://apidock.com/rails/Kernel/silence_stream)
```ruby
def silence_stream(stream)
    old_stream = stream.dup
    stream.reopen(RbConfig::CONFIG['host_os'] =~ /mswin|mingw/ ? 'NUL:' : '/dev/null')
    stream.sync = true
    yield
  ensure
    stream.reopen(old_stream)
    old_stream.close
  end
```

lyttam:
Sorry, I didn't realize this involved using a deprecated method.  Since `silence_stream` has issues (not thread safe) and is removed in Rails 5, I don't think we should start using it.  We may upgrade to Rails 5 soon and have to revert this work - plus, we would risk having mysterious build failures caused by thread-level race conditions. 

It looks like there is no replacement for this method unfortunately:  https://stackoverflow.com/questions/33350725/what-replace-silence-capture-kernel-method-in-rails-5 .  So we may be stuck with excess logging.  Not the worst problem to have.

lyttam:
@redconfetti However, your change from the deprecated `serve_static_assets` to `serve_static_files` is good, and I think it should go in this release.

lyttam:
@redconfetti I'd rather avoid introducing differences between the development and production environments whenever possible. Also, bugs in our build and deploy process can have a high cost in developer hours even if they don't affect the production runtime environment.  Thirdly, we would have to remove this code when we upgrade to Rails 5.

PR #7103, opened 2018-05-07
SISRP-38414 - Show Expected Graduation Term for Law Students - CalCentral Dev

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-38414


app/models/my_academics/graduation.rb
```diff
@@ -3,42 +3,37 @@ class Graduation < UserSpecificModel
...
+      return {} unless HubEdos::UserAttributes.new(user_id: @uid).has_role?(:student) && hub_statuses.present?
+      terms_with_appts = nil
+      appts_in_graduating_term = nil
+      ugrd_grad_term = extract_latest_undergraduate_graduation_term
+      if ugrd_grad_term.present?
+        terms_with_appts = active_terms_with_enrollment_appointments
+        appts_in_graduating_term = appointments_in_graduating_term(ugrd_grad_term, terms_with_appts)
+      end
+
       HashConverter.camelize({
```
nkutub: whats the point of this? aren't you returning a hash already that looks camel case already?

davidlimcheng: @nkutub Yep!  Left this in cause it was already there, but definitely good to take out.  Will amend.

app/models/my_academics/graduation.rb
```diff
@@ -49,25 +44,91 @@ def active_terms_with_enrollment_appointments
...
+      parsed_program_codes = []
+      parsed_plans = []
+      plans.try(:each) do |plan|
+        program_code = plan.try(:[], 'academicPlan').try(:[], 'academicProgram').try(:[], 'program').try(:[], 'code')
+        plan_description = plan.try(:[], 'academicPlan').try(:[], 'plan').try(:[], 'description')
+        expected_grad_term_id = plan.try(:[], 'expectedGraduationTerm').try(:[], 'id')
+        expected_grad_term_name = Berkeley::TermCodes.normalized_english(plan.try(:[], 'expectedGraduationTerm').try(:[], 'name'))
+
+        if parsed_program_codes.include? program_code
+          related_plan_index = parsed_plans.try(:find_index) do |parsed_plan|
```
nkutub: @davidlimcheng , could you not use .find? or select? Not a show stopper

app/models/my_academics/graduation.rb
```diff
@@ -49,25 +44,91 @@ def active_terms_with_enrollment_appointments
...
+          parsed_plans.push(
+            {
+              program: program_code,
+              plans: Array.wrap(plan_description),
+              expectedGradTermIds: Array.wrap(expected_grad_term_id),
+              expectedGradTermNames: Array.wrap(expected_grad_term_name)
+            })
+        end
+      end
+      parsed_plans
```
nkutub: where is the sort happening?

davidlimcheng: @nkutub It's not really a "sort", and more of placing each Plan into a Program bucket.  "Sort" is a misleading name, but `sort_plans_by_program` seemed less confusing than `place_plans_into_program_buckets` at the time.

PR #7104, opened 2018-05-07
SISRP-39478 - Student and Advisor view needs to display Academic status

nkutub:
https://jira.berkeley.edu/browse/SISRP-39478

* add academic standing feature to status and holds card
* new angular controller and factory
* new API route
* new backend controller and model

### DIFF
--- a/app/models/my_academics/my_standings.rb
+++ b/app/models/my_academics/my_standings.rb
@@ -1,0 +1,68 @@
+module MyAcademics
+  class MyStandings < UserSpecificModel
+    include Cache::CachedFeed
+    include Cache::JsonifiedFeed
+    include Cache::UserCacheExpiry
+
+    def get_feed_internal
+      academic_standings = get_academic_standings
+      parsed_standings = parse_standings(academic_standings)
+      {
+        feed: {
+          currentStandings: parsed_standings[:current_standings],
+          standingsHistory: parsed_standings[:standings_history]
+        }
+      }
+    end
+
+    def student_empl_id
+      User::Identifiers.lookup_campus_solutions_id @uid
+    end
+
+    def get_academic_standings
+      academic_standings = EdoOracle::Queries.get_academic_standings(student_empl_id)
+      academic_standings ||= []
+    end
+
+    def parse_standings(academic_standings)
+      parsed_standings = empty_standings
+      current_standing = get_latest_current_standing(academic_standings)
+      return parsed_standings unless current_standing
+      # Its possible to have more than one current standing with same term and action date
+      academic_standings.each do |standing|
+        if standing['term_id'] == current_standing['term_id'] && standing['action_date'] == current_standing['action_date']
+          parsed_standings[:current_standings].push(parse_standing(standing)) unless parsed_standings[:current_standings].any?
+        else
+          parsed_standings[:standings_history].push(parse_standing(standing))
+        end
+      end
+      parsed_standings
+    end
+
+    def empty_standings
+      {
+        current_standings: [],
+        standings_history: []
+      }
+    end
+
+    def parse_standing(standing)
+      term = Berkeley::TermCodes.from_edo_id(standing['term_id'])
+      {
+        acadStandingStatus: standing['acad_standing_status'],
+        acadStandingStatusDescr: standing['acad_standing_status_descr'],
+        acadStandingActionDescr: standing['acad_standing_action_descr'],
+        termId: standing['term_id'],
+        actionDate: Concerns::AcademicsModule.cast_utc_to_pacific(standing['action_date']),
+        termName: Berkeley::TermCodes.to_english(term[:term_yr], term[:term_cd])
+      }
+    end
+
+    def get_latest_current_standing(academic_standings)
+      return nil unless academic_standings.any?
+      academic_standings.sort_by!{|s| [s['term_id'], s['action_date']]}.reverse!
+      academic_standings[0]
+    end
+
+  end
+end
--- a/spec/controllers/my_standings_controller_spec.rb
+++ b/spec/controllers/my_standings_controller_spec.rb
@@ -1,0 +1,22 @@
+describe MyStandingsController do
+  before do
+    allow_any_instance_of(EdoOracle::Queries).to receive(:get_academic_standings).and_return({})
+  end
+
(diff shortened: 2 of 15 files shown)


app/models/my_academics/my_standings.rb
```diff
@@ -0,0 +1,68 @@
...
+      academic_standings ||= []
+    end
+
+    def parse_standings(academic_standings)
+      parsed_standings = empty_standings
+      current_standing = get_latest_current_standing(academic_standings)
+      return parsed_standings unless current_standing
+      # Its possible to have more than one current standing with same term and action date
+      academic_standings.each do |standing|
+        if standing['term_id'] == current_standing['term_id'] && standing['action_date'] == current_standing['action_date']
```
davidlimcheng: @nkutub Are we guaranteed that these columns will be present?  May be a good idea to use `#try` here just in case EdoDB returns nil values.

nkutub: @davidlimcheng because its a query it will always return a column even if the value is nil.

app/models/my_academics/my_standings.rb
```diff
@@ -0,0 +1,68 @@
...
+    def empty_standings
+      {
+        current_standings: [],
+        standings_history: []
+      }
+    end
+
+    def parse_standing(standing)
+      term = Berkeley::TermCodes.from_edo_id(standing['term_id'])
+      {
```
davidlimcheng: @nkutub Not a blocker, but for future reference, I think the [HashConverter](https://github.com/ets-berkeley-edu/calcentral/blob/master/lib/hash_converter.rb) module could be very helpful for these types of things.

app/models/my_academics/my_standings.rb
```diff
@@ -0,0 +1,68 @@
...
+    end
+
+    def parse_standings(academic_standings)
+      parsed_standings = empty_standings
+      current_standing = get_latest_current_standing(academic_standings)
+      return parsed_standings unless current_standing
+      # Its possible to have more than one current standing with same term and action date
+      academic_standings.each do |standing|
+        if standing['term_id'] == current_standing['term_id'] && standing['action_date'] == current_standing['action_date']
+          parsed_standings[:current_standings].push(parse_standing(standing)) unless parsed_standings[:current_standings].any?
```
davidlimcheng: @nkutub In your comment on line 31, you noted that it's possible to have more than one current standing with the same term and action date. But in this line, we're only taking the first one that matches.  Should we not be considering all possible standings for a student?

spec/controllers/my_standings_controller_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+  describe '#get_feed' do
+    let(:feed) { :get_feed }
+
+    it_behaves_like 'an unauthenticated user'
+
+    it 'should get a non-empty feed for an authenticated (but fake) user' do
+      session['user_id'] = '0'
+      get :get_feed
+      json_response = JSON.parse(response.body)
+      puts json_response
```
davidlimcheng: @nkutub Renegade `puts`

src/assets/templates/widgets/academic_standing.html
```diff
@@ -0,0 +1,64 @@
...
+          </td>
+          <td class="cc-table-right">
+            <div data-ng-bind="standing.termName"></div>
+          </td>
+        </tr>
+        <tr data-ng-if="standing.acadStandingStatus !== 'GST'">
+          <td></td>
+          <td colspan="2">
+            <span data-ng-bind="standing.acadStandingActionDescr" class="cc-standing-sub-title"></span><br>
+            <span data-ng-bind="'Please consult with your college advisor for guidence.'" class="cc-standing-sub-title"></span>
```
davidlimcheng: @nkutub Typo in guidance. Also, our current codebase also seems to only use ngBind for variables so that things like {{variable}} doesn't flash on the screen while rendering. For static strings like this, I'd actually prefer it to be outside of ngBind, just for readability:

`<span class="cc-standing-sub-title">Please consult with your college advisor for guidance.</span>`

src/assets/templates/widgets/status_and_holds.html
```diff
@@ -5,6 +5,21 @@ <h2>Status and Holds</h2>
 
   <div class="cc-widget-padding" data-cc-spinner-directive="statusHolds.isLoading">
     <h3 class="cc-status-holds-status-header" data-ng-if="regStatus.show || showResidency">Status</h3>
+     <div data-ng-if="!api.user.profile.academicRoles.law && !api.user.profile.academicRoles.grad">
+      <div class="cc-status-holds-section">
+        <div data-ng-if="residency.official.description">
+          <h4>Academic Standing</h4>
+          <div>
+            <button class="cc-button-link" data-ng-click="academics.showStanding = true" data-ng-show="!academics.showStanding">Show Academic Standing</button>
```
davidlimcheng: @nkutub Not a blocker, but you can also simplify these two lines by using academics.showStanding = !academics.showStanding.

app/models/my_academics/my_standings.rb
nkutub: Yes, the logic is if we find more than one record for the latest term, we only add the first one as the status and skip all others. This loop is necessary to keep the additional latest standings from ending up in the history.

spec/models/my_academics/my_academic_standings_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+      it_behaves_like 'a model receiving no data'
+    end
+    context 'when AcademicStanding query is populated' do
+      before do
+        EdoOracle::Queries.stub(:get_academic_standings) { standings_response }
+      end
+      it 'should successfully return a response' do
+        standings = subject[:feed][:standingsHistory]
+        expect(standings.count).to eq 2
+
```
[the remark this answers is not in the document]
nkutub: the white spaces make it easier to read.

nkutub:
@davidlimcheng all fixed!

src/assets/templates/widgets/status_and_holds.html
redconfetti: @davidlimcheng - How would that simplify it? It seems simple now, and also less error proof.

davidlimcheng:
Thanks @nkutub!

PR #7106, opened 2018-05-08
SISRP-40131 - Update README.md and .gitignore

marcpechaitis:
[SISRP-40131](https://jira.berkeley.edu/browse/SISRP-40131)

Updates the Installation Instructions in README.md and adds items to .gitignore (package-lock.json && Prettier)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,6 +73,9 @@ public/styleguide
 # Npm log
 npm-debug.log
 
+# Npm package-lock.json
+package-lock.json
+
 # Vagrant
 .vagrant/
 
@@ -81,3 +84,6 @@ npm-debug.log
 
 # VScode Javascript config
 jsconfig.json
+
+# Prettier ignore file
+.prettierignore
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@
 
 ## Dependencies
 
+* Administrator privileges
 * [Bundler](http://gembundler.com/rails3.html)
 * [Git](https://help.github.com/articles/set-up-git)
 * [JDBC Oracle driver](http://www.oracle.com/technetwork/database/enterprise-edition/jdbc-112010-090769.html)
@@ -13,14 +14,13 @@
 * [JRuby 9.1.14.0](http://jruby.org/)
 * [Node.js >=0.10.30](http://nodejs.org/)
 * [PostgreSQL](http://www.postgresql.org/)
-* [Rails 3.2.x](http://rubyonrails.org/download)
-* [Rubygems](http://rubyforge.org/frs/?group_id=126)
+* [Rubygems 2.5.1](https://rubygems.org/pages/download)
 * [RVM](https://rvm.io/rvm/install/) - Ruby version managers
 * [xvfb](http://xquartz.macosforge.org/landing/) - xvfb headless browser, included for Macs with XQuartz
 
 ## Installation
 
-1. Install Java 8 JDK:
+1. Install Java 8 JDK ([8u172](http://www.oracle.com/technetwork/java/javase/downloads/index.html)):
 http://www.oracle.com/technetwork/java/javase/downloads/index.html
 
 1. Install Postgres:
@@ -40,7 +40,6 @@ http://www.oracle.com/technetwork/java/javase/downloads/index.html
     ```bash
     brew update
     brew install postgresql
-    initdb /usr/local/var/postgres
     ```
 
     1. __For Mountain Lion & Mavericks users ONLY:__  [Fix Postgres paths](http://nextmarvel.net/blog/2011/09/brew-install-postgresql-on-os-x-lion/).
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -73,6 +73,9 @@ public/styleguide
 # Npm log
 npm-debug.log
 
+# Npm package-lock.json
+package-lock.json
```
redconfetti: @marcpechaitis - Are we really wanting to not include [package-lock.json](https://docs.npmjs.com/files/package-lock.json)? Perhaps we're not using it yet, but I think we should. It would help keep us locked into a stable set of packages that were used with the app during testing.

If we're not locking the versions of the NodeJS packages we're using down to specific versions, our build system might apply a patch update that breaks something right before we cut our release. This could have happened with [SISRP-28750](https://jira.berkeley.edu/browse/SISRP-28750). We created [SISRP-28768](https://jira.berkeley.edu/browse/SISRP-28768) specifically to address this at some point in the future.

README.md
```diff
@@ -14,13 +15,13 @@
 * [Node.js >=0.10.30](http://nodejs.org/)
 * [PostgreSQL](http://www.postgresql.org/)
 * [Rails 3.2.x](http://rubyonrails.org/download)
```
lyttam: @marcpechaitis Hmm, this seems pretty old. We've been on Rails 4.1 for a while, and just upgraded to 4.2.8. Of course I didn't catch this in the PR that upgraded Rails...

I think we should either change this line or remove it - thoughts? While Rails is a dependency, it is managed by bundler - and we don't list any other bundler-managed gems here, so perhaps we should remove it.

@davidlimcheng @redconfetti

lyttam:
Nice work @marcpechaitis!  I don't know if anyone explained the workflow yet, but what Jason said... and we also include a link to the PR in a comment on the Jira.  This helps us keep track of what was done and why (in lieu of an actual Jira-Git-Bamboo integration, which does exist).

.gitignore
marcpechaitis: @redconfetti I wasn't sure if we wanted to include package-lock.json or not.  I'm happy to include it if you think it will be useful.  (TBH, @davidlimcheng said "yes include it in ignore so we can decide as a team".

README.md
marcpechaitis: @lyttam I can do either change or remove it.  I'll leave it up to the team decide which is best option.
cc: @davidlimcheng  @redconfetti

.gitignore
redconfetti: @marcpechaitis - Okay. Ignore it for now then.

README.md
redconfetti: @marcpechaitis - I say update it.

redconfetti:
@marcpechaitis - Yeah, we have a [workflow document](https://wikihub.berkeley.edu/display/SIS/CalCentral+Developer+Workflow) that we forgot to point you to that explains how we contribute code.

README.md
marcpechaitis: @lyttam yeah I'm leaning towards removing it, since it will be getting installed via bundler and keeping it as a "dependency" implies a need for developer to install it intentionally.

.gitignore
davidlimcheng: @redconfetti @marcpechaitis I actually tried implementing `package-lock.json` before, but ran into a bunch of problems.  That was with an external package, and I believe `npm`'s default behavior now is to create a `package-lock.json` file.  For what it's worth, I think it'd be good to try out when we update Node to a version that has that feature, but for now I think including it in `.gitignore` is the way to go.

README.md
davidlimcheng: +1 for removing - I think it creates more confusion explicitly listing it, when that step will be taken care of with `bundle install`.

README.md
```diff
@@ -14,13 +15,13 @@
...
-* [Rubygems](http://rubyforge.org/frs/?group_id=126)
+* [Rubygems 2.5.1](https://rubygems.org/pages/download)
 * [RVM](https://rvm.io/rvm/install/) - Ruby version managers
 * [xvfb](http://xquartz.macosforge.org/landing/) - xvfb headless browser, included for Macs with XQuartz
 
 ## Installation
 
-1. Install Java 8 JDK:
+1. Install Java 8 JDK ([8u172](http://www.oracle.com/technetwork/java/javase/downloads/index.html)):
```
davidlimcheng: What do you guys think of removing the minor version here?  I don't think all of us are actually on the same version, and I'd hate to have to update this every time we update our minor versions.  

@nkutub @lyttam @redconfetti @marcpechaitis

davidlimcheng:
Awesome work @marcpechaitis!

marcpechaitis:
I have amended the PR:

- Removed dependency for Rails b/c its happening automagically via bundler

- Added link to [workflow document](https://wikihub.berkeley.edu/pages/viewpage.action?spaceKey=SIS&title=CalCentral+Developer+Workflow#CalCentralDeveloperWorkflow-CreateYourOwnFork)'s instructions on how to create a fork (as a newbie it took me a couple minutes to figure out what to do here when I went through the process)

README.md
redconfetti: @davidlimcheng - Do we know if the minor version matters to how successfully the app runs?

raydavis:
Thanks for the ping, @redconfetti . So long as your trusty ETS comrades-in-GitHub have access to the relevant Wiki pages, I don't see a problem -- if ETS changes something that impacts documentation we can't edit ourselves, you'd just be stuck with a minor follow-up chore. I imagine maintenance will continue to be an on-and-off affair regardless of whether developer documentation is stored there or GitHub.

redconfetti:
@marcpechaitis - I've created a [Developer Workflow](https://github.com/ets-berkeley-edu/calcentral/wiki/Workflow) wiki doc

marcpechaitis:
@lyttam , @redconfetti , @davidlimcheng , @nkutub

I have amended the PR:

- Changed reference from [CalCentral Confluence](https://wikihub.berkeley.edu/pages/viewpage.action?spaceKey=SIS&title=CalCentral+Developer+Workflow#CalCentralDeveloperWorkflow-CreateYourOwnFork) page to the Github [wiki](https://github.com/ets-berkeley-edu/calcentral/wiki/Workflow)

lyttam:
Thanks @marcpechaitis, looks like this is ready to go. Congrats on the most commented PR, like, probably ever?

PR #7109, opened 2018-05-09
SISRP-38422 - Surfaces cumulative Law units

lyttam:
https://jira.berkeley.edu/browse/SISRP-38422

Changes affect Profile and Academic Summary


app/models/my_academics/gpa_units.rb
```diff
@@ -66,15 +67,32 @@ def parse_cumulative_gpa(statuses)
...
+      unit_totals = EdoOracle::Queries.get_career_unit_totals(@uid)
+      has_active_career = unit_totals.any? do |career|
+        active? career
+      end
+      result = {
+        totalUnits: 0,
+        totalLawUnits: 0
+      }
+      unit_totals.each do |career|
+        result[:totalUnits] += (career['total_cumulative_units'] if active?(career) || !has_active_career).to_f
```
davidlimcheng: @lyttam Just want to clarify what's happening here - is the objective here to pull in units only for the user's active career, unless the user has no active career, in which case we'll pull in all units?

lyttam: @davidlimcheng Yes exactly.  I think that's what the academic-summary API does, thought I can't confirm it.

spec/models/my_academics/gpa_units_spec.rb
```diff
@@ -18,122 +18,147 @@
...
     end
 
-    context 'when the Transfer Credit API provides no data' do
-      before do
-        allow(CampusSolutions::TransferCredit).to receive(:new).and_return double(get: nil)
+    describe '#get_cumulative_units', testext: false do
+      subject { described_class.new(uid).get_cumulative_units }
+
+      context 'when user is active in a program' do
+        let(:uid) { 300216 }
```
davidlimcheng: @lyttam Not a blocker for merge, but maybe we should give these IDs a descriptive variable name so that if they do need to be used elsewhere in the spec, or if another spec needs a similar test case, they'll be easily identifiable.

spec/models/edo_oracle/queries_spec.rb
```diff
@@ -59,6 +59,36 @@
...
   end
 
+  describe '#get_career_unit_totals', testext: false do
+    subject { described_class.get_career_unit_totals(uid) }
+    let(:uid) { 300216 }
+
+    it_behaves_like 'a successful query'
+
+    it 'returns the expected result' do
+      puts subject.to_yaml
```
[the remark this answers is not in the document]
lyttam: @davidlimcheng thanks! fixed.

spec/models/my_academics/gpa_units_spec.rb
lyttam: @davidlimcheng good idea, but I don't know a good way to do it. The variable used on line 143 needs a consistent, generic name because it is used in all the examples in this group.

I could make a set of variables like this at the top of the spec
```
let(:law_uid) { 300216 }
```

and then in the examples, use 
```let(:uid) { law_uid} ```

But I don't know if it's worth the effort, what do you think?

lyttam: @davidlimcheng I think I'd rather let the `context` explain what type of UID it is. In this case, 'when user is active in a program'.

davidlimcheng:
Nice, @lyttam

PR #7111, opened 2018-05-09
SISRP-39160 - UX Dev: Summer-Visitor Plan + Degree-Seeking Plan Student Experience

redconfetti:
https://jira.berkeley.edu/browse/SISRP-39160

* Hides links in Student Resources card from non-degree seeking Summer Visitors
* Hides Academic Career, Level, Terms Information, and Cumulative Units from non-degree seeking Summer Visitors
* Hides residency section of Status & Holds card for non-degree seeking Summer Visitors
* Hides Degree Progress cards from non-degree seeking Summer Visitors
* Hides Advising card from non-degree seeking Summer Visitors


app/models/edo_oracle/queries.rb
```diff
@@ -639,5 +639,26 @@ def self.get_academic_standings (student_id)
...
+          WHERE
+            STUDENT_ID = '#{student_id}'
+        SQL
+        return result
+      else
+        []
+      end
+    end
+
+    def self.campus_solutions_id(uid)
```
davidlimcheng: @redconfetti, what do you think about doing this in `MyAcademics::MyTermCpp`?  I can't think of any serious drawbacks to doing it here, but it would be nice to keep a clean separation of concerns.  You can also include the `User::Identifiers` mixin in `MyAcademics::MyTermCpp` to use the existing `#lookup_campus_solutions_id`

app/models/edo_oracle/queries.rb
```diff
@@ -639,5 +639,26 @@ def self.get_academic_standings (student_id)
...
+      if student_id = campus_solutions_id(uid)
+        result = safe_query <<-SQL
+          SELECT
+            TERM_ID as term_id,
+            ACAD_CAREER_CODE as acad_career,
+            ACAD_PROGRAM as acad_program,
+            ACAD_PLAN as acad_plan
+          FROM SISEDO.UC_TERM_CPPV00_VW
+          WHERE
+            STUDENT_ID = '#{student_id}'
```
lyttam: @redconfetti I haven't done this with any of mine, but I think we should be including `INSTITUTION = 'UCB01'` in all of our queries for better performance.  I mean to go back and add it to the queries I've been working on.

spec/models/edo_oracle/queries_spec.rb
```diff
@@ -20,6 +20,8 @@
     allow(Settings.features).to receive(:hub_term_api).and_return false
   end
 
+  let(:oski_uid) { '61889' }
```
lyttam: @redconfetti I suggest moving this line underneath `describe '.get_student_term_cpp'`, since `oski_uid` is only relevant to that example group.

app/models/edo_oracle/queries.rb
redconfetti: @lyttam - Yes. That was in the works. Ken Hensel made the modifications in DEV, I'll apply that now.

app/models/edo_oracle/queries.rb
redconfetti: @davidlimcheng - Up until this point we've used UID with all queries. This is the first time that STUDENT_ID is used instead. I was trying to keep the same convention.

It does seem like the convention of keeping queries as simple as possible has been broken here, so I'll move ahead with it.

src/assets/templates/academics_student_profile.html
```diff
@@ -71,7 +71,7 @@ <h3 data-ng-if="!collegeAndLevel.isCurrent && api.user.profile.roles.student &&
 
       <div data-ng-include="'widgets/academics/college_and_level.html'"></div>
 
-      <div class="cc-table cc-table-top-border cc-clearfix" data-ng-if="gpaUnits.totalUnits > 0 || gpaUnits.totalLawUnits > 0">
+      <div class="cc-table cc-table-top-border cc-clearfix" data-ng-if="!api.user.profile.academicRoles.nonDegreeSeekingSummerVisitor && (gpaUnits.totalUnits > 0 || gpaUnits.totalLawUnits > 0)">
         <table>
```
redconfetti: @lyttam - This rose as a conflict when I rebased just now. I believe I merged my changes with yours appropriately.

app/models/edo_oracle/queries.rb
```diff
@@ -664,5 +664,19 @@ def self.get_academic_standings (student_id)
...
+    def self.get_student_term_cpp(student_id)
+      result = safe_query <<-SQL
+        SELECT
+          TERM_ID as term_id,
+          ACAD_CAREER_CODE as acad_career,
+          ACAD_PROGRAM as acad_program,
+          ACAD_PLAN as acad_plan
+        FROM SISEDO.STUDENT_TERM_CPPV00_VW
+        WHERE
+          INSTITUTION = 'UCB01' AND
```
marcpechaitis: Hardcoding as straight-up text 'UCB01' makes me cringe.  Is it possible to have a global variable to represent 'UCB01' instead?

lyttam:
@redconfetti This just needs a passing Travis build.

PR #7116, opened 2018-05-10
SISRP-38837 - hides non-Law enrollments from active Law students

lyttam:
https://jira.berkeley.edu/browse/SISRP-38837

- Non-Law students can see all past and present enrollments.
- Active Law students can only see their active career enrollments.
- A dual career law student (active in 2 careers) can see enrollments in both active careers


nkutub:
LGTM @lyttam

PR #7118, opened 2018-05-10
SISRP-40187 - Need SF negative service indicators to display in CalCentral if configured to do so in BCS

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-40187


nkutub:
LGTM @davidlimcheng

PR #7119, opened 2018-05-11
SISRP-39683 - Display grading links for all terms an instructor taught

redconfetti:
https://jira.berkeley.edu/browse/SISRP-39683

**Note:** I know this reduces testing coverage, but I plan on building out the new Rspec test harness for grading in a followup / QA PR that I'll be working on. @nkutub authorized that I develop in this manner to ensure that we get this in by deadline.


app/models/edo_oracle/queries.rb
```diff
@@ -663,5 +663,19 @@ def self.get_academic_standings (student_id)
...
+        SELECT
+          ACAD_CAREER as acad_career,
+          TERM_ID as term_id,
+          SESSION_CODE as session_code,
+          MID_BEGIN_DT as mid_term_begin_date,
+          MID_END_DT as mid_term_end_date,
+          FINAL_BEGIN_DT as final_begin_date,
+          FINAL_END_DT as final_end_date
+        FROM
+          SISEDO.GRADING_DATES_CS_V00_VW
```
lyttam: @redconfetti Is there any way to add a WHERE clause to this query?  Running a query with no qualifiers makes me nervous.  If this view grows in size in the future, it could affect performance.

app/models/my_academics/grading.rb
```diff
@@ -17,126 +23,106 @@ def merge(data)
...
       if has_law?(semester[:classes])
         semester.merge!({
-          gradingAssistanceLinkLaw: grading_period.links.law
+          gradingAssistanceLinkLaw: grading_config[:links][:law].url
         })
       end
     end
 
+    #####################################################################################################################
+    # Add Grading Dates to Semesters and Summer Sections
```
lyttam: @redconfetti Comment unnecessary; the name of the function is self-explanatory

app/models/my_academics/grading.rb
```diff
@@ -17,126 +23,106 @@ def merge(data)
...
-            gradingAssistanceLink: grading_period.links.final
+            gradingAssistanceLinkMidpoint: grading_config[:links][:midterm].url,
+            gradingAssistanceLink: grading_config[:links][:general].url
           })
         end
       end
       if has_law?(semester[:classes])
         semester.merge!({
-          gradingAssistanceLinkLaw: grading_period.links.law
+          gradingAssistanceLinkLaw: grading_config[:links][:law].url
```
lyttam: @redconfetti I would discourage us from continuing to use this pattern of having Hash keys present only when they have a value.  I'd rather see this function construct a Hash with all keys present, each key having a `nil` value when appropriate.

app/models/my_academics/grading.rb
```diff
@@ -17,126 +23,106 @@ def merge(data)
...
+        if (grading_periods)
+          begin_date = grading_periods[:final_begin_date]
+          end_date = grading_periods[:final_end_date]
+          section.merge!({
+            gradingPeriodStartDate: begin_date,
+            gradingPeriodEndDate: end_date,
+            gradingPeriodEndDateFormatted: format_period_end_summer(end_date).to_s
+          })
+        else
+          logger.error "No grading periods found for term: #{term_id}, career: #{GRADING_TYPE_TO_CAREER_MAP[grading_career_type]}"
```
lyttam: @redconfetti Is this really an ERROR? Seems like it should be a WARN or INFO.

app/models/my_academics/grading_config.rb
```diff
@@ -0,0 +1,53 @@
+module MyAcademics
+  class GradingConfig
```
lyttam: @redconfetti Recommend this class be renamed to `Grading`.  'Config' doesn't make sense here.

app/models/my_academics/grading_config.rb
```diff
@@ -0,0 +1,53 @@
...
+        law: nil
+      }
+      faculty_cat = Links::LinkCategory.where(root_level: false, name: 'Faculty').first
+      grading_cat = Links::LinkCategory.where(root_level: false, name: 'Grading').first
+      if !(faculty_cat && grading_cat)
+        Rails.logger.error "'Faculty' or 'Grading' link categories are not present. Refresh links by requesting /api/my/campuslinks/refresh"
+      else
+        link_section = Links::LinkSection.where(link_top_cat_id: faculty_cat.id, link_sub_cat_id: grading_cat.id).first
+        if (link_section_links = link_section.links.to_a)
+          links[:general] = link_section_links.find {|link| link.name == 'Assistance with Grading: General'}
```
lyttam: @redconfetti Would it be possible to find the link based on its ID rather than its name?  Names can and probably will be changed, while the ID is more reliable.

app/models/my_academics/grading_module.rb
```diff
@@ -53,70 +53,58 @@ def grading_status_mapping
...
-        "Q4" => :Q4
+        "1" => 'Q3',
+        "6W1" => 'Q4',
+        "6W2" => 'Q4',
+        "8W" => 'Q4',
+        "10W" => 'Q4',
+        "Q1" => 'Q1',
+        "Q2" => 'Q2',
+        "Q3" => 'Q3',
+        "Q4" => 'Q4'
```
lyttam: @redconfetti I think these are better as symbols (e.g. `:Q4`).  You can use the `#intern` function when you need to compare a string to a symbol.

app/models/my_academics/grading_module.rb
```diff
@@ -53,70 +53,58 @@ def grading_status_mapping
...
       }
     end
 
+    # Returns true if any classes found with Law department
```
lyttam: @redconfetti Comments unnecessary on this and the following function.  The function name says it all.

app/models/my_academics/grading_module.rb
```diff
@@ -53,70 +53,58 @@ def grading_status_mapping
...
 
     def format_period_end(end_date)
+      return nil if end_date.blank?
       return end_date.to_date.strftime('%b %d') if DateTime.now.year == end_date.to_date.year
       end_date.to_date.strftime('%b %d, %Y')
     end
 
-    def format_period_start_summer(start_date)
-      start_date.to_date.strftime('%m/%d')
+    # Returns true if any classes found with non-Law department
```
lyttam: @redconfetti Again, no comment needed here.

lyttam:
@redconfetti It sure is nice to see `grading_period` disappear from Settings.yml! Thanks for taking on this huge task.

app/models/edo_oracle/queries.rb
redconfetti: @lyttam - It's returning a global set of values. There isn't a scenario that exists where we'd limit it. This will be providing us with the grading windows for all terms Fall 2016+. It will be growing, but only by 4 records every Fall and Spring term (4 careers), and no more than 16 records every summer term (due to summer sessions).

app/models/my_academics/grading_config.rb
redconfetti: @lyttam - There is already a MyAcademics::Grading class. This one is providing what used to be configured manually, grading periods and links.

Would it be better if we split it into `MyAcademics::GradingDates` and `MyAcademics::GradingLinks`?

app/models/my_academics/grading_config.rb
```diff
@@ -0,0 +1,53 @@
...
+    def self.grading_links
+      links = {
+        general: nil,
+        midterm: nil,
+        law: nil
+      }
+      faculty_cat = Links::LinkCategory.where(root_level: false, name: 'Faculty').first
+      grading_cat = Links::LinkCategory.where(root_level: false, name: 'Grading').first
+      if !(faculty_cat && grading_cat)
+        Rails.logger.error "'Faculty' or 'Grading' link categories are not present. Refresh links by requesting /api/my/campuslinks/refresh"
```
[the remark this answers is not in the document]
redconfetti: @lyttam - It is an error that I think we should log. The links should be present in our database, loaded from `/public/json/campuslinks.json`. I want to code this as an error logged should the link disappear from the database, so that the Admins will report the issue to us and we can run an HTTP request to /api/my/campuslinks/refresh should that occur.

app/models/my_academics/grading_config.rb
redconfetti: Unfortunately we do not know the ID of links that are imported into the database from [/public/json/campuslinks.json](https://github.com/ets-berkeley-edu/calcentral/blob/4ce28fd4/public/json/campuslinks.json). It likely varies from system to system.

We have prior art for code that does this in the [Financial Resources card](https://github.com/ets-berkeley-edu/calcentral/blob/4ce28fd4/src/assets/javascripts/angular/controllers/widgets/financesLinksController.js#L75). Perhaps we should introduce some sort of ID string as part of that model so that we can refer to them by a fixed ID rather than the name of the link that's configured.

I really don't like the idea of this feature being held up by the need to refactor the way our [Campus Links](https://wikihub.berkeley.edu/display/SIS/CalCentral+My+Campus+Links+API) are modeled. I'm not sure how much time would be required to implement that change.

Would you agree to a follow-up PR? @nkutub any thoughts?

app/models/my_academics/grading_config.rb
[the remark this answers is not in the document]
redconfetti: @lyttam - Are you suggesting that it be a WARN instead?

app/models/my_academics/grading_config.rb
lyttam: @redconfetti yeah, good idea. It's better if each model only has one job.

PR #7120, opened 2018-05-11
SISRP-38422 - Refactors Profile and Academic Summary cards

lyttam:
https://jira.berkeley.edu/browse/SISRP-38422

Includes visual updates per the design mockups.

Makes some slight changes to styling on Profile and Academic Summary to make the two cards consistent with each other.

### DIFF
--- a/src/assets/stylesheets/_academic_summary.scss
+++ b/src/assets/stylesheets/_academic_summary.scss
@@ -2,6 +2,9 @@
   .cc-academic-summary-border-wrapper {
     border: 1px solid $cc-color-very-light-grey;
     margin: 15px;
+    &:first-child {
+      margin-top: 0;
+    }
   }
   .cc-academic-summary-enrollment-header {
     display: inline-block;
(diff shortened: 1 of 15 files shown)


lyttam:
Will fix build failures and amend.

lyttam:
Ready to review/merge.

src/assets/stylesheets/_academic_summary.scss
```diff
@@ -1,7 +1,10 @@
-.cc-academic-summary {
+div.cc-academic-summary {
   .cc-academic-summary-border-wrapper {
```
redconfetti: @lyttam - Is this `div` specifier necessary? What about the `h3`, `table`, and `options` below. Something that we've tried to maintain in the past is that we try to use the [minimum amount of specificity](https://css-tricks.com/strategies-keeping-css-specificity-low/), starting with just applying classes to change the styling of an element... and using two consecutive classes such as `.cc-academic-summary .cc-academic-summary-header` to be more specific, avoiding element and ID specifiers as much as we can.

It also encourages the [OOCSS pattern](https://www.smashingmagazine.com/2011/12/an-introduction-to-object-oriented-css-oocss/) also.

lyttam: Actually, I did introduce it... will amend.

PR #7123, opened 2018-05-15
SISRP-39289 - Create CalCentral Link to Credentials for Law Transcript PDF - UX Dev w/ API

nkutub:
https://jira.berkeley.edu/browse/SISRP-39289
bug fixes


lyttam:
Makes sense.

PR #7126, opened 2018-05-16
NS-46 Fill missing SIS course data for Loch

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-46

There will be followup PRs for Nessie, and then for BOAC.


johncrossman:
Backwards compatible... Just the way we like it.

PR #7127, opened 2018-05-16
NOJIRA: Fixes expected grad term error and makes small style improvement on semesters

lyttam:
There was inconsistency between the Semesters card and the Academic Summary page in showing a term that the student withdrew from.

### DIFF
--- a/src/assets/stylesheets/_academic_summary.scss
+++ b/src/assets/stylesheets/_academic_summary.scss
@@ -17,7 +17,7 @@
     color: $cc-color-dim-gray;
     vertical-align: bottom;
   }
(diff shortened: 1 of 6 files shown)


redconfetti:
@lyttam - I can understand that if we make changes that doesn't affect functionality, such as Rspec tests, or documentation... but shouldn't we be associating work like this with some sort of JIRA?

lyttam:
@redconfetti Perhaps. This was just an error that @nkutub reported in Slack.

redconfetti:
@lyttam - Good call. It sounds like it doesn't change functionality. It's just a bug, and an aesthetic improvement.

PR #7128, opened 2018-05-16
SISRP-40400 - Hides Pts column when semester has no grade points

lyttam:
https://jira.berkeley.edu/browse/SISRP-40401

In order to appease the JS linter, I had to convert some keys from snake to camel case.  Eventually we can convert the entire feed to camel case but I don't want to tackle that whole project at this time.


redconfetti:
@lyttam - Yeah, I've seen `/* jshint camelcase: false */` in the header of some files to over-ride that requirement. We should continue to work towards that ideal as we modify the code.

PR #7131, opened 2018-05-17
NS-46 Fix unintentionally repositioned data item

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-46

Undoing the blunder in the previous commit, https://github.com/ets-berkeley-edu/calcentral/pull/7126


sandeepmjay:
LGTM  @raydavis. Merging.

PR #7132, opened 2018-05-17
SISRP-40400 - Hides Grading Basis from Law students

lyttam:
https://jira.berkeley.edu/browse/SISRP-40400


lyttam:
Please don't merge yet; I need to make another change.

lyttam:
Moved the logic to the front-end, since Law wants the Grading Basis to show on the class page but not on the Semesters card.

PR #7143, opened 2018-05-21
SISRP-38837 - Moves career filtering logic to the Semesters model

lyttam:
https://jira.berkeley.edu/browse/SISRP-38837

Since the query is shared by multiple callers, I had to move the filtering to the view model layer - otherwise, the filtering would sometimes get skipped because of caching.


redconfetti:
@lyttam Looks good. I like that it made the data models simpler too :)

PR #7148, opened 2018-05-22
SISRP-40483: Hides points on GRAD classes from concurrent students

lyttam:
https://jira.berkeley.edu/browse/SISRP-40483

No student should see grade points on a LAW class. This change hides points on GRAD classes also, if the student is in a concurrent GRAD+LAW program.

Also adds a caching layer to the query that is used to determine whether the student is concurrent.


nkutub:
LGTM @lyttam

PR #7153, opened 2018-05-23
SISRP-39248 - Fixes bug with sorting of tasks to order further action needed tasks last

redconfetti:
https://jira.berkeley.edu/browse/SISRP-39248


nkutub:
LGTM @redconfetti

PR #7162, opened 2018-05-23
SISRP-39683 - Exposes grading link when roster is present

redconfetti:
https://jira.berkeley.edu/browse/SISRP-39683

See QA PR #7162


nkutub:
LGTM @redconfetti

PR #7164, opened 2018-05-24
SISRP-40515 - Undergrad non-degree student shouldn't see Academic Standing at all

nkutub:
https://jira.berkeley.edu/browse/SISRP-40515

### DIFF
--- a/src/assets/templates/widgets/status_and_holds.html
+++ b/src/assets/templates/widgets/status_and_holds.html
@@ -5,7 +5,7 @@
 
   <div class="cc-widget-padding" data-cc-spinner-directive="statusHolds.isLoading">
     <h3 class="cc-status-holds-status-header" data-ng-if="regStatus.show || showResidency">Status</h3>
-     <div data-ng-if="api.user.profile.roles.undergrad || (api.user.profile.roles.advisor && targetUser.academicRoles.current.ugrd)">
+     <div data-ng-if="(api.user.profile.academicRoles.ugrd && !api.user.profile.academicRoles.ugrdNonDegree) || (api.user.profile.roles.advisor && targetUser.academicRoles.current.ugrd && !targetUser.academicRoles.current.ugrdNonDegree)">
       <div class="cc-status-holds-section">
         <div>
           <h4>Academic Standing</h4>


src/assets/templates/widgets/status_and_holds.html
```diff
@@ -5,7 +5,7 @@ <h2>Status and Holds</h2>
 
   <div class="cc-widget-padding" data-cc-spinner-directive="statusHolds.isLoading">
     <h3 class="cc-status-holds-status-header" data-ng-if="regStatus.show || showResidency">Status</h3>
-     <div data-ng-if="api.user.profile.roles.undergrad || (api.user.profile.roles.advisor && targetUser.academicRoles.ugrd)">
+     <div data-ng-if="(api.user.profile.academicRoles.ugrd && !api.user.profile.academicRoles.ugrdNonDegree) || (api.user.profile.roles.advisor && targetUser.academicRoles.ugrd && !targetUser.academicRoles.ugrdNonDegree)">
```
lyttam: @nkutub why use the `ugrdNonDegree` role here? That's not the same as UCBX.

The `concurrent` academic role is given to students with UCBX careers.

nkutub: @lyttam , no concurrent is when they have a degree-seeking grad or undergrad and also UCBX (non degree seeking) if you look at the use case in the JIRA you will see its marked as non degree seeking and the my status does not list concurrent = true

lyttam: Also, `academicRoles.ugrdNonDegree` is only assigned if the student is in the undergrad non-degree seeking program. There are other (non-undergrad) non-degree seeking programs.

lyttam: Per discussion with Nader, the Academic Standing should be hidden from undergrad non-degree seeking students, not UCBX students.  I updated the title to reflect this.

PR #7168 introduced breaking changes to academic roles. This PR will need to be amended to account for those changes before it can be merged.

src/assets/templates/widgets/status_and_holds.html
```diff
@@ -5,7 +5,7 @@ <h2>Status and Holds</h2>
 
   <div class="cc-widget-padding" data-cc-spinner-directive="statusHolds.isLoading">
     <h3 class="cc-status-holds-status-header" data-ng-if="regStatus.show || showResidency">Status</h3>
-     <div data-ng-if="api.user.profile.roles.undergrad || (api.user.profile.roles.advisor && targetUser.academicRoles.current.ugrd)">
+     <div data-ng-if="(api.user.profile.academicRoles.ugrd && !api.user.profile.academicRoles.ugrdNonDegree) || (api.user.profile.roles.advisor && targetUser.academicRoles.current.ugrd && !targetUser.academicRoles.current.ugrdNonDegree)">
```
lyttam: @nkutub sorry, didn't notice until after I merged this and #7165 - you missed a couple references to `api.user.profile.academicRoles`. Please update these to `api.user.profile.academicRoles.current`.

PR #7168, opened 2018-05-24
SISRP-40499: Introduces historical academic roles

lyttam:
https://jira.berkeley.edu/browse/SISRP-40499

*Important*:  `MyAcademicRoles` feed has a new structure:

```
academicRoles: {
  current: {
    ...
  },
  historical: {
  ... 
  }
}
```

`academicRoles.current` replaces `academicRoles` (these are the same set of roles derived from the academic-status response that we've been using).

`academicRoles.historical` is a list of all the roles a student has ever had, based on the plan, program, and career of every term they've enrolled in.  This information was needed for displaying the Law transcript link to students who were formerly Law.


davidlimcheng:
LGTM @lyttam, holding off on merge for potential merge conflicts we talked about

redconfetti:
@lyttam - I figured it might come to this. I didn't know if I should have done it by year, or in a group like you did. CPP view has entered the building

app/models/concerns/academic_roles.rb
```diff
@@ -47,6 +47,7 @@ module AcademicRoles
     # Role(s) assigned to a user if they are in a program associated with that role.
     ACADEMIC_PROGRAM_ROLES = [
       {role_code: 'lettersAndScience', match: ['UCLS']},
+      {role_code: 'degreeSeeking', match: [], exclude: ['GNODG', 'LNODG', 'UNODG', 'XCCRT', 'XFPF']},
```
redconfetti: @lyttam Good call making this a role of its own and moving the logic to the front-end based on the combination of both roles.

redconfetti: @lyttam - So this checks to make sure that the user has at least one plan/program that exists, but is not in one of these programs? I'm concerned that we might have a user that has no plans (maybe this won't happen) but that shows as degree seeking (false positive).

lyttam: @redconfetti yes good catch, a person who has never enrolled in a term would get the `degreeSeeking` role because they would have no programs.

I will try to fix this in a follow-up PR.

lyttam: @redconfetti actually, I checked the code and I don't think this scenario can happen. If there is no term, or if there is a term with no program, then we won't assign any roles.

PR #7193, opened 2018-05-31
SISRP-37833: Upgrades nokogiri

lyttam:
https://jira.berkeley.edu/browse/SISRP-37833

Remember to run `bundle install` after pulling this change into your local codebase.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -74,8 +74,7 @@ gem 'jruby-openssl', '0.9.19'
 gem 'addressable', '~> 2.3.4'
 
 # for parsing formatted html
-# Do NOT update until https://github.com/sparklemotion/nokogiri/issues/1114 is fixed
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -75,7 +75,7 @@ gem 'addressable', '~> 2.3.4'
 
 # for parsing formatted html
 # Do NOT update until https://github.com/sparklemotion/nokogiri/issues/1114 is fixed
```
redconfetti: Does this include the fix from sparklemotion/nokogiri#1114 ? If so, shouldn't we remove this comment?

redconfetti: @lyttam - Does this include the fix from sparklemotion/nokogiri#1114 ? If so, shouldn't we remove this comment?

lyttam: @redconfetti Good catch. Yes, the issue was resolved in v1.6.8.rc1.

PR #7208, opened 2018-06-08
SISRP-40307: To Law students, show only active career transfer credits

lyttam:
https://jira.berkeley.edu/browse/SISRP-40307

If student is or was ever a law student, and they have active career(s), show only the transfer credits for the active career(s) - otherwise show all.

In addition, I combined two different queries against `SISEDO.CLC_STUDENT_CAREERV00_VW` into one in order to leverage caching and reduce db calls.


app/models/edo_oracle/transfer_credit.rb
```diff
@@ -26,109 +32,69 @@ def get_feed
...
+        career = credit['career']
 
-    def get_transfer_credit_detailed
-      if (transfer_credits = EdoOracle::Queries.get_transfer_credit_detailed campus_solutions_id)
-        return parse_rows_detailed transfer_credits
+        next unless (career_description = CAREERS.try(:[], career.try(:intern)))
+        next if law_student? && !(career_whitelist.include? career)
+
+        is_law = :LAW == career.try(:intern)
+        result[career_description.downcase] ||= []
```
davidlimcheng: @lyttam Very minor, but since we're using `:downcase` twice, how about throwing it over to line 42?

lyttam: @davidlimcheng Ah yes, here we can.  In `get_transfer_credit_summary`, we can't.  I will amend with the change you suggest.

lyttam:
@davidlimcheng  Amended.

redconfetti:
@lyttam - LGTM

PR #7209, opened 2018-06-08
SISRP-40529: Prevents FPP enrollment call when not authorized

lyttam:
https://jira.berkeley.edu/browse/SISRP-40529


redconfetti:
@lyttam - LGTM

PR #7210, opened 2018-06-08
BOAC-919 Change data source for Course Section Allowed Units

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-919


paulkerschen:
To the max!

PR #7212, opened 2018-06-11
SISRP-40517: Adds new script to install dependencies for bamboo build

lyttam:
Per Ray's comment:  https://jira.berkeley.edu/browse/SISRP-40517?focusedCommentId=567849&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-567849

The script I'm adding is a copy of `bamboo-tests.sh` with the part that runs the tests removed. Once this is merged, I will update the Bamboo configuration to run this script instead of `bamboo-tests.sh`.

### DIFF
--- a/script/bamboo-install.sh
+++ b/script/bamboo-install.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+# Script to install necessary dependencies, for use on Bamboo CI
+
+# set up environment
+export JRUBY_OPTS="-Xcompile.invokedynamic=false -J-Xmx900m -J-Djruby.compile.mode=OFF"
+
+cd $( dirname "${BASH_SOURCE[0]}" )/..
+
+# Enable rvm and use the correct Ruby version and gem set.
+[[ -s "$HOME/.rvm/scripts/rvm" ]] && . "$HOME/.rvm/scripts/rvm"
+source .rvmrc
+GEMSET="$1"
+if [ -z "$1" ]; then
+  GEMSET="calcentral"
+fi
+rvm gemset use $GEMSET
+
+# get Ruby deps
+bundle install --local --retry 3 || { echo "WARNING: bundle install --local failed, running bundle install"; bundle install --retry 3 || { echo "ERROR: bundle install failed"; exit 1; } }
+bundle package --all || { echo "WARNING: bundle package failed"; exit 1; }


script/bamboo-install.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+# Script to install necessary dependencies, for use on Bamboo CI
+
+# set up environment
+export DISPLAY=":99"
+export JRUBY_OPTS="-Xcompile.invokedynamic=false -J-Xmx900m -J-Djruby.compile.mode=OFF"
```
lyttam: @raydavis Can I remove these two lines? I'm not sure what the `DISPLAY` variable is for. I believe the `JRUBY_OPTS` variable is mostly only relevant when running the test suite.

raydavis: * The `DISPLAY` is for front-end tests, I think, and should be droppable.

* Can't say whether the `JRUBY_OPTS` is needed just to run JRuby Bundler, but I notice that it's also set in `build-torquebox.sh`, so I'd probably play it safe myself.

* Speaking of `build-torquebox.sh`, if you want to go all out, you could skip adding this file in favor of adding the two extra {{bundle}} lines to the build script.

script/bamboo-install.sh
```diff
@@ -0,0 +1,22 @@
...
+GEMSET="$1"
+if [ -z "$1" ]; then
+  GEMSET="calcentral"
+fi
+rvm gemset use $GEMSET
+
+# get Ruby deps
+bundle install --local --retry 3 || { echo "WARNING: bundle install --local failed, running bundle install"; bundle install --retry 3 || { echo "ERROR: bundle install failed"; exit 1; } }
+bundle package --all || { echo "WARNING: bundle package failed"; exit 1; }
+./script/front-end-build.sh
```
raydavis: No need for a front-end-build in this script, since `build-torquebox.sh` also runs it.

lyttam: Cool, I'll remove it.

script/bamboo-install.sh
lyttam: Thanks. We could combine this script with `build-torquebox` - maybe once we're sure we don't want to run tests on Bamboo anymore. Having a separate script makes it *slightly* easier to switch back to running tests if we decide to, though I can't imagine why we'd do that.

lyttam:
@raydavis I removed a couple lines per your comments.  Should be ready to merge.

raydavis:
Next stop: Bamboo job configurations!

PR #7214, opened 2018-06-19
SISRP-39215: Improves Terms & Conditions backend validation

lyttam:
https://jira.berkeley.edu/browse/SISRP-39215

### DIFF
--- a/app/models/campus_solutions/terms_and_conditions.rb
+++ b/app/models/campus_solutions/terms_and_conditions.rb
@@ -19,12 +19,19 @@ module CampusSolutions
     end
 
     def valid?(params)
-      aid_years = []
-      aid_years_feed = CampusSolutions::MyAidYears.new(@uid).get_feed
-      aid_years_feed.try(:[], :feed).try(:[], :finaidSummary).try(:[], :finaidYears).try(:each) do |aid_year|
-        aid_years.push(aid_year.try(:[], :id).try(:to_i))
+      valid_aid_years = []
+      aid_years = CampusSolutions::MyAidYears.new(@uid).get_feed
+      aid_years.try(:[], :feed).try(:[], :finaidSummary).try(:[], :finaidYears).try(:each) do |aid_year|
+        unless prior_response_recorded? aid_year
+          valid_aid_years.push(aid_year.try(:[], :id).try(:to_i))
+        end
       end
-      aid_years.include?(params[:aidYear].try(:to_i))
+      valid_aid_years.include?(params[:aidYear].try(:to_i))
+    end
+
+    def prior_response_recorded?(aid_year)
+      response = aid_year.try(:[], :termsAndConditions).try(:[], :approved)
+      response == true || response == false
     end
 
     def request_root_xml_node
--- a/spec/models/campus_solutions/terms_and_conditions_spec.rb
+++ b/spec/models/campus_solutions/terms_and_conditions_spec.rb
@@ -9,11 +9,24 @@ describe CampusSolutions::TermsAndConditions do
       {
         feed: {
           finaidSummary: {
-            finaidYears: [{ id: 2016 }, { id:2015 }]
+            finaidYears: [
+              {
+                id: 2016,
+                termsAndConditions: {
+                  approved: terms_and_conditions_approved
+                }
+              }, {
+                id:2015,
+                termsAndConditions: {
+                  approved: terms_and_conditions_approved
+                }
+              }
+            ]
           }
         }
       }
     }
+    let(:terms_and_conditions_approved) { nil }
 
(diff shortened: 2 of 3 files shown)


spec/models/campus_solutions/terms_and_conditions_spec.rb
```diff
@@ -75,6 +88,16 @@
       it_should_behave_like 'a simple proxy that returns errors'
       it_behaves_like 'a proxy that properly observes the finaid feature flag'
       it_behaves_like 'a proxy that got data successfully'
+
+      context 'when response has already been recorded for this aid year' do
+        let(:terms_and_conditions_approved) { false }
+        it 'raises an error and aborts' do
+          expect{
+            subject
+          }.to raise_error Errors::BadRequestError, /Invalid request: {:response=>"Y", :aidYear=>"2016"}/
```
davidlimcheng: @lyttam Is there a reason why `subject` is wrapped in `{}` brackets instead of `()`?

lyttam: @davidlimcheng The curly brackets allow the statement to be split across multiple lines, though in this case `subject` is just one line. It's not necessary, just more readable. I copied this from line 73 and preserved the style.

app/models/campus_solutions/terms_and_conditions.rb
```diff
@@ -19,12 +19,19 @@ def self.field_mappings
...
+          valid_aid_years.push(aid_year.try(:[], :id).try(:to_i))
+        end
       end
-      aid_years.include?(params[:aidYear].try(:to_i))
+      valid_aid_years.include?(params[:aidYear].try(:to_i))
+    end
+
+    def prior_response_recorded?(aid_year)
+      response = aid_year.try(:[], :termsAndConditions).try(:[], :approved)
+      response == true || response == false
```
redconfetti: @davidlimcheng  - This could be re-written as `[true, false].include? response`, or even using [Object.in?](https://devdocs.io/rails~4.2/object#method-i-in-3F) like `response.in? [true, false]`

app/models/campus_solutions/terms_and_conditions.rb
```diff
@@ -19,12 +19,19 @@ def self.field_mappings
...
-        aid_years.push(aid_year.try(:[], :id).try(:to_i))
+      valid_aid_years = []
+      aid_years = CampusSolutions::MyAidYears.new(@uid).get_feed
+      aid_years.try(:[], :feed).try(:[], :finaidSummary).try(:[], :finaidYears).try(:each) do |aid_year|
+        unless prior_response_recorded? aid_year
+          valid_aid_years.push(aid_year.try(:[], :id).try(:to_i))
+        end
       end
-      aid_years.include?(params[:aidYear].try(:to_i))
+      valid_aid_years.include?(params[:aidYear].try(:to_i))
```
redconfetti: @davidlimcheng - Every time this runs it compares a single `:aidYear` parameter to a newly compiled list of valid aid years? Shouldn't we perhaps compile that list of valid aid years in an instance variable like:
```ruby
def valid_aid_years
  @valid_aid_years ||= lambda do
    valid_aid_years = []
    aid_years = CampusSolutions::MyAidYears.new(@uid).get_feed
    aid_years.try(:[], :feed).try(:[], :finaidSummary).try(:[], :finaidYears).try(:each) do |aid_year|
      unless prior_response_recorded? aid_year
        valid_aid_years.push(aid_year.try(:[], :id).try(:to_i))
      end
    end
    return valid_aid_years
  end
end
```

redconfetti: Where does `CampusSolutions::TermsAndConditions#valid?` get called from BTW? I'm not sure what context.

app/models/campus_solutions/terms_and_conditions.rb
lyttam: @redconfetti do you have a strong preference for one of those options?

app/models/campus_solutions/terms_and_conditions.rb
lyttam: @redconfetti `valid?` is called from `CampusSolutions::PostingProxy#post`, so it only gets called once when the user submits a response to either accept or reject the Terms and Conditions.  In this case, we really do need to compile a fresh list of aid years every time, because the data will change after the first POST is made.  The purpose of this function is to prevent subsequent POST requests from being made (i.e. the user trying to change their previous answer).

app/models/campus_solutions/terms_and_conditions.rb
redconfetti: Not strong.

PR #7216, opened 2018-06-20
SISRP-38459 - BearFacts legacy grade rosters - Faculty access in CalCentral

redconfetti:
https://jira.berkeley.edu/browse/SISRP-38459

### DIFF
--- a/app/models/my_academics/grading.rb
+++ b/app/models/my_academics/grading.rb
@@ -20,8 +20,14 @@ module MyAcademics
       teaching_semesters.try(:each) do |semester|
         term_id = Berkeley::TermCodes.slug_to_edo_id(semester[:slug])
         add_grading_information_links(semester)
-        add_grading_dates(semester, term_id)
-        add_grading_to_classes(semester[:classes], term_id)
+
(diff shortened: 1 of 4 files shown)


app/models/my_academics/grading.rb
```diff
@@ -120,6 +126,43 @@ def add_grading_to_classes(semester_classes, term_id)
       end
     end
 
+    def add_legacy_term_grading_to_classes(semester_classes, term_id)
+      semester_classes.try(:each) do |semester_class|
+        semester_class.try(:[],:sections).try(:each) do |section|
+          if is_primary_section? section
+            has_grade_access = has_grading_access?(section)
+            grading_link = fetch_link('UC_CX_TERM_GRD_LEGACY', { TERM_ID: term_id })
```
lyttam: @redconfetti Can this `fetch_link` call be moved to line 130 so that it doesn't get called for every section in every class?

redconfetti: @lyttam - oooh! good catch

lyttam:
Looks good!

PR #7130, opened 2018-05-17
SISRP-39693 - Academic Summary Transfer Credit section - split by career, add Law-specific data

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-39693

PR #7217, opened 2018-06-21
SISRP-39445: Converts Transfer Credit card to SISEDO data, groups by career

lyttam:
https://jira.berkeley.edu/browse/SISRP-39445
https://jira.berkeley.edu/browse/SISRP-33399

This is tackling 2 birds at the same time by making the Transfer Credit card use SISEDO views as its data source.


app/controllers/transfer_credit_controller.rb
```diff
@@ -1,7 +1,6 @@
 class TransferCreditController < ApplicationController
   include AllowDelegateViewAs
   before_filter :api_authenticate
-  before_filter :authorize_for_enrollments
```
redconfetti: @lyttam - Why are we removing this security check? Also, if we're removing it... shouldn't we also remove the include for `AllowDelegateViewAs`?

src/assets/javascripts/angular/controllers/widgets/academics/transferCreditController.js
```diff
@@ -0,0 +1,37 @@
...
+
+var angular = require('angular');
+var _ = require('lodash');
+
+angular.module('calcentral.controllers').controller('TransferCreditController', function(transferCreditFactory, $scope) {
+
+  $scope.transferCredits = {
+    isLoading: true,
+    law: {
+      show: true
```
davidlimcheng: @lyttam Should this be defaulted to `false`?

config/routes.rb
```diff
@@ -131,6 +131,7 @@
     get '/api/advising/resources/:student_uid' => 'advising_student#resources', :defaults => { :format => 'json' }
     get '/api/advising/student/:student_uid' => 'advising_student#profile', :defaults => { :format => 'json' }
     get '/api/advising/student_success/:student_uid' => 'advising_student#student_success', :defaults => { :format => 'json' }
+    get '/api/advising/transfer_credit/:student_uid' => 'advising_student#transfer_credit', :defaults => { :format => 'json' }
```
redconfetti: @lyttam - I see that you've added a new Advising API end-point for the Transfer Credit card, but I also see you've removed it from the My Academics Merged feed. Are we removing this card for students from the My Academics page?

app/controllers/transfer_credit_controller.rb
lyttam: @redconfetti good catch; I didn't meant to remove it.

config/routes.rb
redconfetti: Oh, I think I see that we moved Transfer Credit data to it's own API in #7130. You're performing cleanup by removing it from the My Academics Merged feed eh?

lyttam: @redconfetti No, the card on the My Academics page will get that data by calling `/api/academics/transfer_credits`

src/assets/javascripts/angular/controllers/widgets/academics/transferCreditController.js
lyttam: @davidlimcheng not here - `show = true` means the section will be expanded by default. I can see how that's confusing. Perhaps I will change it to `expand`.

lyttam: Never mind, I can't change the name of the variable.

davidlimcheng:
LGTM @lyttam  will merge after Travis

lyttam:
@davidlimcheng ready for merging.

PR #7219, opened 2018-06-26
SISRP-41072 - Task card need to show recent due date task on top

redconfetti:
https://jira.berkeley.edu/browse/SISRP-41072


src/assets/javascripts/angular/services/tasksService.js
```diff
@@ -145,13 +145,15 @@ angular.module('calcentral.services').service('tasksService', function(apiServic
           break;
         }
       }
-      var incompleteSectionTasks = _.clone($scope.lists.incomplete.filter(taskFilter).sort(sortByDueDate));
+      var incompleteSectionTasks = _.clone($scope.lists.incomplete.filter(taskFilter)).sort(sortByDueDate);
+      var furtherActionNeededTasks = _.remove(incompleteSectionTasks, isCsFurtherActionNeededTask);
+      var incompleteSortedSectionTasks = _.concat(incompleteSectionTasks, furtherActionNeededTasks);
```
lyttam: @redconfetti I believe it's a JavaScript best practice to declare variables at the top of the scope to avoid [variable hoisting](https://developer.mozilla.org/en-US/docs/Glossary/Hoisting). It doesn't make a big difference here, more of a style preference.

redconfetti: @lyttam - I've declared them at the top, so now they're hoisted anyway.

PR #7222, opened 2018-06-29
[DO NOT MERGE] SISRP-39902: Removes profile edit functionality

lyttam:
https://jira.berkeley.edu/browse/SISRP-39903

We are removing the ability to edit certain profile information in CalCentral, and pointing the user to CS to make changes. CalCentral will still display the information.

This work isn't quite finished - I'm waiting for new links to be configured in CS so I can add the IDs in CalCentral.  Getting this out now because it's a large PR.

This PR is broken into 4 commits.  Feel free to review and not merge; I will be making a 5th commit that adds the link IDs.


app/models/hub_edos/my_student.rb
```diff
@@ -33,6 +34,13 @@ def get_feed_internal
         end
       end
 
+      # TODO: replace link ID with the correct one
```
redconfetti: @lyttam - You have a TODO here to replace these links with the correct link ID?

src/assets/javascripts/angular/controllers/widgets/profilePopoverController.js
```diff
@@ -0,0 +1,22 @@
...
+var angular = require('angular');
+var _ = require('lodash');
+
+angular.module('calcentral.controllers').controller('ProfilePopoverController', function(csLinkFactory, $scope) {
+  $scope.profilePopover = {
+    isLoading: true
+  };
+
+  var loadLink = function() {
+    // TODO: change this to the correct link ID
```
redconfetti: @lyttam - It appears you have the same note here to change it to the correct Link ID also

app/models/hub_edos/my_student.rb
redconfetti: @lyttam - Nevermind, I just noticed your comment about this just being phase 1 while you wait for the links to be setup.

src/assets/javascripts/angular/controllers/widgets/profilePopoverController.js
redconfetti: @lyttam - Nevermind. I noticed that the CS Links are still pending.

PR #7224, opened 2018-07-03
SISRP-37919 - New Admit Card Expiration & Performance Updates (needed in PRD before 8/22/18) - UX Dev

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-3791

- removing hardcoded date dependency, adding edodb date dependency
- major refactor of SIR process to reduce calls to edodb
- refactor to front-end due to new date dependency


app/models/campus_solutions/new_admit_resources.rb
```diff
@@ -139,22 +139,16 @@ def get_section_links(link_collection)
...
       @new_admit_attributes ||= {}.tap do |new_admit_attributes|
         new_admit_statuses = CampusSolutions::Sir::SirStatuses.new(@uid).get_feed
-        undergraduate_status = new_admit_statuses.try(:[], :sirStatuses).try(:find) do |status|
-          status.try(:[], :isUndergraduate)
+        if (undergraduate_status = new_admit_statuses.try(:[], :sirStatuses).try(:find) {|status| status.try(:[], :isUndergraduate)})
+          new_admit_attributes.merge!({
+            admitTerm: undergraduate_status.try(:[], :newAdmitAttributes).try(:[], :admitTerm),
+            applicationNbr: undergraduate_status.try(:[], :checkListMgmtAdmp).try(:[], :admApplNbr).try(:to_s),
+            roles: undergraduate_status.try(:[], :newAdmitAttributes).try(:[], :roles),
+          })
```
lyttam: @davidlimcheng I actually liked this better the old way, where all the keys on lines 143-146 were present in the Hash (and `nil` if `undergraduate_status` doesn't exist).

app/models/edo_oracle/queries.rb
```diff
@@ -598,24 +598,25 @@ def self.get_pnp_unit_count (student_id)
...
+          EXPIRE_DT_AD as expiration_date,
+          PROG_ACTION as admit_action,
           PROG_STATUS as admit_status
         FROM
-          SISEDO.APPLICANT_ADMIT_DATA_UGV00_VW
+          SISEDO.APPLICANT_ADMIT_DATAV00_VW
         WHERE
           STUDENT_ID = '#{student_id}' AND
-          APPLICATION_NBR = '#{application_nbr}'
+          PROG_ACTION != 'DATA'
```
lyttam: @davidlimcheng Is `!=` allowed here?  I don't think that's valid SQL. The `<>` operator is SQL for "not equals".

app/models/campus_solutions/new_admit_resources.rb
davidlimcheng: @lyttam May I ask why?  I agree that it would make more sense if this object was being merged with an object that will be served to the front-end, but this `attributes` method returns an object that should only be used within this this particular class, and not sent to the front-end.

I nested this in an `if` so that Ruby wouldn't use any resources to look for keys in `undergraduate_status` if `undergraduate_status == nil`

app/models/edo_oracle/queries.rb
davidlimcheng: @lyttam Thanks for the clarification!  PR amended

app/models/campus_solutions/new_admit_resources.rb
lyttam: @davidlimcheng my preference is for a more object-oriented approach, where objects are treated like instances of a class (i.e. they always have the same properties/same shape). It's mostly a style preference. There are cases where that approach can reduce the chance of bugs or code misuse. 

You could code it this way, or just let Ruby do the `try` (I don't believe it's a resource-intensive operation), but I'll leave it up to you to decide. I don't have strong feelings about this particular case. 
```
          new_admit_attributes.merge!({
            admitTerm: undergraduate_status ? undergraduate_status.try(:[], :newAdmitAttributes).try(:[], :admitTerm) : nil,
            applicationNbr: undergraduate_status ? undergraduate_status.try(:[], :checkListMgmtAdmp).try(:[], :admApplNbr).try(:to_s) : nil,
            roles: undergraduate_status ? undergraduate_status.try(:[], :newAdmitAttributes).try(:[], :roles) : nil,
          })
```

davidlimcheng: Makes sense @lyttam, amended to just let Ruby run the `:try`'s - should be good to go now.

PR #7227, opened 2018-07-05
SISRP-37469: Show class topic to instructors

lyttam:
https://jira.berkeley.edu/browse/SISRP-37469

Certain types of classes all share the same title, but the topic indicates what the class is actually about. Students can already see the topic. Now we'll show it to instructors too.

I deviated somewhat from the design after coming to an agreement with the BSA.

### DIFF
--- a/app/models/my_academics/grading.rb
+++ b/app/models/my_academics/grading.rb
@@ -36,18 +36,18 @@ module MyAcademics
         # A termCode of 'C' denotes a summer term. Every non-law term has midpoint grading, except summer.
         if is_summer_semester? semester
           semester.merge!({
-            gradingAssistanceLink: grading_info_links[:general].url
+            gradingAssistanceLink: grading_info_links[:general].try(:url)
           })
         else
           semester.merge!({
-            gradingAssistanceLinkMidpoint: grading_info_links[:midterm].url,
-            gradingAssistanceLink: grading_info_links[:general].url
+            gradingAssistanceLinkMidpoint: grading_info_links[:midterm].try(:url),
+            gradingAssistanceLink: grading_info_links[:general].try(:url)
           })
         end
       end
       if has_law?(semester[:classes])
         semester.merge!({
-          gradingAssistanceLinkLaw: grading_info_links[:law].url
+          gradingAssistanceLinkLaw: grading_info_links[:law].try(:url)
         })
       end
     end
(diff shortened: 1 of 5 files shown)


app/models/my_academics/grading.rb
```diff
@@ -36,18 +36,18 @@ def add_grading_information_links(semester)
         # A termCode of 'C' denotes a summer term. Every non-law term has midpoint grading, except summer.
         if is_summer_semester? semester
           semester.merge!({
-            gradingAssistanceLink: grading_info_links[:general].url
+            gradingAssistanceLink: grading_info_links[:general].try(:url)
```
lyttam: This just fixes an error I noticed while testing.

src/assets/javascripts/angular/controllers/pages/academicsController.js
```diff
@@ -261,6 +269,7 @@ angular.module('calcentral.controllers').controller('AcademicsController', funct
       }
       $q.all(requests).then(filterWidgets);
     }
+    console.dir($scope);
```
davidlimcheng: @lyttam Got a `console` in the wild

src/assets/templates/academics_semester_classes.html
```diff
@@ -171,7 +171,8 @@ <h4 class="cc-academics-course-standard-grading-header cc-academics-course-gradi
           <thead>
             <th width="15%" scope="col" class="cc-academics-course-number">Course</th>
             <th width="10%" scope="col" class="show-for-medium-up">Class Number</th>
-            <th data-ng-attr-width="{{(containsMidpointclass || isSummerSemester) ? '35%' : '45%'}}" scope="col">Title</th>
+            <th data-ng-attr-width="{{(45 - (10*(containsMidpointClass ? 1 : 0)) - (10*(isSummerSemester ? 1 : 0)) - (15*(teachingCoursesHaveTopics ? 1 : 0))) + '%'}}" scope="col">Title</th>
```
davidlimcheng: @lyttam Can you talk me through a bit about what's going on here?

src/assets/javascripts/angular/controllers/pages/academicsController.js
lyttam: Thanks, will amend.

src/assets/templates/academics_semester_classes.html
davidlimcheng: Ahhh I see.  Nice!  This pattern seems very useful, I'll have to remember it.

lyttam: @davidlimcheng welll it's the best I could come up with, but it's confusing to read. Hopefully we don't have to specify column widths on tables too often. The only reason we have to do it here is because this table has too many columns (not good for mobile).

lyttam:
This should be ready to merge now.

davidlimcheng:
Looks good @lyttam

PR #7231, opened 2018-07-12
SISRP-41163: Converts Housing card to read from new SISEDO view

lyttam:
https://jira.berkeley.edu/browse/SISRP-41163

The Housing card was previously parsing and extracting data from the Financial Aid Data feed. Now it will be populated from a SISEDO view instead.


lyttam:
@davidlimcheng I removed the testext parameter from those two specs.

app/controllers/my_housing_controller.rb
```diff
@@ -0,0 +1,10 @@
+class MyHousingController < ApplicationController
+  include AllowDelegateViewAs
+  before_filter :api_authenticate_401
+  before_filter :authorize_for_financial
+
+  def get_feed
+    options = params.permit :aid_year
```
redconfetti: @lyttam - Shouldn't we put some sort of filter in place here that forces this value to be numeric. I see this value ends up inserted into the query directly.

redconfetti:
It's nice to see that we undid that hack I had to apply to get housing data.

app/controllers/my_housing_controller.rb
lyttam: @redconfetti Good idea, I'll add that.

lyttam:
@redconfetti I incorporated your suggestion - see `Finaid::MyHousing#my_aid_year`

redconfetti:
@lyttam - I would have preferred it happen closer to the controller layer, where the content provided by the client/user is received by the system, but this is still good.

I see there is a conflict now with your commits though, so I can't merge this.

PR #7232, opened 2018-07-13
SISRP-39575 - CPP VIEW: CalCentral Development (replace academic status API for Enrollment card)

nkutub:
https://jira.berkeley.edu/browse/SISRP-39575

### DIFF
--- a/app/models/my_academics/class_enrollments.rb
+++ b/app/models/my_academics/class_enrollments.rb
@@ -5,6 +5,7 @@ module MyAcademics
     include Cache::UserCacheExpiry
     include CampusSolutions::EnrollmentCardFeatureFlagged
     include Concerns::AcademicStatus
+    include Concerns::AcademicRoles
     include LinkFetcher
 
     ENROLLMENT_DECK_KEYS = ['fpf', 'haasFullTimeMba', 'haasEveningWeekendMba', 'haasExecMba', 'summerVisitor', 'courseworkOnly', 'law', 'concurrent']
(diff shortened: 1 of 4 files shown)


app/models/my_academics/class_enrollments.rb
```diff
@@ -28,23 +29,50 @@ def grouped_student_plan_roles
...
+        role = get_enrollment_plan_role(term_cpp)
+        role_key = [role, career_code]
+        grouped_roles[:data][role_key] = { role: role, career_code: career_code, academic_plans: []} if grouped_roles[:data][role_key].blank?
+        grouped_roles[:data][role_key][:academic_plans] << build_plan(role, term_cpp['acad_plan'])
+        grouped_roles[:metadata][:includes_fpf] = true if role == 'fpf'
       end
       grouped_roles
     end
 
+    def get_current_cpp
```
redconfetti: @nkutub - Should we perhaps rename this as `get_current_and_future_cpp`? It's grabbing CPP data for this term and the next one right?

nkutub: yup @redconfetti , makes sense

spec/models/my_academics/class_enrollments_spec.rb
```diff
@@ -53,6 +53,8 @@
     }
   end
   let(:holds) { [] }
+  let(:term_id) { '2168' }
+  let(:term_cpp) { :ugrd_nutritional_science_plan_term_cpp }
```
lyttam: @nkutub Why not create these test cases in the H2 database (via `PopulateSisedosH2`) instead of mocking them here?  Benefits of having them in H2 are easier maintenance and other specs can use them too.

nkutub: Not sure I like using a common H2 database to test integration/functional level testing. This creates a single point of dependency that all test would have to ensure they don't step on other tests data configuration. for example, if I had two tests that required different 2168 results, I would not be able to use the H@ database without using another ID to target only the rows I need. This would not be true to the code and introduce test specific data that could corrupt the testing integrity  @lyttam

nkutub: The H@ database should only be used for unit testing the Data Layer Unit tests.

lyttam:
Discussed with Nader and agreed to leave the Hash fixtures in place.

PR #7234, opened 2018-07-14
SISRP-38885 - Terms API Refactor to SISEDO View for CLC - Development

nkutub:
https://jira.berkeley.edu/browse/SISRP-38885

### DIFF
--- a/app/models/berkeley/terms.rb
+++ b/app/models/berkeley/terms.rb
@@ -128,6 +128,22 @@ module Berkeley
       cs_terms
     end
 
+    def load_terms_from_edo_db
+      cs_terms = []
+      edo_db_terms = EdoOracle::Queries.get_undergrad_terms
+      edo_db_terms.each_with_index do |term, index|
+        new_term = Berkeley::Term.new.from_edo_db(term)
+        cs_terms << new_term
+        # Only keep 2 future terms in array
+        cs_terms.shift if is_future_term?(new_term) && index > 1
+      end
+      cs_terms
+    end
+
+    def is_future_term?(new_term)
+      new_term.start > @current_date
+    end
+
     def fetch_terms_from_api
       # Unlike the legacy database view, the HubTerm API does not support bulk queries. We have to
       # loop through enough API calls to find:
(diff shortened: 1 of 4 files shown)


spec/models/berkeley/terms_spec.rb
```diff
@@ -3,6 +3,164 @@
   before { allow(Settings.hub_term_proxy).to receive(:fake).and_return true }
   subject {Berkeley::Terms.fetch(options)}
 
+  let(:term_data) do
```
lyttam: @nkutub Why not create these test cases in the H2 database (via `PopulateSisedosH2`) instead of mocking them here? Benefits of having them in H2 are easier maintenance and other specs can use them too

app/models/berkeley/terms.rb
```diff
@@ -128,6 +128,22 @@ def load_terms_from_file
       cs_terms
     end
 
+    def load_terms_from_edo_db
+      cs_terms = []
+      edo_db_terms = EdoOracle::Queries.get_undergrad_terms
```
redconfetti: @nkutub - I'm glad we're making it explicit somewhere that our entire system relies on the Undergraduate terms schedule, not Grad or Law.

lyttam:
Discussed with Nader and agreed to leave the Hash fixtures in place.

app/models/edo_oracle/queries.rb
```diff
@@ -753,5 +753,29 @@ def self.get_transfer_credit_detailed (person_id)
...
+          CLASS_BEGIN_DT as class_begin_date,
+          CLASS_END_DT as class_end_date,
+          INSTRUCTION_END_DT as instruction_end_date,
+          GRADES_ENTERED_DT as grades_entered_date,
+          FINAL_EXAM_WK_BEGIN_DT as final_exam_week_start_date,
+          END_DROP_ADD_DT as end_drop_add_date,
+          IS_SUMMER as is_summer
+        FROM  SISEDO.CLC_TERMV00_VW
+        WHERE ACADCAREER_CODE = 'UGRD'
+          AND TERM_TYPE IS NOT NULL
```
lyttam: @nkutub If you add INSTITUTION to the WHERE clause, you may get better performance from this query (and it will protect us in the future if they ever add another institution).

nkutub:
@redconfetti

PR #7246, opened 2018-07-19
SISRP-38606 - Loan History Card - CLC Dev

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-38606

* Back-end was done this way for the ability to mix-and-match user-specific feeds and globally-cached feeds into one merged feed.
* Most changes to `routes.rb` is just alphabetizing


app/models/edo_oracle/queries.rb
```diff
@@ -788,6 +788,7 @@ def self.get_undergrad_terms
       SQL
     end
 
+<<<<<<< HEAD
```
davidlimcheng: I don't know why this happened during rebasing, but all of these merge conflicts are removed in a later commit.

config/routes.rb
```diff
@@ -71,44 +71,48 @@
...
+    get '/api/my/tasks' => 'my_tasks#get_feed', :as => :my_tasks, :defaults => { :format => 'json' }
+    get '/api/my/textbooks_details' => 'my_textbooks#get_feed', :as => :my_textbooks, :defaults => { :format => 'json' }
+    get '/api/my/up_next' => 'my_up_next#get_feed', :as => :my_up_next, :defaults => { :format => 'json' }
+    get '/api/photo/:uid' => 'photo#photo', :as => :photo, :defaults => {:format => 'jpeg' }
     get '/api/service_alerts' => 'service_alerts#get_feed', :as => :service_alerts, :defaults => { :format => 'json' }
     get '/campus/:campus_course_id/photo/:person_id' => 'campus_rosters#photo', :defaults => { :format => 'jpeg' }, :action => 'show'
-    get '/api/media/:term_yr/:term_cd/:dept_name/:catalog_id' => 'mediacasts#get_media', :defaults => { :format => 'json' }
-    get '/api/my/committees' => 'my_committees#get_feed', :defaults => { :format => 'json' }
-    get '/api/my/committees/photo/member/:member_id' => 'my_committees#member_photo', :defaults => { :format => 'jpeg' }
-    get '/api/my/committees/photo/student/:student_id' => 'my_committees#student_photo', :defaults => { :format => 'jpeg' }
```
redconfetti: Alphabetized!

image

app/models/financials/loan_history/loans_aid_years.rb
```diff
@@ -0,0 +1,123 @@
...
+      end
+
+      def get_loan_data
+        data = EdoOracle::Queries.get_loan_history_categories_aid_years campus_solutions_id
+        parse_loan_aid_years HashConverter.camelize(data)
+      end
+
+      def parse_loan_aid_years(loans_data)
+        # Put each aid_year from the returned payload into an array, remove duplicates, and convert it to a hash
+        loans_hash = loans_data.map { |loan| loan.try(:[], :aidYear) }.uniq.map { |aid_year| [aid_year, { aidYear: aid_year, aidYearFormatted: format_aid_year_name(aid_year), loans: [] }] }.to_h
```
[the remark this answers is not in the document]
davidlimcheng: @redconfetti I think it's more of a style preference, since all this line is really doing is initializing a hash, I thought it worked well as a one-liner.

app/models/concerns/loan_history_module.rb
```diff
@@ -0,0 +1,43 @@
+module Concerns
+  module LoanHistoryModule
+    extend self
+
+    def calculate_estimated_monthly_payment(interest_rate, principal_value, repayment_period)
+      return nil unless interest_rate && principal_value && repayment_period && (interest_rate > 0)
+      # Interest rate is provided as a percentage, so we'll need to convert it to a value
+      # Furthermore, we'll divide it by 12 since it is an annual interest rate, and repayment_period is in months
```
lyttam: @davidlimcheng I don't think the comments are necessary here; your code is easy enough to understand.

redconfetti: @davidlimcheng - It almost seems like the second line needs more explanation to me. We're doing an exponential calculation against a negative, then dividing against it... Huh?

app/models/edo_oracle/queries.rb
```diff
@@ -800,5 +800,161 @@ def self.get_housing(person_id, aid_year)
         ORDER BY HSG.TERM_ID
       SQL
     end
+    def self.is_loan_history_active (student_id)
```
lyttam: @davidlimcheng Since this function doesn't actually return a boolean, the `is_` prefix is a little confusing. Perhaps you could name it something like `loan_history_status` instead?

app/models/financials/loan_history/loans_aid_years.rb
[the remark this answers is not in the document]
redconfetti: @davidlimcheng - My suggestion was about readability. If you want to leave it as is, that's fine.

app/models/edo_oracle/queries.rb
```diff
@@ -800,5 +800,161 @@ def self.get_housing(person_id, aid_year)
...
+      message_codes_sql = message_codes.join("\',\'")
+      result = safe_query <<-SQL
+        SELECT
+          MESSAGE_TYPE_CD as code,
+          MESSAGE_TITLE as title,
+          MESSAGE_TEXT as description
+        FROM
+          SISEDO.CLC_FA_LNHST_MESSAGES
+        WHERE
+          INSTITUTION = 'UCB01' AND
```
lyttam: @davidlimcheng There is a `and_institution` function you can call instead of hard-coding it in the WHERE clause.

app/models/financials/loan_history/loans_aid_years.rb
```diff
@@ -0,0 +1,123 @@
...
+      def get_child_loan_details(view_name)
+        EdoOracle::Queries.get_loan_history_aid_years_details(campus_solutions_id, view_name)
+      end
+
+      def campus_solutions_id
+        @cs_id = lookup_campus_solutions_id
+      end
+
+      def show_interest_rate?(loan)
+        loan.try(:[], :interestRate) != 'UNK'
```
lyttam: @davidlimcheng do you happen to know what UNK means? It would be nice to have it as a constant with a name that explains what it is:
```
FRIENDLY_ENGLISH_DESCRIPTION = 'UNK'
```

app/models/financials/loan_history/loans_cumulative.rb
```diff
@@ -0,0 +1,118 @@
...
+              amountOwed: nil,
+              estMonthlyPayment: nil,
+              loanType: loan.try(:[], :loanType),
+              sequence: loan.try(:[],:childSequence).try(:to_i)
+            }
+
+          unless loans_hash[loan_category].try(:[], :hasDescr)
+            loan_descr = enrolled_pre_fall_2016?(campus_solutions_id) ? loan.try(:[], :categoryDescr) + ' ' + loan.try(:[], :categoryDescrPre2168) : loan.try(:[], :categoryDescr)
+            loans_hash[loan_category][:descr] = loan_descr
+            loans_hash[loan_category][:hasDescr] = true
```
lyttam: @davidlimcheng what is the purpose of `hasDescr`?  Couldn't you just check for the existence of `descr`?

app/models/financials/loan_history/loans_aid_years.rb
davidlimcheng: @lyttam Good suggestion, I'll add `UNKNOWN = 'UNK'` as a constant and use that.

app/models/financials/loan_history/loans_cumulative.rb
```diff
@@ -0,0 +1,118 @@
...
+            monthly_payment = parsed_amount > 0 ? parse_owed_value(choose_monthly_payment(parsed_est_monthly_payment, minimum_monthly_payment)) : 0
+
+            add_amounts_to_running_total(loans_hash, loan_category, parsed_amount, monthly_payment)
+            loan_obj.merge!(
+              {
+                amountOwed: parsed_amount,
+                estMonthlyPayment: monthly_payment
+              })
+          else
+            logger.error("#{loan.try(:[], :loanType)} child view missing for UID #{@uid}")
```
lyttam: @davidlimcheng Is this really an ERROR, or is it more of a WARN situation?  IMO an ERROR should be logged when an operation fails. If it's a case of missing or misconfigured data that needs to be flagged, WARN is more appropriate.

app/models/financials/loan_history/loans_cumulative.rb
davidlimcheng: @lyttam My thinking here is that these fields come from a config page in CS, where optional and required fields mix and match, and `hasDescr` was a way to keep track of loan categories that have already been parsed.

app/models/edo_oracle/queries.rb
redconfetti: @davidlimcheng - Maybe we should update it to support an optional alias.

```
    def self.and_institution(object_alias)
      alias_prefix = object_alias.present? ? object_alias + '.' : nil
      <<-SQL
        AND #{alias_prefix}INSTITUTION = 'UCB01'
      SQL
    end
```

app/models/financials/loan_history/loans_cumulative.rb
```diff
@@ -0,0 +1,118 @@
...
+        end
+        loans_summary = { amountOwed: parse_owed_value(cumulative_amount), estMonthlyPayment: parse_owed_value(cumulative_monthly_payment) }
+        remove_temporary_properties(loans_hash, loans_summary)
+      end
+
+      def remove_temporary_properties(loans_hash, loans_summary)
+        loans_hash.each_value do |loan_category|
+          loan_category.delete(:hasDescr)
+        end
+        convert_loans_to_array(loans_hash, loans_summary)
```
lyttam: @davidlimcheng this line seems out of place here, given the name of this function.

app/models/financials/loan_history/loans_cumulative.rb
davidlimcheng: @lyttam I was struggling with this. The way the loans are architected, if one of the child views are missing, then there is definitely an error _in the EDODB layer_. Not a CalCentral error - so I think this is where I ask for a point of clarification - should we only log something as an `error` if it's a CalCentral error? 

We're so closely coupled with CS it's hard to separate concerns between the two.

app/models/financials/loan_history/loans_cumulative.rb
davidlimcheng: @lyttam This line is simply calling the next method in the pipeline - I can run `convert_loans_to_array` in `get_feed_internal` if that makes more sense - but in my opinion I think it's okay to pass values from method to method in a pipeline.

app/models/financials/loan_history/loans_summary.rb
```diff
@@ -0,0 +1,23 @@
...
+    class LoansSummary < UserSpecificModel
+      include Cache::CachedFeed
+      include Cache::UserCacheExpiry
+      include Cache::RelatedCacheKeyTracker
+      include Concerns::LoanHistoryModule
+      include User::Identifiers
+
+      def get_feed_internal
+        loan_history_active = { active: is_loan_history_active?(lookup_campus_solutions_id), amountOwed: nil }
+        loan_history_active[:active] ? loan_history_active.merge(get_cumulative_amount_owed) : loan_history_active
```
lyttam: @davidlimcheng this seems a little weird. Why not just return this?
```
{
  active: false,
  amountOwed: nil
}
```
or
```
{
  active: true,
  amountOwed: <some dollar amount>
}
```

app/models/financials/loan_history/loans_aid_years.rb
redconfetti: U.N. Owen

app/models/financials/loan_history/loans_summary.rb
davidlimcheng: @lyttam That's essentially what this is doing - the `get_cumulative_amount_owed` returns a hash that looks like 
```
{
    amountOwed: <some dollar amount>
}
```
So the `merge` just overwrites the `amountOwed: nil` with the actual value.

spec/models/concerns/loan_history_module_spec.rb
```diff
@@ -0,0 +1,154 @@
+describe Concerns::LoanHistoryModule do
+
+  context '#calculate_estimated_monthly_payment' do
```
lyttam: @davidlimcheng I like to use `describe` instead of `context` when grouping specs that test a particular function.

```
describe '#calculate_estimated_monthly_payment' do
  ...
  context 'when only given some of the parameters' do ... end
  context 'when given all required parameters' do ... end
  ...
end
```

app/models/edo_oracle/queries.rb
davidlimcheng: @lyttam @redconfetti I started to do this, and realized that `and_institution` also returns SQL that has an `AND`, and isn't suitable if `INSTITUTION = 'UCB01'` is the only clause in `WHERE`. 

Making the returned `AND` dynamic is certainly possible, but far enough from the scope of this PR that I'm going to propose that we leave that for a future PR.

app/models/financials/loan_history/loans_cumulative.rb
redconfetti: @davidlimcheng - Why would an error with a EDODB view be cause in this class, instead of at the EdoOracle::Queries level? Shouldn't that raise an exception, which bubbles up the call stack to the controller resulting in an HTTP 500 (internal server) error?

davidlimcheng: @redconfetti An error wouldn't be thrown, but _functionally_ it's an error, because the child view *should* always be there.

app/models/financials/loan_history/loans_summary.rb
davidlimcheng: @lyttam As for returning the following object:
```
{
  active: loan_history_active,
  amountOwed: loan_history_active ? get_cumulative_amount_owed : nil
}
```
This will run `#loan_history_active` twice, and the response of that is not cached, so we will have to assign it to memory to refer to it:
```
active = loan_history_active
{
  active: active,
  amountOwed: active ? get_cumulative_amount_owed : nil
}
```
Essentially the same pattern as the one in `get_feed_internal`, but with 3 lines instead of 2.

src/assets/templates/widgets/loan_history/aid_year_overview.html
```diff
@@ -0,0 +1,81 @@
...
+        <thead>
+          <tr>
+            <th>Loan Type</th>
+            <th>Loan Name</th>
+            <th class="cc-table-right"><span>Borrowed<a data-ng-click="scrollToDefinition('BRWD', $event)"><i class="cc-icon fa fa-info-circle"></i></a></span></th>
+            <th class="cc-table-right"><span>Interest Rate<a data-ng-click="scrollToDefinition('INTR', $event)"><i class="cc-icon fa fa-info-circle"></i></a></span></th>
+            <th class="cc-table-right"><span>Est. Monthly Payment<a data-ng-click="scrollToDefinition('DSCL', $event)"><i class="cc-icon fa fa-info-circle"></i></a></span></th>
+          </tr>
+        </thead>
+        <tbody data-ng-repeat="loan in aidYear.loans | orderBy:'sequence'">
```
lyttam: @davidlimcheng I think you want the `ng-repeat` on the `<tr>` element, not the `<tbody>`.  Unless you're dividing a table into sub-tables for some reason, it only needs one tbody.

src/assets/templates/widgets/loan_history/cumulative_overview.html
```diff
@@ -0,0 +1,67 @@
...
+            <th class="cc-table-right">
+              <span data-ng-if="showAmountOwedVerbiage(loanCategory.category)">Amount Owed<a class="hide-for-small" data-ng-click="scrollToDefinition('AMNT', $event);"><i class="cc-icon fa fa-info-circle"></i></a></span>
+              <span data-ng-if="!showAmountOwedVerbiage(loanCategory.category)">Borrowed<a class="hide-for-small" data-ng-click="scrollToDefinition('BRWD', $event);"><i class="cc-icon fa fa-info-circle"></i></a></span>
+            </th>
+            <th class="cc-table-right">
+              <span>Est. Monthly Payment<a class="hide-for-small" data-ng-click="scrollToDefinition('DSCL', $event)"><i class="cc-icon fa fa-info-circle"></i></a></span>
+            </th>
+          </tr>
+        </thead>
+        <tbody data-ng-if="showLoanList(loanCategory.category)" data-ng-repeat="loan in loanCategory.loans | orderBy:'sequence'">
```
lyttam: @davidlimcheng same as above; I think the `ng-repeat` should be on the `<tr>`. Otherwise it looks like you could wind up with multiple `<tbody>`s, some of them potentially empty.

app/models/financials/loan_history/loans_cumulative.rb
davidlimcheng: Enough to convince me!

davidlimcheng:
@lyttam @redconfetti amendments pushed

app/models/financials/loan_history/loans_cumulative.rb
lyttam: @davidlimcheng I would place it at the end of `calculate_cumulative_totals`. Just to avoid future confusion ("why is this "remove" function doing a "convert"?)

lyttam:
@davidlimcheng Just the one outstanding thing in `remove_temporary_properties` that I think should change... otherwise, looks great.

app/models/financials/loan_history/loans_cumulative.rb
davidlimcheng: Ahhh got it - makes sense.  Will do and amend @lyttam .

davidlimcheng:
@lyttam, amended.  Thanks for the thorough review!

lyttam:
Nice!

PR #7252, opened 2018-07-23
SISRP-41072 - Task card need to show recent due date task on top.

redconfetti:
https://jira.berkeley.edu/browse/SISRP-41072

### DIFF
--- a/src/assets/templates/widgets/tasks.html
+++ b/src/assets/templates/widgets/tasks.html
@@ -106,7 +106,7 @@
               </div>
 
               <div data-ng-show="taskSection.show" class="cc-widget-tasks-list-wrapper">
-                <div class="cc-widget-tasks-list-header">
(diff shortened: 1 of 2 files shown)


src/assets/templates/widgets/tasks.html
```diff
@@ -106,7 +106,7 @@ <h2 class="cc-left">Tasks</h2>
               </div>
 
               <div data-ng-show="taskSection.show" class="cc-widget-tasks-list-wrapper">
-                <div class="cc-widget-tasks-list-header">
+                <div class="cc-widget-tasks-list-header" data-ng-if="taskSection.tasks.incomplete.length > 0">
```
lyttam: @redconfetti It would be safer to use `data-ng-if="taskSection.tasks.incomplete.length"` here, since that would catch both the case where `taskSection.tasks.incomplete` is an empty array, and the case where `taskSection.tasks.incomplete` is not an array/does not have a property called `length`.

PR #7253, opened 2018-07-23
SISRP-41163: Hides Update Housing link if housing period has passed

lyttam:
https://jira.berkeley.edu/browse/SISRP-41163

If student has housing options for Fall and Spring, but it's past the end of the Spring housing period, it's too late for them to update their housing.

Also adds the housing card model to the expiry processor so that changes to housing data in CS will be reflected immediately on the Housing card.


app/models/finaid/my_housing.rb
```diff
@@ -15,10 +16,11 @@ class MyHousing < UserSpecificModel
 
     def get_feed_internal
       return {} unless is_feature_enabled
-      feed = EdoOracle::Queries.get_housing(@uid, my_aid_year)
+      @feed = EdoOracle::Queries.get_housing(@uid, my_aid_year)
+      puts @feed
```
redconfetti: @lyttam - You've got a rogue `puts` statement

app/models/finaid/my_housing.rb
```diff
@@ -82,7 +84,14 @@ def continuing_undergrad?
       is_undergrad && is_continuing
     end
 
-    def my_aid_year
+    def can_update_housing?
+      max_term = @feed.try(:last)
+      if (housing_period_end_date = max_term.try(:[], 'housing_end_date'))
+        (Settings.terms.fake_now || DateTime.now) < cast_utc_to_pacific(housing_period_end_date)
```
redconfetti: @lyttam - Are you missing an `end` for the `if` statement here?

lyttam: @redconfetti Looks like tab formatting got messed up. I'll fix and amend.

PR #7256, opened 2018-07-24
SISRP-41444 - If Transfer Credit data is not posted yet then the student should not see Transfer Credit Card on My Academics page

nkutub:
https://jira.berkeley.edu/browse/SISRP-41444


redconfetti:
@nkutub - LGTM

PR #7258, opened 2018-07-24
SISRP-41286 - Enrollment Card -> Class Adjustment Section -> When Column is not displaying correctly

nkutub:
https://jira.berkeley.edu/browse/SISRP-41286


lyttam:
Same issue with Travis builds timing out, but I'll merge anyway.

PR #7279, opened 2018-08-04
SISRP-41772 - Remove/replace 'testext' tests v2

redconfetti:
https://jira.berkeley.edu/browse/SISRP-41772


redconfetti:
@raydavis @johncrossman - I'm not sure if I'm removing TestExt tests you think should be retained for manual use (from local command line). Let me know if I went overboard here.

raydavis:
Thanks, @redconfetti . It's true that a lot of OEC and Google support is currently only exercised in "testext" mode, but personally I'm OK with ETS bringing optional versions of such tests back on an as-needed basis.

app/models/my_academics/my_academic_roles.rb
```diff
@@ -59,5 +57,9 @@ def map_roles(default_roles, assigned_roles)
         end
       end
     end
+
+    def term_cpp
+      MyAcademics::MyTermCpp.new(@uid).get_feed
```
lyttam: @redconfetti do you want to assign this to a class instance variable?
```@term_cpp = MyAcademics::MyTermCpp.new(@uid).get_feed```
That way we aren't creating a new instance of `MyAcademics::MyTermCpp` every time this method is called.

lyttam:
@redconfetti Very nice. Just one small suggestion.

app/models/my_academics/my_academic_roles.rb
redconfetti: @lyttam - Good call

PR #7281, opened 2018-08-09
SISRP-41926 - Instructor 3030794638 Class Details Not Displaying in CalCentral

redconfetti:
https://jira.berkeley.edu/browse/SISRP-41926


app/models/edo_oracle/queries.rb
```diff
@@ -28,7 +28,7 @@ class Queries < Connection
       sec."primary" AS "primary",
       sec."sectionNumber" AS section_num,
       sec."component-code" as instruction_format,
-      sec."primaryAssociatedSectionId" as primary_associated_section_id,
+      TO_CHAR(sec."primaryAssociatedSectionId") as primary_associated_section_id,
```
redconfetti: This is the primary fix, which will require regression testing.

src/assets/javascripts/angular/controllers/pages/academicsController.js
```diff
@@ -7,7 +7,7 @@ var angular = require('angular');
 /**
  * Academics controller
  */
-angular.module('calcentral.controllers').controller('AcademicsController', function(academicsFactory, academicsService, academicStandingsFactory, holdsFactory, apiService, badgesFactory, linkService, registrationsFactory, userService, $q, $routeParams, $scope, $location) {
+angular.module('calcentral.controllers').controller('AcademicsController', function(academicsFactory, academicsService, academicStandingsFactory, holdsFactory, apiService, linkService, registrationsFactory, $q, $routeParams, $scope, $location) {
```
redconfetti: These services were not used in the controller, and thus should be removed.

src/assets/javascripts/angular/services/academicsService.js
```diff
@@ -77,7 +77,7 @@ angular.module('calcentral.services').service('academicsService', function() {
    */
   var courseCollectionHasTopics = function(courses) {
     return !!_.find(courses, function(course) {
-      return course.topics.length > 0;
+      return (_.get(course, 'topics.length') > 0);
```
redconfetti: This is fix would cause the issue to fail more gracefully if it happens again (the class page displays), but is not as whole of a fix as ensuring that the primary section IDs match during processing within `academicsService.getClassesSections`.

nkutub:
LGTM

PR #7287, opened 2018-08-24
SISRP-40088 - Move from Gulp to Webpack for Front-End Build

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-40088

This PR will have the following effects:
1. Replaces Gulp with webpack and webpack-serve for development server
2. Replaces JSHint and JSCS with ESLint
3. Introduces Babel to the pipeline so that we can start using ES6 syntax
4. Separates production and development dependencies, so that we can build a production bundle without all the tools we use just for development

**Reading**
CLC Developer's Guide to moving from Gulp to Webpack - **Read before merging!**
https://wikihub.berkeley.edu/display/SIS/Guide+to+Moving+from+Gulp+to+Webpack

Webpack Documentation Concepts (long but IMO the best explanation of webpack)
https://webpack.js.org/concepts/

Great article from one company's move from Angular 1.x to React
(not saying we're going to take the same route, but it's good to see the how the team planned their migration)
https://www.smartly.io/blog/how-to-migrate-from-angular-to-react-without-a-massive-rewrite


lyttam:
@davidlimcheng this is awesome! Nice work.

src/assets/javascripts/index.js
```diff
@@ -1,9 +1,55 @@
+'use strict';
+
+// Angular
+require('angular');
```
redconfetti: @davidlimcheng - I take it this is why we're able to remove `var angular = require('angular');` from so many scripts, because they're all part of the same package namespace? We're making 'angular' a module that can be injected in our directives/controllers as needed?

README-VAGRANT.md
```diff
@@ -1,105 +0,0 @@
-# CalCentral with Vagrant
```
redconfetti: @davidlimcheng - Why is it that you chose to remove the Vagrant/Docker stuff with this? Was it because it conflicted with your Webpack configuration?

redconfetti:
Just wanted to note that I am excited about the prospect of using Redux for global state, because it might help us break up the My Academics feed dependency and isolate, while still being able to share the data.

README-VAGRANT.md
redconfetti: David let me know it's because we'd have to rewrite Gulp dependency work within Vagrant/Docker config. It was better to just remove it.

src/assets/javascripts/index.js
redconfetti: David clarified for me in person that there is a dependency tree, instead of files floating in space that need to be required.

davidlimcheng: @redconfetti Exactly - this gives us a global `angular` declaration that we could use, due to the dependency graph that Webpack creates using this `index.js` as an entrypoint.

davidlimcheng:
I think I accidentally opened this PR from `ets/SISRP-4008` rather than `davidlimcheng/SISRP-40088`, so I'm going to close this PR and reopen.

PR #7288, opened 2018-08-24
SISRP-40774 - Final Exams Schedule (Instructors) - UX Dev

redconfetti:
https://jira.berkeley.edu/browse/SISRP-40774

### DIFF
--- a/app/models/edo_oracle/queries.rb
+++ b/app/models/edo_oracle/queries.rb
@@ -313,30 +313,30 @@ module EdoOracle
     # No Campus Oracle equivalent.
     def self.get_section_final_exams(term_id, section_id)
       safe_query <<-SQL
-        SELECT
-          sec."term-id" AS term_id,
-          sec."session-id" AS session_id,
-          sec."id" AS section_id,
-          sec."finalExam" AS exam_type,
-          exam."EXAM_DT" AS exam_date,
-          exam."EXAM_START_TIME" AS exam_start_time,
-          exam."EXAM_END_TIME" AS exam_end_time,
-          exam."EXAM_EXCEPTION" as exam_exception,
-          exam."FACILITY_DESCR" AS location,
-          exam."FINALIZED" AS finalized
-        FROM
-          SISEDO.CLC_FINAL_EXAM_INFOV00_VW exam
-        LEFT JOIN SISEDO.CLASSSECTIONALLV01_MVW sec ON (
-          exam."CRSE_ID" = sec."cs-course-id" AND
-          exam."STRM" = sec."term-id" AND
-          exam."SESSION_CODE" = sec."session-id" AND
-          exam."CRSE_OFFER_NBR" = sec."offeringNumber" AND
-          exam."CLASS_SECTION" = sec."sectionNumber"
-        )
-        WHERE
-          sec."term-id" = '#{term_id}' AND
-          sec."id" = '#{section_id}'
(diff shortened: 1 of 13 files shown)


app/models/edo_oracle/queries.rb
```diff
@@ -326,7 +326,7 @@ def self.get_section_final_exams(term_id, section_id)
           exam."FINALIZED" AS finalized
         FROM
           SISEDO.CLC_FINAL_EXAM_INFOV00_VW exam
-        LEFT JOIN SISEDO.CLASSSECTIONALLV01_MVW sec ON (
+        RIGHT JOIN SISEDO.CLASSSECTIONALLV01_MVW sec ON (
```
lyttam: @redconfetti why use a `RIGHT JOIN` here? Do you want to select class sections that don't have an exam?

redconfetti: @lyttam - Yes. We need the 'exam_type' field from the section table, even if there isn't an exam. I've adjusted the code to account for blank exam fields.

lyttam: @redconfetti ok. Just asking because that's a fundamental change to how the views are joined... before it was "select exams regardless of whether they have a section" and now it's "select sections regardless of whether they have an exam".

I would swap the views around and do a left join, just because I think that's more common and easier to read:
```
SELECT ...
FROM
  SISEDO.CLASSSECTIONALLV01_MVW sec
LEFT JOIN SISEDO.CLC_FINAL_EXAM_INFOV00_VW exam
...
```

app/models/my_academics/exams.rb
```diff
@@ -133,22 +156,33 @@ def parse_exam(exam)
...
           return nil
         end
       end
       date = exam[:exam_date]
-      date && date.strftime('%a %b %-d')
+      format = '%a %b %-d'
+      if instructor_format
+        format = '%a, %b %-d'
+        format = '%b %-d, %Y' if !time_is_current_year?(date)
+      end
```
lyttam: @redconfetti you could reduce the number of assignment operations being done here. One way to do it:

```
format = '%a %b %-d' unless instructor_format
format = '%a, %b %-d' if instructor_format && time_is_current_year?(date)
format = '%b %-d, %Y' if instructor_format && !time_is_current_year?(date)
```

redconfetti:
@lyttam - I've implemented the adjustments you suggested. I've also introduced a new directive for the display of Final Exam schedules on the Class Page.

src/assets/javascripts/angular/directives/academicsClassInfoFinalExamsDirective.js
```diff
@@ -0,0 +1,43 @@
+/* jshint camelcase: false */
+'use strict';
+
+var angular = require('angular');
+var _ = require('lodash');
+
+angular.module('calcentral.directives').directive('ccAcademicsClassInfoFinalExamsDirective', function() {
```
redconfetti: @lyttam - I didn't want to continue putting logic in `academicsController` just to service this one area of the class info page, so I made a directive. We should probably try to do this anywhere we can within the "My Academics" domain of pages/widgets (main page, semesters page, class page), so that the scope of dependencies is isolated and that we're closer to being ready to migrate to React components.

Would you agree?

lyttam:
@redconfetti Cool, I like the use of a directive. Looks good.

PR #7289, opened 2018-08-27
SISRP-41161 - Refactors FASO card and switches to SISEDO data

lyttam:
https://jira.berkeley.edu/browse/SISRP-41161

This PR covers most of the refactor, but I anticipate there will be more changes/refinements.  I also hope to come back and get the Selenium script working as time permits.


src/assets/javascripts/angular/controllers/widgets/finaid/finaidSummaryController.js
```diff
@@ -7,105 +7,84 @@ var _ = require('lodash');
...
+
+  var loadFinancialAidSummary = function() {
+    return finaidFactory.getFinancialAidSummary()
+      .then(
+        parseFinancialAidSummary,
+        function errorCallback() {
+          $scope.financialAidSummary.errored = true;
+        }
+      )
+    .then(loadTasksIncompleteCount)
```
redconfetti: @lyttam - Is there a reason you decided to make this happen linearly, rather than asynchronously?
The count is based on the tasks API call, and doesn't rely on the Financial Aid Summary data AFAIK.

lyttam: @redconfetti Good point, I will make it synchronous. I don't think the tasks call needs its own `isLoading` spinner, do you?

redconfetti: @lyttam - No. That would be very odd. We had it setup to do both at the same time, and only set `isLoading` to false once both asynchronous jobs were complete, using the `$q.all` mechanism.

redconfetti:
@lyttam Look good other than the one note.

lyttam:
@redconfetti I implemented your suggestion to call the tasks API asynchronously. I also made a number of other changes.

app/models/concerns/query_helper.rb
```diff
@@ -0,0 +1,20 @@
...
+
+    def self.included(target)
+      target.extend ClassMethods
+    end
+
+    module ClassMethods
+
+      def and_institution(object_alias)
+        <<-SQL
+          AND #{object_alias}.INSTITUTION = 'UCB01'
```
davidlimcheng: @lyttam, did you mean to use the previously defined `UC_BERKELEY` here?

davidlimcheng:
Looks great @lyttam.  Just one question for you and some merge conflicts.

app/models/concerns/query_helper.rb
lyttam: @davidlimcheng yes, good call. I will amend.

src/assets/javascripts/angular/controllers/widgets/finaid/finaidSummaryController.js
```diff
@@ -6,96 +6,83 @@ var _ = require('lodash');
...
+    finaidFactory.getFinancialAidSummary()
+    .then(
+      parseFinancialAidSummary,
+      function errorCallback() {
+        $scope.financialAidSummary.errored = true;
       }
-    );
+    )
+    .then(
+      loadTasksIncompleteCount
```
lyttam: fyi @redconfetti , I had to move `loadTasksIncompleteCount` back into the chain of promises because it depends on the value of `finaidService.options.finaidYear.id`, which is set in `parseFinancialAidSummary`.

lyttam:
@davidlimcheng ready to merge.

PR #7295, opened 2018-08-30
SISRP-42338: Adds option to source campus links from file

lyttam:
https://jira.berkeley.edu/browse/SISRP-42338

Will be adding a new feature flag around using Postgres for campus links.


redconfetti:
@lyttam - Looks good.

PR #7297, opened 2018-09-04
SISRP-39065 - AA: UX Dev - Advisor Student Lookup

redconfetti:


src/assets/javascripts/angular/controllers/widgets/userSearchController.js
```diff
@@ -70,6 +74,20 @@ angular.module('calcentral.controllers').controller('UserSearchController', func
     return users;
   };
 
+  var decorateUserName = function(user) {
+    var legalFirst = _.trim(_.get(user, 'firstNameLegal'));
+    var legalMiddle = _.trim(_.get(user, 'middleNameLegal'));
+    var legalLast = _.trim(_.get(user, 'lastNameLegal'));
+    var preferredFirst = _.trim(_.get(user, 'firstNamePreferred'));
+    var preferredMiddle = _.trim(_.get(user, 'middleNamePreferred'));
+    console.log(preferredMiddle);
```
davidlimcheng: wild `console`

app/models/user/search_users_by_name.rb
```diff
@@ -20,18 +12,17 @@ def only_special_characters?(name)
...
-      ldap_users = CalnetLdap::UserAttributes.get_attributes_by_name(name, !!opts[:include_guest_users])
-      ldap_users.each do |ldap_user|
-        # Allow CalNet attributes to be overridden by other data sources.
-        uid = ldap_user[:ldap_uid]
-        if (user = User::SearchUsersByUid.new(opts.merge(id: uid)).search_users_by_uid)
-          users << user
+      sisedo_users = EdoOracle::Queries.search_students(name)
+      sisedo_users.each do |sisedo_user|
+        if uid = sisedo_user['campus_uid']
+          if (ihub_user = User::SearchUsersByUid.new(opts.merge(id: uid)).search_users_by_uid)
```
davidlimcheng: @redconfetti Minor, but it looks like we're assigning `ihub_user` to whatever is being returned by `User::SearchUsersByUid`, but we're not using it.  Instead of assigning the `ihub_user` variable to whatever is returned, can we use `!!` to coerce the returned object to a boolean instead?  Would that make sense in this context?  I'm not exactly sure what `User::SearchUsersByUid#search_users_by_uid` is supposed to return, maybe you can help shine a light on that.

redconfetti: @davidlimcheng - Yes. Great point! We are performing this search now just to make sure that the user does exist within the boundaries of the `opts` hash. We don't need that variable. I'll update that now.

redconfetti:
@davidlimcheng - Fixes applied

davidlimcheng:
LGTM @redconfetti

PR #7298, opened 2018-09-05
CLC-6899, OEC sis-import fix; google-drive gem update and better session mgmt

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-6899

My latest comment in the Jira describes the change.

I ran this on my laptop and it hummed along for 4+ hours until crashing on `Timeout error; url: https://apis.berkeley.edu/uat/sis/v1/terms`.  38 dept sheets were created in course-eval's `2018-D/Step 1: SIS Import` folder. The sis-data in those 38 looks legit. I would like to get this change on `junction-qa` and give it a proper run.  cc: @raydavis

PR #7300, opened 2018-09-05
CLC-6899, OEC sis-import fix; google-drive gem update and better session mgmt

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-6899

QA PR (merged) is #7298

I suggest we merge this _after_ it's been verified in `jqa` environment.


raydavis:
It has been verified.

PR #7304, opened 2018-09-07
SISRP-41780 - CalCentral Cal 1 Card Meal Plan Update

redconfetti:
https://jira.berkeley.edu/browse/SISRP-41780

New design calls for static meal plan links, now that the API no longer provides accurate information.

image


lyttam:
If the meal plan data is inaccurate and no longer being used by the UI, then we should remove it from the feed. Merging this PR but will submit a follow-up PR, since Jason is out.

PR #7307, opened 2018-09-10
SISRP-41856 - Add DSP Links to CalCentral - UX Dev

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-41856


app/models/campus_solutions/student_resources.rb
```diff
@@ -44,6 +49,22 @@ def get
...
+      ucbx_only_links = [:disabled_student_services, :register_to_vote_ca, :register_to_vote_non_ca]
+
+      if is_ucbx_only? roles
+        link_settings.select { |link| ucbx_only_links.include? link[:feed_key] }
+      else
+        link_settings
+      end
+    end
+
+    def is_ucbx_only?(roles)
```
nkutub: @davidlimcheng , should this be a method at a common class related to roles shared module vs this specialized use?

davidlimcheng: @nkutub - we spoke and agreed that it's a potentially useful shared method, but we have a preference for not making shared methods until it's necessary to do so.

PR #7312, opened 2018-09-12
SISRP-42360: Feature flags recording of user visits

lyttam:
https://jira.berkeley.edu/browse/SISRP-42360

### DIFF
--- a/app/models/user/user_visits_feature_flagged.rb
+++ b/app/models/user/user_visits_feature_flagged.rb
@@ -1,0 +1,17 @@
+module User
+  module UserVisitsFeatureFlagged
+    def is_feature_enabled
+      Settings.features.user_visits
+    end
+
+    def self.included(base)
+      base.extend ClassMethods
+    end
+
+    module ClassMethods
+      def is_feature_enabled
+        Settings.features.user_visits
+      end
+    end
+  end
+end
--- a/spec/models/user/visit_spec.rb
+++ b/spec/models/user/visit_spec.rb
@@ -1,15 +1,24 @@
-describe "User::Visit" do
+describe User::Visit do
 
-  before do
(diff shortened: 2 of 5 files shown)


app/models/user/user_visits_feature_flagged.rb
```diff
@@ -0,0 +1,17 @@
+module User
+  module UserVisitsFeatureFlagged
+    def is_feature_enabled
```
davidlimcheng: @lyttam I'm not sure if I'm correct about this, but since only the class method is used in `User::Visit`, can this instance method be removed?

spec/models/user/visit_spec.rb
```diff
@@ -1,15 +1,27 @@
...
+    let(:uid) { rand(9999999).to_s }
 
-    saved = User::Visit.where(:user_id => @user_id)
-    saved.should_not be_nil
+    context 'when feature flag is on' do
+      let(:user_visits_feature_flag) { true }
+      it 'should record a user\'s visit' do
+        expect(subject).to be_truthy
+      end
+      it 'should record a user\'s visit' do
```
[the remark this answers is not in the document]
lyttam: @davidlimcheng no, I forgot to rename the example. Will amend

app/models/user/user_visits_feature_flagged.rb
lyttam: @davidlimcheng I thought about it, but decided to make this feature flagged module conform to all the other ones... which all have an instance method called `is_feature_enabled`.

lyttam:
@davidlimcheng Amended. Ready to merge unless you want to fight about `is_feature_enabled`.

davidlimcheng:
LGTM!

PR #7313, opened 2018-09-12
SISRP-41516 - Delegates - Financial Resource Card Endless Spinning Bug

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-41516

### DIFF
--- a/app/controllers/campus_solutions/financial_resources_controller.rb
+++ b/app/controllers/campus_solutions/financial_resources_controller.rb
@@ -1,6 +1,5 @@
 module CampusSolutions
   class FinancialResourcesController < ApplicationController
-
     before_action :api_authenticate
 
     def get_emergency_loan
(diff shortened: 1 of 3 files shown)


davidlimcheng:
Thanks @lyttam, title changed, but please hold off on merging as something just popped up in the JIRA.

davidlimcheng:
OK fixed the issue @lyttam, I had to actually remove delegate access to `FinancialResources`. Should be good to go upon re-review.

davidlimcheng:
Ahh scratch that.  One more requirement change, sorry.

davidlimcheng:
OK, really good to go now, @lyttam.  Made it so that delegates won't trigger any API calls that don't need to be made on the front-end, and removed the delegate access I previously gave to those links.

lyttam:
Looks great.

PR #7317, opened 2018-09-13
SISRP-42335: Fixes misc. NoMethodErrors

lyttam:
https://jira.berkeley.edu/browse/SISRP-42335

### DIFF
--- a/app/models/concerns/academics_module.rb
+++ b/app/models/concerns/academics_module.rb
@@ -84,6 +84,7 @@ module Concerns
     end
 
     def time_bucket(term_yr, term_cd)
+      return nil unless current_term
       term_yr = term_yr.to_i
       if term_yr < current_term.year || (term_yr == current_term.year && term_cd < current_term.code)
         bucket = 'past'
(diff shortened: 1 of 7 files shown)


app/models/concerns/academics_module.rb
```diff
@@ -84,6 +84,7 @@ def terms
     end
 
     def time_bucket(term_yr, term_cd)
+      return nil unless current_term
```
davidlimcheng: @lyttam, it looked like the intended logic on line 94 was to have `bucket` default to 'current'.  Will returning `nil` instead cause any problems for downstream methods?

lyttam: @davidlimcheng yeah, that's assuming the term exists. If the term doesn't exist (I suspect that's happening when there's a problem fetching the terms), then I'm not sure we want to default to 'current' - that could result in multiple 'current' terms.  I don't believe it will cause problems if `time_bucket` returns `nil`, beyond the term failing to display (which is pretty graceful imo).

davidlimcheng: Makes sense - thanks for the clarification.

davidlimcheng:
LGTM!

PR #7320, opened 2018-09-17
SISRP-42554: Retires Cal 1 Card card

lyttam:
https://jira.berkeley.edu/browse/SISRP-42554

Moved the links from the card into campuslinks.json.

If we ever go back to reading the campus links from Postgres, we'll have to write a migration to populate those three links in the database from the JSON file.


lyttam:
@raydavis Are you relying on the Cal 1 Card status and meal plan data for anything?  We just got word that the API won't be providing accurate information anymore.

nkutub:
LGTM @lyttam

raydavis:
Thank you for the heads-up, @lyttam ! ETS only relies on Cal1Card for official student photos, so it sounds like we'd be OK.

PR #7323, opened 2018-09-19
CLC-6908 Only instructors in current terms should see Create Course Site

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6908


raydavis:
I amended the PR to fix a buggy test (which wasn't noticed before thanks to the buggy code).

johncrossman:
Nice work, @raydavis  I'll merge when Travis gives the go.

PR #7324, opened 2018-09-20
SISRP-39065 - Adds support for case insensitive search, distinct query results. Fixes Saved/Recent tab functionality

redconfetti:

### DIFF
--- a/app/models/edo_oracle/queries.rb
+++ b/app/models/edo_oracle/queries.rb
@@ -784,20 +784,20 @@ module EdoOracle
 
     def self.search_students(search_string)
       result = safe_query <<-SQL
-        SELECT
-        STUDENT_ID as student_id,
-        CAMPUS_ID as campus_uid,
-        FIRST_NAME as first_name_legal,
-        MIDDLE_NAME as middle_name_legal,
-        LAST_NAME as last_name_legal,
-        UC_PRF_FIRST_NM as first_name_preferred,
-        UC_PRF_MIDDLE_NM as middle_name_preferred,
-        EMAIL_ADDR as email,
-        ACAD_PROG as academic_programs
-      FROM SISEDO.CLC_STDNT_LOOKUP_V00_VW
-      WHERE
-        UC_SRCH_CRIT LIKE '%#{search_string}%' AND
-        rownum < 31
+        SELECT DISTINCT
+          STUDENT_ID as student_id,
+          CAMPUS_ID as campus_uid,
+          FIRST_NAME as first_name_legal,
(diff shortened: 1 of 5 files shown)


davidlimcheng:
LGTM @redconfetti

app/models/edo_oracle/queries.rb
```diff
@@ -784,20 +784,20 @@ def self.get_undergrad_terms
...
+          MIDDLE_NAME as middle_name_legal,
+          LAST_NAME as last_name_legal,
+          UC_PRF_FIRST_NM as first_name_preferred,
+          UC_PRF_MIDDLE_NM as middle_name_preferred,
+          EMAIL_ADDR as email,
+          ACAD_PROG as academic_programs
+        FROM SISEDO.CLC_STDNT_LOOKUP_V00_VW
+        WHERE
+          upper(UC_SRCH_CRIT) LIKE upper('%#{search_string}%') AND
+          rownum < 31
```
lyttam: Just FYI, `DISTINCT` applies to all columns in the `SELECT`. So if a person has two email addresses, this query will return two rows.

app/models/user/search_users_by_name.rb
```diff
@@ -14,10 +14,14 @@ def only_special_characters?(name)
 
     def search_sisedo(name, opts)
       users = []
+      search_string = name.to_s.gsub(/[^0-9a-z ]/i, '')
```
lyttam: @redconfetti Why remove special characters (non-alphanumeric) from a name?  If you haven't already, it might be worth finding out what restrictions CS and/or the SISEDO database apply to person names. If those sources allow special characters, then we should too.

https://www.w3.org/International/questions/qa-personal-names

app/models/edo_oracle/queries.rb
redconfetti: @lyttam - That's not expected, but if it did happen I'm satisfied with the result.

app/models/user/search_users_by_name.rb
redconfetti: @lyttam - Yeah. I was actually thinking about that. Names like "O'Brien" might need the apostrophe. I'm checking on this with the BSA right now. I'll make a follow-up PR if things change.

lyttam: @redconfetti Perhaps we should err on the side of caution and allow all Unicode characters - leaving it up to the user to spell the person's name right (including any non-alphanumeric characters).  Name spelling can be a sensitive issue, and is a very common problem that is often solved poorly in the tech industry.

PR #7325, opened 2018-09-20
CLC-6901 Official Sections needs to let the spinning wheel ride

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6901


paulkerschen:
A little reassurance is all we need.

PR #7326, opened 2018-09-20
SISRP-42629: adds EMPLID param to Shopping Sheet link

lyttam:
https://jira.berkeley.edu/browse/SISRP-42629

### DIFF
--- a/app/models/edo_oracle/financial_aid/queries.rb
+++ b/app/models/edo_oracle/financial_aid/queries.rb
@@ -172,7 +172,8 @@ module EdoOracle
 
       def self.get_financial_aid_summary(person_id, aid_year)
         result = safe_query <<-SQL
-        SELECT UC_COST_ATTENDANCE,
+        SELECT STUDENT_ID,
+          UC_COST_ATTENDANCE,
           UC_GIFT_AID_WAIVER,
           UC_NET_COST,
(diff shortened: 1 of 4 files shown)


marcpechaitis:
@lyttam (and @davidlimcheng) ,
Late to the party here, but what happens if student clicks on shopping sheet link, then manually updates the EMPLID value in the URL to someone else's EMPLID -- does she see the shopping sheet for a different student?

lyttam:
@marcpechaitis Isn't there a known issue regarding that?  I'm not sure what's been done about it at this point.  

When the user clicks a link to CS or hits any CS URL, they are supposed to get a CAS authentication page first.  Once authenticated, CS should only let them see their own data.  If that's not the case, then something needs to be done about it on the CS side.

marcpechaitis:
@lyttam ,
I'm not familiar with how the Shopping Sheet was built (its delivered and not custom), and whether it checks that the EMPLID of the current user matches that which is being displayed on the page?

In our custom pages we do include safeguards to check that the current user is only viewing her own data.

I do think it is worth a quick test to see what happens.

lyttam:
@marcpechaitis Agreed, I will ask the BSA/QA team to test that (it requires a user who fits the criteria for seeing the Shopping Sheet link to be set up in test CAS so that we can log in as that user)

PR #7328, opened 2018-09-21
CLC-6880 Allow more time for Project Site creation

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6880


paulkerschen:
Fifteenth verse! Same as the first!

PR #7330, opened 2018-09-24
**DO NOT MERGE** SISRP-39065 - AA: UX Dev - Advisor Student Lookup

davidlimcheng:

We are passing the search string to the back-end, but doing so with special characters will stop the process, so we'll have to remove all special characters in the front-end first.

EDIT: Feature has been pulled from release, eventually going to close this PR.


lyttam:
@davidlimcheng I'm concerned about removing special characters from names.  There are lots of cases where non-alphanumeric characters are part of someone's name. I think those characters need to be encoded rather than removed.

davidlimcheng:
@lyttam The view we are relying on (`SISEDO.CLC_STDNT_LOOKUP_V00_VW`) actually strips special characters to provide the search criteria.  For the name "O'brien," for example, the view will provide more accurate results stripping the apostrophe.

That being said, @marcpechaitis can speak to this better than I can.

lyttam:
@davidlimcheng @marcpechaitis Is that the preferred policy? I keep raising this because people's names are very personal and culturally significant, and this issue deserves sensitivity and careful consideration.

If non-alphanumeric characters are removed, how does that impact searching for names like Björk, René, Carreño, etc?

davidlimcheng:
I've included this work in: 
https://github.com/davidlimcheng/calcentral/tree/SISRP-39065_all_commits

We can `cherry-pick` it to a new branch when we're ready to include it into `master`.  This is a good discussion, and I will include a link to this PR in the JIRA to facilitate discussion.

marcpechaitis:
@lyttam, Yes, in the PeopleSoft PS_NAMES table there are two fields LAST_NAME_SRCH and FIRST_NAME_SRCH.  These fields are automatically populated when the LAST_NAME and FIRST_NAME have values, and the values are converted to UPPER case and remove the special characters.

So in your example, Björk would have a FIRST_NAME_SRCH value of BJORK.  So if the user enters 'Björk' or 'Bjork' as a search term, the CalCentral search would picked it up on account of removing the 'ö' and converting to UPPER.

I'm also going to propose that (when we do end up implementing this code) we also strip out spaces.  The LAST_NAME_SRCH value for "Del Rio" is "DELRIO".  If CalCentral user searches on "Del Rio" (having a space) the current approach would not find a match from Campus Solutions.

cc: @davidlimcheng and @redconfetti

PR #7339, opened 2018-09-28
SISRP-39065 - filters input name to match campus solutions search criteria

davidlimcheng:

Follows a Campus Solutions function to prepare an inputted name to use as a query condition for SISEDO.  
CS code can be found in [Confluence.](https://wikihub.berkeley.edu/display/SIS/CalCentral+View+As+Card)

### DIFF
--- a/app/models/user/search_users_by_name.rb
+++ b/app/models/user/search_users_by_name.rb
@@ -14,8 +14,8 @@ module User
 
     def search_sisedo(name, opts)
       users = []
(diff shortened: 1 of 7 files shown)


lyttam:
@davidlimcheng this looks great! I'm glad to see those RSpec test cases.

Side note, it would be interesting to build a string that contains all of the Unicode characters (or a representative sample of them) and run it through both the CalCentral and CS functions, and compare the output.

redconfetti:
Looks great @davidlimcheng !

PR #7347, opened 2018-10-04
 CLC-6918 Add SIS ID changes CSV import as alternative to API loop

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6918
https://jira.ets.berkeley.edu/jira/browse/CLC-6920

First steps to improve the bCourses daily maintenance job.

Once we've tested the CSV-based SIS User ID import on a Junction server, the next step will be to put all the CSVs into one zip archive to be imported as one large job.


johncrossman:
A quality first step. Thanks @raydavis

PR #7350, opened 2018-10-10
CLC-6919 Use zip archive for daily SIS Imports to bCourses

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6919

Also takes care of CLC-6920, since `canvas_proxy.import_zipped_csvs` can be configured `false` to restore the old loop-APIs-then-loop-CSV-imports behavior.


johncrossman:
Oh, fancy, this PR has a zipper!

PR #7357, opened 2018-10-15
SISRP-43078 - Update Concerns::AdvisingResources to use Oracle View for CPP

redconfetti:
https://jira.berkeley.edu/browse/SISRP-43078

### DIFF
--- a/app/controllers/concerns/advising_resources.rb
+++ b/app/controllers/concerns/advising_resources.rb
@@ -55,8 +55,8 @@ module AdvisingResources
 
   def self.lookup_student_career(user_id)
     return nil if user_id.blank?
-    career = Concerns::AcademicStatus.newest_career(Concerns::AcademicStatus.academic_statuses MyAcademics::MyAcademicStatus.new(user_id).get_feed)
-    career.try(:[], 'code')
+    term_cpp = MyAcademics::MyTermCpp.new(user_id).get_feed
+    latest_career = term_cpp.first.try(:[], 'acad_career')
   end
 
   def self.fetch_links(link_config)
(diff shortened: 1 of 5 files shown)


app/controllers/concerns/advising_resources.rb
```diff
@@ -55,8 +55,10 @@ def self.empl_id(user_id)
 
   def self.lookup_student_career(user_id)
     return nil if user_id.blank?
-    career = Concerns::AcademicStatus.newest_career(Concerns::AcademicStatus.academic_statuses MyAcademics::MyAcademicStatus.new(user_id).get_feed)
-    career.try(:[], 'code')
+    cs_id = empl_id(user_id)
+    term_cpp = EdoOracle::Queries.get_student_term_cpp(cs_id)
+    latest_career_term = term_cpp.sort {|x,y| y['term_id'] <=> x['term_id'] }.first
```
lyttam: @redconfetti Is it possible to move the sorting into the query itself?  i.e. add an `ORDER BY` clause to the SQL.

It depends on what else uses this query and how the results are being cached.  But if it's feasible to do (maybe with some light refactoring), I'd recommend it since the database engine is optimized for sorting.

redconfetti: That's a great question @lyttam.
I just realized that [MyAcademics::MyTermCpp](https://github.com/ets-berkeley-edu/calcentral/blob/33271b3b/app/models/my_academics/my_term_cpp.rb#L8) is the primary consumer of this query, which provides it to others. So I should probably update this code to rely on that model, because it performs cacheing of the results.

In fact, [MyAcademics::ClassEnrollments](https://github.com/ets-berkeley-edu/calcentral/blob/33271b3b/app/models/my_academics/class_enrollments.rb#L98,L102) makes use of that same model already, so there's room for this update to obtain those records once and share them.

The ordering of the results doesn't appear to matter in [MyAcademics::MyAcademicRoles](https://github.com/ets-berkeley-edu/calcentral/blob/33271b3b/app/models/my_academics/my_academic_roles.rb#L62) either, so I'll update the query to sort by term ID descending (future to past). This strikes me as the most useful sorting for any future dependencies.

lyttam:
Looks good!

PR #7366, opened 2018-10-22
CLC-6929 Treat whitespace as null for OEC location checks

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6929


paulkerschen:
Stepping into the mind of Travis, I don't think he'd have a problem with this.

PR #7368, opened 2018-10-24
SISRP-40675 - Improve display and visibility of Early Drop Deadline Courses

redconfetti:
https://jira.berkeley.edu/browse/SISRP-40675

### DIFF
--- a/spec/models/my_academics/class_enrollments_spec.rb
+++ b/spec/models/my_academics/class_enrollments_spec.rb
@@ -583,4 +583,45 @@ describe MyAcademics::ClassEnrollments do
       end
     end
   end
+
+  context '#parse_early_drop_deadline_classes' do
+    let(:instruction) do
+      {
+        enrolledClasses: [
+          {
+            edd: 'Y',
+            subjectCatalog: 'MATH 1A',
+            title: 'CALCULUS',
+          },
+          {
+            edd: 'N',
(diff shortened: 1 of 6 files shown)


spec/models/my_academics/class_enrollments_spec.rb
```diff
@@ -583,4 +583,45 @@
       end
     end
   end
+
+  context '#early_drop_deadline_classes' do
```
lyttam: @redconfetti the function under test is called `parse_early_drop_deadline_classes`, can you change this line to match?

redconfetti: @lyttam - Done. I also updated the references to the API to expect 'EDD' instead of 'EARLY_DROP'.

lyttam: @redconfetti Hmm why change that? I'd rather have it spelled out. I'm not going to be able to remember what EDD stands for.

redconfetti: @lyttam - I didn't choose to change it. It was implemented on the API end with that. I suggested `EARLY_DROP`, but that's not what was implemented.

lyttam: @redconfetti got it. Thanks for fighting the good fight.

PR #7371, opened 2018-10-26
CLC-6931 OEC shouldn't worry so much about BIOLOGY

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6931


paulkerschen:
It doesn't need to know about a science book.

PR #7372, opened 2018-10-26
CLC-5189 E-Grades Export needs to spin more

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-5189

Everybody loves a spinner.


johncrossman:
What's better than spin? More spin.

PR #7380, opened 2018-11-02
SISRP-43372: adds package-lock to git and forces upgrade of errant dependency

lyttam:
https://jira.berkeley.edu/browse/SISRP-43372


raydavis:
THANK YOU!

PR #7386, opened 2018-11-06
SISRP-39698 - Split CalCentral Financial Aid Checklist Items on by Aid Year Dev

redconfetti:
https://jira.berkeley.edu/browse/SISRP-39698


src/assets/javascripts/angular/services/tasksService.js
```diff
@@ -55,6 +52,35 @@ angular.module('calcentral.services').service('tasksService', function(apiServic
     };
   };
 
+  var csFinaidFilterFactory = function(taskSection) {
+    return function(task) {
```
nkutub: @redconfetti, fanciness happening here, any reason, you just create a single function(taskSection, task) ?

redconfetti: @nkutub - It's a factory method that returns another method. The method returned is used with `filter` method as a callback that only accepts a single `task` argument.

Because there can be a variable number of `taskSection` objects (one for 2017-2018, another for 2018-2019, etc), we need a function returned that is tailored specifically for a specific task section.

I'm using a trick here to ensure that the `taskSection.id` and `taskSection.aidYearId` is associated within the scope of the function returned... because the function itself MUST accept only a single `task` argument. I can't make the `filter` support passing the `id` and `aidYearId`. So this is a trick to accomplish that.

nkutub: got it

PR #7396, opened 2018-11-14
SISRP-42013 - Academic Status updates (UX Dev)

nkutub:
https://jira.berkeley.edu/browse/SISRP-42013

* bug fixes

### DIFF
--- a/src/assets/templates/widgets/academic_standing.html
+++ b/src/assets/templates/widgets/academic_standing.html
@@ -17,7 +17,7 @@
               Please consult with your college advisor for guidance.
             </span>
             <button class="cc-button-link cc-standing-sub-title" data-ng-click="academics.showLearnMore = true" data-ng-show="!academics.showLearnMore">Learn More</button>
-            <span data-ng-bind="standings.learnMoreMessage" class="cc-standing-sub-title" data-ng-show="academics.showLearnMore"></span>
+            <span data-ng-bind-html="standings.learnMoreMessage" class="cc-standing-sub-title" data-ng-show="academics.showLearnMore"></span>
           </td>
         </tr>
         <tr data-ng-if="standings.standingsHistory.length">
@@ -34,7 +34,7 @@
                 <thead>
                   <tr>
                     <th scope="col">Status History</th>
-                    <th scope="col" class="cc-table-right cc-table-right-spacing">Term Ending</th>
+                    <th scope="col" class="cc-table-right">Term Ending</th>
                   </tr>
                 </thead>
                 <tbody data-ng-repeat="standing in standings.standingsHistory">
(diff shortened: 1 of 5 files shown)


src/assets/templates/widgets/academic_standing.html
```diff
@@ -17,7 +17,7 @@
               Please consult with your college advisor for guidance.
             </span>
             <button class="cc-button-link cc-standing-sub-title" data-ng-click="academics.showLearnMore = true" data-ng-show="!academics.showLearnMore">Learn More</button>
-            <span data-ng-bind="standings.learnMoreMessage" class="cc-standing-sub-title" data-ng-show="academics.showLearnMore"></span>
+            <span ng-bind-html="standings.learnMoreMessage" class="cc-standing-sub-title" data-ng-show="academics.showLearnMore"></span>
```
redconfetti: @nkutub - Can we use the `data-` prefix (i.e. `data-ng-bind-html`)? It's a practice we've kept in place to avoid any sort of issues with invalid HTML5 syntax. I know it's not strictly required, but we've kept up that standard for so long, I don't see why we should abandon it now.

src/assets/templates/widgets/status_and_holds.html
```diff
@@ -4,8 +4,8 @@ <h2>Status and Holds</h2>
   </div>
 
   <div class="cc-widget-padding" data-cc-spinner-directive="statusHolds.isLoading">
-    <h3 class="cc-status-holds-status-header" data-ng-if="regStatus.show || showResidency">Status</h3>
-     <div data-ng-if="api.user.profile.features.standings && ((api.user.profile.academicRoles.current.ugrd && !api.user.profile.academicRoles.current.ugrdNonDegree) || (api.user.profile.roles.advisor && targetUser.academicRoles.current.ugrd && !targetUser.academicRoles.current.ugrdNonDegree))">
+     <div data-ng-if="api.user.profile.features.standings && ((hasStandingAlert || api.user.profile.academicRoles.current.ugrd) && !api.user.profile.academicRoles.current.ugrdNonDegree || (api.user.profile.roles.advisor && (targetUser.academicRoles.current.ugrd || hasStandingAlert) && !targetUser.academicRoles.current.ugrdNonDegree))">
```
redconfetti: @nkutub - Is it possible we could break this down into methods in the `$scope` defined in the controller, and give it an easier to understand definition. I'm trying to pseudo code this and I'm still confused.

I'm thinking something like `api.user.profile.features.standing && (isVisibleToStudent() || isVisibleToAdvisorViewingStudent())`.

nkutub: yup, I was thinking that I will do that

src/assets/templates/widgets/academic_standing.html
nkutub: ya , missed that

PR #7405, opened 2018-11-17
SISRP-43016 - Switches filenaming to use feed section list. Adds course name column for crosslisted courses

redconfetti:
https://jira.berkeley.edu/browse/SISRP-43016

See QA PR #7406

### DIFF
--- a/app/models/rosters/csv.rb
+++ b/app/models/rosters/csv.rb
@@ -1,11 +1,16 @@
 module Rosters
   class Csv < Common
-    def initialize(students, opts = {})
-      @students = students
+    def initialize(rosters_feed, opts = {})
+      @students = rosters_feed.try(:[], :students)
+      @sections = rosters_feed.try(:[], :sections)
       @campus_course_id = opts[:campus_course_id] if opts[:campus_course_id].present?
       @section_id = opts[:section_id] if opts[:section_id].present?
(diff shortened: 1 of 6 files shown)


app/models/rosters/csv.rb
```diff
@@ -1,13 +1,20 @@
...
       @section_id = opts[:section_id] if opts[:section_id].present?
       @enroll_option = opts[:enroll_option] if opts[:enroll_option].present?
       filter_students_by_options
     end
 
+    def is_crosslisted_course?
+      @is_crosslisted_course ||= Proc.new do
+        !!@sections.to_a.find { |sec| !!sec.try(:[], :cross_listing) }
+      end.call
+    end
```
davidlimcheng: @redconfetti, is there a difference between this and something that might be more readable, like so?
```
@is_crosslisted_course ||= !!@sections.to_a.find { |sec| !! sec.try(:[], :cross_listing) }
```
In other words, what is the `Proc` definition adding?

redconfetti: @davidlimcheng - You have a point, the Proc isn't needed to ensure that it's only run once.

redconfetti: @davidlimcheng - It's updated now

davidlimcheng:
Thanks @redconfetti , LGTM, will merge once Travis gives the green light.

PR #7416, opened 2018-11-28
SISRP-43143 - Rewrite Student Resources Card in React

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-43143

Kicks off our React refactor!  Still need to create a roadmap, but this will get us going on new development efforts and refactoring at the lowest level (controllers/templates).

@nkutub, @redconfetti, this is the same functionality that I demoed a month or so ago. It's majorly refactored though to be more React-friendly. Would it be helpful to do another demo, both as a refresher and to cover new material?


redconfetti:
@davidlimcheng - Yes. Going over this again for 30-60 minutes is a good idea. I notice a lot of things that we reviewed before, but I'm not sure what you've changed since last time.

spec/models/campus_solutions/student_resources_spec.rb
```diff
@@ -1,17 +1,130 @@
 describe CampusSolutions::StudentResources do
+  require('pp')
```
redconfetti: @davidlimcheng - Is this needed?

src/react/components/StudentResources.js
```diff
@@ -0,0 +1,21 @@
...
+    );
+  }
+}
+StudentResources.propTypes = {
+  renderLinkSection: PropTypes.func.isRequired,
+  resources: PropTypes.array.isRequired,
+  widgetConfig: PropTypes.object.isRequired
+};
+
+export default StudentResources;
```
redconfetti: @davidlimcheng - So it looks like you're using a WidgetContainer component to wrap the widget, which will make it possible to implement shared "widget" functionality that applies to all the "widgets" or "cards" we show. This is the closest things to "inheritance" we can get with React right... creating components that are shared for the same purpose.

davidlimcheng: Yep. Javascript does actually offer a form of inheritance via their classes - for example we could do something like `class StudentResources extends Widget` instead of `class StudentResources extends React.Component`, but React docs actually recommend not doing that:
https://reactjs.org/docs/composition-vs-inheritance.html

So yes, using what they call "Containment", is the best way we can do things that mimic inheritance within React.

spec/models/campus_solutions/student_resources_spec.rb
davidlimcheng: @redconfetti  amended - nice catch!

PR #7420, opened 2018-12-06
SISRP-42744 - CalCentral Access Fringe Case Review & Analysis

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-42744

Changes Authentication Validator logic to accommodate users utilizing Single Sign-On from Slate. To sum it up:
- if you are an undergraduate student/applicant,
  - check to see if you've authenticated via Slate SSO
    - if so, deny if the `ADMT_UX` affiliation has not been provisioned
    - continue login if otherwise
- if you are not an undergraduate student/applicant,
  - use current login logic

cc @raydavis - Wanted to ping you since this functionality is a dependency of `CanvasLtiController`. This will change the way undergraduates are handled - the assumption was that if you are able to authenticate via CAS (as opposed to Slate), then you belong in CalCentral (line 64 in `authentication_validator.rb`). I hope this doesn't conflict with the Canvas Authentication process - if so, let's talk it through!

### DIFF
--- a/app/models/user/authentication_validator.rb
+++ b/app/models/user/authentication_validator.rb
@@ -7,12 +7,17 @@ module User
 
     attr_reader :auth_uid
 
-    def initialize(auth_uid)
+    def initialize(auth_uid, auth_handler=nil)
+      @auth_handler = auth_handler
       @auth_uid = auth_uid
     end
 
     def feature_enabled?
       Settings.features.authentication_validator
+    end
+
+    def slate_auth_handler
+      @slate_handler ||= Settings.authentication_handlers.slate.to_s.downcase
     end
 
     def validated_user_id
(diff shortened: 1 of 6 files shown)


app/models/user/authentication_validator.rb
```diff
@@ -47,20 +52,31 @@ def held_applicant?
...
+        cs_roles = roles_from_cs_affiliations(affiliations)
+        if is_undergraduate?
+          user_auth_handler = @auth_handler.to_s.downcase
+          # check to see if the student was a SSO user from Slate
+          if user_auth_handler.include?(slate_auth_handler)
+            # if they were, only let them in if they have the `ADMT_UX` role provisioned from Campus Solutions
+            return !cs_roles[:releasedAdmit]
+          else
+            # otherwise, they were able to authenticate via Calnet credentials, so let them in
+            return false
```
davidlimcheng: If the user did not authenticate via `slateAuthenticationHandler`, we assume that they were authenticated via another secure route and should be allowed into CalCentral.  Welcoming extra scrutiny of that logic here!

app/models/user/authentication_validator.rb
```diff
@@ -7,14 +7,19 @@ class AuthenticationValidator
...
+      @auth_handler = auth_handler
       @auth_uid = auth_uid
     end
 
     def feature_enabled?
       Settings.features.authentication_validator
     end
 
+    def slate_auth_handler
+      @slate_handler ||= Settings.authentication_handlers.slate.to_s.downcase
```
raydavis: {{settings.yml}} doesn't include an {{authentication_handlers}} section. Missing from the PR?

app/models/user/authentication_validator.rb
```diff
@@ -69,11 +85,11 @@ def has_ldap_affiliations?
       ldap_roles.has_value?(true)
     end
 
-    def is_graduate_applicant?
+    def is_undergraduate?
```
raydavis: @davidlimcheng , I don't think ETS Junction has access to the Checklist API.

Due to short-circuiting on `held`, it looks like the existing master-branch logic will only get to the next line of code if the CS affiliations have already told us that this is an unreleased-applicant and if there are no CalNet affiliations. If I've got that right, then Junction/bCourses wouldn't end up trying to call the API, since we don't allow CAS login through Slate.

But it looks like this PR moves the SirStatuses/Checklist call way up in the decision tree, so that the API gets called for anyone with CS affiliations. 

Do I have that right, or did I get lost in the brush?

raydavis:
@davidlimcheng , if it turns out that my reading is correct & this PR would cause the CS Checklist API to be called for most logins, we might be able to work around the conflict through configurations.

Despite its impressively general name, `User::AuthenticationValidator` only does one thing: block "partially" (as in not yet announced) admitted new students from getting into CalCentral and learning their status early.

But sometime after the CalCentral/Junction split, ETS set up Junction-specific CAS integration with Custom Access Groups which included Guest accounts but not unreleased-applicants. As a result, in our case the AuthenticationValidator should be a no-op and should be OK to disable.

Obviously we'll need some time to test that, though, so do let me know what you think.

app/models/user/authentication_validator.rb
davidlimcheng: Ah, I added it on my `local.yml` but I should also add it to actual `Settings.yml` - will amend.

davidlimcheng:
Thanks for the concise summary of the issue, @raydavis. To further add on to your concern of the early calling of the `Checklist` API, I'm also not too thrilled about having to make potentially 3 network calls for the non-undergraduate flow.  

Let me check to see if there's any other way we can distinguish undergraduate applicants vs non-undergraduate applicants besides hitting `Checklists`. If that is not possible, though, would you want to test disabling `AuthenticationValidator` before merging this PR?  I believe we have a release scheduled for 1/30 for this change, so we've got some wiggle room with regards to timeline.

app/models/user/authentication_validator.rb
davidlimcheng: Totally - responded to you in PR-level comments!

raydavis:
Thanks, @davidlimcheng & friends -- looks like that should be an improvement for SIS CalCentral, too!

On the side, I'll create a Junction ticket for this sprint to disable ETS dependency on that class.

davidlimcheng:
@raydavis - Amended with updated logic, added `authentication_handlers` to `settings.yml`, and was able to clean up some of the testing in `authentication_validator_spec.rb`. Thanks for getting us started down the path of more efficient code!

app/models/user/authentication_validator.rb
raydavis: Thanks, @davidlimcheng -- looks like your rewrite removed the only call to this method, so you might not even need it anymore.

redconfetti:
@davidlimcheng - Why is the logic concerning Graduate Applicants being dropped?

PR #7235, opened 2018-07-15
SISRP-36756 - Final Exams Schedule (Students)

redconfetti:
https://jira.berkeley.edu/browse/SISRP-36756

PR #7421, opened 2018-12-07
SISRP-44157 - Remove unused final exam week attribute from Berkeley::Term

redconfetti:
https://jira.berkeley.edu/browse/SISRP-44157

### DIFF
--- a/app/models/berkeley/term.rb
+++ b/app/models/berkeley/term.rb
@@ -103,8 +103,6 @@ module Berkeley
         @classes_start = term['classes_start'].to_date.in_time_zone.to_datetime
         @instruction_end = term['instruction_end'].to_date.in_time_zone.to_datetime
         @classes_end = term['classes_end'].to_date.in_time_zone.to_datetime
-        @final_exam_week_start = term['final_exam_week_start'].to_date.in_time_zone.to_datetime
-        @final_exam_cs_data_available = @end.advance(days: -56)
         @end_drop_add =term['end_drop_add'] if term['end_drop_add'].present?
       end
       self
@@ -129,8 +127,6 @@ module Berkeley
         @classes_start = db_row['class_begin_date'].to_date.in_time_zone.to_datetime  if db_row['class_begin_date'].present?
         @instruction_end = db_row['instruction_end_date'].to_date.in_time_zone.to_datetime if db_row['instruction_end_date'].present?
         @classes_end = db_row['class_end_date'].to_date.in_time_zone.to_datetime if db_row['class_end_date'].present?
-        @final_exam_week_start = db_row['final_exam_week_start_date'].to_date.in_time_zone.to_datetime if db_row['final_exam_week_start_date'].present?
-        @final_exam_cs_data_available = @end.advance(days: -56)
         @end_drop_add =db_row['end_drop_add_date'] if db_row['end_drop_add_date'].present?
       end
       self
@@ -189,7 +185,7 @@ module Berkeley
     def to_h
       methods = [:campus_solutions_id, :name, :year, :code, :slug, :to_english,
         :start, :end, :classes_start, :classes_end, :instruction_end, :grades_entered,
-        :is_summer, :legacy?, :final_exam_week_start, :end_drop_add]
+        :is_summer, :legacy?, :end_drop_add]
       Hash[methods.collect {|m| [m, send(m)]}]
     end
 
--- a/app/models/edo_oracle/queries.rb
+++ b/app/models/edo_oracle/queries.rb
@@ -773,7 +773,6 @@ module EdoOracle
           CLASS_END_DT as class_end_date,
           INSTRUCTION_END_DT as instruction_end_date,
           GRADES_ENTERED_DT as grades_entered_date,
-          FINAL_EXAM_WK_BEGIN_DT as final_exam_week_start_date,
           END_DROP_ADD_DT as end_drop_add_date,
           IS_SUMMER as is_summer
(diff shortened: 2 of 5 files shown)


app/models/berkeley/term.rb
```diff
@@ -189,7 +189,7 @@ def grades_entered
     def to_h
       methods = [:campus_solutions_id, :name, :year, :code, :slug, :to_english,
         :start, :end, :classes_start, :classes_end, :instruction_end, :grades_entered,
-        :is_summer, :legacy?, :final_exam_week_start, :end_drop_add]
```
davidlimcheng: @redconfetti , can you expand a bit on why removing this symbol enables `to_h` to work, and what `to_h` is used for?  I see that `final_exam_week_start` is still being used [elsewhere in Berkeley::Term](https://github.com/ets-berkeley-edu/calcentral/blob/master/app/models/berkeley/term.rb#L106), will removing it here cause errors in other places that are still depending on it?

redconfetti: @davidlimcheng - Because we removed `:final_exam_week_start` and `:final_exam_cs_data_available` in #7235, so when #to_h would try to reference it, it would get an error.

redconfetti:
Honestly my goal was to fix Berkeley::Term so that when you're viewing one of the object in the Rails console, it would show information about the object in the console. It turns out that `ActiveAttrModel` seems to be interrupting this behavior. At least now we can call `term.to_h` or `term.to_s` to see which term the object represents.

```ruby
terms = Berkeley::Terms.fetch
terms.to_s
 => "#<Berkeley::Terms:0x5a654e05>"

current_term = terms.current
 => #<Berkeley::Term>

current_term.to_s
NoMethodError: undefined method `final_exam_week_start' for #<Berkeley::Term>
	from org/jruby/RubyBasicObject.java:1657:in `method_missing'
	from /Users/jasonmiller/.rvm/gems/jruby-9.1.14.0@calcentral/gems/activemodel-4.2.8/lib/active_model/attribute_methods.rb:433:in `method_missing'
	from /Users/jasonmiller/Projects/calcentral/app/models/berkeley/term.rb:193:in `block in to_h'
```

app/models/berkeley/term.rb
davidlimcheng: Ah - thanks for the pointer to #7235. The removal of `:final_exam_week_start` and `:final_exam_cs_data_available` in #7235 implies that we no longer need that data for any functionality that relies on `Berkeley::Term`.  Can you confirm that?  And if that is the case, then can we also remove the usages of those variables in lines [106+107](https://github.com/ets-berkeley-edu/calcentral/blob/master/app/models/berkeley/term.rb#L106) and [132+133](https://github.com/ets-berkeley-edu/calcentral/blob/master/app/models/berkeley/term.rb#L132) in `Berkeley::Term`?  

If they are indeed still necessary, adding `attr_reader :final_exam_week_start` also fixes the issue, since it allows us to get the value:
```
# after adding attr_reader :final_exam_week_start back to Berkeley::Term
jruby-9.1.14.0@calcentral :001 > >> terms = Berkeley::Terms.fetch

jruby-9.1.14.0@calcentral :002 > >> terms.to_s
terms.to_s
 => "#<Berkeley::Terms:0x147c00c7>" 

jruby-9.1.14.0@calcentral :004 > >> current_term = terms.current
current_term = terms.current

jruby-9.1.14.0@calcentral :005 > >> current_term.to_s
current_term.to_s
 => "#<Berkeley::Term> {:campus_solutions_id=>\"2188\", :name=>\"Fall\", :year=>2018, :code=>\"D\", :slug=>\"fall-2018\", :to_english=>\"Fall 2018\", :start=>Wed, 15 Aug 2018 00:00:00 -0700, :end=>Fri, 14 Dec 2018 00:00:00 -0800, :classes_start=>Wed, 22 Aug 2018 00:00:00 -0700, :classes_end=>Fri, 07 Dec 2018 00:00:00 -0800, :instruction_end=>Fri, 07 Dec 2018 00:00:00 -0800, :grades_entered=>Fri, 11 Jan 2019 00:00:00 -0800, :is_summer=>false, :legacy?=>false, :final_exam_week_start=>Mon, 10 Dec 2018 00:00:00 -0800, :end_drop_add=>2018-09-12 00:00:00 UTC}" 

jruby-9.1.14.0@calcentral :006 > >> current_term.to_h
current_term.to_h
 => {:campus_solutions_id=>"2188", :name=>"Fall", :year=>2018, :code=>"D", :slug=>"fall-2018", :to_english=>"Fall 2018", :start=>Wed, 15 Aug 2018 00:00:00 -0700, :end=>Fri, 14 Dec 2018 00:00:00 -0800, :classes_start=>Wed, 22 Aug 2018 00:00:00 -0700, :classes_end=>Fri, 07 Dec 2018 00:00:00 -0800, :instruction_end=>Fri, 07 Dec 2018 00:00:00 -0800, :grades_entered=>Fri, 11 Jan 2019 00:00:00 -0800, :is_summer=>false, :legacy?=>false, :final_exam_week_start=>Mon, 10 Dec 2018 00:00:00 -0800, :end_drop_add=>2018-09-12 00:00:00 UTC} 
```

redconfetti: @davidlimcheng - Good catch. These were originally added in #5833 in anticipation of the Final Exam Schedule cards. When I removed them in #7235 I left support for these fields.

As we discussed in person, I'm going to remove support for them as they are not in use in our code.

davidlimcheng:
Thanks @redconfetti!

PR #7422, opened 2018-12-12
SISRP-41477 - On Profile card Student Overview Page is showing Default photo When there is no photo of a student.

frankiepun:
https://jira.berkeley.edu/browse/SISRP-41477

### DIFF
--- a/src/assets/javascripts/angular/controllers/pages/userOverviewController.js
+++ b/src/assets/javascripts/angular/controllers/pages/userOverviewController.js
@@ -66,6 +66,8 @@ angular.module('calcentral.controllers').controller('UserOverviewController', fu
   };
   $scope.standingIsVisible = false;
 
+  $scope.profilePicture = {};
+
   var parseAdvisingResources = function(response) {
     var links = $scope.ucAdvisingResources.links;
     angular.extend(links, _.get(response, 'data.feed'));
--- a/src/assets/templates/widgets/toolbox/user_preview.html
+++ b/src/assets/templates/widgets/toolbox/user_preview.html
@@ -13,7 +13,16 @@
           <tbody>
           <tr>
             <th>
-              <div data-cc-profile-image-directive data-name="targetUser.fullName" data-uid="targetUser.ldapUid"></div>
+             <!-- <div data-cc-profile-image-directive data-name="targetUser.fullName" data-uid="targetUser.ldapUid"></div> -->
+
+              <img data-cc-load-error-directive="profilePicture" data-ng-if="!profilePicture.loadError" class="cc-widget-profile-img" height="96" width="72"
+                   data-ng-attr-alt="{{ targetUser.fullName && targetUser.fullName.trim ? targetUser.fullName.trim() + '\'s photo' : 'profile photo' }}"
+                   data-ng-src="api/photo/{{ targetUser.ldapUid }}">
+              <div data-ng-if="profilePicture.loadError"  class="cc-widget-profile-picture-not-available">
+                <i class="fa fa-user"></i>
+              </div>
+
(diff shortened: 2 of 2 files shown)


src/assets/templates/widgets/toolbox/user_preview.html
```diff
@@ -13,7 +13,16 @@ <h3 data-ng-bind="targetUser.error.summary"></h3>
           <tbody>
           <tr>
             <th>
-              <div data-cc-profile-image-directive data-name="targetUser.fullName" data-uid="targetUser.ldapUid"></div>
+             <!-- <div data-cc-profile-image-directive data-name="targetUser.fullName" data-uid="targetUser.ldapUid"></div> -->
```
redconfetti: @frankiepun - You need to remove this comment.

src/assets/templates/widgets/toolbox/user_preview.html
```diff
@@ -13,7 +13,16 @@ <h3 data-ng-bind="targetUser.error.summary"></h3>
...
+
+              <img data-cc-load-error-directive="profilePicture" data-ng-if="!profilePicture.loadError" class="cc-widget-profile-img" height="96" width="72"
+                   data-ng-attr-alt="{{ targetUser.fullName && targetUser.fullName.trim ? targetUser.fullName.trim() + '\'s photo' : 'profile photo' }}"
+                   data-ng-src="api/photo/{{ targetUser.ldapUid }}">
+              <div data-ng-if="profilePicture.loadError"  class="cc-widget-profile-picture-not-available">
+                <i class="fa fa-user"></i>
+              </div>
+
```
marcpechaitis: @frankiepun extra whitespace?

src/assets/javascripts/angular/controllers/pages/userOverviewController.js
```diff
@@ -66,6 +66,8 @@ angular.module('calcentral.controllers').controller('UserOverviewController', fu
   };
   $scope.standingIsVisible = false;
 
+  $scope.profilePicture = {};
```
redconfetti: @frankiepun - I don't see why we're declaring this in the UserOverviewController. There isn't functionality within that controller that populates a value in that property of the `$scope`.

src/assets/templates/widgets/toolbox/user_preview.html
```diff
@@ -13,7 +13,12 @@ <h3 data-ng-bind="targetUser.error.summary"></h3>
           <tbody>
           <tr>
             <th>
-              <div data-cc-profile-image-directive data-name="targetUser.fullName" data-uid="targetUser.ldapUid"></div>
+              <img data-cc-load-error-directive="profilePicture" data-ng-if="!profilePicture.loadError" class="cc-widget-profile-img" height="96" width="72"
```
redconfetti: @frankiepun - Also, because the `profilePicture` property isn't being populated with a value via the `$scope`, I don't see the `data-cc-load-error-directive` nor the `data-ng-if` attribute as being needed here.

It looks like this code simply relies on the `data-ng-src` attribute to hit API behind `/api/photo/:ldap_uid`, which returns the users photo.

frankiepun: @redconfetti Thanks for the great review. ccLoadErrorDirective will attach a callback to the img tag to handle the image load error and set the controller's $scope.profilePicture.loadError to true. That's why we initialize $scope.profilePicture = {} to an empty object. I totally understand the confusion of the ccLoadErrorDirective because it took me quite some time to understand how to use it. I think it may be easier to schedule a meeting to review it together. Again, thanks for the peer review.

PR #7425, opened 2018-12-13
CLC-6957 Short-circuit bCourses refresh on DB errors

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6957

These two queries play too central a role for the show to go on without them.

The PR also fixes end-user reporting of background job errors.


johncrossman:
`fallible_query` wins method name of the month.

PR #7426, opened 2018-12-13
CLC-6957 Interrupt bCourses import if too many enrollment drops

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6957

Would you like a belt with those suspenders?


redconfetti:
@raydavis - LGTM

johncrossman:
Ditto. Thanks @raydavis

PR #7157, opened 2018-05-23
SISRP-37193 - Reserve Capacity: Show in CalCentral - Student view

nkutub:
https://jira.berkeley.edu/browse/SISRP-37193
* add feature flag - not active for this release

PR #7430, opened 2018-12-18
SISRP-39112: use buttons for ng-click functions

developish:
https://jira.berkeley.edu/browse/SISRP-39112

### DIFF
--- a/src/assets/templates/academics_semester_classes.html
+++ b/src/assets/templates/academics_semester_classes.html
@@ -87,8 +87,8 @@
                 </td>
                 <td data-ng-if="api.user.profile.features.reservedCapacity">
                   <div data-ng-if="section.waitlistPosition && section.hasReservedSeats">
-                    <span data-ng-click="course.capacity = section.capacity; course.showSeats = true" data-ng-show="!course.showSeats"><a href="#">Show</a></span>
-                    <span data-ng-click="course.showSeats = false" data-ng-show="course.showSeats"><a href="#">Hide</a></span>
+                    <button data-ng-click="course.capacity = section.capacity; course.showSeats = true" data-ng-show="!course.showSeats" class="cc-button-link">Show</button>
+                    <button data-ng-click="course.showSeats = false" data-ng-show="course.showSeats" class="cc-button-link">Hide</button>
                   </div>
                   <div data-ng-if="section.waitlistPosition && !section.hasReservedSeats">
                     N/A
--- a/src/assets/templates/directives/academics_class_info_enrollment.html
+++ b/src/assets/templates/directives/academics_class_info_enrollment.html
@@ -28,40 +28,40 @@
 
       <div class="visible-for-small-only cc-academics-class-enrollment-mobile-button-spacer"></div>
 
-      <a class="cc-button cc-class-enrollment-formatted-content-show-button"
+      <button class="cc-button cc-class-enrollment-formatted-content-show-button"
           data-ng-disabled="displayedSection === 'addresses' || noStudentsSelected()"
           data-ng-click="displaySection('addresses')">
         <i class="fa fa-clipboard" aria-hidden="true"></i>
         <span class="hidden-for-small-only">Copy Addresses</span>
-      </a>
+      </button>
 
-      <a class="cc-button cc-class-enrollment-formatted-content-show-button"
+      <button class="cc-button cc-class-enrollment-formatted-content-show-button"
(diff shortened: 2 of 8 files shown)


src/assets/templates/academics_semester_classes.html
```diff
@@ -87,8 +87,8 @@ <h3>Wait Lists</h3>
                 </td>
                 <td data-ng-if="api.user.profile.features.reservedCapacity">
                   <div data-ng-if="section.waitlistPosition && section.hasReservedSeats">
-                    <span data-ng-click="course.capacity = section.capacity; course.showSeats = true" data-ng-show="!course.showSeats"><a href="#">Show</a></span>
-                    <span data-ng-click="course.showSeats = false" data-ng-show="course.showSeats"><a href="#">Hide</a></span>
+                    <button data-ng-click="course.capacity = section.capacity; course.showSeats = true" data-ng-show="!course.showSeats">Show</button>
```
davidlimcheng: @developish Since this instance doesn't have any explicit CSS, did this change have any impact on the actual UI of the button?

davidlimcheng:
@developish As `CLC` is the JIRA instance that ETS uses, can we change both the PR title and the commit name to `SISRP-39112` to keep confusion to a minimum?

src/assets/templates/academics_semester_classes.html
redconfetti: Yeah. Expecting `cc-button-link` class used here.

src/assets/templates/directives/academics_class_info_enrollment.html
```diff
@@ -2,12 +2,12 @@
   <div class="cc-widget-title">
     <h2 class="cc-left" data-ng-bind="title"></h2>
     <div class="cc-academics-class-enrollment-options cc-right">
-      <a class="cc-button hidden-for-small-only" target="_self"
+      <button class="cc-button hidden-for-small-only" target="_self"
```
davidlimcheng: @developish This might have been better left as an `<a>`. `<button>` might work, but `target="_self"` may have to be changed to `formtarget="_self"`. I'm not too familiar with the context here, this one might need to change back.

Button attributes: https://www.w3schools.com/tags/tag_button.asp

davidlimcheng:
In addition to what @redconfetti mentioned, smoke testing the functionality of the changes is also important. Testing the button locally and comparing it to what the same "button" is doing in CC-QA, things like that.

developish:
@redconfetti you copied [Developer Workflow](https://wikihub.berkeley.edu/display/SIS/CalCentral+Developer+Workflow) to the GitHub Wiki, I think because my clearance hadn't come through. But I referenced it instead of the GitHub version where you updated the CLC references to SISPR. We should probably only have it one place.

src/assets/templates/directives/academics_class_info_enrollment.html
developish: This change wasn't supposed to be included. It's intended to be a link.

redconfetti:
@developish - Thanks for pointing that out. CalCentral was originally tracked using the [ETS JIRA system](https://jira.ets.berkeley.edu/jira/secure/Dashboard.jspa), using the 'CLC' prefix for tasks. When CalCentral ownership was moved over to the Student Information Systems Replacement Project (SISRP), all the tasks switched from `CLC` to `SISRP` in the new system.

Now we're simply the Student Information Systems department, so SISRP is a left over representation that we're keeping in place.

I've updated the article to reflect the SISRP prefix, to avoid confusion in the future (even though the images still reference CLC JIRAs).

developish:
I think you've mentioned that to me before, I even started everything with SISRP but then changed it to CLC when I read through the guide 😬

developish:
@redconfetti I didn't go through the changes in the UI to verify, if you could give me a hand with that, I'd appreciate it.

src/assets/templates/academics_semester_classes.html
redconfetti: @davidlimcheng - I reviewed this with @developish and we concluded that this code is used by a feature that is currently feature flagged as off. See #7157

davidlimcheng: @redconfetti @developish Thanks guys. I would think though, that it's better to ensure this looks as expected now, rather than potentially forgetting about it and turning on the feature flag in the future. What do you guys think?

developish:
I went through and verified the UI for all the changes, and rebased onto current master

redconfetti:
@developish - Looks good

PR #7433, opened 2018-12-21
SISRP-41873 FA API Refactor #1 - No Terms & Conditions redirect

marcpechaitis:
https://jira.berkeley.edu/browse/SISRP-41873

### DIFF
--- a/src/assets/javascripts/angular/controllers/pages/termsAndConditionsController.js
+++ b/src/assets/javascripts/angular/controllers/pages/termsAndConditionsController.js
@@ -10,10 +10,7 @@ angular.module('calcentral.controllers').controller('TermsAndConditionsControlle
     isLoading: true
   };
 
-  var getTermsAndConditions = function() {
-    var options = {
-      finaidYear: $routeParams.finaidYearId
-    };
+  var getTermsAndConditions = function(options) {
     return termsAndConditionsFactory.getTermsAndConditions(options).then(
       function successCallback(response) {
         var termsAndConditions = _.get(response, 'data.termsAndConditions');
@@ -23,9 +20,12 @@ angular.module('calcentral.controllers').controller('TermsAndConditionsControlle
     );
   };
 
-  getTermsAndConditions();
+  getTermsAndConditions({ finaidYear: $routeParams.finaidYearId });
 
   $scope.$on('calcentral.custom.api.finaid.approvals', function() {
-    termsAndConditionsFactory.getTermsAndConditions();
+    termsAndConditionsFactory.getTermsAndConditions({
+      finaidYear: $routeParams.finaidYearId,
+      refreshCache: true
+    });
   });
 });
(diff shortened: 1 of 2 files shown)


marcpechaitis:
Note to reviewers:  We have removed ability for user to "decline" Terms & Conditions, so I also removed some dead code that is no longer necessary.

src/assets/javascripts/angular/controllers/pages/termsAndConditionsController.js
```diff
@@ -23,9 +25,9 @@ angular.module('calcentral.controllers').controller('TermsAndConditionsControlle
     );
   };
 
-  getTermsAndConditions();
+  getTermsAndConditions(options);
```
redconfetti: @marcpechaitis - Because T&C are global data, we only need to retrieve the data set once?

marcpechaitis: @redconfetti incorrect, Terms & Conditions are agreed to on an _per Aid Year_ basis

redconfetti: Discussed this with @marcpechaitis and clarified that the `$scope.$on` callback on lines 30-32 needs to not only grab a fresh data copy (like it does now) but also put the data retrieved into the `$scope.termsAndConditions`. Additionally this approach is refreshing the cache on each page load, which isn't necessary.

src/assets/javascripts/angular/controllers/widgets/finaid/finaidApprovalsController.js
```diff
@@ -3,28 +3,18 @@
...
   $scope.sendResponseTC = function(finaidYearId, response) {
-    finaidFactory.postTCResponse(finaidYearId, response).then(function(data) {
-      if (response === 'N') {
-        showDeclineMessage(data);
-      } else {
-        sendEvent();
-        $location.path('/finances');
-      }
-    });
+    finaidFactory.postTCResponse(finaidYearId, response).then(sendEvent()).then($route.reload());
```
redconfetti: @marcpechaitis - Nice use of `$route.reload()`. I didn't know that existed for refreshing the current route context.

So we're just simply sending a single request now for when the user accepts the terms and conditions, and then reloading the current page. Did this work well in your local testing? I'm imagining that the `sendEvent()` call might be asynchronously called along with `$route.reload()`, thus introducing a possible race condition.

redconfetti: Actually, even if the page reloads before the refresh of data happens, it will end up triggering the `$scope` within those other controllers to refresh the data in the page. So a race condition probably doesn't exist.

marcpechaitis: Thanks for the analysis and feedback @redconfetti

src/assets/javascripts/angular/controllers/pages/termsAndConditionsController.js
marcpechaitis: I implemented the changes suggested by @redconfetti , thanks Jason!

PR #7435, opened 2018-12-21
SISRP-40280 - Apply UGRD and GRAD grading window support to Grading links

redconfetti:
https://jira.berkeley.edu/browse/SISRP-40280


developish:
For posterity: I reviewed and gave @redconfetti some refactoring ideas on this via slack

app/models/my_academics/grading/applicator.rb
```diff
@@ -0,0 +1,398 @@
+module MyAcademics
```
redconfetti: `MyAcademics::Grading` is now `MyAcademics::Grading::Applicator`. It was moved and refactored to make use of the other new models under the `MyAcademics::Grading` module.

app/models/my_academics/grading/applicator.rb
```diff
@@ -0,0 +1,398 @@
...
+            add_grading_dates_law(semester, term_id) if has_career_class?('LAW', semester[:classes])
+          end
+        end
+      end
+
+      def add_grading_dates_general(semester, term_id)
+        has_ugrd_classes = has_career_class?('UGRD', semester[:classes])
+        has_grad_classes = has_career_class?('GRAD', semester[:classes])
+
+        ugrd_session = MyAcademics::Grading::Session.get_session(term_id, 'UGRD')
```
redconfetti: @developish - This is where the new logic for determining the grading dates was applied using the new models

app/models/my_academics/grading/helpers.rb
```diff
@@ -0,0 +1,121 @@
+module MyAcademics
+  module Grading
+    module Helpers
```
redconfetti: This is exactly the same code as `MyAcademics::GradingModule`, now called `MyAcademics::Grading::Helpers`. Helpers no longer used are `#format_period_start` and `#format_period_end` which have been moved to `MyAcademics::Grading::Period`.

app/models/my_academics/grading/info_links.rb
```diff
@@ -0,0 +1,29 @@
+module MyAcademics
+  module Grading
+    class InfoLinks
```
redconfetti: This was previously MyAcademics::GradingInfoLinks

developish:
Assuming we can get the build to pass, this looks OK to me. It seems like it might be a connection issue, but I don't know how to trigger a rebuild.

developish:
@davidlimcheng I mean, it _could_ be that... Any objections to merging now that everything's passing?

redconfetti:
@developish - It's passing the tests now. I missed one small update to MyAcademics::Merged spec (changing `MyAcademics::Grading` to `MyAcademics::Grading::Applicator`)

davidlimcheng:
No objections here!  Thanks guys.

app/models/my_academics/grading/applicator.rb
```diff
@@ -0,0 +1,398 @@
...
+        grad_session = MyAcademics::Grading::Session.get_session(term_id, 'GRAD')
+
+        applicable_midterm_periods = Array.new.tap do |array|
+          if ugrd_session.midterm_period == grad_session.midterm_period
```
davidlimcheng: @redconfetti @developish So this `==` is actually calling the `==` that is defined in `MyAcademics::Grading::Period`?  This method here:
```
      def ==(other)
        self.start_date == other.start_date && self.due_date == other.due_date
      end
```
Is that right? So defining a method called `==` on the `Period` class pretty much overwrites the native `==` comparator?

app/models/my_academics/grading/applicator.rb
```diff
@@ -0,0 +1,398 @@
...
+            array << ugrd_session.midterm_period
+          else
+            array << ugrd_session.midterm_period if has_ugrd_classes
+            array << grad_session.midterm_period if has_grad_classes
+          end
+        end
+
+        applicable_final_periods = Array.new.tap do |array|
+          if ugrd_session.final_period == grad_session.final_period
```
redconfetti: Had a discussion with @davidlimcheng about the `#==` operator/method. I cited [Everything You Need to Know About Ruby Operators](https://www.rubyguides.com/2018/07/ruby-operators/) to explain the Ruby-ism at play here.

spec/models/my_academics/grading/info_links_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  before do
+    my_campus_links_stub = double(:links_my_campus_links, get_feed: campus_links)
+    allow(Links::MyCampusLinks).to receive(:new).and_return(my_campus_links_stub)
+  end
+
+  describe '.fetch' do
+    it 'returns faculty grading info links' do
+      grading_info_links = described_class.fetch
+      puts "result: #{grading_info_links.inspect}"
```
davidlimcheng: renegade `puts`

app/models/my_academics/grading/applicator.rb
redconfetti: Yes. And on a custom class that is appropriate, because the [BasicObject](https://ruby-doc.org/core-2.6/BasicObject.html) class is the parent class of all classes in Ruby. [BasicObject#==](https://devdocs.io/ruby~2.3/basicobject#method-i-3D-3D) returns true only if both objects are the same exact object (same Object Id).

```ruby
class Foo; end;
class Bar; end;

f = Foo.new
 => #<Foo:0x5de32e00>
b = Bar.new
 => #<Bar:0x219ec4b6>

f == b
 => false

f == f
 => true

f2 = Foo.new
 => #<Foo:0xcdc09d>
f == f2
 => false
```

Obviously this behavior is overridden by Ruby itself in classes such as [FixNum#==](https://devdocs.io/ruby~2.3/fixnum#method-i-3D-3D), [Array#==](https://devdocs.io/ruby~2.3/array#method-i-3D-3D), [Range#==](https://devdocs.io/ruby~2.3/range#method-i-3D-3D), [Hash#==](https://devdocs.io/ruby~2.3/hash#method-i-3D-3D), and many others.

In [Everything You Need to Know About Ruby Operators](https://www.rubyguides.com/2018/07/ruby-operators/) it's explained that Ruby operators of many types (Logical, Arithmetic, Assignment, Unary, Splat, Matching, Ternary, Shovel, Triple Equals) are supported in this manner where they are defined as methods in your classes, and can be used with the "method" syntax (e.g. `this_period.==(that_period)`) or with the operator syntax (e.g. `this_period == that_period`).

```ruby
2 > 3
 => false
2.>(3)
 => false

# doesn't work for non-operator methods
2.to_f
 => 2.0
2 to_f
SyntaxError: (irb):14: syntax error, unexpected tIDENTIFIER
```

davidlimcheng: Thanks for the explanation!

PR #7440, opened 2019-01-03
SISRP-41873 Removes FinaidApprovalsController and refactors TermsAndConditionsController & Title4Controller

marcpechaitis:
https://jira.berkeley.edu/browse/SISRP-41873


src/assets/javascripts/angular/controllers/pages/termsAndConditionsController.js
```diff
@@ -2,14 +2,20 @@
...
     isLoading: true
   };
 
+  var sendEvent = function() {
+    $rootScope.$broadcast('calcentral.custom.api.finaid.approvals');
+  };
+
+  $scope.sendResponseTC = function(finaidYearId, response) {
+    $scope.termsAndConditions.isLoading = true;
+    finaidFactory.postTCResponse(finaidYearId, response).then(sendEvent).then($route.reload);
```
davidlimcheng: @marcpechaitis, I think it would actually be better to stick `$route.reload` as a promise within the `$scope.$on` block, after the call to `getTermsAndConditions()`. This means you'll have to "promise-ify" `getTermsAndConditions()` using Angular, as [documented here.](https://docs.angularjs.org/api/ng/service/$q)

Doing this will reload the page only after we're sure the new fetch has been completed, and $scope has been updated. This might actually be the race condition we were looking for.

davidlimcheng: Thinking it through a bit more - I wonder if the `$route.reload` is necessary, since we can fetch the new data values and update  `$scope` without reloading at all.  This may take some trial/error, but I would try it both with and without the `reload()`. 

From the `$route.reload()` [documentation:](https://docs.angularjs.org/api/ngRoute/service/$route#reload)

>Causes $route service to reload the current route even if $location hasn't changed.
As a result of that, ngView creates new scope and reinstantiates the controller.

When the controller is reinstantiated, it will run ` getTermsAndConditions({ finaidYear: $routeParams.finaidYearId });` as specified on line 29. This is missing the `{ refreshCache: true }` option, so it may actually give us stale data. I think this is also another cause of the race condition.

In other words:
1. Upon receiving the `'calcentral.custom.api.finaid.approvals'` broadcast, the controller will refetch `termsAndConditionsFactory.getTermsAndConditions` with the `refreshCache: true`
2. At the same time, the controller is receiving `$route.reload`, which tells is to fetch `termsAndConditionsFactory.getTermsAndConditions` again, but this time without `refreshCache: true`

So, how about we try scrapping `$route.reload()`, and letting the `$scope.$on` block handle all of the data fetching duties and `$scope` updating duties. I think this should actually be enough to get the intended outcome.

marcpechaitis: Thanks for the suggestions @davidlimcheng!  I have removed `$route.reload` and everything works as expected.

src/assets/templates/finaid_tc.html
```diff
@@ -8,7 +8,7 @@ <h1 class="cc-heading-page-title cc-hidden cc-print-show">
         Financial aid for <span data-ng-bind="api.user.profile.fullName"></span>
       </h1>
     </div>
-    <div data-ng-controller="FinaidApprovalsController" class="columns">
+    <div data-ng-controller="TermsAndConditionsController" class="columns">
```
davidlimcheng: @marcpechaitis, I see that this controller is already defined for this template in [routeConfiguration.js](https://github.com/ets-berkeley-edu/calcentral/blob/master/src/assets/javascripts/angular/configuration/routeConfiguration.js#L94), so we actually don't need to define it again here. I believe this will then spin up two controllers:  one for the page and another one within the template, when we only need one controller.

src/assets/javascripts/angular/controllers/pages/termsAndConditionsController.js
```diff
@@ -2,14 +2,20 @@
...
     isLoading: true
   };
 
+  var sendEvent = function() {
+    $rootScope.$broadcast('calcentral.custom.api.finaid.approvals');
+  };
+
+  $scope.sendResponseTC = function(finaidYearId, response) {
+    $scope.termsAndConditions.isLoading = true;
+    finaidFactory.postTCResponse(finaidYearId, response).then(sendEvent);
```
davidlimcheng: @marcpechaitis This one is not a blocker, but it's a bit weird to have the `POST` function live in `finaidFactory` while the `GET` function live in `termsAndConditionsFactory`. I think moving `postTCResponse` to `termsAndConditionsFactory` makes more sense.  What do you think?

davidlimcheng:
@marcpechaitis Regarding the use of `$broadcast`, I see that the only place we are broadcasting the `api.calcentral.custom.api.finaid.approvals` message is within `TermsAndConditionsController` and `Title4Controller`. These are then triggering a refetch of data in **both** controllers. 

My question is: Why does `Title4Controller` need to know about a user accepting Terms and Conditions, and vice versa? 

If there is a need for them to know about the status of the other, I would expect an initial data fetch from `TermsAndConditionsController` for Title4 data, and vice versa. Since there's nothing like that defined, I'd assume that `TermsAndConditionsController` doesn't really care about Title 4 authorization, and `Title4Controller` does not really care about Terms & Conditions.

This can allow us to remove the use of the `$broadcast()` and rely solely on `Promises` for fetching data at the appropriate time. Removing `$broadcast()`s is a significant boost to performances, since it [uses for for() loop to hit every child scope.](https://stackoverflow.com/questions/20245972/are-angular-broadcast-and-on-expensive/25000618) 

In addition, this will allow us to cut down on an unnecessary fetch. When a user accepts their T&C and `$broadcast('calcentral.custom.api.finaid.approvals')` is run, it kicks off a fetch in both `TermsAndConditionsController` **and** `Title4Controller` (if it is already instantiated), even if `Title4Controller` doesn't really need to since only T&C data has been updated.


That being said, I know the functional BSA is currently out and we need to test this for the 1/27 release.  Can you create a 1/27 JIRA to follow-up on that and add it to this PR? Every little optimization we can get is significant!

src/assets/templates/finaid_tc.html
marcpechaitis: Thanks @davidlimcheng, I have made the changes we discussed in Slack.

src/assets/javascripts/angular/controllers/pages/termsAndConditionsController.js
marcpechaitis: Thanks @davidlimcheng, I have made the changes we discussed in Slack.  Note, I made similar changes to refactor the `title4Factory` in the same manner.

davidlimcheng:
Looks great, thanks for the follow-ups @marcpechaitis

PR #7448, opened 2019-01-04
SISRP-39745: Fix visual bug and prevent label from wrapping

developish:
https://jira.berkeley.edu/browse/SISRP-39745

### DIFF
--- a/src/assets/stylesheets/_myfinances.scss
+++ b/src/assets/stylesheets/_myfinances.scss
@@ -62,7 +62,6 @@
     margin-top: 5px;
   }
   .cc-page-myfinances-due-date {
-    display: block;
     .cc-page-myfinances-due-date-icons-medium-up {
       margin-left: 10px;
     }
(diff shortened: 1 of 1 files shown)


src/assets/stylesheets/_myfinances.scss
```diff
@@ -62,7 +62,6 @@
     margin-top: 5px;
   }
   .cc-page-myfinances-due-date {
-    display: block;
```
davidlimcheng: @developish Curious why we're removing this?

developish: It's the cause of the visual bug: the class is applied to a `td` tag, overriding the default of `table-cell` display style, so that if the row gets taller for any reason (anything wrapping, like the finances-type or the description), the bottom border doesn't adjust appropriately on that particular cell.

davidlimcheng: Gotcha, thanks @developish

PR #7455, opened 2019-01-12
SISRP-43941 - 500 error on 'textbooks' proxy when viewing-as instructor

nkutub:
https://jira.berkeley.edu/browse/SISRP-43941


redconfetti:
@nkutub LGTM

PR #7467, opened 2019-01-16
SISRP-44440 - Removes unused link details

redconfetti:
https://jira.berkeley.edu/browse/SISRP-44440

Removes left-over hard coded link details, after CS Link API was implemented for this link in #7454


davidlimcheng:
@redconfetti, Why are we removing the `$scope.taxForm.viewFormLink` property but not the [logic that depends on it?](https://github.com/ets-berkeley-edu/calcentral/blob/master/src/assets/javascripts/angular/controllers/widgets/financesLinksController.js#L151)

redconfetti:
@davidlimcheng I just updated the description above to note that the link details are left over after we replaced them with CS Link API driven link data.

davidlimcheng:
Ahh got it, thanks for clearing that up @redconfetti

PR #7470, opened 2019-01-16
SISRP-44316 - SIR Card Does Not Display for LAW students selecting Maybe

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-44316

For the purposes of cherry-picking commits to prepare the `qa` branch for release, this commit should follow **after** all commits related to [SISRP-42744](https://jira.berkeley.edu/browse/SISRP-42744) (commits documented in JIRA ticket).

### DIFF
--- a/app/models/campus_solutions/sir/sir_statuses.rb
+++ b/app/models/campus_solutions/sir/sir_statuses.rb
@@ -65,7 +65,32 @@ module CampusSolutions
         checklist_items.try(:each) do |checklist_item|
           sir_checklist_items.push(checklist_item) if checklist_item.try(:[], :adminFunc) == 'ADMP' && is_active_offer?(checklist_item)
         end
-        map_sir_configs(sir_checklist_items)
+        filtered_sir_items = filter_duplicate_items(sir_checklist_items)
+        map_sir_configs(filtered_sir_items)
+      end
+
+      def filter_duplicate_items(sir_checklist_items)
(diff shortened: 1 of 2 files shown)


app/models/campus_solutions/sir/sir_statuses.rb
```diff
@@ -65,7 +65,32 @@ def extract_sir_checklist_items(checklist_items)
         checklist_items.try(:each) do |checklist_item|
           sir_checklist_items.push(checklist_item) if checklist_item.try(:[], :adminFunc) == 'ADMP' && is_active_offer?(checklist_item)
         end
-        map_sir_configs(sir_checklist_items)
+        filtered_sir_items = filter_duplicate_items(sir_checklist_items)
+        map_sir_configs(filtered_sir_items)
+      end
+
+      def filter_duplicate_items(sir_checklist_items)
+        unique_sir_checklist_items = sir_checklist_items.uniq { |item| item[:checkListMgmtAdmp][:admApplNbr] }
```
redconfetti: @davidlimcheng - I'm curious if you could have made this more efficient by iterating over the array once and grouping the elements in a hash using the `:admApplNbr` value as the key. During that loop you could detect if there are duplicates present. You could loop over it once more to collect only the ones with the highest sequence number.

Are we sure there will be only two duplicates at most? Or might there be 3-4 sometimes?

redconfetti: Discussed with @davidlimcheng over Slack and concluded that the way it's implemented is best

PR #7475, opened 2019-01-22
SISRP-41162 FA API Refactor #2 - Finaid Profile

marcpechaitis:
https://jira.berkeley.edu/browse/SISRP-41162


app/models/edo_oracle/financial_aid/queries.rb
```diff
@@ -241,6 +241,93 @@ def self.get_terms_and_conditions(person_id, aid_year)
         SQL
         result.first
       end
+
+      def self.get_finaid_profile_status(person_id, aid_year)
+        result = safe_query <<-SQL
+        SELECT
+          UC.AID_YEAR             AS AID_YEAR,
```
redconfetti: @marcpechaitis - Your queries are too aesthetically pleasing.

app/controllers/my_finaid_profile_controller.rb
```diff
@@ -0,0 +1,10 @@
+class MyFinaidProfileController < ApplicationController
+  include AllowDelegateViewAs
+  before_filter :api_authenticate_401
+  before_filter :authorize_for_financial
+
+  def get_feed
+    options = params.permit :aid_year
```
davidlimcheng: @marcpechaitis , Any reason behind using `permit` instead of `require`? I presume because in `my_finaid_profile`, aid year can also be grabbed from another model?
```
@my_aid_year ||= (@options[:aid_year] || FinancialAid::MyAidYears.new(@uid).default_aid_year).to_i.to_s
```
But in this case, isn't it possible that `FinancialAid::MyAidYears#default_aid_year` can still return `nil`?

app/models/financial_aid/my_finaid_profile.rb
```diff
@@ -0,0 +1,224 @@
...
+        }
+      end
+    end
+
+    def itemGroupsProfile
+      @itemGroupsProfile ||= [
+        [
+          {
+            title: 'Academic Career',
+            value: status['acad_career_descr']
```
davidlimcheng: @marcpechaitis , the logic within `success?`:
```
    def success?
      (status || level || enrollment || residency || isir) && (title4 && terms_and_conditions)
    end
```
To me this infers that any one of `status`, `level`, `enrollment`, `residency`, or `isir` can return something like `nil` or `false`. If that is the case, we'll want to use a `try` here, so Rails won't throw us an error when it tries to access `nil['acad_career_descr']`.

davidlimcheng: This also applies to anywhere else relying on any of those values.

app/models/financial_aid/my_finaid_profile.rb
```diff
@@ -0,0 +1,224 @@
...
+        ]
+      ]
+    end
+
+    def categories
+      @categories ||= [
+        {
+          title: status['title'],
+          itemGroups: itemGroupsProfile
+        },{
```
davidlimcheng: Missing a line break here

spec/models/edo_oracle/financial_aid/queries_spec.rb
```diff
@@ -172,4 +172,74 @@
...
 
+  describe '#get_finaid_profile_status' do
+    subject { described_class.get_finaid_profile_status(uid, aid_year) }
+    let(:uid) { 61889 }
+    let(:aid_year) { 2018 }
+
+    it_behaves_like 'a successful query that returns one result'
+
+    it 'returns the expected result' do
+      puts subject
```
davidlimcheng: renegade `puts`

spec/models/edo_oracle/financial_aid/queries_spec.rb
```diff
@@ -172,4 +172,74 @@
...
+      expect(subject.count).to eq 12
+      expect(subject).to have_keys(%w(aid_year acad_career_descr exp_grad_term sap_status verification_status award_status acad_holds candidacy filing_fee berkeley_pc title message))
+    end
+  end
+
+  describe '#get_finaid_profile_acad_level' do
+    subject { described_class.get_finaid_profile_acad_level(uid, aid_year) }
+    let(:uid) { 61889 }
```
davidlimcheng: @marcpechaitis Not a blocker for merge, but since basically this whole spec uses UID 61889 and SID 11667051, we can just define them once at the root-level, and they don't need to be redefined at every level. Again not a blocker/requirement, it's just a bit cleaner if you have the bandwidth for it.

spec/models/edo_oracle/financial_aid/queries_spec.rb
```diff
@@ -172,4 +172,74 @@
...
+
+  describe '#get_finaid_profile_isir' do
+    subject { described_class.get_finaid_profile_isir(uid, aid_year) }
+    let(:uid) { 61889 }
+    let(:aid_year) { 2018 }
+
+    it_behaves_like 'a successful query that returns one result'
+
+    it 'returns the expected result' do
+      puts subject
```
davidlimcheng: renegade `puts`

spec/models/financial_aid/my_finaid_profile_spec.rb
```diff
@@ -0,0 +1,156 @@
+describe FinancialAid::MyFinaidProfile do
+  before do
+    allow_any_instance_of(FinancialAid::MyAidYears).to receive(:default_aid_year).and_return '2019'
+  end
+
+  describe '#get_feed' do
+    subject { described_class.new(uid, {aid_year: aid_year}).get_feed }
+    let(:uid) { 61889 }
```
davidlimcheng: @marcpechaitis Same thing here, you can define `uid` on the outer-scoped block and it should apply to every `context` block.

app/controllers/my_finaid_profile_controller.rb
redconfetti: @davidlimcheng - I'm reviewing this PR with @marcpechaitis over a Google hangout. Marc indicated that the API end-point shouldn't require the aid year, it should use the default one when not present. So `#permit` seems to be the proper configuration.

It very much could [return a nil value](https://github.com/ets-berkeley-edu/calcentral/blob/f6243dac/app/models/financial_aid/my_aid_years.rb#L24), but that's certainly an exception. We only query this API for students that have some sort of financial aid data.

app/models/financial_aid/my_finaid_profile.rb
marcpechaitis: Thanks @davidlimcheng I have fixed this wherever I was using it.

spec/models/financial_aid/my_finaid_profile_spec.rb
marcpechaitis: Thanks @davidlimcheng I have made the changes you suggested.

spec/models/edo_oracle/financial_aid/queries_spec.rb
marcpechaitis: Thanks @davidlimcheng I have removed the renegade!

spec/models/edo_oracle/financial_aid/queries_spec.rb
marcpechaitis: Thanks @davidlimcheng I have removed the renegade!

app/models/financial_aid/my_finaid_profile.rb
marcpechaitis: Thanks @davidlimcheng I have implemented your suggestion.

spec/models/edo_oracle/financial_aid/queries_spec.rb
marcpechaitis: Thanks @davidlimcheng I have implemented your suggestion.

app/models/financial_aid/my_finaid_profile.rb
```diff
@@ -0,0 +1,229 @@
...
+    def title4
+      @title4 ||= EdoOracle::FinancialAid::Queries.get_title4(@uid)
+    end
+
+    def terms_and_conditions
+      @terms_and_conditions ||= EdoOracle::FinancialAid::Queries.get_terms_and_conditions(@uid, my_aid_year)
+    end
+
+    def success?
+      (status || level || enrollment || residency || isir) && (title4 && terms_and_conditions)
```
davidlimcheng: @marcpechaitis @redconfetti  Since `my_aid_year` is a dependency on most of these, can we add a check here in the case `my_aid_year == nil`?  I like that we're using `success?` as a check in `finaid_profile`:
```
    def finaid_profile
      return nil unless success?
```
But, if we know that `my_aid_year` is nil, I don't think we should kick off all of these EDO queries, since we know they're going to fail without that input parameter. 

So, my suggestion is to add a nil check for `my_aid_year` before doing anything else in `success?`. This way, we can stop the query chain from kicking off if we already know it's doomed to fail without an `aid_year`.  Does that make sense?

app/controllers/my_finaid_profile_controller.rb
davidlimcheng: @marcpechaitis @redconfetti Thanks guys, I added a follow-up comment in `MyFinaidProfile#success?`.

PR #7482, opened 2019-01-31
SISRP-44405 - Link to Campus Logic from CalCentral

redconfetti:
https://jira.berkeley.edu/browse/SISRP-44405

### DIFF
--- a/src/assets/javascripts/angular/controllers/widgets/finaid/finaidCommunicationsController.js
+++ b/src/assets/javascripts/angular/controllers/widgets/finaid/finaidCommunicationsController.js
@@ -5,8 +5,9 @@ var _ = require('lodash');
 /**
  * Finaid Communications controller
  */
-angular.module('calcentral.controllers').controller('FinaidCommunicationsController', function($q, $scope, activityFactory, finaidFactory, finaidService, linkService, tasksFactory, tasksService) {
+angular.module('calcentral.controllers').controller('FinaidCommunicationsController', function($q, $scope, activityFactory, csLinkFactory, finaidFactory, finaidService, linkService, tasksFactory, tasksService) {
   $scope.communicationsInfo = {
+    csLinks: {},
     aidYear: '',
     isLoading: true,
     counts: {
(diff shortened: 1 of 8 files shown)


app/controllers/my_campus_links_controller.rb
```diff
@@ -1,6 +1,5 @@
 class MyCampusLinksController < ApplicationController
   extend Cache::Cacheable
-  include AllowDelegateViewAs
```
redconfetti: I noticed that this controller doesn't use any of the methods in this module. This is just a clean-up modification.

src/assets/javascripts/angular/controllers/widgets/finaid/finaidCommunicationsController.js
```diff
@@ -64,12 +65,31 @@ angular.module('calcentral.controllers').controller('FinaidCommunicationsControl
     });
   };
 
-  var getFinaidYearInfo = function(options) {
-    return finaidFactory.getFinaidYearInfo({
-      finaidYearId: options.finaidYearId
+  var getVerificationAndAppealsLink = csLinkFactory.getLink({
```
redconfetti: I don't like implementing additional requests from the front-end just to get these links, but this is a feature that was given to me last minute in the sprint that they needed with high priority.

I created [SISRP-44764](https://jira.berkeley.edu/browse/SISRP-44764) as a task to refactor this card so that it uses a single API end-point to source all this data.

In fact, I started to do that work until I noticed that I didn't have enough time to do it this release, hence the controller update above, and also the patch file in SISRP-44764

redconfetti: I don't like how these controllers are how bloated further, making more API calls to render the card. This was a must-have deliverable for this release however.

I have created two JIRAS ([SISRP-44764](https://jira.berkeley.edu/browse/SISRP-44764) and [SISRP-44811](https://jira.berkeley.edu/browse/SISRP-44811)) to ensure that clean-up of these cards is performed at some point in the future.

app/controllers/my_campus_links_controller.rb
davidlimcheng: @redconfetti, I think we need to keep this in, as it has a [method](https://github.com/ets-berkeley-edu/calcentral/blob/master/app/controllers/concerns/allow_delegate_view_as.rb#L3) `allow_if_delegate_view_as?` that overwrites a boolean [defined in](https://github.com/ets-berkeley-edu/calcentral/blob/master/app/controllers/application_controller.rb#L59) `ApplicationController`. This boolean is subsequently used in `ApplicationController#deny_if_filtered`.

redconfetti: Aah. Good catch @davidlimcheng . I'll bring it back.

davidlimcheng:
LGTM

PR #7485, opened 2019-02-04
SISRP-41162 FA Profile API Refactor - followup

marcpechaitis:
https://jira.berkeley.edu/browse/SISRP-41162


marcpechaitis:
Many thanks to the team for the help to get this one over the line!

app/models/financial_aid/my_finaid_profile.rb
```diff
@@ -62,7 +62,7 @@ def terms_and_conditions
     end
 
     def success?
-      my_aid_year.present? ? (status || level || enrollment || residency || isir) && (title4 && terms_and_conditions) : false
+      my_aid_year.present? && !status.empty? ? true : false
```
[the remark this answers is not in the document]
marcpechaitis: @redconfetti `status` is an array, and `.present?` is not a valid method for [Array](https://ruby-doc.org/core-2.2.2/Array.html)

src/assets/templates/widgets/finaid_profile.html
```diff
@@ -1,60 +1,116 @@
-<div class="cc-widget cc-widget-finaid-profile" data-ng-controller="FinaidProfileController">
+<div
+  class="cc-widget cc-widget-finaid-profile"
+  data-ng-controller="FinaidProfileController"
+>
```
redconfetti: @marcpechaitis - What's character width do you use as a cut-off to decide to format a tag like this? 80 characters?

I don't like this format, but I'm fine with it to make it easier to read. It didn't seem that necessary here.

marcpechaitis: Its what [Prettier](https://prettier.io/) recommends.

src/assets/templates/widgets/finaid_profile.html
```diff
@@ -1,60 +1,116 @@
...
+          <div class="cc-left">
+            <strong>Title IV</strong>
+            <a
+              ng-if="api.user.profile.actAsOptions.canSeeCSLinks"
+              href="/finances/finaid/t4/auth"
+            >
+              Update
+              <span class="cc-visuallyhidden"
+              >Update Title IV Authorization</span
+              >
```
redconfetti: Yeah. I don't see how this helps the readability.

marcpechaitis:
d'oh I forgot to update some specs to account for my API changes

I'll be pushing those up soon...

app/models/financial_aid/my_finaid_profile.rb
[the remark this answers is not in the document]
marcpechaitis: @redconfetti it turns out that `!status.empty?` won't work (if `status` is undefined)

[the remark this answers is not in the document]
redconfetti: @marcpechaitis - Really? If something is undefined it's `nil`, which is an object in itself that also usually includes methods available to other types. The opposite of `#present?` is `#blank?`.

Keep in mind that if you're referencing pure Ruby documentation, you're not going to see `#present?` or `#blank?` because they're added to the Object class by the Rails library called ActiveSupport.

See [Object#present?](https://devdocs.io/rails~4.2/object#method-i-present-3F)

```ruby
[1,2,3].present?
 => true
[].present?
 => false
nil.present?
 => false
'dsfd'.present?
 => true
{}.present?
 => false
{something: nil}.present?
 => true
```

src/assets/templates/widgets/finaid_profile.html
redconfetti: @marcpechaitis - Okay. I'm not going to block you on this, but I highly recommend that we implement a dotfile of some sort so that all of us share the same config (instead of switching the formatting back and forth).

@davidlimcheng - What do you say we implement a [.prettierrc.yaml](https://prettier.io/docs/en/configuration.html) file? @nkutub @developish, comments?

marcpechaitis: @redconfetti, to be clear we are not formally using Prettier (yet?) -- the impact on the codebase if/when we decide to implement it will be huge.... I simply used it as a standalone tool to help me figure out my extra `</div>` tag and I left it in that format.

I can un-prettify it manually if you want?

redconfetti:
@marcpechaitis - Let me know when the specs are present and I'll re-review. 👍

app/models/financial_aid/my_finaid_profile.rb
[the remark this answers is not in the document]
marcpechaitis: Thanks @redconfetti , I didn't realize that about ActiveSupport.

`.present?` works for this case so I have just adjusted the code to use that.

src/assets/templates/widgets/finaid_profile.html
redconfetti: @marcpechaitis - We can move forward with it. I want to discuss this further as a team though.

marcpechaitis:
> @marcpechaitis - Let me know when the specs are present and I'll re-review. 👍

@redconfetti I have updated the specs and they are ready for re-review.

redconfetti:
Looks good @marcpechaitis

PR #7487, opened 2019-02-05
NOJIRA Fix errant bitwise OR

raydavis:
I spotted this long-standing typo while working on something else, but now that we know about it we might as well fix it.


raydavis:
Whoops, not a typo and not a bitwise-OR, just an array-union.

PR #7489, opened 2019-02-05
SISRP-41853 - CalCentral P/NP Limit Calculator - UX Dev

davidlimcheng:
https://jira.berkeley.edu/browse/SISRP-41853

Also introduces a bunch of changes to the React stack:
- Presentational components are all now functional components (as opposed to classes) (thanks @developish!)
- Starts a CSS repository in `src/react/stylesheets`
  - These follow the [Block Element Modifier](http://getbem.com/naming/) naming approach, which @developish suggested. I think this is a great idea and we kinda/sorta do it already with our custom class names, but IMO this is a cleaner approach. Happy to discuss
- Changes some details on how the `Widget` component works, but it's mostly the same idea
- Removes some components, since we are creating a separate CSS repo for the React stack, I don't think we need components like `BulletList` since `lists.scss` should be self-explanatory, and hopefully we'll be able to maintain our stylesheets to be straightforward enough.

Thanks again Brandon for the great suggestions! 

I will schedule a meeting sometime this week to review these changes and how they will affect development moving forward.


app/models/campus_solutions/pnp_calculator/calculator_values.rb
```diff
@@ -0,0 +1,25 @@
+module CampusSolutions
+  module PnpCalculator
+    class CalculatorValues < UserSpecificModel
+      include Cache::CachedFeed
+      include Cache::UserCacheExpiry
+      include Cache::RelatedCacheKeyTracker
```
redconfetti: @davidlimcheng - Why are we including Cache::RelatedCacheKeyTracker? Isn't that only needed when the cache key includes some other variable in addition to the UID? Such as when data specific to a term is being cached?

app/models/campus_solutions/pnp_calculator/calculator_values.rb
```diff
@@ -0,0 +1,25 @@
...
+      include User::Identifiers
+
+      def get_feed_internal
+        values
+      end
+
+      def campus_solutions_id
+        @campus_solutions_id ||= lookup_campus_solutions_id(@uid)
```
redconfetti: @davidlimcheng - Why can't we simply just call `User::Identifiers.lookup_campus_solutions_id` directly, so that it's not ambiguous which methods are being used by this class?

redconfetti:
Merging now, @developish will follow up with @davidlimcheng next week to properly review coding approach.

PR #7492, opened 2019-02-06
SISRP-44520 - Terms not appearing in Student Success card for Advisor

nkutub:
https://jira.berkeley.edu/browse/SISRP-44520

### DIFF
--- a/app/models/student_success/term_gpa.rb
+++ b/app/models/student_success/term_gpa.rb
@@ -9,7 +9,13 @@ module StudentSuccess
     end
 
     def merge(data={})
-      data[:termGpa] = get_term_gpas
+      term_gpas = get_term_gpas
+      data[:termGpaWithZero] = term_gpas.clone
+      data[:termGpa] = remove_zero_gpas(term_gpas)
+    end
+
+    def remove_zero_gpas(term_gpas)
+      term_gpas.delete_if {|term| term[:termGpaUnits].to_i == 0}
     end
 
     def get_term_gpas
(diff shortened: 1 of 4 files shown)


davidlimcheng:
@nkutub, looking at the JIRA comments, it seems like there might have been design considerations for the issue.  Was there more communication done offline for this solution?

app/models/student_success/term_gpa.rb
```diff
@@ -9,7 +9,13 @@ def initialize(opts={})
     end
 
     def merge(data={})
-      data[:termGpa] = get_term_gpas
+      term_gpas = get_term_gpas
+      data[:termGpaWithZero] = get_term_gpas.clone
```
davidlimcheng: @nkutub I think this should be `term_gpas.clone`

app/models/student_success/term_gpa.rb
```diff
@@ -9,7 +9,13 @@ def initialize(opts={})
...
+      data[:termGpa] = remove_zero_gpas(term_gpas)
```
davidlimcheng: If this is no longer being used anywhere, can we just get rid of it?

nkutub: @davidlimcheng I don't understand ? both are being used

app/models/student_success/term_gpa.rb
nkutub: good catch

nkutub:
> @nkutub, looking at the JIRA comments, it seems like there might have been design considerations for the issue. Was there more communication done offline for this solution?

@davidlimcheng not sure what you are referring to

app/models/student_success/term_gpa.rb
davidlimcheng: I see now that it is still being used in `chartGpaTrend()`.

davidlimcheng:
@nkutub I'm just asking if for clarification on the fix for this.  Seems like there was an open-ended discussion on the JIRA ticket, but you must have communicated with the designer/BSA outside of the JIRA ticket.

PR #7499, opened 2019-02-07
SISRP-42206 - Display Current Term Appointment Data in CalCentral

developish:
https://jira.berkeley.edu/browse/SISRP-42206

### DIFF
--- a/app/models/edo_oracle/employment_appointment_queries.rb
+++ b/app/models/edo_oracle/employment_appointment_queries.rb
@@ -1,0 +1,37 @@
+module EdoOracle
+  class EmploymentAppointmentQueries < Connection
+    include ActiveRecordHelper
+    include ClassLogger
+    include Concerns::QueryHelper
+  
+    def self.get_terms_taught(uid)
+      result = safe_query <<-SQL
+        SELECT UC_TERMS_TAUGHT as terms_count
+        FROM SYSADM.PS_UCC_GRTRMTAUGHT
+        WHERE CAMPUS_ID = '#{uid}'
+      SQL
+  
+      if result.empty?
+        0
+      else
+        result.first['terms_count'].to_i
+      end
+    end
+
+    def self.get_appointments(uid)
+      safe_query <<-SQL
+        SELECT
+          DESCR as description,
+          DISTRIB_BEGIN_DT as start_date,
+          DISTRIB_END_DT as end_date,
+          DIST_PCT as distribution_percentage,
+          COMPRATE as compenstation,
+          UC_L4_DESCR as unit,
+          JOBCODE as job_code,
+          STEP
+        FROM SYSADM.PS_UCC_GRCURAPTVW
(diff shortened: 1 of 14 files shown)


redconfetti:
More reading suggested by Brandon:

* [Class vs Stateless Components](https://itnext.io/react-component-class-vs-stateless-component-e3797c7d23ab)
* [Spread operator (ES6)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax)

src/react/components/EmploymentAppointments/AppointmentRow.js
```diff
@@ -0,0 +1,76 @@
...
+            <tbody>
+              <tr>
+                <td>{ this.props.unit }</td>
+                <td>{ parseInt(this.props.step) }</td>
+                <td className="alignRight">${ parseFloat(this.props.compenstation).toFixed(2) }</td>
+                <td className="alignRight">{ parseFloat(this.props.distributionPercentage).toFixed(2) }%</td>
+              </tr>
+            </tbody>
+          </table>
+        </div> : <React.Fragment></React.Fragment>
```
redconfetti: @developish - You can do this with the same syntax above using the `&&` instead of the ternary. I'm so smart. You totally didn't help me discover this.

app/models/edo_oracle/employment_appointment_queries.rb
```diff
@@ -0,0 +1,37 @@
+module EdoOracle
```
marcpechaitis: Are we now going to start splitting up the `xyz_queries.rb` files more atomically?

I like it, but it does not conform to our current structure.

src/react/components/EmploymentAppointments.js
```diff
@@ -0,0 +1,138 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { react2angular } from 'react2angular';
+import WidgetHeader from './base/widget/WidgetHeader';
+import AppointmentRow from './EmploymentAppointments/AppointmentRow';
+import InstructingAppointments from './EmploymentAppointments/InstructingAppointments';
+
+const APILink = (props) => {
```
marcpechaitis: @developish I have concerns about the a11y of these links?  Will they provide the usual accessibility features we should use?  (For example: hidden text that causes a screen reader to announce "this link will open in a new window")

src/react/components/EmploymentAppointments.js
```diff
@@ -0,0 +1,138 @@
...
+            { this.state.termsTaught > 0 &&
+              <p>
+                <strong>{this.state.termsTaught}</strong> Terms Teaching
+                (past, current, and future)
+              </p>
+            }
+  
+            <h3>Appointments <span>(current and future)</span></h3>
+  
+            { this.state.appointments.length
```
marcpechaitis: I love this!

PR #7502, opened 2019-02-08
CLC-6979 Define queries to collect applicant data

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6979

Another day, another database....


johncrossman:
Enter EDW.

PR #7505, opened 2019-02-11
SISRP-42529 - Show where the transfer units are added into the semester totals

developish:


spec/controllers/advising_student_controller_spec.rb
```diff
@@ -227,4 +227,30 @@
...
+
+      context 'when not viewing a student' do
+        it_behaves_like 'an endpoint refusing a request'
+      end
+    end
+
+    context 'when advisor authorized' do
+      let(:student) { true }
+
+      context 'when not viewing a student' do
```
davidlimcheng: @developish The context for this test is the same as the one before it - but different behavior is expected. Typo?

src/react/components/AcademicSummary/AcademicSummary.js
```diff
@@ -0,0 +1,130 @@
...
+  Promise.all([status, profile, academics, transferCredit]).then((responses) => {
+    resolve(responses);
+  });
+});
+
+class AcademicSummary extends Component {
+  constructor(props) {
+    super(props);
+
+    fetchAcademics.then((responses) => {
```
davidlimcheng: @developish, is there an advantage to doing this in the `constructor` rather than in something like `componentDidMount`? 

I know that initializing the state should always take way less time than `fetchAcademics`, but putting the data fetch in `componentDidMount` should remove any possibility of a race condition, right?

src/react/components/AcademicSummary/AcademicSummary.js
```diff
@@ -0,0 +1,130 @@
...
+      academicsPermitted: false
+    };
+  }
+
+  printPage() {
+    window.print();
+  }
+
+  showEnrollment() {
+    return this.showSemesters() || this.state.showTransferCredit;
```
davidlimcheng: @developish, I'm curious how this knows what `this.showSemesters()` is when `showSemesters` was never bound to `self`? No need to answer if I'm way off-base here, but I thought that class methods had to be bound to the class (e.g. `this.showSemesters = this.showSemesters.bind(this)` due to scope issues in React.

Aside from my curiosity, I do think it'd be a bit cleaner to remove `showSemesters()` altogether and just have this method hold all the logic. Just IMO and not a blocker for merge.

src/react/components/AcademicSummary/AcademicSummary.js
developish: No, there's not really an advantage. I was avoiding linter warnings that came up because I was using `componentWillMount`. `componentDidMount` is better.

src/react/components/AcademicSummary/AcademicSummary.js
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  showSemesters() {
+    return this.state.academics.semesters.length && this.state.user.hasStudentHistory;
+  }
+
+  render() {
+    if (this.state.isPermitted) {
+      return (
+        <div className="cc-page-academics">
```
davidlimcheng: @developish, do you mind creating a JIRA to refactor the CSS to the react-specific classes in the future?

src/react/components/AcademicSummary/AcademicSummary.js
```diff
@@ -0,0 +1,130 @@
...
+              <a href="/academics">My Academics</a> &raquo; Academic Summary
+            </h1>
+          </div>
+          <div className="row">
+            <div className="medium-10 medium-offset-1 column">
+              <div className="cc-widget cc-academic-summary">
+                <div className="cc-widget-title">
+                  <h2 className="cc-left">Academic Summary</h2>
+                  <button className="cc-button cc-button-blue cc-right cc-widget-title-button" onClick={this.printPage}>
+                    <FontAwesomeIcon name="print" /> Print
```
davidlimcheng: Nice, I made components for each Icon but I think this is much cleaner.

src/react/components/AcademicSummary/AcademicSummary.js
```diff
@@ -0,0 +1,130 @@
...
+class AcademicSummary extends Component {
+  constructor(props) {
+    super(props);
+
+    fetchAcademics.then((responses) => {
+      responses.forEach(response => {
+        this.setState(response);
+      });
+
+      this.setState({ loaded: true });
```
davidlimcheng: @developish  I've been using `immutability-helper` to update the state in every case, but I don't think that was the right approach. Using it only when necessary (basically within `shouldComponentUpdate`) sounds cleaner. Not really picking out anything here, just thinking out loud.

src/react/components/AcademicSummary/AcademicSummary.js
developish: The binding issue has to do with events and making `this` work specifically in a callback. It's not necessary here as we're not dealing with any events.

Good call on `showSemesters`. It was relevant before the semesters components were extracted.

src/react/components/AcademicSummary/AcademicSummary.js
developish: Gladly

src/react/components/AcademicSummary/Enrollment.js
```diff
@@ -0,0 +1,61 @@
+import React, { Component, Fragment } from 'react';
+import PropTypes from 'prop-types';
+
+import HubTermLegacyNote from './HubTermLegacyNote';
+import TransferCredit from '../TransferCredit/TransferCredit';
+import Semester from './Semester';
+import SemestersSummary from './SemestersSummary';
+
+class Enrollment extends Component {
```
davidlimcheng: @developish  Looks like this could be changed to a functional component

src/react/components/AcademicSummary/Enrollment.js
```diff
@@ -0,0 +1,61 @@
...
+            semesters={this.props.semesters}
+            totalLawUnits={this.props.gpaUnits.totalLawUnits}
+            totalUnits={this.props.gpaUnits.totalUnits}
+          />
+        }
+      </Fragment>
+    );
+  }
+
+  showSemesters() {
```
davidlimcheng: @developish  I'm not against alphabetizing method names within classes, but I do think `render` should always be last.  What do you think? 

cc @marcpechaitis @redconfetti

marcpechaitis: I agree with @davidlimcheng

src/react/components/AcademicSummary/Semester.js
```diff
@@ -0,0 +1,185 @@
...
+        <div className="summary-semester__header" style={{ overflow: 'hidden' }}>
+          {this.enrolledAndNotWithdrawn() &&
+            <h4>
+              <a className="cc-left" href={`/academics/semester/${this.props.slug}`}>
+                {this.props.name}
+              </a>
+            </h4>
+          }
+
+          {(this.withdrawn() || !this.enrolled() || this.notEnrolledHasStudy()) &&
```
davidlimcheng: Really digging this pattern!

src/react/components/AcademicSummary/Semester.js
```diff
@@ -0,0 +1,185 @@
...
+import PropTypes from 'prop-types';
+import PrimarySection from './PrimarySection';
+
+import './Semester.scss';
+
+class Semester extends Component {
+  render() {
+    return (
+      <div className="summary-semester">
+        <div className="summary-semester__header" style={{ overflow: 'hidden' }}>
```
davidlimcheng: @developish I can't seem to find where these element classnames are defined - I expected them to be in `Semester.scss`, but there's not much there. Can you shed some light?

src/react/components/AcademicSummary/Semester.js
```diff
@@ -0,0 +1,185 @@
...
+                  <th>Title</th>
+                  <th className="cc-table-right cc-academic-summary-table-units">Un.</th>
+
+                  {this.props.totalLawUnits && 
+                    <th className="cc-table-right cc-academic-summary-table-units">Law Un.</th>
+                  }
+
+                  <th>Gr.</th>
+                  <th>
+                    {this.showPoints() && <Fragment>Pts.</Fragment>}
```
davidlimcheng: @developish I'm not 100% sure about the design of this, but if `showPoints() === false`, there will be nothing here, but the column will still exist I believe?  Is that the intention? Just making sure that the intended behavior is not to dynamically remove the column if points should not be shown.

src/react/components/AcademicSummary/ValueOrDash.js
```diff
@@ -0,0 +1,14 @@
+import React, { Fragment } from 'react';
+import PropTypes from 'prop-types';
+
+const ValueOrDash = (props) => (
+  props.value
+    ? props.value
+    : <Fragment>&mdash;</Fragment>
```
davidlimcheng: @developish Bit nitpick-ish but would it be safe to wrap `props.value` around `<Fragment>` as well?  Something like 

```
value = props.value ? props.value : `&mdash;`
return <Fragment>value</Fragment>
```

src/react/components/AcademicSummary/Enrollment.js
developish: I think we should discuss. Alphabetical is just coincidental here, I think render should actually be at the top because that's what you need to see to understand what the class does. Methods below are implementation details.

src/react/components/AcademicSummary/ValueOrDash.js
```diff
@@ -0,0 +1,14 @@
...
+);
+
+ValueOrDash.propTypes = {
+  value: PropTypes.string
```
davidlimcheng: @developish, might be a good idea to force this to be a string **or** a number, and make it required
```
value: PropTypes.oneOfType([
  PropTypes.string,
  PropTypes.number
]).isRequired
```

src/react/components/FontAwesomeIcon.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const FontAwesomeIcon = (props) => (
```
davidlimcheng: @developish, what do you think about using an object to do a lookup on the icon name? That way, we'll have a place where we can see all of the icons CalCentral commonly uses, and if there was a typo entering the name and an icon doesn't exist, we'll have error-handling to tell us so.

src/react/components/AcademicSummary/Semester.js
developish: Yes, that's the intention and what the current implementation does.

src/react/components/AcademicSummary/ValueOrDash.js
developish: with `return <Fragment>{value}</Fragment>`

It would be safe, but wouldn't add anything, while obscuring the reason for the `Fragment`.

I'd personally prefer my future self to have one logical statement to read with an implicit return instead of adding the complexity of reading through assignment, return, braces within JSX and back-ticks

src/react/components/AcademicSummary/ValueOrDash.js
developish: I don't think so: this is a case where the API can return null and it actually means something, then this component formats the UI for that null value. I see your point, but I think making that happen would require modifying the data for presentational purposes far away from the actual point of presentation (backend API or a data processing step), which I would try to avoid if possible.

src/react/components/AcademicSummary/Enrollment.js
marcpechaitis: I've only seen `render()` at the end of the class and assumed its convention / best practice?

There is even an eslint rule to enforce this:
https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/sort-comp.md

Discussions:
https://github.com/airbnb/javascript/tree/master/react#ordering
https://github.com/airbnb/javascript/issues/912#issuecomment-225606567

src/react/components/StudentProfile/Degrees.js
```diff
@@ -0,0 +1,71 @@
...
+    <tr>
+      <th>
+        {degrees.length === 1 ? 'Degree' : 'Degrees'} Conferred
+      </th>
+      <td>
+        {degrees.map((degree, index) => (
+          <div className="Degree" key={index}>
+            <div className="Degree__icon-container">
+              {degree.majors[0].type === 'CRT'
+                ? <i className="cc-icon cc-icon-blue cc-icon-ribbon"></i>
```
davidlimcheng: Can these leverage the `<FontAwesomeIcon>` component?

src/react/components/FontAwesomeIcon.js
developish: I didn't think it was worth it for this case, but yes. I have an idea for that which isn't an object, but could use named imports like:

`import FontAwesomeIco, { FA_PRINT } from './FontAwesomeIcon'`

`<FontAwesomeIcon name={FA_PRINT} />`

So it wouldn't even compile without a defined icon type. I'm curious to see the implementation you had in mind

src/react/components/StudentProfile/Degrees.js
```diff
@@ -0,0 +1,71 @@
...
+  const honors = (degree) => (degree.honors && degree.honors.honors) || [];
+
+  return ( 
+    <tr>
+      <th>
+        {degrees.length === 1 ? 'Degree' : 'Degrees'} Conferred
+      </th>
+      <td>
+        {degrees.map((degree, index) => (
+          <div className="Degree" key={index}>
```
davidlimcheng: @developish Any reason why this class is capitalized?

src/react/components/StudentProfile/Degrees.js
```diff
@@ -0,0 +1,71 @@
...
+            <div className="Degree__icon-container">
+              {degree.majors[0].type === 'CRT'
+                ? <i className="cc-icon cc-icon-blue cc-icon-ribbon"></i>
+                : <i className="cc-icon cc-icon-blue fa fa-graduation-cap"></i>
+              }
+            </div>
+            <div className="Degree__body">
+              <div className="Degree__description">
+                {description(degree)} <span>in</span> {degreeMajors(degree)}
```
davidlimcheng: Very cool, I didn't know you can mix&match JSX fragments and variables like this.  Another, possibly cleaner approach is to use template literals:
```
`${description(degree)} in ${degreeMajors(degree)}`
```

src/react/components/TransferCredit/ExamRow.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const ExamRow = ({ name, value }) => {
```
davidlimcheng: Instead of using destructuring here, would it be safer to just use `props`, especially since these props are not required?

src/react/components/StudentProfile/Degrees.js
developish: Yes, trying on a CSS pattern I've seen in create-react-app. Styling the components as the "block" and matching naming. It's low effort to change in the future when we fully standardize

src/react/components/StudentProfile/Degrees.js
developish: Not yet, but I'm all for it once we get there (also, removing our need to care about where an icon is from in order to use it)

src/react/components/AcademicSummary/ValueOrDash.js
davidlimcheng: Understood - I do think it'd be a good idea to allow a number here as well just for flexibility, but we can always add that in the future when the need arises.

src/react/components/AcademicSummary/AcademicSummary.js
davidlimcheng: Wow, sounds like I have a fair amount of refactoring to do!

src/react/components/AcademicSummary/PrimarySection.js
```diff
@@ -0,0 +1,71 @@
+import React, { Component, Fragment } from 'react';
+import PropTypes from 'prop-types';
+import ValueOrDash from './ValueOrDash';
+
+class PrimarySectionRow extends Component {
```
davidlimcheng: Looks like this could also be a functional component

src/react/components/TransferCredit/ExamRow.js
developish: No, it wouldn't make a difference at all

src/react/components/AcademicSummary/ValueOrDash.js
developish: Agreed

spec/controllers/advising_student_controller_spec.rb
```diff
@@ -227,4 +227,30 @@
...
+      let(:session_user_is_advisor) { true }
+
+      context 'when not viewing a student' do
+        it_behaves_like 'an endpoint refusing a request'
+      end
+
+      context 'when viewing a student' do
+        let(:student) { true }
+  
+        context 'when not viewing a student' do
```
davidlimcheng: @developish I think this block should be removed, and the test itself nested within `when viewing a student`.

src/react/components/APILink.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import qs from 'querystringify';
+
+const propTypes = {
```
davidlimcheng: I like that these are defined on top.  +1

src/react/components/APILink.js
```diff
@@ -0,0 +1,34 @@
...
+  name: PropTypes.string,
+  title: PropTypes.string,
+  ucFrom: PropTypes.string,
+  ucFromLink: PropTypes.string,
+  ucFromText: PropTypes.string,
+  url: PropTypes.string,
+  urlId: PropTypes.string
+};
+
+const APILink = ({ title, name, url: url, ucFrom, ucFromText }) => {
```
davidlimcheng: the default value for `url` should be removed here

src/react/components/AcademicSummary/AcademicSummary.js
davidlimcheng: @developish Can this actually be inserted within a `finally()`  block instead, so that we won't get an infinite spinner if one of the fetches time out?

src/react/components/AcademicSummary/Semester.js
```diff
@@ -0,0 +1,187 @@
...
+import PrimarySection from './PrimarySection';
+
+import './Semester.scss';
+
+const propTypes = {
+  canViewGrades: PropTypes.bool.isRequired,
+  transferCredit: PropTypes.object
+};
+
+class Semester extends Component {
```
davidlimcheng: @developish  I _think_ this can be a functional component?

src/react/components/Spinner.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Spinner = (props) => (
```
davidlimcheng: @developish There is an existing component in `react/components/base/Spinner.js` that is pretty much the same thing.

src/react/components/StudentProfile/ExpectedGraduation.js
```diff
@@ -0,0 +1,23 @@
...
+const propTypes = {
+  graduation: PropTypes.object.isRequired,
+  isAdvisingStudentLookup: PropTypes.bool.isRequired
+};
+
+const ExpectedGraduation = (props) => {
+  return (
+    <Fragment>
+      <GenericGraduation {...props} />
+      <LawGraduation terms={props.graduation.gradLaw.expectedGraduationTerms || []} />
```
davidlimcheng: @developish Not a blocker for merge, but I think it's cleaner to define `defaultProps` within the component in most cases:
```
LawGraduation.defaultProps = {
  graduation: {
    gradLaw: {
      expectedGraduationTerms: []
    }
  }
}
```

src/react/components/StudentProfile/StudentProfile.js
```diff
@@ -0,0 +1,157 @@
...
+const propTypes = {
+  academics: PropTypes.object.isRequired,
+  studentProfile: PropTypes.object.isRequired,
+  user: PropTypes.object.isRequired
+};
+
+class StudentProfile extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
```
davidlimcheng: @developish  I haven't looked at what component is passing props to this one, but why are these props being set to state in this component? That would create two different states with the same values that both have to be managed. 

Shouldn't these values be managed by the parent state, or wherever they originate? That way we are assured that each re-render passes the same values down the chain.

src/react/components/TransferCredit/ExamRow.js
davidlimcheng: I had assumed that JS would throw an error if a key wasn't there, but it looks like it'd just get set to `undefined`.

src/react/components/TransferCredit/GenericTransferCredit.js
```diff
@@ -0,0 +1,93 @@
...
+const hasGradePoints = (details) => details && details.find(transfer => transfer.gradePoints);
+
+const propTypes = {
+  detailed: PropTypes.array,
+  summary: PropTypes.object,
+  isStudent: PropTypes.bool,
+  reportLink: PropTypes.object
+};
+
+class GenericTransferCredit extends Component {
```
davidlimcheng: This can be changed to functional component

src/react/components/APILink.js
developish: Agreed

src/react/components/StudentProfile/StudentProfile.js
developish: This is basically to wrangle the API into something more manageable

src/react/components/AcademicSummary/Semester.js
developish: Could be, I don't think it's worth it at this time

src/react/components/Spinner.js
developish: I know, this is a simpler version that I would have recommended if you had been in on Friday. It's nice that we can have both and nothing breaks

src/react/components/StudentProfile/ExpectedGraduation.js
developish: Let's talk about this more later, I'd like to investigate more

src/react/components/TransferCredit/GenericTransferCredit.js
developish: Word

spec/controllers/advising_student_controller_spec.rb
developish: Yep, copy/paste error

davidlimcheng:
Thanks for the mammoth effort, @developish !

PR #7510, opened 2019-02-13
CLC-6983 More departments sign onto OEC

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6983

Dear Diary....


johncrossman:
Come on in, the water's fine.

PR #7514, opened 2019-02-15
SISRP-44030: Compliance Requirement: Cal Grant Acknowledgement needed during Online Registration

developish:

### DIFF
--- a/app/models/cal_grant/acknowledgement_expiry.rb
+++ b/app/models/cal_grant/acknowledgement_expiry.rb
@@ -1,0 +1,7 @@
+module CalGrant
+  module AcknowledgementExpiry
+    def self.expire(uid=nil)
+      CalGrant::Acknowledgement.expire uid
+    end
+  end
+end
--- a/app/models/cal_grant/queries.rb
+++ b/app/models/cal_grant/queries.rb
@@ -1,0 +1,18 @@
+module CalGrant
+  class Queries < ::EdoOracle::Connection
+    include ActiveRecordHelper
+    include ClassLogger
+    include Concerns::QueryHelper
+
+    def self.get_activity_guides(uid)
+      safe_query <<-SQL
+        SELECT
+          PTAI_ITEM_ID as id,
+          STRM as term_id,
+          ACKNLDG_STATUS as status
+        FROM SYSADM.PS_UCC_AG_FA001_VW
+        WHERE CAMPUS_ID = '#{uid}'
+      SQL
+    end
+  end
+end 
--- a/src/assets/javascripts/angular/factories/calGrantsFactory.js
+++ b/src/assets/javascripts/angular/factories/calGrantsFactory.js
@@ -1,0 +1,11 @@
+'use strict';
+
+angular.module('calcentral.factories').factory('calGrantsFactory', function(apiService) {
+  const url = '/api/my/calgrant_acknowledgements';
+
+  const getCalGrants = function(options) {
+    return apiService.http.request(options, url);
+  };
+
+  return { getCalGrants };
+});
--- a/src/react/components/CalGrantAcknowledgement/HoldListItem.js
+++ b/src/react/components/CalGrantAcknowledgement/HoldListItem.js
@@ -1,0 +1,66 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+
+import APILink from '../APILink';
+
+const propTypes = {
+  children: PropTypes.node,
+  title: PropTypes.string,
+  state: PropTypes.string,
+  actionLink: PropTypes.object,
+  serviceIndicator: PropTypes.object
+};
+
+class HoldListItem extends Component {
+  constructor(props) {
+    super(props);
+    this.state = { expanded: false };
+  }
+
+  toggle() {
+    this.setState({ expanded: !this.state.expanded });
+  }
+
+  classNames() {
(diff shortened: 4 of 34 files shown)


app/controllers/calgrant_acknowledgements_controller.rb
```diff
@@ -0,0 +1,5 @@
+class CalgrantAcknowledgementsController < ApplicationController
```
marcpechaitis: I see this file is named `calgrant_ack...` yet you're using `cal_grant` in the other folders?  should these be consistent?

marcpechaitis: specifically `cal_grant` vs `calgrant`?

app/models/cal_grant/queries.rb
```diff
@@ -0,0 +1,18 @@
+module CalGrant
+  class Queries < ::EdoOracle::Connection
```
davidlimcheng: We should discuss when it's appropriate to create a new model for queries, and when to use `EdoOracle::Queries`.

src/assets/javascripts/angular/factories/calGrantsFactory.js
```diff
@@ -0,0 +1,13 @@
...
+  const url = '/api/my/calgrant_acknowledgements';
+
+  const getCalGrants = function(options) {
+    return apiService.http.request(options, url);
+  };
+
+  return { getCalGrants };
+};
+
+angular.module('calcentral.factories').factory('calGrantsFactory', calGrantsFactory);
```
davidlimcheng: @developish Can this be moved to the top for consistency?

davidlimcheng: I also notice that it's different from the "function-block" (for lack of a better term) style the way our factories are usually defined - I'd like to keep these consistent as much as possible.

app/models/cal_grant/acknowledgement_expiry.rb
```diff
@@ -0,0 +1,7 @@
+module CalGrant
```
marcpechaitis: Why is this not part of `CampusSolutions` module?

src/react/helpers/forTermId.js
```diff
@@ -0,0 +1,3 @@
+const forTermId = (list, termId) => list.find(item => item.termId === termId);
```
davidlimcheng: @developish I don't think we should create files for one-off helpers. This seems like a term-specific helper function, so can we rename the file to "terms" or "termHelper.js" (need to discuss to stay consistent) so that in the future, we'll know to put any term-specific helper functions here?

Also would like to avoid using `export default`, and making it a named export instead.

src/react/components/CalGrantAcknowledgement/HoldListItem.js
```diff
@@ -0,0 +1,66 @@
+import React, { Component } from 'react';
```
marcpechaitis: Coming into it with the Campus Solutions background, I would suggest we should organize our folders with module-specific sub-folders:

I.e., instead of 
`src/react/components/CalGrantAcknowledgement/HoldListItem.js` 
use 
` src/react/components/fa/CalGrantAcknowledgement/HoldListItem.js`

app/controllers/calgrant_acknowledgements_controller.rb
developish: Yes

src/react/components/EnrollmentLinks/ConcurrentDrop.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
```
marcpechaitis: Same as my comment above... I think we need more organization:
`src/react/components/EnrollmentLinks/ConcurrentDrop.js`
becomes
`src/react/components/sr/EnrollmentLinks/ConcurrentDrop.js`

otherwise this flat structure has potential to get really messy?

app/models/cal_grant/acknowledgement_expiry.rb
developish: I'm open to that, I was mostly trying to stay away from messing with existing stuff and the concern is really more about enrollment eligibility (despite what causes it being related to financial aid)

app/models/cal_grant/queries.rb
developish: Agreed. I'm trying to intentionally suggest changes here. EdoOracle::Queries is a junk drawer. I'm trying organize by business concern rather than implementation detail.

src/assets/javascripts/angular/factories/calGrantsFactory.js
developish: Yeah

app/models/cal_grant/queries.rb
davidlimcheng: @developish Understood, but I think it would be helpful to discuss suggestions earlier, rather than having a quick discussion before rushing it into the codebase for testing purposes.

src/react/helpers/forTermId.js
developish: I would recommend "terms" rather than "termHelper"

developish: This is shared logic between components, which is why I extracted it to an importable file. But I agree with the not using default exports on the helper collections

PR #7519, opened 2019-02-16
SISRP-42529: Bugfix, failure to always show transfer credit in semester detail

developish:

That's the original JIRA, bug is in the comments WRT "Transfer Units Posted"

### DIFF
--- a/src/react/components/AcademicSummary/Enrollment.js
+++ b/src/react/components/AcademicSummary/Enrollment.js
@@ -15,8 +15,8 @@ const propTypes = {
 };
 
 const Enrollment = ({ gpaUnits, semesters, transferCredit, user, transferReportLink }) => {
-  const showSemesters = semesters.length && user.hasStudentHistory;
-  const showSummary = showSemesters && (user.roles.law || gpaUnits.totalLawUnits > 0);
+  const showSemesters = !!(semesters.length && user.hasStudentHistory);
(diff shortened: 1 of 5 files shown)


src/react/components/AcademicSummary/Enrollment.js
```diff
@@ -15,8 +15,8 @@ const propTypes = {
 };
 
 const Enrollment = ({ gpaUnits, semesters, transferCredit, user, transferReportLink }) => {
-  const showSemesters = semesters.length && user.hasStudentHistory;
-  const showSummary = showSemesters && (user.roles.law || gpaUnits.totalLawUnits > 0);
+  const showSemesters = semesters.length && user.hasStudentHistory ? true : false;
```
nkutub: @developish  I think this can be done shorter with !!(semesters.length && user.hasStudentHistory)

developish: Is that the preferred style? I'm ok with it, just find it personally a bit more cryptic

nkutub: we are used to it in other parts of the code

nkutub: so it is sorta perferred

nkutub:
LGTM @developish

PR #7521, opened 2019-02-19
SISRP-42529: Bugfix, prevent undefined credit report link exception

developish:

### DIFF
--- a/src/react/components/AcademicSummary/AcademicSummary.js
+++ b/src/react/components/AcademicSummary/AcademicSummary.js
@@ -72,6 +72,13 @@ class AcademicSummary extends Component {
     return (this.state.academics.semesters.length && this.state.user.hasStudentHistory) || this.state.showTransferCredit;
   }
 
+  tcReportLink() {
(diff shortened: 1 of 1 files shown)


src/react/components/AcademicSummary/AcademicSummary.js
```diff
@@ -106,7 +113,7 @@ class AcademicSummary extends Component {
                           semesters={this.state.academics.semesters}
                           user={this.state.user}
                           transferCredit={this.state.transferCredit}
-                          transferReportLink={this.state.academics.studentLinks.tcReportLink}
+                          transferReportLink={this.tcReportLink()}
```
marcpechaitis: @developish ... Just asking the question to see if I understand... what happens in `<Enrollment />` component if `this.tcReportLink()` is _null_?

developish: Nothing, it's only ever attempted to be rendered within `TransferCredit/TransferCreditTitle.js` if the user is a student. The bug was that `studentLinks` could have been undefined if the user is actually an advisor and calling tcReportLink on it raised an error.

PR #7526, opened 2019-02-21
CLC-6979 Copy applicants data to Loch

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6979

This also refactors our code to support the new range of Data Loch tasks, moving query-specific logic out of the S3 and Zipper classes.

Non-term-specific data about students in the advisee population will be copied to a 'sis-data/advisees' folder in S3.


app/models/data_loch/stocker.rb
```diff
@@ -14,46 +14,77 @@ def s3_from_names(targets)
...
+                Rails.logger.error "Got unknown job name #{job}!"
+              end
+            end
+          end
           previous_sids = sids
         end
-        s3.upload('demographics', demographics_path)
-        Rails.logger.info "Demographics snapshot complete."
+        job_paths.each do |job, path|
+          s3.upload("advisees/#{job}", path) if path
```
johncrossman: There it is, `advisees`!

johncrossman:
Looks good @raydavis, zippers and all. Thanks for the specs.

PR #7527, opened 2019-02-21
SISRP-44030: Allow expiration at the server with refresh=true

developish:


redconfetti:
Looks good @developish

PR #7531, opened 2019-02-25
SISRP-44030: Expire financial aid profiles on acknowledgement

developish:


app/controllers/my_holds_controller.rb
```diff
@@ -3,7 +3,7 @@ class MyHoldsController < ApplicationController
   before_filter :api_authenticate_401
 
   def get_feed
-    if params[:expireCache] == true
+    if params[:expireCache]
```
redconfetti: @developish - Thanks for throwing these in while you were at it

image

PR #7533, opened 2019-02-25
CLC-6956 Remove Course Captures dependency on legacy Oracle DB

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6956
https://jira.berkeley.edu/browse/SISRP-36642

Attention SIS CalCentral crew: This drops CalCentral's last dependency on the DB2-era DB (except as the home of CalDap data). After the merge, a config of `features.allow_legacy_fallback=false` should be equivalent to pulling academics data solely from the CS-era EDO DB.


johncrossman:
Thank you @raydavis  Ties severed without loss.

PR #7537, opened 2019-02-27
[QA] SISRP-42529: Changes requested by law

developish:

Also addresses some duplicate key warnings and prevents the component
from fetching data until it's actually mounted

See DEV PR #7536


redconfetti:
@developish - Looks good. I like that we can still do multiple requests using `Promise` that resolve only after they're all asynchronously finished.

PR #7551, opened 2019-03-06
SISRP-36642 CLC-7003 Add support for SISEDO CalDap view

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-7003
https://jira.berkeley.edu/browse/SISRP-36642

The new SISEDO.CALCENTRAL_PERSON_INFO_VW will only be referred to if the application is configured with `features.legacy_caldap=false`.

### DIFF
--- a/app/models/campus_oracle/queries.rb
+++ b/app/models/campus_oracle/queries.rb
@@ -2,25 +2,6 @@ module CampusOracle
   class Queries < Connection
     include ActiveRecordHelper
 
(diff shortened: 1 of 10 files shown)


raydavis:
@redconfetti is the developer on the SIS side of the task, and so he's the preferred reviewer. Other opinions are always welcome, though!

redconfetti:
Looks great @raydavis. Thanks for applying that to User::BasicAttributes.

PR #7554, opened 2019-03-07
NOJIRA: Use puma webserver in development

developish:


redconfetti:
@developish - Oh yeah. I'd always thought of this as being something to use in production more than local dev. Good call

PR #7569, opened 2019-04-16
 [QA] CLC-7013 Exclude secondary section enrollments for late withdrawals

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-7013

The OEC slowdown may make this code unusable in practice, but this should at least be a short-term functional fix.


johncrossman:
Thanks @raydavis

PR #7573, opened 2019-04-18
[QA] CLC-7034 Another attempt to thread the late-withdrawal needle

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-7034

This version appears to fulfill the test cases at CLC-7034, CLC-7033, and CLC-7013. Let's see how many more we can trip over.


paulkerschen:
I put my cap on and I thinked,
"All grade_marks must be distinct."

PR #7575, opened 2019-04-24
[QA] CLC-7036 Work around order-dependent Oracle issue in WHERE clause

raydavis:
QA-first workaround for https://jira.ets.berkeley.edu/jira/browse/CLC-7036

It's hard to interpret this as anything other than an Oracle bug, although it would be nice to get DBA confirmation. (Oracle has recently had at least one other bug relating to old-style WHERE-clause implicit joins.)

### DIFF
--- a/app/models/edo_oracle/bulk.rb
+++ b/app/models/edo_oracle/bulk.rb
@@ -59,8 +59,10 @@ module EdoOracle
           TRIM(crs."title") AS course_title,
           cls."allowedUnitsMaximum" AS allowed_units
         FROM
-          SISEDO.CLASSSECTIONALLV01_MVW sec,
-          SISEDO.EXTENDED_TERM_MVW term
+          SISEDO.CLASSSECTIONALLV01_MVW sec
+        JOIN SISEDO.EXTENDED_TERM_MVW term1 ON (
+          term1.STRM = sec."term-id" AND
+          term1.ACAD_CAREER = 'UGRD')
         LEFT OUTER JOIN SISEDO.DISPLAYNAMEXLATV01_MVW xlat ON (xlat."classDisplayName" = sec."displayName")
         LEFT OUTER JOIN SISEDO.API_COURSEV01_MVW crs ON (xlat."courseDisplayName" = crs."displayName")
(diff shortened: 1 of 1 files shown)


paulkerschen:
Q: Where does an 800-pound gorilla put its JOIN clause?
A: Anywhere it wants.

marcpechaitis:
@raydavis ,
I don't have insight into the issue as reported by JIRA, and I'd be interested in learning more about the problem this solves?

As you know we're using the same sql over on the sis-side...are we subject to same bug?  Can we connect (via Slack?) on Thursday?

raydavis:
Hi, @marcpechaitis -- yes, I figured I'd get in touch with you all today after doing a test on your code from https://github.com/sis-berkeley-edu/calcentral/pull/10/files

 Briefly, though, I originally mimicked your old-school implicit join in our code, except for putting the "term" table _second_ in the FROM clause:

```
        FROM
          SISEDO.CLASSSECTIONALLV01_MVW sec,
          SISEDO.EXTENDED_TERM_MVW term
        LEFT OUTER JOIN SISEDO.DISPLAYNAMEXLATV01_MVW xlat ON (xlat."classDisplayName" = sec."displayName")
        LEFT OUTER JOIN SISEDO.API_COURSEV01_MVW crs ON (xlat."courseDisplayName" = crs."displayName")
        LEFT OUTER JOIN SISEDO.CLASSV00_VW cls ON ( ...
```

This had already passed QA when we happened to hit the code again yesterday, and Oracle began throwing an error:
```
ORA-00904: "SEC"."displayName": invalid identifier
```

If I swapped the tables to put "SISEDO.EXTENDED_TERM_MVW term" first, the error went away. But order-dependent parsing in an implicit join seems so freakish that I chose to play it safe with an explicit join instead.

The underlying cause remains a very annoying mystery. So far I haven't come up with an explanation other than "bug in this version of Oracle," but maybe you can do better?

marcpechaitis:
@raydavis,
That is bizarre 🧐 ... when I get some time I want to play around with the SQL and see if I can replicate the error on our side.

cc: @developish , @davidlimcheng , @redconfetti , @nkutub , @frankiepun

PR #7579, opened 2019-05-01
CLC-7042 Exclude inactive CalNet populations from bulk UID search

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-7042


raydavis:
I'll fix the spec after I get back from my appointment.

raydavis:
Looks like this is ready for review.

paulkerschen:
The past isn't inactive, it's not even past.

PR #7581, opened 2019-05-02
NS-482 Load list of L&S students into Nessie S3

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-482

The first domino topples....


paulkerschen:
Into the crate!

raydavis:
Also noted at CLC-7047.

PR #7586, opened 2019-05-22
CLC-7028: Adds script to load advising notes data into S3

lyttam:
https://jira.ets.berkeley.edu/jira/browse/CLC-7028


lyttam:
@raydavis @paulkerschen After learning about stored procedures for a while, I finally figured out how to avoid splitting up the CLOB:  simply remove the DISTINCT from the query.  I think DISTINCT causes it to implicitly convert CLOB to VARCHAR.  

I'll separate note records from topic records on the Nessie side instead.

paulkerschen:
Congratulations on finding a way through, @lyttam. I'm impressed (if "impressed" is the word) that our H2 test db is willing to handle CLOB.

PR #7593, opened 2019-06-05
CLC-7053 Add Data Loch access check for sys admin use

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-7053

Also adds advisor note data to SISEDO DB view checker.


lyttam:
Thanks for adding the views!

PR #7595, opened 2019-06-05
[QA] CLC-7053 Add Data Loch access check for sys admin use

raydavis:
QA version of https://github.com/ets-berkeley-edu/calcentral/pull/7593


johncrossman:
You gotta test the waters before you jump in the Data Lake.

PR #7599, opened 2019-06-21
CLC-7063 Use SIS as an alternative provider of Gender/Ethnicity data

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-7063


johncrossman:
SIS, the provider of choice. Looks good @raydavis

PR #7602, opened 2019-07-01
CLC-7065 Junction can live without the legacy Oracle DB

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-7065

Another step toward shutdown.


paulkerschen:
Yet I fear the Oracle DB will be unable to live without Junction.

PR #7605, opened 2019-07-03
CLC-7059: Increases heap size for advisor data upload

lyttam:
When I tried to run the script on junction-dev, I ran into the old `Java::JavaLang::OutOfMemoryError`. Increasing the heap size resolved this.

https://jira.ets.berkeley.edu/jira/browse/CLC-7059


raydavis:
After ten minutes, Travis reports: "The build *created*. This is a change from the previous build, which *passed*."

The heck with you, Travis.

PR #7606, opened 2019-07-03
CLC-7059: Intifies new advisor columns

lyttam:
https://jira.ets.berkeley.edu/jira/browse/CLC-7059


paulkerschen:
From intractacle to intelligible.

PR #7611, opened 2019-07-10
CLC-7070 Flip executable bits on shell scripts

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-7070


lyttam:
Thank you for the fix.

PR #7617, opened 2019-07-12
CLC-7072 Continue to support pre-CS Course Captures in bCourses

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-7072

... dropping a bit of CalCentral-specific support code on the way.


johncrossman:
When it comes to Course Capture, I prefer the old stuff. So... thank you @raydavis!

PR #7618, opened 2019-07-15
CLC-7071 Update paths for temp directories in configs

paulkerschen:
Adjusts the data_loch path added in #7615, and the canvas_proxy path it was presumably copied from (which is already getting overridden in local configs, but no need to have a confusing default hanging around).


raydavis:
That train has left the station.

PR #7622, opened 2019-07-18
 CLC-6677 Remove MyModels, others

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6677

CampusSolutions to come...


johncrossman:
(Not) this year's models

PR #7623, opened 2019-07-22
CLC-6677 Adieu, Crosswalk, CampusSolution, FinancialAid

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-6677

This should be the last big deletion. Next will be dangling `lib` files, settings, etc. I'd like to leave the fixtures around a while longer, though.


paulkerschen:
The last zebra has crossed onto the ark.

PR #7625, opened 2019-07-23
CLC-7081 Send list of all active undergraduates to Data Loch

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-7081


paulkerschen:
Let's active.

PR #7633, opened 2019-08-07
CLC-7089: Increases max heap size for notes script

lyttam:
https://jira.ets.berkeley.edu/jira/browse/CLC-7089


raydavis:
Thanks, @lyttam !

PR #7640, opened 2019-08-21
CLC-7101 CLC-7103 Send term definitions to Data Loch

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-7101
https://jira.ets.berkeley.edu/jira/browse/CLC-7103

Also upgrades a  Gem and deletes some obsolete code.


paulkerschen:
Nokogiri, it's been a while.

PR #7645, opened 2019-09-11
CLC-7113 Log fatal exceptions from Rake shell scripts

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-7113

This solves a few puzzles from the past few years...


paulkerschen:
bash: still an innovation leader!

PR #7646, opened 2019-09-11
CLC-7110 Remove American Studies from OEC

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-7110


paulkerschen:
America had a good run.

PR #7647, opened 2019-09-19
CLC-7090 Handle Data Loch term transitions automatically

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-7090


paulkerschen:
It's about time we got a manager involved; I sure am tired of us doing it ourselves.

PR #7652, opened 2019-10-07
CLC-7130, load canvas-customization when 'load' event is fired

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-7130

In regard to issue above, R1 peer says, "...you should be able to fix it by wrapping in function: window.addEventListener('load', () => { ... })".  This PR follows that recommendation.

I suggest you append `?w=1` to URL of this page.

Note: The `DOMContentLoaded` approach described in https://github.com/instructure/canvas-lms/commit/48c055ed0e3c4b8e11a1103566e86ae99b03e54e is "not reliable".

### DIFF
--- a/public/canvas/index.js
+++ b/public/canvas/index.js
@@ -1,25 +1,29 @@
 (function(window, $) {
   'use strict';
 
-  window.CALCENTRAL = 'https://junction.berkeley.edu';
+  var loadCanvasCustomization = function() {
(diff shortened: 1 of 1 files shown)


raydavis:
Thanks, @johncrossman . I'll merge now, but of course someone will need to manually configure it inside ucberkeley.beta for testing.

PR #7654, opened 2019-10-09
CLC-7125 Add more logging for connection pool leakage

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-7125

Probably a temporary measure, but we'll see....

### DIFF
--- a/app/models/edo_oracle/connection.rb
+++ b/app/models/edo_oracle/connection.rb
@@ -12,7 +12,21 @@ module EdoOracle
     def self.query(sql, opts={})
       result = []
       use_pooled_connection do
+        if Rails.logger.debug?
+          Rails.logger.debug("#{self.name} working with connection #{connection}, object_id #{connection.object_id}, from pool #{connection.pool}")
+          pool_connections = connection.pool.connections
+          Rails.logger.debug("Current pool size #{pool_connections.size}")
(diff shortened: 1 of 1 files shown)


redconfetti:
@raydavis - Yeah. I think we're still dealing with the connection pool issue still also. Last time it came up I devised a theory, but we haven't followed up on it yet.

* [ActiveRecord Concurrency in Rails4: Avoid leaked connections!](https://bibwild.wordpress.com/2014/07/17/activerecord-concurrency-in-rails4-avoid-leaked-connections/)

When there is a new Thread (Thread.new) called, it makes a connection but might not check it back into the pool. It's possible our JMS workers are creating leaked connections. See [
[JmsWorker ](https://github.com/sis-berkeley-edu/calcentral/blob/master/lib/services/jms_worker.rb#L21).

> But if you create your own threads to use ActiveRecord (inside or outside a Rails app, doesn’t matter), you absolutely do.  If you proceed blithly to use AR like you are used to in Rails, but have created Threads yourself — then connections will be automatically checked out to you when needed…. and never checked back in.

You could try adding `ActiveRecord::Base.forbid_implicit_checkout_for_thread!` to the Thread calls and compare connection counts after dispatching JMS calls.

redconfetti:
On second thought, you're not using `Thread.new` any longer in your code. Maybe this isn't the case.

PR #7658, opened 2019-10-17
CLC-7125 Two-for-one experimental Oracle connection methods

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-7125

Every which way but loose.


paulkerschen:
There must be some way to shut off this thing.

PR #7664, opened 2019-10-28
CLC-7150 Add intended majors to Data Loch stocker runs

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-7150


johncrossman:
Another tributary of the data-lake.

PR #7668, opened 2019-11-04
CLC-7135 Whitelist to block inactivation of bCourses accounts

raydavis:
https://jira.ets.berkeley.edu/jira/browse/CLC-7135

This PR also removes the obsolete `author` account type, which was originally used to support campus-link-list editing.


johncrossman:
Thank you for tossing out the obsolete and bringin' in the goodness, @raydavis

PR #7684, opened 2020-01-03
CLC-7146, query S3 for latest knob file and make a record of its timestamp

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-7146

Logic to pull from S3 and check against the version currently running in Junction environment.


paulkerschen:
Makes sense!

PR #7695, opened 2020-01-22
CLC-7202, s3-based deploy scripts get dedicated dir: scripts/deploy

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-7202

Having one set of scripts to manage both old and new deploy strategies is too messy.


paulkerschen:
Congratulations indeed.

PR #7702, opened 2020-01-30
CLC-7205, boilerplate for is_deploy_necessary.sh script

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-7205


paulkerschen:
In a perfect world, deploy would never be necessary.

PR #7705, opened 2020-02-03
CLC-7213, in prod, Ops might need to deploy same knob-id more than once

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-7213

prod deploys will use `deploy-to-current-node.sh`, which tolerates same-knob-id deploys. Other environments use the more restrictive `deploy-to-all-nodes.sh`. Use `?w=1` to review.


johncrossman:
cc: @paulkerschen @lyttam

PR #7710, opened 2020-02-21
CLC-7204 Update secure_headers gem

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-7204

The gem API has substantially changed. See https://github.com/github/secure_headers/blob/master/docs/upgrading-to-3-0.md: "secure_headers 3.0 is a near-complete rewrite."


johncrossman:
Secure the humans and the headers!

PR #7711, opened 2020-02-21
CLC-7238 Get robust about encoding in CSV generation

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-7238


paulkerschen:
Travis and I are finally on good terms again.

PR #7712, opened 2020-02-24
BOAC-3314 Add academic standing to nightly loch jobs

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3314

Though academic standing entries come with a term attached, the BOA UX won't be thinking of them as per-term values and the update schedule doesn't fit with the rest of our per-term data. Accordingly we'll just grab the most recent standing available per SID and treat it as non-term-specific data.


johncrossman:
Looks good @paulkerschen  Waiting on Travis

PR #7733, opened 2020-04-06
CLC-7285: Updates advisor permission query to reflect SIS view change

lyttam:
https://jira.ets.berkeley.edu/jira/browse/CLC-7285


paulkerschen:
Thanks @lyttam. Looking at the Nessie code on the other side:

https://github.com/ets-berkeley-edu/nessie/blob/master/nessie/sql_templates/create_advisor_schema.template.sql#L89

it looks like the Nessie SQL template will have to be tweaked as soon as this hits production to avoid failure. So either we'd need to coordinate a patch to Nessie code, or else have Junction ship off some placeholder value for DISPLAY_ONLY until we can tell Nessie to stop expecting it. Which of those do you think would be less awkward?

paulkerschen:
@lyttam Or am I wrong, and if Redshift gets to the end of the row without finding a value for DISPLAY_ONLY, it will just set it to null and move on? The schema also seems to be looking for an AUTH_ACTIONS column that isn't there any more.

lyttam:
@paulkerschen I think Redshift will just ignore the missing columns.  I should be able to verify that with a test file before this PR is merged.

lyttam:
@paulkerschen I confirmed that the job completes and that DISPLAY_ONLY and AUTH_ACTIONS will be null when those columns are missing from the file that Junction creates. Thanks for checking! I hadn't considered potential release dependency problems.

paulkerschen:
Thanks for confirming, @lyttam!

PR #7755, opened 2020-04-16
CLC-7320 When you give Rails bad syntax, it just smiles, nods and ignores you

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-7320

Follow-up to #7754.


lyttam:
That's rather too polite.

PR #7758, opened 2020-04-22
CLC-7307 CLC-7312 Rework Canvas enrollment refresh to accommodate recent updates only

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-7307
https://jira.ets.berkeley.edu/jira/browse/CLC-7308
https://jira.ets.berkeley.edu/jira/browse/CLC-7312
https://jira.ets.berkeley.edu/jira/browse/CLC-7313

Some TODOs still in the mix and more tests needed, but this should get the basic logic in place. Our Canvas refresh job now has three flavors, represented as subclasses of a base job:
- the "all" flavor which continues to do a comprehensive update every night;
- the "accounts" flavor which we run as a manual refresh every so often;
- the new "recent" flavor which we expect to pick up new memberships (but, in general, not drop old ones) every hour or so.


johncrossman:
Thanks @paulkerschen  Looks good. Let's give it a go.

PR #7766, opened 2020-05-04
CLC-7344 bCourses mailing lists: retry and failure notifications

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-7344


johncrossman:
If life gives you lemons then `retries += 1`

PR #7778, opened 2020-06-09
CLC-7386 Let's try batching that student attributes query

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-7386

The big data got too big for the JVM.


johncrossman:
Batch it good!

PR #7788, opened 2020-06-17
[JQA] CLC-7392, webcast/recordings init fix: 'fake' false in non-test env

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-7392

`fixtures/course_capture/legacy_recordings.json` is being used in non-test envs. We _should_ be using the file under `~/.calcentral_config/`


johncrossman:
master PR will follow

PR #7812, opened 2020-08-13
CLC-7440 Remove bConnected card, non-OEC OAuth wiring

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-7440

Also take out a lot of dummy JSON no longer relevant to the kind of dummies we are nowadays.


johncrossman:
Dummies of the world unite! Waiting on T.

PR #7815, opened 2020-08-19
CLC-7450 Respect Canvas gradebook override in E-Grades exports

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-7450


johncrossman:
Freedom is having the right to override.

PR #7829, opened 2020-09-03
CLC-7459 Work around JRuby 9.2 DateTime serialization issue

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-7459

Devs who have noticed a slew of “invalid offset is ignored” warnings in our recent spec runs might be interested in https://github.com/jruby/jruby/issues/6385. It hits us whenever we try to stash term dates in the Rails cache, and we can work around it with a simple wrapper class using Ruby delegation.


johncrossman:
Thanks for the well written summary in JRuby ticket.

redconfetti:
@paulkerschen - Thanks for finding the fix to this

PR #7830, opened 2020-09-03
CLC-7459 Remove deprecated filter syntax

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-7459

A filter by any other name...


johncrossman:
Actions speak louder than filters. Waiting on T.

PR #7834, opened 2020-09-03
CLC-7467, auth api from Vue; allow cross-site when local; mixin/utils

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-7467


paulkerschen:
oxfordJoin!

PR #7836, opened 2020-09-04
CLC-7467, /toolbox view with ApiTest card

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-7467

Wire in more API endpoints. Bootstrap icons.


paulkerschen:
Moving right along.

PR #7847, opened 2020-09-14
CLC-7493, solve the CORS problem when running front-end on diff port

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-7493


paulkerschen:
I’m convinced.

PR #7850, opened 2020-09-15
NOJIRA, Junction README: dev setup for front-end (Vue) work

johncrossman:
Devs doing Junction dev on local must have:
```yaml
 vue:
  localhost_base_url: 'http://localhost:8080'
```


paulkerschen:
Thanks for the user manual!

PR #7853, opened 2020-09-15
CLC-7501, 'Act As' component and correspond API calls (work in progress)

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-7501

Shared component for `saved` and `recent` users. More to come.


paulkerschen:
Act naturally.

PR #7856, opened 2020-09-16
CLC-7501, when acting-as the footer is fixed with 'Stop acting' button

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-7501

`?w=1` might help.


paulkerschen:
Cut!

PR #7858, opened 2020-09-16
CLC-7540 Port over global Junction/bCourses SCSS

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-7540

Work in progress. We probably don’t need all of this, but the dependencies are complex.


johncrossman:
@paulkerschen Thank you for porting over the styles. Makes sense. Npm complains about "corrupted" package. Any idea what's up?

paulkerschen:
Thanks, @johncrossman, looking into it. It was behaving itself locally.

paulkerschen:
@johncrossman I have failed to solve the corrupted package mystery, but the problem appears to have resolved itself.

PR #7862, opened 2020-09-17
CLC-7507, run OEC task, tail log, and then announce success or fail

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-7507


paulkerschen:
OEC! OEC!

PR #7866, opened 2020-09-18
CLC-7516, /roster view and header component; fix participating-depts in OEC widget

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-7516

Let's get it started.

### DIFF
--- a/src/views/Login.vue
+++ b/src/views/Login.vue
@@ -20,7 +20,7 @@
       </div>
     </div>
     <div>
-      <Footer />
+      <Footer :show-build-summary="true" />
     </div>
   </div>
 </template>
--- a/src/views/StandaloneLti.vue
+++ b/src/views/StandaloneLti.vue
@@ -8,7 +8,7 @@
         </div>
       </b-col>
     </b-row>
-    <Footer />
+    <Footer :show-build-summary="true" />
   </div>
 </template>
 
--- a/src/assets/styles/_bootstrap_variables.scss
+++ b/src/assets/styles/_bootstrap_variables.scss
@@ -1,0 +1,23 @@
+.btn-cc-blue {
+  background: #3b7ea1;
+  border: 1px solid #265269;
+  color: #fff;
+}
+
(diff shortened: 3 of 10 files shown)


src/views/Login.vue
```diff
@@ -20,7 +20,7 @@
       </div>
     </div>
     <div>
-      <Footer />
+      <Footer :show-build-summary="true" />
```
johncrossman, line 23: I assume we don't want to show build-info in LTI views.

src/views/StandaloneLti.vue
```diff
@@ -1,6 +1,5 @@
 <template>
   <div>
-    <ToolboxHeader />
```
johncrossman, line 3: No toolbox-header in LTI views?

paulkerschen: @johncrossman If we want to follow current Junction behavior, then this `StandaloneLti` view is equivalent to the current https://junction-dev.berkeley.edu/canvas/user_provision - what we use to look at LTI tools in a CalCentral (not LTI) context. Maybe this view needs a better name. Anyhow, in this context it’s probably useful to keep around the Junction-style tool header, footer build summary and so on.

When viewing LTI tools in the actual LTI context, we may need to use a completely different view (equivalent to the difference between `src/bcourses_embedded.html` and `src/index-junction.html` on the current master branch).

johncrossman: Thanks @paulkerschen Makes sense. I'll revert that line.

src/views/Login.vue
paulkerschen: What do you think about restoring the build summary here too? In the actual LTI context there won’t be a footer component at all.

johncrossman: @paulkerschen My comment was misplaced. We _are_ showing build-info in this case.

paulkerschen: Oh, right, of course.

PR #7867, opened 2020-09-18
CLC-7548, if error then forward to /error; tweaks to axios interceptor

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-7548


paulkerschen:
Thanks, @johncrossman, let’s get this in. If we’re lucky, catchall error handling will meet our needs for the LTI tools; if not, hopefully not hard to tweak.

PR #7869, opened 2020-09-21
CLC-7515, roster photos: sections dropdown and student listing

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-7515


src/views/Roster.vue
```diff
@@ -0,0 +1,121 @@
...
+  mixins: [Context, Util],
+  components: {RosterPhoto},
+  data: () => ({
+    context: 'canvas',
+    courseId: undefined,
+    roster: undefined,
+    section: null
+  }),
+  mounted() {
+    // this.courseId = this.$route.query.id || 1493973
```
paulkerschen, line 108: @johncrossman Stray debugging line?

johncrossman: Ah yes, removing.

paulkerschen: FYI, if we want to preserve the current routing in https://github.com/ets-berkeley-edu/calcentral/blob/master/src/assets/javascripts/angular/configuration/routeConfiguration.js, then the non-LTI route for this view would be `canvas/rosters/:canvas_course_id`. That way you can inject any course id you like during development.

johncrossman: @paulkerschen Did you see the change in `router.ts`?  What you describe was my intention.

paulkerschen: Oh yeah, there it is. I thought it wasn’t there before, but clearly I’m senile.

PR #7875, opened 2020-09-22
[QA] CLC-7552 Update column name in EDO decompressor

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-7552


johncrossman:
Decompressors are essential.

PR #7885, opened 2020-09-24
NOJIRA, a few dev node dependencies are required to build WAR bundle

johncrossman:
This will add a little weight to the WAR but not much.


paulkerschen:
(What is it good for?)

PR #7886, opened 2020-09-24
CLC-7557, warbler must include 'dist'

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-7557

Warblers gonna warble.


config/warble.rb
```diff
@@ -1,6 +1,6 @@
 Warbler::Config.new do |config|
   config.override_gem_home = false
-  config.dirs = %w(app bin config db dist lib log public script vendor versions tmp)
+  config.dirs = %w(app bin config dist db lib log public script vendor versions tmp)
```
paulkerschen, line 3: @johncrossman Looks like this isn’t an addition but a rearrangement. Still needed?

johncrossman: Ha! Okay, this will need a bit more investigation. I'll close and investigate our WAR'ing mechanism.

PR #7890, opened 2020-09-27
CLC-7557, git mv public/dummy (and similar) out of 'public' dir

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-7557

The `public/` dir should only have `index.html` (Vue) and `favicon.ico`. It currently contains test data, a stock image, and a few maintenance HTML files. The first step in migrating stuff out of that dir: test data and stock image. The goal is to move all web assets into the `src/` dir. The rest of it will live under `test/`, `docs/`, or `config/`.


johncrossman:
Silly Travis, one successful build a all we need. Cc: @paulkerschen

paulkerschen:
Makes sense.

PR #7895, opened 2020-09-29
CLC-7573, remove old 'CalCentral' maintenance pages; deploy script need not copy index_maintenance.html

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-7573

`index_maintenance.html` is bundled in the `dist` dir and does not need to be copied by deploy scripts. The Apache config will need an update.


johncrossman:
Explaining the diff of this PR:
* removed `public/maintenance` (old CalCentral maintenance page)
* moved  `public/junction-maintenance` to `public/maintenance`
* renamed `public/maintenance/index_maintenance.html` to `public/maintenance/index.html`

PR #7906, opened 2020-10-01
CLC-7556 Accessibility announcer

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-7556

Old-school DOM manipulation here, but seems like it did the job in Angular land.


johncrossman:
@paulkerschen - Thank you for the announce-ability. Linter is a bit grumpy.

PR #7913, opened 2020-10-01
CLC-7580 Consistent markup for outbound links

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-7580


johncrossman:
Components-R-Us!

PR #7916, opened 2020-10-02
CLC-7525, migrate '_canvas_user_*.scss' files

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-7525

Including create-course-site styles.


paulkerschen:
@johncrossman FWIW, I’ve been trying to follow Vue patterns by adding these styles as scoped to the components themselves. course_grade_export, for example: https://github.com/ets-berkeley-edu/calcentral/blob/master/src/components/bcourses/CourseGradeExport.vue

PR #7917, opened 2020-10-02
CLC-7561 Find a Person to Add: Vue template layout

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-7561


src/assets/styles/lib/foundation.css
```diff
@@ -31,74 +31,6 @@ select {
...
-}
-
-[class*="column"] + [class*="column"]:last-child {
-  float: right;
-}
-
-[class*="column"] + [class*="column"].end {
-  float: left;
-}
-
```
paulkerschen, line 101: Foundation’s `row` and `columns` styling conflicts with classes used by bootstrap-vue and causes unpredictable spacing. I don’t think we can have both in the same project.

When porting over the Angular templates, I’ve been replacing `<div class=“row”>` or `<div class=“row collapse”>` with `<b-row no-gutters`>, and that seems to translate well.

Likewise, `<div class=“md-3 columns”>` would become `<b-col md=“3”>`.

johncrossman:
@paulkerschen Looks good. Although Travis failed on `@import "row_collapse_bug"`

paulkerschen:
Thanks, @johncrossman, fixed.

PR #7918, opened 2020-10-02
CLC-7526, CourseSectionsTable component is Vue equivalent of Angular data-bc-sections-table

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-7526


paulkerschen:
It’s cellular, modular, interactive-odular.

PR #7921, opened 2020-10-05
CLC-7525, break CreateCourseSite.vue into components per 'workflowStep'

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-7525

An all-in-one `CreateCourseSite` component is unwieldy. Breaking up its template code based on steps in workflow makes sense, methinks.


lyttam:
Makes sense to me!

PR #7923, opened 2020-10-05
CLC-7524, fix create-site-step-1 formatting after ouster of foundation CSS

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-7524

And moved some global styles to more appropriate 'scoped' realm.


paulkerschen:
Ousted!

PR #7924, opened 2020-10-05
CLC-7563 CLC-7564 Legacy Course Capture LTI

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-7563
https://jira.ets.berkeley.edu/jira/browse/CLC-7564


src/assets/styles/junction-global.scss
```diff
@@ -47,9 +47,6 @@
 // Academics
 @import "roster";
 
-// Webcasts
-@import "webcast";
-
```
paulkerschen, line 52: Contents of this file are moved into the CourseCaptures component.

johncrossman:
Ah, webcasts. Things were so simple then.

PR #7944, opened 2020-10-08
CLC-7596 Make error routing available to Vue components

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-7596

We love our vue-router error handling. But then again, sometimes components need to deal with a 401 or a 403 themselves. But then again again, not every component knows what to do with a 401 or a 403. Maybe this pattern will help.


johncrossman:
I like this pattern. You found the middle ground.

PR #7945, opened 2020-10-08
CLC-7526, instructors go directly to 'selecting sections' step in create-course-site workflow

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-7526


paulkerschen:
Onward!

PR #7946, opened 2020-10-08
CLC-7528, in select-sections step, select some or all; style fixes

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-7528


paulkerschen:
Templates within templates...

PR #7947, opened 2020-10-08
CLC-7530, kick off progress bar with selected CCNs

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-7530


paulkerschen:
Well, show me the way to the next progress bar...

PR #7959, opened 2020-10-09
CLC-7605, use OutboundLink for mailto links in Roster Photos

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-7605


src/components/util/OutboundLink.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <a :href="href" target="_blank">
+  <a :id="id || `link-to-${href.replace(/\W/g, '')}`" :href="href" target="_blank">
```
johncrossman, line 2: For fallback id, remove non-alphanumeric from `href`.

PR #7961, opened 2020-10-10
CLC-7607, Roster download filename when 'embedded'

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/CLC-7607

### DIFF
--- a/src/views/Roster.vue
+++ b/src/views/Roster.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="cc-page-roster">
+  <div v-if="!loading" class="cc-page-roster">
     <b-container v-if="roster && !error" fluid>
       <b-row align-v="start" class="cc-page-roster cc-print-hide cc-roster-search" no-gutters>
         <b-col class="pb-2 pr-2" sm="3">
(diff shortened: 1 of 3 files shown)


src/views/Roster.vue
```diff
@@ -132,7 +132,19 @@ export default {
     viewMode: 'photos'
   }),
   created() {
+    this.$loading()
     this.getCanvasCourseId()
+    if (this.canvasCourseId === 'embedded') {
+      getCourseUserRoles(this.canvasCourseId).then(
+        data => {
+          this.canvasCourseId = data.courseId
```
johncrossman, line 140: Do we have a preferred way of getting actual canvasCourseId when embedded?

paulkerschen: A lot of tools don't need to know the actual canvasCourseId on the front end - they just stick with "embedded" and let server-side logic pull the course from session properties when API calls are made.

If the front end genuinely needs to know the course ID, I think this is a fine way to do it - I did the same thing in CourseGradeExport.vue because the front end needs the actual course ID to build Canvas URLs.

PR #7964, opened 2020-10-20
[QA] CLC-7626 Use connection pooling with Dalli memcached client

paulkerschen:
DO NOT MERGE until @pfarestveit greenlights.

https://jira.ets.berkeley.edu/jira/browse/CLC-7626


config/settings.yml
```diff
@@ -180,6 +180,7 @@ cache:
   maximum_expires_in: <%= 29.days %>
   race_condition_ttl: <%= 10.seconds %>
   stale_connection_flush_interval: <%= 30.seconds %>
+  pool: 20
```
paulkerschen, line 183: We may need to tweak this setting in production, but let’s start by assuming that our memcached transactions are quick enough that we can use a smaller pool than Postgres (95) or SISEDO (200).

PR #7979, opened 2020-11-16
CLC-7601 CLC-7603 Standardize on old-style spinner

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-7601
https://jira.ets.berkeley.edu/jira/browse/CLC-7603

This is the quickest way to get consistent; if we want to move to the fancy new Bootstrap spinner we’ll have to implement it across the board.


lyttam:
Spinners gonna spin!

PR #7982, opened 2020-11-17
CLC-7615 Fix keyboard navigation in Vue forms

paulkerschen:
Fixes https://jira.ets.berkeley.edu/jira/browse/CLC-7615 and similar problems elsewhere. Using `type=“submit”` on buttons and `@submit.prevent` in forms is the accessibility lesson of the day.


johncrossman:
@paulkerschen - Travis did not wake up for this one. I think a minor amend will do the trick.

paulkerschen:
Travis is dead to the world, but scout’s honor my local linter passes.

PR #7983, opened 2020-11-17
CLC-7604, printable roster with proper stylings

johncrossman:


johncrossman:
I amended this PR and still no Travis. Local lint was happy.

PR #7987, opened 2020-11-17
CLC-7618 Client-side validation for user search

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-7618


johncrossman:
Looks good. Waiting on T.

PR #7995, opened 2020-11-19
CLC-7602 Proper embedded paths for site creation

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-7602


pfarestveit:
Looks right proper to me.

PR #8004, opened 2020-11-30
CLC-7645 DATA science for OEC

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-7645


db/schema.rb
```diff
@@ -1,97 +1,90 @@
-# encoding: UTF-8
```
paulkerschen, line 1: Changes to `schema.rb` formatting are not significant; it’s just updated to Rails 6 formatting as part of the db migration.

johncrossman:
One more PDO app gets Data Science. Looks good. Waiting on T.

PR #8014, opened 2020-12-03
CLC-7657 Message when no students in roster

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/CLC-7657


johncrossman:
Yes we have no students.

PR #8016, opened 2020-12-03
CLC-7604 Tweak roster photo grid display

paulkerschen:


paulkerschen:
Hardcoded sizes look inelegant but the whole grid is just too volatile without them.

PR #8017, opened 2020-12-04
CLC-7604 One more go-round on the roster grid

paulkerschen:


johncrossman:
Thank you for wrestling with this, @paulkerschen   I had a hard time with that CSS.

